PR #1, opened 2026-06-04
feat: implement config loading + session store (T4 + T5)

lml2468:
## Summary

Implements T4 (config system) and T5 (SQLite persistence layer) for cc-channel-octo.

### T4: Config (`src/config.ts` + `config.example.json`)

- `loadConfig()` with three-level priority: env vars (`CC_OCTO_*`) > `config.json` > defaults
- Strict integer parsing for numeric env vars, CSV parsing for list fields
- Validates required fields (`botToken`, `apiUrl`) — throws on missing
- `config.example.json` updated with safe/full mode documentation via `_comment` field

### T5: DB Adapter + Session Store (`src/db-adapter.ts` + `src/session-store.ts`)

- `DbAdapter` / `PreparedStatement` / `RunResult` interfaces — thin abstraction enabling future `node:sqlite` migration
- `BetterSqliteAdapter`: auto-creates parent dirs, WAL mode, foreign keys, 5s busy timeout
- `SessionStore` with prepared statements for all operations:
  - `getOrCreate` — creates or touches existing session
  - `appendUser` / `appendAssistant` — insert message + touch session `updated_at`
  - `buildHistoryPrefix` — last N messages (default 40) ordered by AUTOINCREMENT `id`
  - `cleanExpired` — removes sessions inactive > 7 days (`ON DELETE CASCADE` for messages)
  - `deleteSession` — explicit session removal
- Schema: `sessions` + `messages` (AUTOINCREMENT PK for stable ordering) + `group_members`

### Files changed (4 only, as scoped)

- `src/config.ts`
- `src/db-adapter.ts`
- `src/session-store.ts`
- `config.example.json`

### Verification

```
```

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -1,4 +1,5 @@
 {
+  "_comment": "Safe mode (default): allowedTools restricted to read-only tools. Full mode: add Write, Edit, Bash to allowedTools. Env vars (CC_OCTO_*) override these values. See README.md.",
   "botToken": "bf_YOUR_BOT_TOKEN",
   "apiUrl": "https://your-octo-instance.com",
   "cwd": "/path/to/your/project",
--- a/config.example.json
+++ b/config.example.json
@@ -1,5 +1,6 @@
 {
-  "_comment": "Safe mode (default): allowedTools restricted to read-only tools. Full mode: add Write, Edit, Bash to allowedTools. Env vars (CC_OCTO_*) override these values. See README.md.",
+  "_comment_safe": "Safe mode (default): read-only tools, settingSources=['user'] only. This is the recommended starting point.",
+  "_comment_full": "Full mode: add Write, Edit, Bash to allowedTools. Adding 'project' to settingSources loads cwd/.claude/settings.json — WARNING: malicious repos can override allowedTools via project settings hooks. Only use 'project' with trusted codebases.",
   "botToken": "bf_YOUR_BOT_TOKEN",
   "apiUrl": "https://your-octo-instance.com",
   "cwd": "/path/to/your/project",
@@ -8,7 +9,7 @@
   "sdk": {
     "allowedTools": ["Read", "Glob", "Grep", "WebFetch", "WebSearch"],
     "permissionMode": "bypassPermissions",
-    "settingSources": ["user", "project"]
+    "settingSources": ["user"]
   },
 
   "rateLimit": {
--- a/src/config.ts
+++ b/src/config.ts
@@ -48,7 +48,7 @@ function defaults(): Config {
     sdk: {
       allowedTools: ['Read', 'Glob', 'Grep', 'WebFetch', 'WebSearch'],
       permissionMode: 'bypassPermissions',
-      settingSources: ['user', 'project'],
+      settingSources: ['user'],
     },
     rateLimit: {
       maxPerMinute: 5,
@@ -60,12 +60,18 @@ function defaults(): Config {
   };
 }
 
-function readConfigFile(path: string): PartialConfig {
-  if (!existsSync(path)) {
+function readConfigFile(configFilePath: string): PartialConfig {
+  if (!existsSync(configFilePath)) {
     return {};
   }
-  const raw = readFileSync(path, 'utf-8');
-  const parsed = JSON.parse(raw) as Record<string, unknown> & PartialConfig;
+  const raw = readFileSync(configFilePath, 'utf-8');
+  let parsed: Record<string, unknown> & PartialConfig;
+  try {
+    parsed = JSON.parse(raw) as Record<string, unknown> & PartialConfig;
+  } catch (err) {
+    const msg = err instanceof Error ? err.message : String(err);
+    throw new Error(`Failed to parse config file ${configFilePath}: ${msg}`);
+  }
   // Strip top-level keys starting with "_" (e.g. _comment).
   const cleaned: Record<string, unknown> = {};
   for (const [k, v] of Object.entries(parsed)) {
@@ -104,9 +110,12 @@ function parseCsv(value: string): string[] {
 }
 
 function parseIntStrict(value: string, name: string): number {
-  const n = Number(value);
-  if (!Number.isFinite(n) || !Number.isInteger(n)) {
+  if (!/^\d+$/.test(value)) {
     throw new Error(`Invalid integer for ${name}: ${value}`);
+  }
+  const n = Number(value);
+  if (!Number.isFinite(n) || n < 1) {
+    throw new Error(`Invalid integer for ${name}: ${value} (must be >= 1)`);
   }
   return n;
 }
@@ -136,6 +145,9 @@ function applyEnv(cfg: Config): Config {
     next.sdk.maxTurns = parseIntStrict(env.CC_OCTO_SDK_MAX_TURNS, 'CC_OCTO_SDK_MAX_TURNS');
   }
   if (env.CC_OCTO_SDK_SYSTEM_PROMPT) next.sdk.systemPrompt = env.CC_OCTO_SDK_SYSTEM_PROMPT;
+  if (env.CC_OCTO_SDK_SETTING_SOURCES) {
+    next.sdk.settingSources = parseCsv(env.CC_OCTO_SDK_SETTING_SOURCES);
+  }
 
   if (env.CC_OCTO_RATE_LIMIT_MAX_PER_MINUTE) {
     next.rateLimit.maxPerMinute = parseIntStrict(
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,9 +1,9 @@
 /**
  * Configuration loading.
  * Three-level priority: env > config.json > defaults.
- *
- * Will be implemented in T4.
  */
+
+import { readFileSync, existsSync } from 'node:fs';
 
 export interface Config {
   botToken: string;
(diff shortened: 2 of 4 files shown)


Jerry-Xin (changes requested):
Summary: Relevance gate passed; this PR is in scope for `cc-channel-octo`, but the config environment override contract is incomplete.

🔴 Blocking

- 🔴 Critical — `sdk.settingSources` cannot be overridden by environment variables, breaking the documented `env > config.json > defaults` priority.
  `Config` includes `sdk.settingSources` at [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ife_v9em/repo/src/config.ts:19), and `config.example.json` documents it as a configured list at [config.example.json](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ife_v9em/repo/config.example.json:11). However, `applyEnv()` handles other SDK fields and CSV list fields but never reads something like `CC_OCTO_SDK_SETTING_SOURCES` at [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ife_v9em/repo/src/config.ts:114). That means a value from `config.json` silently wins over env for this field, contrary to the PR’s core T4 behavior. Add the missing env override using `parseCsv()` and cover it with a config test.

💬 Non-blocking

- 🟡 Warning — `parseIntStrict()` is not very strict for human-facing config input. At [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ife_v9em/repo/src/config.ts:106), `Number()` accepts forms like `1e3`, `0x10`, and whitespace-padded values. It also allows `0` and negative values for limits. Consider a decimal integer regex plus range checks for `maxPerMinute`, `maxContextChars`, `historyLimit`, and `maxTurns`.

- 🟡 Warning — `messages` has no index for the main history query. `buildHistoryPrefix()` queries by `session_id` and orders by `id` at [src/session-store.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ife_v9em/repo/src/session-store.ts:94). Add an index such as `(session_id, id)` before sessions accumulate enough rows for this to matter.

- 🔵 Suggestion — Add focused tests for `loadConfig()` precedence, env CSV parsing, invalid integer rejection, required-field failures, `getOrCreate()`, append/history ordering, and cascade deletion. The current test suite only covers Octo protocol behavior.

✅ Highlights

- The SQLite schema uses FK cascade for session cleanup, and `BetterSqliteAdapter` enables WAL, foreign keys, and a busy timeout.
- Prepared statements are initialized once and reused, which is a good fit for the synchronous `better-sqlite3` adapter.
- Verified after `npm ci`: `npm run type-check` passed, `npm test` passed with 37 tests, and `npm run lint` had warnings only in pre-existing Octo socket code.

yujiawei (changes requested):
# Code Review — PR #1 (cc-channel-octo)

Independent review of T4 (config loading) + T5 (SQLite persistence). This is a security-sensitive PR (`needs-human-review`), so I weighted the configuration defaults heavily.

**Verdict: Changes requested** — one security-default item should be hardened before merge. The rest is solid work; everything else below is non-blocking.

## Summary

Clean, well-structured infra PR. Verified locally against head `c47560d`:

- ✅ `npx tsc --noEmit` — passes, zero errors
- ✅ `npx eslint src/` — 0 errors (13 warnings, all pre-existing in `src/octo/socket.ts`, untouched by this PR)
- ✅ `loadConfig` precedence (env > file > defaults), `_`-prefix stripping, CSV/int parsing, required-field validation — logic is correct
- ✅ `FOREIGN KEY ... ON DELETE CASCADE` actually fires — confirmed `foreign_keys = ON` is set on the single better-sqlite3 connection, so `cleanExpired`/`deleteSession` cascade to `messages` as intended
- ✅ `buildHistoryPrefix` DESC-then-`reverse()` yields chronological order — correct
- ✅ Scope is exactly the 4 files described; no consumers wired yet (`agent-bridge.ts`, `session-router.ts`, `group-context.ts` remain stubs)

No correctness bugs, no data-loss bugs, no build breakage.

---

## 🔴 Blocking — harden the zero-config security default

**`src/config.ts:48-51` (and mirrored in `config.example.json`)**

```ts
sdk: {
  allowedTools: ['Read', 'Glob', 'Grep', 'WebFetch', 'WebSearch'],
  settingSources: ['user', 'project'],
},
```

The shipped default combines `permissionMode: 'bypassPermissions'` with `settingSources: ['user', 'project']`. When this reaches the Agent SDK `query()`, loading the `'project'` setting source means the SDK reads `.claude/settings.json` from the resolved `cwd` (default `process.cwd()`), which can define **hooks** (`PreToolUse`/`PostToolUse`/…) that auto-execute shell commands. Hooks run **outside** the `allowedTools` allowlist, so the read-only "safe mode" tool list does **not** mitigate this. Net effect once wired: any `.claude/settings.json` present in the working directory can run arbitrary commands with the bot's privileges, with no user gate — even in the "safe" profile.

This is **inert in this diff** (the SDK is not called yet — `agent-bridge.ts` is a stub for T10), which is why it is not an active RCE today. But it is being baked in **now** as the shipped default and the documented `config.example.json`, and will go live unchanged the instant T10 wires `query()`. Defaults are the security boundary, and this is the cheap moment to fix them.

**Requested change (pick one, minimally):**
- Default `settingSources` to `[]` (or drop `'project'`), forcing operators to opt into hook auto-exec explicitly; and/or
- Default `permissionMode` to a non-bypass mode and require explicit opt-in for `bypassPermissions`.

If the team deliberately wants this posture for the MVP, that's a legitimate call — but it should be an explicit, documented decision rather than the silent zero-config default, and I'd want a human security owner to sign off (this is exactly what `needs-human-review` is for).

---

## 🟡 Non-blocking — recommended (P2)

1. **No index on `messages(session_id)`** (`src/session-store.ts:37-44`). `selectRecentMessages` (`WHERE session_id = ? ORDER BY id DESC`) and every cascade delete locate child rows by `session_id`; SQLite does not auto-index the FK child side, so both full-scan. Cheapest highest-value schema fix, and now (before data/consumers) is the right time:
   ```sql
   CREATE INDEX IF NOT EXISTS idx_messages_session_id ON messages(session_id);
   ```
2. **Duplicated `historyLimit` default with a divergence trap** (`config.ts:58` sets `context.historyLimit = 40`; `session-store.ts:55` independently defines `DEFAULT_HISTORY_LIMIT = 40`). `buildHistoryPrefix` falls back to its own constant and never reads `Config.context.historyLimit`, so an operator who sets `historyLimit` will silently still get 40 unless every future caller threads the config value through. Decide the source of truth now — have `SessionStore` take the limit from config at init/call time and drop the local constant.
3. **`group_members` table has no CRUD methods** (`src/session-store.ts:46-52`) — created and `init()`-run, but no producer/consumer anywhere (its owner `group-context.ts` is a T9 stub). Either add the accessors with the table, or defer the `CREATE TABLE` until T9 so schema and accessors land together.
4. **Bad `config.json` throws a raw `SyntaxError`** (`config.ts:67-68`) with no file path. `JSON.parse` is unguarded while the required-field errors right below are clear and actionable. For an operator-hand-edited, security-sensitive config, wrap and rethrow as `Failed to parse config file <path>: <message>`.
5. **No tests for any of the three new modules.** `vitest` is already wired and `SessionStore`/`loadConfig` are DI-friendly (adapter injected, path injectable). The high-value pure units — `parseIntStrict` edge cases, env>file>defaults precedence, `buildHistoryPrefix` ordering, `cleanExpired` cutoff math, `getOrCreate` touch-vs-insert — have no coverage, and `tsc` won't catch an inverted `reverse()` or off-by-one cutoff. The protocol layer already established a test convention; recommend matching it before more modules build on top.
6. **Prompt-injection surface (forward-looking)** (`src/session-store.ts:132-137`). `buildHistoryPrefix` concatenates raw user/assistant content into a flat `[role]: content` prefix. Harmless as stored text today, but whoever consumes it in `agent-bridge` (T10) must treat it as untrusted — flagging now so the trust boundary is explicit when query() is wired.

## ⚪ Nits (optional)

- **`parseIntStrict` accepts hex/scientific/negative/zero** (`config.ts:106-112`) — `Number('0x10') → 16`, `Number('1e3') → 1000`, `'-1'`/`'0'` pass. Not exploitable in this diff (values are unconsumed), but for `rateLimit.maxPerMinute`, `maxTurns`, `maxContextChars`, `historyLimit` these should ultimately be range-validated (`>= 1`, etc.). Note `historyLimit` eventually reaches SQL `LIMIT ?`, where SQLite treats a negative as "no limit" — worth a lower-bound clamp when this gets consumed.
- **`permissionMode` is an unvalidated `string`** (`config.ts:16`, `132-134`) — a typo passes through silently. Validating against the known enum and throwing on unknown would be good hygiene for a security-sensitive field. (Lower priority since the SDK's unknown-value fallback is the restrictive `'default'`, i.e. it fails safe.)
- **`allowedTools` can gain `Bash`/`Write`/`Edit` via env/config with no warning** (`config.ts:129-131`) — expected design (README documents "full mode"), but a startup log line when write/exec tools are enabled under `bypassPermissions` would help operators notice the posture.
- **`SELECT *` in `selectSession`** (`session-store.ts:86`) — list columns explicitly to match `selectRecentMessages` (harmless today; `rowToSession` maps fields explicitly).
- **`applyEnv`/`mergeConfig` share array references** (`config.ts:95`, `114-121`) — `botBlocklist`/`allowedTools`/`settingSources` alias the parsed-JSON/defaults arrays. No live mutation today; a defensive copy makes `Config` a true value object.
- **Truthy env checks block empty-string overrides** (`config.ts:123-162`) — `CC_OCTO_SDK_SYSTEM_PROMPT=""` is ignored. Fine for tokens/URLs; if intentional-empty matters for `systemPrompt`, use `!== undefined`.
- **`_`-prefix stripping is root-only** (`config.ts:69-73`) — a nested `_comment` under `sdk`/`context` would survive the spread into `Config.sdk`. Harmless; document root-only or strip recursively.

---

## For the human reviewer

The only item I'd insist on before this becomes load-bearing is the **`bypassPermissions` + `settingSources: ['user','project']` default**. It's inert in this diff but ships as the zero-config posture and config example. Please confirm the intended security default for the bot before/with T10. Everything else is quality/maintainability and can land as follow-ups.

lml2468 (commented):
## cc-channel-octo PR #1 Review — c47560d836fc

### Summary

Config loading (T4) and session store (T5) implementation. Clean code structure — `DbAdapter` abstraction, three-level config priority, prepared statements. But there's a P0 security issue in the defaults that blocks merge.

### Findings

**P0 — `settingSources: ['user', 'project']` default is a security hole** (`config.ts:51`)

The default allows Claude Code SDK to load `.claude/settings.json` from the *project* directory. Combined with `permissionMode: 'bypassPermissions'`, a malicious repo can ship a `.claude/settings.json` that adds `Bash` / `Write` / `Edit` to `allowedTools`, silently escalating beyond the safe read-only defaults. This is not hypothetical — it's the exact attack vector that `settingSources` controls.

Fix: default to `['user']` only. Let operators explicitly opt into `['user', 'project']` via config or env when they trust the project source.

**P1 — no env override for `settingSources`** (`config.ts`, `applyEnv()`)

`applyEnv()` handles `CC_OCTO_SDK_ALLOWED_TOOLS` and `CC_OCTO_SDK_PERMISSION_MODE` but not `settingSources`. This breaks the `env > config > defaults` contract documented in the `_comment` and file header. Add `CC_OCTO_SDK_SETTING_SOURCES` handling.

**P1 — missing index on `messages.session_id`** (`session-store.ts`)

`selectRecentMessages` runs `WHERE session_id = ? ORDER BY id DESC LIMIT ?` — without an index on `(session_id, id)` this is a full table scan per history load. Add:
```sql
CREATE INDEX IF NOT EXISTS idx_messages_session_id ON messages(session_id, id);
```

**P2 — `DEFAULT_HISTORY_LIMIT` duplicates config** (`session-store.ts:54`)

`buildHistoryPrefix` defaults to its own `DEFAULT_HISTORY_LIMIT = 40`, but `config.ts` also defaults `historyLimit: 40`. Two separate sources of truth for the same value. The session store should always receive the limit from config — drop the hardcoded default parameter.

### Verdict

CHANGES_REQUESTED — P0 security issue must be fixed before merge. Agree with Jerry and yujiawei's findings on the same vectors.

lml2468 (commented):
## cc-channel-octo PR #1 Re-review — 8719213fcdde (incremental from c47560d836fc)

| 1 | P0 | `settingSources` default includes `project` — injection risk | ✅ Fixed: default changed to `['user']` only |
| 2 | P1 | `applyEnv` missing `CC_OCTO_SDK_SETTING_SOURCES` | ✅ Fixed: handler added |
| 3 | P1 | messages table missing `(session_id, id)` index | ✅ Fixed: `CREATE INDEX IF NOT EXISTS idx_messages_session_id` added |
| 4 | P2 | `DEFAULT_HISTORY_LIMIT` duplicates config | ✅ Fixed: constant removed, `buildHistoryPrefix` now requires explicit `limit` parameter |

### Additional Improvements (beyond my findings)

- `parseIntStrict` hardened: regex `^\d+$` gate + `>= 1` floor (addresses Jerry's feedback)
- `readConfigFile` now catches JSON parse errors with descriptive message instead of raw throw
- `config.example.json` split into `_comment_safe` / `_comment_full` with explicit security warning about `project` in `settingSources`
- Parameter renamed `path` → `configFilePath` to avoid shadowing

### Verdict

APPROVED — all P0/P1/P2 items resolved. Security posture is correct: safe by default, opt-in to dangerous modes with explicit warning.

Jerry-Xin (approved):
The PR is in scope for this repository and the implementation is broadly mergeable; I found no blocking defects, but there are a few follow-up risks worth addressing.

💬 Non-blocking

🟡 Warning — `src/config.ts:63-80`, `src/config.ts:83-102`: `config.json` is parsed with type assertions but no runtime shape validation. Invalid file values such as `"historyLimit": "-1"`, `"allowedTools": "Read"`, or `"rateLimit": { "maxPerMinute": 0 }` can enter the app despite the `Config` type. Env integers are strictly parsed, but file values are trusted. Consider validating/coercing the file config too, especially numeric bounds and array fields.

🟡 Warning — `src/session-store.ts:133-137`: `buildHistoryPrefix()` has no default limit even though the PR description says default 40. It also does not clamp or reject `limit <= 0`; in SQLite, negative limits can remove the cap and return the full message history. A default parameter plus a `>= 1` guard would avoid accidental oversized context.

🔵 Suggestion — `src/session-store.ts:127-130`: message insert and session touch are separate statements. With the current adapter this is probably fine, but wrapping `append()` in a transaction would keep message persistence and `updated_at` updates atomic if the store gains more logic later.

🔵 Suggestion — `src/session-store.ts`: there are no tests covering schema creation, cascade delete, history ordering, expiry cleanup, or init-before-use behavior. Focused unit tests with a temporary SQLite database would give good coverage for a small amount of code.

✅ Highlights

The PR keeps the config and persistence scope clean, uses prepared statements for runtime values, enables foreign keys/WAL/busy timeout in `src/db-adapter.ts`, and changes the example config to a safer default `settingSources` profile.

I could not locally run the verification commands because this checkout has no installed dependencies: `vitest` and `tsc` were not found.

PR #2, opened 2026-06-04
test(session-store): add SessionStore + DbAdapter tests (T6)

lml2468:
### Coverage (33 tests)

**CRUD round-trip (7 tests)**
- getOrCreate creates new / returns existing with touched updatedAt
- appendUser + appendAssistant → buildHistoryPrefix chronological order
- deleteSession with CASCADE, nonexistent session safety
- Multiple sessions are independent

**History window truncation (5 tests)**
- Default 40-message limit with correct most-recent selection
- Custom limit parameter
- Fewer-than-limit returns all messages
- Empty history / nonexistent session returns empty string

**Expired session cleanup (5 tests)**
- 7-day threshold with backdated sessions
- CASCADE deletes messages when session is cleaned
- Multiple expired sessions in batch
- Boundary test (just under 7 days = NOT cleaned)
- cleanExpired returns 0 when nothing to clean

**DbAdapter interface contract (9 tests)**
- exec runs raw SQL
- PreparedStatement: run/get/all
- RunResult.lastInsertRowid
- WAL mode and foreign keys enabled
- close prevents further operations
- transaction wraps atomically

**Edge cases (7 tests)**
- Special characters (CJK, emoji, newlines, quotes)
- Very long messages (100K chars)
- append touches session updatedAt
- Role constraint rejects invalid roles
- FK constraint prevents orphan messages
- AUTOINCREMENT ordering after deletions
- init() is idempotent

### Verification
- `npm test` → 70 tests all passing (37 protocol + 33 session-store)
- `npx tsc --noEmit` → zero errors

Depends on #1 (T4 + T5).

### DIFF
--- a/src/__tests__/session-store.test.ts
+++ b/src/__tests__/session-store.test.ts
@@ -1,0 +1,442 @@
+/**
+ * Session Store tests.
+ *
+ * Coverage:
+ *  - CRUD round-trip (getOrCreate, append, buildHistoryPrefix, deleteSession)
+ *  - History window truncation (40 messages default, custom limit)
+ *  - Expired session cleanup (7-day threshold, CASCADE delete of messages)
+ *  - DbAdapter interface contract (exec, prepare, close, transaction)
+ *  - Edge cases (empty history, concurrent getOrCreate, message ordering)
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { join } from "node:path";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { createAdapter, type DbAdapter } from "../db-adapter.js";
+import { SessionStore, type Session } from "../session-store.js";
+
+// ─── Test Helpers ───────────────────────────────────────────────────────────
+
+let tmpDir: string;
+let adapter: DbAdapter;
+let store: SessionStore;
+
+function setup() {
+  tmpDir = mkdtempSync(join(tmpdir(), "cc-octo-test-"));
+  adapter = createAdapter(join(tmpDir, "test.db"));
+  store = new SessionStore(adapter);
+  store.init();
+}
+
+function teardown() {
+  try {
+    store.close();
+  } catch {
+    /* already closed */
+  }
+  rmSync(tmpDir, { recursive: true, force: true });
+}
+
+// ─── 1. CRUD Round-Trip ─────────────────────────────────────────────────────
+
+describe("SessionStore CRUD", () => {
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it("getOrCreate creates a new session", () => {
+    const session = store.getOrCreate("s1", "ch1", 1);
+    expect(session.id).toBe("s1");
+    expect(session.channelId).toBe("ch1");
+    expect(session.channelType).toBe(1);
+    expect(session.createdAt).toBeGreaterThan(0);
+    expect(session.updatedAt).toBeGreaterThanOrEqual(session.createdAt);
+  });
+
+  it("getOrCreate returns existing session and touches updatedAt", async () => {
+    const s1 = store.getOrCreate("s1", "ch1", 1);
+    // Small delay to ensure updatedAt can differ
+    await new Promise((r) => setTimeout(r, 5));
+    const s2 = store.getOrCreate("s1", "ch1", 1);
+
+    expect(s2.id).toBe("s1");
+    expect(s2.channelId).toBe("ch1");
+    expect(s2.createdAt).toBe(s1.createdAt);
+    expect(s2.updatedAt).toBeGreaterThanOrEqual(s1.updatedAt);
+  });
+
+  it("appendUser + appendAssistant stores messages", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    store.appendUser("s1", "hello");
+    store.appendAssistant("s1", "hi there");
+
+    const history = store.buildHistoryPrefix("s1");
+    expect(history).toContain("[user]: hello");
+    expect(history).toContain("[assistant]: hi there");
+  });
+
+  it("buildHistoryPrefix returns messages in chronological order", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    store.appendUser("s1", "msg1");
+    store.appendAssistant("s1", "msg2");
+    store.appendUser("s1", "msg3");
+
+    const history = store.buildHistoryPrefix("s1");
+    const lines = history.split("\n");
+    expect(lines[0]).toBe("[user]: msg1");
+    expect(lines[1]).toBe("[assistant]: msg2");
+    expect(lines[2]).toBe("[user]: msg3");
+  });
+
+  it("deleteSession removes session and its messages (CASCADE)", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    store.appendUser("s1", "hello");
+    store.appendAssistant("s1", "hi");
+
+    store.deleteSession("s1");
+
+    // History should be empty after deletion
+    const history = store.buildHistoryPrefix("s1");
+    expect(history).toBe("");
+  });
+
+  it("deleteSession on nonexistent session does not throw", () => {
+    expect(() => store.deleteSession("nonexistent")).not.toThrow();
+  });
+
+  it("multiple sessions are independent", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    store.getOrCreate("s2", "ch2", 2);
+
+    store.appendUser("s1", "s1-msg");
+    store.appendUser("s2", "s2-msg");
+
+    const h1 = store.buildHistoryPrefix("s1");
+    const h2 = store.buildHistoryPrefix("s2");
+
+    expect(h1).toBe("[user]: s1-msg");
+    expect(h2).toBe("[user]: s2-msg");
+    expect(h1).not.toContain("s2-msg");
+    expect(h2).not.toContain("s1-msg");
+  });
+});
+
+// ─── 2. History Window Truncation ───────────────────────────────────────────
+
+describe("History window truncation", () => {
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it("default limit is 40 messages", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    for (let i = 0; i < 50; i++) {
+      store.appendUser("s1", `msg-${i}`);
+    }
+
+    const history = store.buildHistoryPrefix("s1");
+    const lines = history.split("\n");
+    expect(lines.length).toBe(40);
+    // Should contain most recent 40 messages (msg-10 through msg-49)
+    expect(lines[0]).toBe("[user]: msg-10");
+    expect(lines[39]).toBe("[user]: msg-49");
+  });
+
+  it("custom limit works", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    for (let i = 0; i < 20; i++) {
+      store.appendUser("s1", `msg-${i}`);
+    }
+
+    const history = store.buildHistoryPrefix("s1", 5);
+    const lines = history.split("\n");
+    expect(lines.length).toBe(5);
+    // Most recent 5: msg-15 through msg-19
+    expect(lines[0]).toBe("[user]: msg-15");
+    expect(lines[4]).toBe("[user]: msg-19");
+  });
+
+  it("returns all messages when fewer than limit", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    store.appendUser("s1", "only-one");
+
+    const history = store.buildHistoryPrefix("s1", 40);
+    const lines = history.split("\n");
+    expect(lines.length).toBe(1);
+    expect(lines[0]).toBe("[user]: only-one");
+  });
+
+  it("empty history returns empty string", () => {
+    store.getOrCreate("s1", "ch1", 1);
+    const history = store.buildHistoryPrefix("s1");
+    expect(history).toBe("");
+  });
+
+  it("history for nonexistent session returns empty string", () => {
+    const history = store.buildHistoryPrefix("nonexistent");
+    expect(history).toBe("");
+  });
+});
+
+// ─── 3. Expired Session Cleanup ─────────────────────────────────────────────
+
+describe("Expired session cleanup", () => {
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it("cleanExpired removes sessions older than 7 days", () => {
+    // Create a session and manually backdate its updated_at
+    store.getOrCreate("old-session", "ch1", 1);
+    store.appendUser("old-session", "old message");
+
+    // Backdate: set updated_at to 8 days ago
+    const eightDaysAgo = Date.now() - 8 * 24 * 60 * 60 * 1000;
+    adapter.prepare("UPDATE sessions SET updated_at = ? WHERE id = ?").run(
+      eightDaysAgo,
+      "old-session",
+    );
+
+    // Create a fresh session
+    store.getOrCreate("fresh-session", "ch2", 1);
+    store.appendUser("fresh-session", "fresh message");
+
+    const cleaned = store.cleanExpired();
+    expect(cleaned).toBe(1);
+
+    // Old session's messages should be gone (CASCADE)
+    const oldHistory = store.buildHistoryPrefix("old-session");
+    expect(oldHistory).toBe("");
+
+    // Fresh session should still exist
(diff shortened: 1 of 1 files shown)


Jerry-Xin (approved):
The PR is relevant to this repository and adds useful coverage for the session persistence and SQLite adapter layers.

**💬 Non-blocking**
- 🟡 Warning: [src/__tests__/session-store.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-iihz9vnu/repo/src/__tests__/session-store.test.ts:56) says `getOrCreate` “touches updatedAt”, but `toBeGreaterThanOrEqual` would still pass if `updatedAt` never changed. Prefer a controlled clock or direct backdated DB row and assert a strict increase.
- 🟡 Warning: [src/__tests__/session-store.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-iihz9vnu/repo/src/__tests__/session-store.test.ts:387) does not actually isolate `append` touching `updatedAt`, because the later `getOrCreate` call also touches the session. Query `sessions.updated_at` directly after `appendUser` instead.
- 🟡 Warning: [src/__tests__/session-store.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-iihz9vnu/repo/src/__tests__/session-store.test.ts:347) tests only successful transaction execution, not atomic rollback. Add a case that throws mid-transaction and asserts prior inserts are rolled back.
- 🔵 Suggestion: [src/__tests__/session-store.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-iihz9vnu/repo/src/__tests__/session-store.test.ts:17) imports `Session` but never uses it, producing a lint warning.

**✅ Highlights**
- Good in-scope coverage for CRUD, history truncation, expiry cleanup, FK/CASCADE behavior, and adapter basics.
- Tests isolate SQLite state with per-test temp directories.
- Verification after `npm ci`: `npm test` passed, `npm run type-check` passed, and `npm run lint` had 0 errors with existing warnings plus the unused `Session` import.

lml2468 (commented):
## cc-channel-octo PR #2 Review — e7304d5740ff

### Summary

442 lines of tests for SessionStore + DbAdapter (T6). Real SQLite via temp dirs — no mocks for the persistence layer, which is exactly right for a DB adapter.

### Assessment

**Good coverage of core paths:**
- CRUD round-trip, history truncation (default 40 + custom limit), expired cleanup with CASCADE verification, DbAdapter contract (WAL, FK, prepare/run/get/all), boundary test at 7-day expiry edge
- Edge cases are well-chosen: special chars, 100K message, AUTOINCREMENT ordering after deletion, FK constraint enforcement, role CHECK constraint, idempotent init

**Minor observations (non-blocking):**

1. `group_members` table is created in the schema but has no SessionStore methods or tests yet — presumably coming in a later task. Fine for now.

2. The `setTimeout(r, 5)` timing in async tests works because the assertions use `toBeGreaterThanOrEqual` (not strict `>`), so no flakiness risk.

3. The "default limit is 40" test hardcodes the value, coupling it to the `DEFAULT_HISTORY_LIMIT` constant. If PR #1 addresses the P2 about removing that constant in favor of config-provided limit, this test should update accordingly — but that's a PR #1 concern.

### Verdict

APPROVED — solid real-DB test suite. No issues found.

yujiawei (approved):
# Code Review — PR #2 (cc-channel-octo)

**Verdict: APPROVED** — solid, well-organized test coverage for `SessionStore` + `DbAdapter`. No blocking issues. A few non-blocking suggestions below to strengthen assertions and fix a doc/lint nit.

## Summary

This test-only PR adds `src/__tests__/session-store.test.ts` (442 lines, 33 tests) covering the existing (unmodified) `session-store.ts` and `db-adapter.ts`. I checked out the head SHA, installed deps, and verified locally:

- ✅ `vitest run` — **33/33 pass** (~0.4s)
- ✅ `tsc --noEmit` — clean (rc=0)
- ⚠️ `eslint` — 1 warning only (unused import, see N1)

Coverage is genuinely good: CRUD round-trip, history-window truncation (default 40 / custom limit), 7-day expiry + CASCADE cleanup, the full `DbAdapter` contract (WAL, FK pragma, prepared-statement run/get/all, transactions), and a strong edge-case set (unicode/newline/100k-char content, role CHECK constraint, FK orphan rejection, AUTOINCREMENT ordering, idempotent `init`). Test isolation via per-test temp dirs + `beforeEach/afterEach` is correct.

## Findings

All findings are non-blocking (P2 / nit).

### P2 — `updatedAt` "touch" tests are tautological (false confidence)

Both tests that claim to verify the touch-on-write invariant pass even when the touch is removed.

- `append touches session updatedAt` (~L393-402): reads the value back through `getOrCreate`, which **itself** touches `updated_at`. So the assertion passes regardless of whether `append`'s own touch ran.
- `getOrCreate returns existing session and touches updatedAt` (~L62-72): `getOrCreate` returns `{ ...existing, updated_at: now }` in-object, so `toBeGreaterThanOrEqual` is satisfied even if the DB write is a no-op.

I confirmed this empirically by mutation testing: removing `this.touchSession.run(...)` from `append()` — and separately from the `getOrCreate` existing-session branch — leaves **all 33 tests green**. A future regression in either touch would slip through silently.

Suggested fix (no added flakiness, a real 5ms delay is already present):
- Read `updated_at` directly via `adapter.prepare('SELECT updated_at FROM sessions WHERE id = ?').get(id)` instead of through a touching `getOrCreate`.
- Assert strictly `>` rather than `>=`.

### nit — Header doc claims "concurrent getOrCreate" coverage that doesn't exist

The file header (L15) lists `concurrent getOrCreate` under Edge cases, but no concurrency test exists (the closest, `multiple sessions are independent`, is isolation, not concurrency). Since `better-sqlite3` is synchronous, true concurrency isn't meaningfully testable here — recommend dropping the phrase so the coverage note matches reality.

### nit — "exactly 7 days" boundary test uses 7 days − 1 second

`session at exactly 7 days is NOT cleaned (boundary)` (~L267-276) backdates to `now - (7d - 1000ms)`, i.e. just under the threshold, not at it. The 1s margin is sensible (it guards against `cleanExpired` recomputing `Date.now()` a few ms later, which would make an exact-equality assertion flaky), so the test is sound — only the name overclaims. Suggest renaming to "just under 7 days". A true equality boundary can't be tested reliably without injecting the clock into `cleanExpired`.

### nit — Unused `Session` import

L17 imports `type Session` but it's never used (the sole eslint warning). `noUnusedLocals` is `false`, so it doesn't break the build, but drop it to keep the lint clean.

### nit — Test dir excluded from `tsc`

`tsconfig.json` excludes `src/__tests__/**/*`, so `tsc --noEmit` never type-checks the test file (vitest's esbuild transform strips types without checking). Low risk and a common deliberate trade-off; just noting that "tsc passes clean" gives no type signal for this file. Optional: a `tsconfig.test.json` if you want type safety on tests.

## Conclusion

Good, well-structured coverage that meaningfully exercises both the store and the adapter contract. None of the above blocks merge. The only one worth addressing soon is the tautological touch assertions (P2), since they advertise a guarantee they don't actually enforce. Approving.

lml2468 (commented):
## cc-channel-octo PR #2 Re-review — 75bb362bfbfa (incremental from e7304d5740ff)

### Incremental Changes (+40 -18)

All review feedback from the previous round addressed:

1. **updatedAt assertions fixed** — replaced `setTimeout(5ms)` + `toBeGreaterThanOrEqual` with backdate-to-60s-ago + direct DB query + strict `toBeGreaterThan`. No more false-positive risk.
2. **Transaction rollback test added** — inserts inside a transaction that throws, then verifies count is 0. Covers the gap Jerry flagged.
3. **Boundary test renamed** — "session at exactly 7 days" → "session just under 7 days". Accurate.
4. **Unused `Session` import removed** — clean.
5. **File header updated** — removed "concurrent getOrCreate" claim.

### Verdict

APPROVED — all previous feedback incorporated cleanly.

Jerry-Xin (approved):
Summary: Project relevance gate passed; these tests directly cover this repository’s `SessionStore` and `DbAdapter` persistence layer, and I found no blocking issues.

💬 Non-blocking
- 🔵 Suggestion: `src/__tests__/session-store.test.ts:263` uses the real clock for the “just under 7 days” boundary with only a 1-second margin. This is likely fine, but fake timers or an injectable clock would make the boundary test fully deterministic under slow CI pauses.
- 🔵 Suggestion: PR description says 33 session-store tests, but the current suite reports 34 tests in `src/__tests__/session-store.test.ts`. Update the PR text to avoid stale coverage metadata.

✅ Highlights
- Good coverage across CRUD, history truncation, cleanup, FK/CHECK constraints, adapter contract, transactions, and long/special message content.
- The tests use isolated temporary database files and validate real SQLite behavior, including WAL and foreign keys.
- Verification passed locally after installing dependencies: `npm test` passed 71 tests, `npm run type-check` passed, and `npm run lint` had only pre-existing warnings in `src/octo/socket.ts`.

lml2468 (commented):
## cc-channel-octo PR #2 Re-review — 7657bdcb91f2 (incremental from 75bb362bfbfa)

### Incremental Changes

Rebase onto updated PR #1 base (`8719213fcdde`). Test content unchanged from previous approval — only difference is all `buildHistoryPrefix` calls now pass explicit `limit` parameter (required after PR #1 removed the default). Config and session-store changes are PR #1's fixes flowing through the rebase.

### Verdict

APPROVED — rebase-only, no new concerns.

Jerry-Xin (approved):
The PR is in scope and adds meaningful SessionStore/DbAdapter coverage; I found no blocking issues.

💬 Non-blocking
- 🟡 Warning: [src/__tests__/session-store.test.ts:240](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-lzqqfg3o/repo/src/__tests__/session-store.test.ts:240) calls `getOrCreate("fresh-1", ...)` to verify the fresh session survived cleanup, but `getOrCreate` would recreate it if it had been deleted. Prefer a direct `SELECT` or give the fresh session a message and assert its history remains.
- 🔵 Suggestion: [src/__tests__/session-store.test.ts:267](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-lzqqfg3o/repo/src/__tests__/session-store.test.ts:267) uses a real-time 1-second boundary. It is probably fine, but fake timers or a wider margin would make the boundary test less sensitive to slow CI pauses.
- 🔵 Suggestion: [src/__tests__/session-store.test.ts:32](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-lzqqfg3o/repo/src/__tests__/session-store.test.ts:32) catches all `close()` errors in teardown. Since one test intentionally closes the adapter, this is understandable, but a narrower helper/state flag would avoid hiding unexpected teardown failures.

✅ Highlights
- Good coverage of SQLite constraints, foreign keys, WAL mode, rollback behavior, CASCADE cleanup, and history ordering/truncation.
- Tests use isolated temp databases per case, which keeps state leakage low.

Verification note: I attempted `npm test`, `npm run type-check`, and `npm run lint`, but this checkout does not have dependencies installed, so `vitest`, `tsc`, and `eslint` were not found.

yujiawei (approved):
# Code Review — PR #2 (cc-channel-octo)

**Scope:** Test-only PR adding `src/__tests__/session-store.test.ts` (464 lines, vitest). Reviewed against the implementations under test (`src/session-store.ts`, `src/db-adapter.ts`) at head SHA `7657bdc`.

## Verdict

**APPROVED.** No P0/P1 issues. Every non-trivial assertion was cross-checked against the actual implementation and matches its behavior; no assertion would fail against the real code. The findings below are coverage gaps and nits — worth a follow-up, but none blocks merge.

## 1. Verification (assertions vs. implementation)

All confirmed correct against `session-store.ts` / `db-adapter.ts`:

- ✅ **Truncation math** — 50 msgs / limit 40 → `ORDER BY id DESC LIMIT 40` takes ids 50..11 (`msg-49`..`msg-10`), reversed to `msg-10`..`msg-49`. 20 / limit 5 → `msg-15`..`msg-19`. Correct.
- ✅ **Chronological ordering** — DESC-then-reverse yields insertion order.
- ✅ **AUTOINCREMENT ordering after mid-delete** — gaps don't reorder; ids 1,3,4 stay ascending.
- ✅ **CASCADE** — `foreign_keys=ON` + `ON DELETE CASCADE` correctly delete messages on both `deleteSession` and `cleanExpired`.
- ✅ **Constraints** — `CHECK(role IN ('user','assistant'))` rejects `'system'`; FK rejects orphan messages.
- ✅ **PRAGMA shapes** — `journal_mode` → `'wal'` (lowercase), `foreign_keys` → `1`.
- ✅ **Transactions** — commit + rollback delegate correctly to `better-sqlite3`'s `db.transaction`.
- ✅ **Special-char / 100k-char round-trips** — full-string compare (not split on `\n`), so the embedded newline is safe.

## 2. Findings

No P0/P1. The following are **P2 coverage gaps** (non-blocking suggestions):

### P2 — `DbAdapter.inTransaction` is never tested
`inTransaction` is a public readonly member of the `DbAdapter` interface (`db-adapter.ts:25`, getter at `64-66`), but the "DbAdapter interface contract" block never references it. The natural test (assert `false` outside, `true` inside the `transaction()` callback) is trivial to add. Low intrinsic risk (one-line passthrough to better-sqlite3), but it's a gap in a PR whose purpose is contract coverage.

### P2 — `getOrCreate` touch path doesn't verify `createdAt` is preserved
The "returns existing session and touches updatedAt" test (lines 62-74) only re-reads `updated_at` and asserts it advanced. It never asserts `created_at` is left unchanged on the second `getOrCreate`, and discards the returned `Session` object. A regression that reset `created_at` on touch would go undetected. Suggest asserting the raw row's `created_at` is unchanged and inspecting the returned `Session`'s `createdAt`/`updatedAt`.

### P2 — `cleanExpired` boundary tested only on one side; strict `<` vs `<=` not pinned
`cleanExpired` uses `DELETE ... WHERE updated_at < cutoff`. The boundary test (lines 269-278) backdates by "7 days minus 1 second" (i.e. `updated_at = cutoff + 1000ms`) and asserts not-cleaned — but this passes identically under both `<` and `<=`. There's no test at the exact cutoff (`updated_at == cutoff` must NOT be cleaned). Flipping the operator to `<=` would break no test. A two-sided boundary block would close this.

### P2 — No public-API test that `appendUser` on a nonexistent session is rejected
The FK-orphan test (lines 434-439) inserts via raw SQL, proving the schema constraint but not the public path. The realistic misuse is `store.appendUser("nonexistent", ...)`, which routes through `append()` with no existence check and relies on the FK. A public-API test here is more meaningful than the raw-SQL one.

## 3. Nits (optional)

- **Boundary test 1s wall-clock margin** (lines 269-278): `expect(cleaned).toBe(0)` holds only if `<1s` elapses between the test's `Date.now()` and `cleanExpired`'s own `Date.now()`. Practically safe (gap is sub-ms; both statements are pre-prepared), but a GC pause or starved CI runner could theoretically flip it. Widening the backdate margin (e.g. `now - (7d - 60_000)`) or injecting a fixed clock removes even the theoretical risk.
- **Tautological assertion** (lines 246-248): `getOrCreate("fresh-1", ...)` then `expect(s.id).toBe("fresh-1")` recreates the row if absent, so it passes even if `fresh-1` were wrongly deleted. The `cleaned === 5` assert (line 243) carries the real signal; the id check adds nothing. Query the row / `buildHistoryPrefix` to actually prove survival.
- **Redundant CASCADE check** (lines 250-267): CASCADE-on-`cleanExpired` is already observed via the public API at lines 213-219 (`buildHistoryPrefix` returns `""`); the raw `SELECT COUNT(*)` re-verifies the same effect and couples to internal table/column names. The role/FK raw-SQL tests, by contrast, are justified (unreachable via the public API).
- **Module-level shared globals** (`tmpDir`/`adapter`/`store`): standard vitest pattern and safe under file-level parallelism, but would break if any block is ever converted to `.concurrent`. Scoping handles per-test would future-proof it.
- **WAL sidecar cleanup** relies entirely on `rmSync(tmpDir)` + per-test `mkdtemp`; fine today, but would silently pollute if a shared DB path were ever introduced.

## Summary

Solid, well-organized test suite with accurate assertions and good edge-case coverage (special chars, long messages, idempotent `init`, transaction rollback). The gaps above are incremental hardening, not blockers. Approving.

PR #3, opened 2026-06-04
feat: SessionRouter + GroupContext (T8 + T9)

lml2468:
### T8: SessionRouter (`src/session-router.ts`)
- Per-session-key serial promise chain (`inboundQueues`) — same key serial, different keys concurrent
- Session key: DM → `from_uid`; Group/CommunityTopic → `channel_id:from_uid`
- Message filters: self (`robotId`), DM blocklist, group blocklist
- Mention gate (group only): `mention.uids` includes robotId, OR `mention.all`, OR `mention.ais`
- Non-text messages → reply "暂不支持此类消息，请发送文字", returns `shouldProcess: false`
- Token-bucket rate limiter (per session key), configurable `maxPerMinute`; rejection replies "请稍后再试"
- All API errors caught and logged, never thrown

### T9: GroupContext (`src/group-context.ts`)
- In-memory sliding window per channel (cap 100), shifts from front on overflow
- `learnMember` short-circuits when name unchanged; writes to `group_members` table via upsert
- `refreshMembers` throttled to 1/hour; tolerant of API failures
- `buildContext` walks newest→oldest accumulating lines until `maxContextChars` budget exhausted, then reverses to chronological order
- `resolveMentions` parses `@name` with progressive suffix-strip fallback for trailing punctuation; dedupes UIDs
- `loadMembersFromDb` populates in-memory maps from SQLite on startup
- `fetchAndLearnUser` for on-demand single-uid lookup

## Verification
- `npx tsc --noEmit` — zero errors ✅
- `npm test` — 37/37 passing ✅
- Only `src/session-router.ts` and `src/group-context.ts` changed (2 files, +321 lines)

### DIFF
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -1,0 +1,193 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { GroupContext } from '../group-context.js';
+import { createAdapter } from '../db-adapter.js';
+import type { DbAdapter } from '../db-adapter.js';
+
+// Mock the Octo API
+vi.mock('../octo/api.js', () => ({
+  getGroupMembers: vi.fn().mockResolvedValue([]),
+  fetchUserInfo: vi.fn().mockResolvedValue(null),
+}));
+
+import { getGroupMembers, fetchUserInfo } from '../octo/api.js';
+
+function createTestAdapter(): DbAdapter {
+  const adapter = createAdapter(':memory:');
+  adapter.exec(`
+    CREATE TABLE IF NOT EXISTS group_members (
+      group_id TEXT NOT NULL,
+      uid TEXT NOT NULL,
+      name TEXT NOT NULL,
+      updated_at INTEGER NOT NULL,
+      PRIMARY KEY(group_id, uid)
+    );
+  `);
+  return adapter;
+}
+
+describe('GroupContext', () => {
+  let adapter: DbAdapter;
+  let ctx: GroupContext;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    adapter = createTestAdapter();
+    ctx = new GroupContext(adapter, 6000);
+  });
+
+  // --- pushMessage + cache ---
+
+  it('pushMessage adds to cache and learns member', () => {
+    ctx.pushMessage('ch1', 'u1', 'Alice', 'hello', Date.now());
+    expect(ctx.getName('u1', 'ch1')).toBe('Alice');
+  });
+
+  it('cache window respects maxWindowSize', () => {
+    for (let i = 0; i < 110; i++) {
+      ctx.pushMessage('ch1', 'u1', 'Alice', `msg-${i}`, Date.now());
+    }
+    const context = ctx.buildContext('ch1');
+    // Should have at most 100 messages (maxWindowSize default)
+    const lines = context.split('\n').filter((l) => l.startsWith('Alice'));
+    expect(lines.length).toBeLessThanOrEqual(100);
+  });
+
+  // --- buildContext ---
+
+  it('buildContext respects maxContextChars budget', () => {
+    const smallCtx = new GroupContext(adapter, 50);
+    for (let i = 0; i < 10; i++) {
+      smallCtx.pushMessage('ch1', 'u1', 'A', `message number ${i}`, Date.now());
+    }
+    const output = smallCtx.buildContext('ch1');
+    // Total length should be within budget + header/trailer
+    expect(output.length).toBeLessThanOrEqual(50);
+  });
+
+  it('buildContext returns empty string for empty cache', () => {
+    expect(ctx.buildContext('ch1')).toBe('');
+  });
+
+  it('buildContext formats correctly', () => {
+    ctx.pushMessage('ch1', 'u1', 'Alice', 'hello', 1000);
+    ctx.pushMessage('ch1', 'u2', 'Bob', 'world', 2000);
+    const output = ctx.buildContext('ch1');
+    expect(output).toContain('[Recent group messages]');
+    expect(output).toContain('Alice：hello');
+    expect(output).toContain('Bob：world');
+    // Alice should come before Bob (chronological)
+    expect(output.indexOf('Alice')).toBeLessThan(output.indexOf('Bob'));
+  });
+
+  // --- Per-channel isolation ---
+
+  it('memberMap is per-channel — same name different channels', () => {
+    ctx.learnMember('ch1', 'uid-a', 'Alice');
+    ctx.learnMember('ch2', 'uid-b', 'Alice');
+    // Each channel maps "Alice" to a different uid
+    expect(ctx.getName('uid-a', 'ch1')).toBe('Alice');
+    expect(ctx.getName('uid-b', 'ch2')).toBe('Alice');
+    // getName should NOT find uid-b in ch1
+    expect(ctx.getName('uid-b', 'ch1')).toBeUndefined();
+  });
+
+  // --- learnMember rename ---
+
+  it('learnMember removes old nameToUid entry on rename', () => {
+    ctx.learnMember('ch1', 'u1', 'OldName');
+    expect(ctx.resolveMentions('@OldName hello', 'ch1')).toEqual(['u1']);
+
+    ctx.learnMember('ch1', 'u1', 'NewName');
+    // Old name should no longer resolve
+    expect(ctx.resolveMentions('@OldName hello', 'ch1')).toEqual([]);
+    // New name should resolve
+    expect(ctx.resolveMentions('@NewName hello', 'ch1')).toEqual(['u1']);
+  });
+
+  // --- resolveMentions ---
+
+  it('resolveMentions strips trailing punctuation', () => {
+    ctx.learnMember('ch1', 'u1', 'Alice');
+    expect(ctx.resolveMentions('@Alice，你好', 'ch1')).toEqual(['u1']);
+    expect(ctx.resolveMentions('@Alice!', 'ch1')).toEqual(['u1']);
+    expect(ctx.resolveMentions('@Alice。', 'ch1')).toEqual(['u1']);
+  });
+
+  it('resolveMentions does NOT do progressive prefix matching', () => {
+    ctx.learnMember('ch1', 'u1', 'A');
+    // @Alice should NOT match "A" via prefix
+    expect(ctx.resolveMentions('@Alice', 'ch1')).toEqual([]);
+  });
+
+  it('resolveMentions per-channel isolation', () => {
+    ctx.learnMember('ch1', 'uid-1', 'Alice');
+    ctx.learnMember('ch2', 'uid-2', 'Alice');
+    // ch1 resolves to uid-1
+    expect(ctx.resolveMentions('@Alice', 'ch1')).toEqual(['uid-1']);
+    // ch2 resolves to uid-2
+    expect(ctx.resolveMentions('@Alice', 'ch2')).toEqual(['uid-2']);
+  });
+
+  it('resolveMentions returns empty for unknown names', () => {
+    expect(ctx.resolveMentions('@Nobody', 'ch1')).toEqual([]);
+  });
+
+  it('resolveMentions deduplicates UIDs', () => {
+    ctx.learnMember('ch1', 'u1', 'Alice');
+    expect(ctx.resolveMentions('@Alice and @Alice again', 'ch1')).toEqual(['u1']);
+  });
+
+  // --- refreshMembers ---
+
+  it('refreshMembers throttles within 1h', async () => {
+    vi.mocked(getGroupMembers).mockResolvedValue([{ uid: 'u1', name: 'Alice', role: 0 }]);
+    await ctx.refreshMembers('ch1', 'http://api', 'token');
+    expect(getGroupMembers).toHaveBeenCalledTimes(1);
+
+    // Second call within 1h should be skipped
+    await ctx.refreshMembers('ch1', 'http://api', 'token');
+    expect(getGroupMembers).toHaveBeenCalledTimes(1);
+  });
+
+  it('refreshMembers does not record lastRefresh on failure', async () => {
+    vi.mocked(getGroupMembers).mockRejectedValueOnce(new Error('network'));
+    await ctx.refreshMembers('ch1', 'http://api', 'token');
+
+    // Should retry immediately since lastRefresh was not set
+    vi.mocked(getGroupMembers).mockResolvedValue([{ uid: 'u1', name: 'Alice', role: 0 }]);
+    await ctx.refreshMembers('ch1', 'http://api', 'token');
+    expect(getGroupMembers).toHaveBeenCalledTimes(2);
+    expect(ctx.getName('u1', 'ch1')).toBe('Alice');
+  });
+
+  // --- loadMembersFromDb ---
+
+  it('loadMembersFromDb populates in-memory maps', () => {
+    // Write directly to DB
+    adapter.prepare(
+      'INSERT INTO group_members (group_id, uid, name, updated_at) VALUES (?, ?, ?, ?)',
+    ).run('ch1', 'u1', 'Alice', Date.now());
+    adapter.prepare(
+      'INSERT INTO group_members (group_id, uid, name, updated_at) VALUES (?, ?, ?, ?)',
+    ).run('ch1', 'u2', 'Bob', Date.now());
+
+    ctx.loadMembersFromDb('ch1');
+    expect(ctx.getName('u1', 'ch1')).toBe('Alice');
+    expect(ctx.getName('u2', 'ch1')).toBe('Bob');
+    expect(ctx.resolveMentions('@Alice @Bob', 'ch1')).toEqual(['u1', 'u2']);
+  });
+
+  // --- fetchAndLearnUser ---
+
+  it('fetchAndLearnUser caches after first lookup', async () => {
+    vi.mocked(fetchUserInfo).mockResolvedValue({ uid: 'u1', name: 'Alice' });
+    const name1 = await ctx.fetchAndLearnUser('u1', 'ch1', 'http://api', 'token');
+    expect(name1).toBe('Alice');
+    expect(fetchUserInfo).toHaveBeenCalledTimes(1);
+
+    // Second call should return cached, no API call
+    const name2 = await ctx.fetchAndLearnUser('u1', 'ch1', 'http://api', 'token');
+    expect(name2).toBe('Alice');
+    expect(fetchUserInfo).toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -104,6 +104,31 @@ describe('GroupContext', () => {
     expect(ctx.resolveMentions('@NewName hello', 'ch1')).toEqual(['u1']);
   });
 
+  it('rename does not clobber another user with the same old display name', () => {
+    // Both users start with different names
+    ctx.learnMember('ch1', 'u1', 'Alice');
+    ctx.learnMember('ch1', 'u2', 'Alice'); // u2 takes over 'Alice'
+
+    // Now u2 renames — should NOT delete 'Alice' → u2 if still valid,
+    // but since u2 owns 'Alice', it should be cleaned up
+    ctx.learnMember('ch1', 'u2', 'Bob');
+    // 'Alice' is no longer mapped to anyone (u1 was overwritten by u2,
+    // u2 renamed to Bob)
+    // But u1's memberMap entry still says 'Alice'
+    expect(ctx.getName('u1', 'ch1')).toBe('Alice');
+    expect(ctx.getName('u2', 'ch1')).toBe('Bob');
+  });
+
+  it('duplicate display name: rename does not delete mapping owned by other uid', () => {
+    ctx.learnMember('ch1', 'u1', 'SharedName');
+    ctx.learnMember('ch1', 'u2', 'SharedName'); // u2 takes over reverse mapping
+    // Now u1 renames — should NOT delete 'SharedName' because it points to u2
+    ctx.learnMember('ch1', 'u1', 'UniqueName');
+    // SharedName should still resolve to u2
+    expect(ctx.resolveMentions('@SharedName', 'ch1')).toEqual(['u2']);
+    expect(ctx.resolveMentions('@UniqueName', 'ch1')).toEqual(['u1']);
+  });
+
   // --- resolveMentions ---
 
   it('resolveMentions strips trailing punctuation', () => {
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -189,9 +189,48 @@ describe('SessionRouter', () => {
     await router.route(makeMsg({ message_id: '1', channel_type: ChannelType.DM }));
     await router.route(makeMsg({ message_id: '2', channel_type: ChannelType.DM }));
 
-    // Should be rate limited
+    // Should be rate limited — first rejection sends notification
     const result = await router.route(makeMsg({ message_id: '3', channel_type: ChannelType.DM }));
     expect(result!.shouldProcess).toBe(false);
+    expect(sendMessage).toHaveBeenCalledWith(
+      expect.objectContaining({ content: '请稍后再试' }),
+    );
+  });
+
+  it('rate limit debounce: only notifies once per window', async () => {
+    const cfg = makeConfig({ rateLimit: { maxPerMinute: 1 } });
+    router = new SessionRouter(cfg, ROBOT_ID);
+
+    // Consume the single token
+    await router.route(makeMsg({ message_id: '1', channel_type: ChannelType.DM }));
+    vi.mocked(sendMessage).mockClear();
+
+    // First rejection — notified
+    await router.route(makeMsg({ message_id: '2', channel_type: ChannelType.DM }));
+    expect(sendMessage).toHaveBeenCalledTimes(1);
+
+    // Subsequent rejections — debounced, no additional notification
+    await router.route(makeMsg({ message_id: '3', channel_type: ChannelType.DM }));
+    await router.route(makeMsg({ message_id: '4', channel_type: ChannelType.DM }));
+    expect(sendMessage).toHaveBeenCalledTimes(1);
+  });
+
+  it('rate limit applies to non-text messages too', async () => {
+    const cfg = makeConfig({ rateLimit: { maxPerMinute: 1 } });
+    router = new SessionRouter(cfg, ROBOT_ID);
+
+    // Consume the single token with a text message
+    await router.route(makeMsg({ message_id: '1', channel_type: ChannelType.DM }));
+
+    // Non-text message should be rate limited, not replied with "暂不支持"
+    vi.mocked(sendMessage).mockClear();
+    const result = await router.route(makeMsg({
+      message_id: '2',
+      channel_type: ChannelType.DM,
+      payload: { type: MessageType.Image },
+    }));
+    expect(result!.shouldProcess).toBe(false);
+    // Should get rate limit notification, not the non-text notice
     expect(sendMessage).toHaveBeenCalledWith(
       expect.objectContaining({ content: '请稍后再试' }),
     );
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -1,0 +1,219 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// Mock the Octo API before importing SessionRouter
+vi.mock('../octo/api.js', () => ({
+  sendMessage: vi.fn().mockResolvedValue(undefined),
+}));
+
+import { SessionRouter } from '../session-router.js';
+import type { BotMessage } from '../octo/types.js';
+import { ChannelType, MessageType } from '../octo/types.js';
+import type { Config } from '../config.js';
+import { sendMessage } from '../octo/api.js';
+
+const ROBOT_ID = 'bot-001';
+
+function makeConfig(overrides?: Partial<Config>): Config {
+  return {
+    botToken: 'test-token',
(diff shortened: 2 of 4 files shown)


lml2468 (commented):
## cc-channel-octo PR #3 Review — fde892e5ca61

### Summary

SessionRouter (T8): routing, per-session promise-chain serialization, mention gate, token bucket rate limiting, blocklist.
GroupContext (T9): in-memory message ring buffer per channel, SQLite-persisted member map, char-budget context builder, `@name` → uid resolver with trailing-punctuation tolerance.

Clean separation — no coupling between the two modules; the orchestrator wires them.

### Findings

**P1 — `tokenBuckets` unbounded memory growth** (`session-router.ts`)

The `tokenBuckets` map grows with every unique session key and is never pruned. For a long-running bot across many users/channels, this leaks. Add periodic cleanup (e.g. in a maintenance timer alongside `SessionStore.cleanExpired()`) or use an LRU with a size cap.

**P1 — Non-text messages trigger a reply for event messages** (`session-router.ts:106`)

`msg.payload.type !== MessageType.Text` catches ALL non-text messages including system events (group join/leave, channel updates via `payload.event`). The bot would reply "暂不支持此类消息" to invisible server events. Event-type messages should be silently dropped, not replied to. Suggestion: add an early return for event payloads before the text-type check.

**P2 — `nameToUid` collision on duplicate display names** (`group-context.ts`)

Two users with the same display name → last writer wins. `resolveMentions("@Alice")` would resolve to the wrong uid. Edge case in practice, but worth a comment acknowledging the limitation.

**P2 — `buildContext` uses fullwidth colon `：` as separator** (`group-context.ts:135`)

`${m.fromName}：${m.content}` uses a Chinese colon. If this context is fed to the SDK as a system prompt prefix, the LLM may interpret the formatting inconsistently. Consider using ASCII `:` for parsing robustness, or document the choice.

### Non-blocking observations

- Session key for groups is `channel_id:from_uid` — per-user isolation is the right default for Claude Code sessions (no cross-user context bleed).
- Promise-chain concurrency control is simple and correct — no lock contention, no deadlock risk.
- `resolveMentions` prefix fallback is a nice touch for handling `@Alice,` or `@Alice.` patterns.
- `refreshMembers` 1-hour interval is reasonable. No persist for `lastRefresh` means warm-up after restart — acceptable.

### Verdict

CHANGES_REQUESTED — P1 items should be addressed: tokenBuckets leak and event-message reply.

Jerry-Xin (changes requested):
The PR is in scope for this repository, but two routing/context correctness bugs need to be fixed before merge.

**🔴 Blocking**
- 🔴 Critical: `SessionRouter` treats `mention.all` as a bot mention in group chats (`src/session-router.ts:67-73`). The existing type contract says `all` is a humans-only legacy signal and should not wake bot adapters (`src/octo/types.ts:67-73`). This will make the bot respond to broad human mentions even when it was not directly addressed. Gate on explicit `uids` or `ais`, and only include `all` if the repository contract is intentionally changed everywhere.

- 🔴 Critical: `GroupContext` stores member identity and mention resolution globally instead of per channel (`src/group-context.ts:24-25`, `src/group-context.ts:70-79`, `src/group-context.ts:147-159`, `src/group-context.ts:173-180`). Two groups with the same display name can resolve to the wrong UID, and `learnMember()` can skip the DB upsert for a second group when the same UID/name was already seen elsewhere. Since the DB schema is keyed by `(group_id, uid)`, the in-memory maps and `resolveMentions` should also be channel-scoped.

**💬 Non-blocking**
- 🟡 Warning: `refreshMembers()` records the throttle timestamp before the API call succeeds (`src/group-context.ts:84-103`). A transient failure prevents retry for an hour. Consider updating `lastRefresh` only after success, or keeping a shorter failure backoff.

- 🟡 Warning: There are no tests for the new modules; the only current test file is protocol-related. Please add coverage for mention gating, same-key serialization, rate limiting, per-channel member resolution, stale rename behavior, and DB load/upsert behavior.

- 🔵 Suggestion: `tokenBuckets`, `messageCache`, `lastRefresh`, and member maps grow for every observed session/channel (`src/session-router.ts:24-25`, `src/group-context.ts:23-26`). For a long-running gateway, add TTL cleanup or prune on session expiration.

**✅ Highlights**
- Same-session promise chaining is simple and should preserve ordering while allowing different keys to run concurrently.
- API reply failures are contained in `replySafe`, avoiding router crashes from notification failures.

yujiawei (changes requested):
# Code Review — PR #3 (cc-channel-octo)

**Scope:** `src/session-router.ts` (T8) + `src/group-context.ts` (T9), +321/-4 over 2 files. Reviewed at head `fde892e`.

**Verdict: CHANGES_REQUESTED.** The build is clean (`tsc --noEmit` zero errors) and 37/37 tests pass, the per-session-key serial queue is correct, and the overall structure is solid. However, the mention gate and mention-resolution logic — the security-sensitive core of this PR — contain confirmed correctness bugs that contradict this repo's own documented invariants. These should be fixed before merge so they don't get wired into the gateway (T7) as-is.

> Note: both classes are currently *dead code* (not yet instantiated; `gateway.ts`/`index.ts` are Phase-2 stubs), so there is no live production impact today. But T8/T9 are being reviewed precisely so these are correct before integration — that's why the logic bugs below are treated as blocking rather than deferred.

---

## P1 — Blocking

### P1-1 `isMentioned()` triggers on `mention.all`, but `all` means *humans-only* (`@所有人`), not AIs
`src/session-router.ts:67-73`
```ts
private isMentioned(msg: BotMessage): boolean {
  const mention = msg.payload.mention;
  if (!mention) return false;
  if (mention.uids?.includes(this.robotId)) return true;
  if (mention.all) return true;   // ← bug
  if (mention.ais) return true;
  return false;
}
```
This repo's own type contract is explicit (`src/octo/types.ts:66-83`): the server double-writes legacy `all=1` as a **humans-only** signal (`@所有人`), and the AI-targeting flag is `ais=1` (`@所有AI`). The adapter is documented as *not* deciding semantics — the server is authoritative and rewrites `all` into canonical form. By treating `mention.all` as a trigger, an AI bot will respond to `@所有人` (a human-only broadcast), which is exactly the opposite of the documented intent.

**Fix:** delete `if (mention.all) return true;`. The `ais` check on the next line already handles legitimate `@所有AI` targeting. (If there is a real product requirement to also answer `@所有人`, document it explicitly and reconcile it with `types.ts` — but as written this silently violates the stated invariant.)

### P1-2 `resolveMentions()` prefix-stripping mis-resolves `@name` to the wrong user
`src/group-context.ts:154-167`
```ts
let uid = this.nameToUid.get(name);
if (!uid) {
  // Try progressively shorter prefixes — handles trailing punctuation.
  for (let end = name.length - 1; end > 0 && !uid; end--) {
    uid = this.nameToUid.get(name.slice(0, end));
  }
}
```
The progressive prefix-strip is too aggressive: it doesn't stop at punctuation, it strips *any* trailing characters. If `张` is a registered member and someone writes `@张三` (a different person, not registered), this resolves `@张三` to `张`'s uid — a confident mis-resolution to the wrong user. In a group bot, an `@`-target driving downstream behavior toward the wrong person is a real correctness/trust issue.

**Fix:** only strip a bounded set of trailing punctuation (the stated intent), e.g. match `@(\S+?)([punct]*)$` or `name.replace(/[,.!?;:，。！？；：、)\]]+$/,'')`, rather than looping over every shorter prefix. Prefer exact match; never fall back to an arbitrary shorter name.

### P1-3 `GroupContext` member maps are global, not per-channel → cross-group identity contamination
`src/group-context.ts:29-47, 154`

`GroupContext` is constructed as `new GroupContext(adapter, maxContextChars)` with no `channelId`, and `memberMap` / `nameToUid` are single instance-wide maps. Every method takes `channelId` as an argument (implying one shared instance for all channels), but `resolveMentions(text)` takes **no** `channelId` and resolves names against the global `nameToUid`. Display-name collisions across groups are common, so an `@name` in group A can resolve to a uid that only exists in group B. The DB layer is correctly keyed by `(group_id, uid)` (`session-store.ts:46`), but the in-memory index throws that scoping away.

**Fix (pick one):** key the in-memory maps per channel (`Map<channelId, Map<...>>`), thread `channelId` through `resolveMentions`/`getName`, or instantiate one `GroupContext` per channel. Please decide this *before* T7 wires it up, since the chosen shape affects the gateway integration.

### P1-4 New routing/mention logic ships with zero unit tests
The 321 new lines (mention gate, rate limiter, serial queue, mention resolution) have **no dedicated tests** — the only test file is the pre-existing `src/__tests__/octo/protocol.test.ts`. For `security_sensitive`-classified routing logic, the gate (`isMentioned`, blocklist, DM/group branching), the token-bucket refill math, and `resolveMentions` should each have table-driven tests. The bugs above (P1-1/P1-2) are precisely the kind a few unit tests would have caught.

**Fix:** add unit tests for `SessionRouter.route()` (self-skip, DM/group blocklist, mention gate including the `all` vs `ais` distinction, non-text reply, rate-limit reject) and `GroupContext` (`buildContext` budget/order, `resolveMentions` exact + punctuation + no-false-positive, window cap).

---

## P2 — Non-blocking (please address with the above)

- **Stale `nameToUid` key on rename** (`group-context.ts:77-89`): `learnMember` writes `nameToUid[newName]` on a name change but never deletes the old name key, so the previous display name keeps resolving forever. Delete the old mapping (when it still points at this uid) before re-inserting.
- **Unbounded in-memory growth** (`group-context.ts`: `memberMap`, `nameToUid`, `lastRefresh`; `session-router.ts`: `tokenBuckets`): these grow without eviction for a long-lived process. `messageCache` (capped at 100/channel) and `inboundQueues` (cleaned up in the `finally`) are fine; the others are not. Consider periodic pruning or an LRU bound, especially for `tokenBuckets` keyed by `channelId:from_uid`.

---

## Things verified as correct ✅

- **Serial queue** (`session-router.ts:32-50`): same-key serial / different-key concurrent; the gate promise only ever resolves (never rejects), so a thrown `processMessage` cannot wedge the chain; the `inboundQueues.get(key) === gate` guard correctly avoids deleting a newer tail. Clean.
- **Token-bucket refill** math (`checkRateLimit`) is sound (lazy, clamped to max).
- **`replySafe` / API error handling**: all outbound calls are wrapped; errors are logged, never thrown — matches the PR description.
- **`buildContext`**: newest→oldest accumulation under `maxContextChars`, then reversed to chronological — correct, and budget underflow is guarded.
- **int64 `message_id`** protection and `client_msg_no` idempotency in `octo/api.ts` are preserved.

---

## For human reviewer (security-sensitive)

Please manually confirm the intended product semantics of `@所有人` (`all`) vs `@所有AI` (`ais`) for this bot — P1-1 is a clear contradiction with `types.ts` and is the single most important item to settle, since it governs when the bot speaks unsolicited in group chats.

lml2468 (commented):
## cc-channel-octo PR #3 Re-review — ca8b09613f1e (incremental from fde892e5ca61)

| 1 | P1 | `tokenBuckets` unbounded growth | ✅ Fixed: `cleanStaleBuckets()` removes entries idle >5min |
| 2 | P1 | Non-text messages reply to system events | ✅ Fixed: `if (msg.payload.event) return null` before text-type check |
| 3 | P2 | `nameToUid` collision across channels | ✅ Fixed: per-channel maps (`memberMapByChannel`, `nameToUidByChannel`) |
| 4 | P2 | Fullwidth colon separator | Kept as-is (non-blocking) |

### Fixes from other reviewers also incorporated

- `mention.all` no longer triggers bot response — only `mention.uids` and `mention.ais` (Jerry/Steve/yujiawei P0)
- `resolveMentions` takes `channelId` — per-channel isolation (yujiawei)
- Progressive prefix matching replaced with proper regex + trailing punctuation strip (yujiawei)
- `refreshMembers` only records `lastRefresh` on success, allowing retry on failure (Jerry)
- `learnMember` deletes old reverse name mapping on rename — prevents stale name→uid entries
- `fetchAndLearnUser` now requires `channelId` for proper channel scoping

### New tests (+412 lines)

Solid coverage for both SessionRouter and GroupContext:
- `mention.all` rejection explicitly tested ✓
- Per-channel member isolation ✓
- Rename old-name cleanup ✓
- Progressive prefix matching removed (test confirms `@Alice` does NOT match member "A") ✓
- System event silent drop ✓
- Rate limit enforcement + stale bucket cleanup ✓
- Serial queue ordering ✓
- `refreshMembers` failure retry ✓

### Verdict

APPROVED — all P1 issues resolved, tests validate the fixes. Significant quality improvement.

Jerry-Xin (changes requested):
Project relevance gate passes: this PR implements in-scope session routing and group-context functionality for `cc-channel-octo`, but one core concurrency behavior is not actually enforced.

**🔴 Blocking**
- 🔴 Critical — `SessionRouter` releases the per-session queue before downstream message handling can run. In `src/session-router.ts:43-50`, the queue only wraps `processMessage()`, which performs filtering, replies, and rate limiting, then returns `RouteResult`. Any caller that receives `shouldProcess: true` will run the actual agent/session work after `route()` has already resolved, so a second message with the same session key can start downstream processing concurrently. This breaks the PR’s stated guarantee: “same key serial, different keys concurrent.” The router needs to wrap the full async handling path, for example by accepting an async handler/callback under the per-key lock, or by returning an explicit lock/release primitive that integration code must hold until processing completes.

**💬 Non-blocking**
- 🟡 Warning — `GroupContext` cannot safely handle duplicate display names in one group. `nameToUidByChannel` stores only one UID per name, and rename handling deletes the old name unconditionally at `src/group-context.ts:94-100` and `src/group-context.ts:122-127`. If two users are named `Alice`, then one user renames away from `Alice`, the remaining `Alice` mapping can be removed. Consider `Map<string, Set<string>>`, or at least delete only when the reverse mapping still points to the UID being renamed.
- 🟡 Warning — non-text messages bypass the rate limiter because the non-text reply happens before `checkRateLimit()` at `src/session-router.ts:100-108`. A user can spam images/files and force unlimited outbound “unsupported message” replies. If the limiter is intended to protect bot/API usage per session, apply it before user-facing rejection replies too.
- 🔵 Suggestion — `resolveMentions()` only supports a narrow token regex at `src/group-context.ts:191-199`, so names containing spaces, hyphens, dots, or other valid display-name characters will not resolve. If Octo display names are unrestricted, prefer matching against known member names or using payload mention entities where available.

**✅ Highlights**
- The PR is well-scoped and relevant to the repository.
- API reply failures are caught in the router, and group member refresh failures are tolerated.
- Added tests cover the main happy paths and several edge cases.
- Verification after installing dependencies: `npm test -- --run` passed 68 tests, and `npx tsc --noEmit` passed.

yujiawei (changes requested):
# Code Review — PR #3 (cc-channel-octo)

Reviewed `SessionRouter` (T8) and `GroupContext` (T9) at head `ca8b0961`. Cross-checked against `octo/types.ts`, `octo/api.ts`, `config.ts`, `session-store.ts`, and `db-adapter.ts`. Build and tests reproduce clean locally (`tsc --noEmit` zero errors, `vitest run` 68/68 green — note this exceeds the PR description's "37/37", the suite also includes the existing protocol tests).

This PR is security-relevant (rate limiting + mention gating). Overall the structure is solid — per-channel member isolation, fail-soft API error handling, and the humans-vs-AI mention split are all done correctly. I'm requesting changes on one defect in the rate-limiting control plus a correctness bug in mention resolution; the rest are P2/nits for follow-up.

## Verification

- ✅ Session key: DM → `from_uid`; group → `channel_id:from_uid` (`session-router.ts:54-59`)
- ✅ Filters: self-skip, DM/group blocklist, system-event skip (`session-router.ts:80-98`)
- ✅ Mention gate correctly treats `mention.all` as humans-only and only responds to `uids` containing the bot or `ais` (`session-router.ts:69-76`) — matches the server semantics documented in `types.ts:64-83`
- ✅ Per-channel member maps avoid cross-group name collisions (`group-context.ts:25-26, 52-68`)
- ✅ `refreshMembers` records `lastRefresh` only on success, allowing retry after failure (`group-context.ts:113-137`)
- ✅ DB schema for `group_members` matches between `session-store.ts:46-52` and the prepared statements in `group-context.ts:44-49` (no cross-file drift)
- ⚠️ Rate-limiting control has outbound-reply gaps (P1, below)
- ⚠️ Mention reverse-map corrupts on duplicate display names (P2, below)

## P1 — Rate limiter does not bound its own outbound replies

`processMessage` (`session-router.ts:78-113`) sends notice replies on two paths that escape the token bucket:

1. **Non-text reply runs before the rate-limit check.** The non-text branch (`:101-104`) calls `replySafe('暂不支持此类消息…')` and returns at `:103`, so it never reaches `checkRateLimit` (`:107`). A flood of non-text messages produces one outbound `sendMessage` per inbound, completely un-throttled.
2. **The over-limit notice has no cooldown.** When `checkRateLimit` returns false, `:107-108` unconditionally calls `replySafe('请稍后再试')`. `checkRateLimit` (`:115-133`) returns false at `:130` with no "already notified" state, so every over-limit message still triggers an outbound reply.

DMs bypass the mention gate (gated by `isGroupLike`, `:65-67, 92-95`), so a DM flood drives a 1:1 outbound message rate regardless of `maxPerMinute`. The expensive path (agent invocation via `shouldProcess: true`) *is* correctly gated, so this is not a full bypass — but for a PR whose stated purpose is a rate limiter, the limiter failing to throttle its own reply traffic is a defect in the security control and worth fixing before merge.

Suggested direction: move the non-text check after the rate-limit check (or count it against the bucket), and suppress/debounce the "请稍后再试" notice (e.g. send at most once per bucket window).

## P2 — Duplicate display names corrupt the @mention reverse map

In `learnMember` (`group-context.ts:89-107`) the reverse map `nameToUid` is keyed by name with no handling for two distinct uids sharing a name. Trace:

- `learnMember(ch, A, "Bob")` → `nameMap = {Bob→A}`
- `learnMember(ch, B, "Bob")` → `:99-100` overwrites → `nameMap = {Bob→B}`
- `learnMember(ch, A, "Robert")` → `:96-97` deletes the shared `Bob` entry → `nameMap = {Robert→A}`, `memberMap = {A→Robert, B→Bob}`

Net result: user **B is still named "Bob" but `@Bob` no longer resolves** (`resolveMentions` reads `nameMap`, `:185-207`). A's rename silently breaks an unrelated user's mention. Degrades gracefully (the mention just isn't expanded, no security impact), but it's a real correctness bug in shared-name groups, which are common in practice. Consider storing `name → Set<uid>` or only deleting the reverse entry when it still points at the uid being updated.

## P2 — `mention.entities[].uid` is not consulted in the mention gate

`isMentioned` (`session-router.ts:69-76`) checks `mention.uids` and `mention.ais` only; it never inspects `mention.entities` (`types.ts:55-66`), where each entity carries its own `uid`. Both `uids` and `entities` are independently optional in the type. If the server ever delivers a direct @bot via `entities` without also populating `uids`, the group mention gate (`:93-95`) drops the message and the bot silently won't respond. This is fail-safe (a miss, not a false trigger), so not a blocker, **but it depends on a server-side invariant that isn't verifiable from this repo.** Please confirm against the octo-server contract that `uids` is always populated (and is a superset of `entities[].uid`) for direct mentions; if not, fold `entities[].uid` into `isMentioned`. Flagging for human verification given the security-sensitive classification.

## P2 — Serial queue does not serialize the expensive downstream work

The per-key promise chain in `route` (`session-router.ts:34-52`) releases the gate in `finally` (`:47`) as soon as `processMessage` resolves. `processMessage` is almost entirely synchronous (only `replySafe` on reject paths awaits; the success path returns synchronously at `:112`). The actual agent invocation happens in the caller *after* `route` returns. So same-key messages are **not** serialized through the heavy work by this code — the lock only serializes cheap routing. This may be intended to be completed by the T12 wiring, but the PR description's "same key serial" claim overstates what's currently enforced. Recommend the lock span the downstream invocation when T12 lands (or document that serialization is the caller's responsibility).

## P2 — Unbounded memory growth in GroupContext

`messageCache`, `memberMapByChannel`, `nameToUidByChannel`, and `lastRefresh` (`group-context.ts:23-27`) are keyed by `channelId` and never evicted. `window.shift()` (`:83-85`) bounds each per-channel array to 100 but never removes channel entries from the top-level maps. A long-running bot active across many channels grows memory without bound — unlike `SessionRouter.cleanStaleBuckets` (`session-router.ts:135-143`), which has the analogous cleanup. Consider an idle-TTL eviction mirroring the token-bucket approach.

## Nits

- `resolveMentions` (`group-context.ts:198`): the trailing-punctuation `replace` is dead code. The capture class `@([\w一-鿿…]+)` (`:193`) shares zero code points with the strip class, so the captured name can never end in stripped punctuation (the `@` match already terminates at punctuation). Harmless, but misleading — either remove it or widen the capture class if astral/other punctuation handling was actually intended.
- `buildContext` per-line cost uses `line.length` (UTF-16 code units, `:174-175`) as a proxy for the `maxContextChars` budget. Fine for a char budget; just note it's code units, not codepoints/tokens, if this later feeds a token-bounded context.

## Summary

Clean, well-tested code with correct mention semantics and good fail-soft behavior. Requesting changes for the rate-limiter reply gap (P1) and the duplicate-name mention corruption (P2). The `entities` mention question (P2) needs human confirmation against the server contract before merge given the security-sensitive label.

Jerry-Xin (changes requested):
Summary: The PR is relevant to `cc-channel-octo`, but the `SessionRouter` locking API has a blocking deadlock/integration flaw that should be fixed before merge.

**🔴 Blocking**
- 🔴 Critical: `withSessionLock` is documented as the API callers should use to serialize the full message handling pipeline, but `route()` also acquires the same lock internally. A caller following the comment at [src/session-router.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-etmfppfm/repo/src/session-router.ts:36) would naturally do something like `withSessionLock(key, async () => { const r = await route(msg); ... })`; that deadlocks because `route()` waits on the outer gate at [src/session-router.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-etmfppfm/repo/src/session-router.ts:60). If callers avoid `withSessionLock` and call `route()` first, only routing is serialized and the downstream agent/session-store work can still run concurrently for the same session. The router needs one clear non-reentrant contract: either make `route()` non-locking and require callers to wrap the full pipeline, or expose a single `withRoutedSession(msg, handler)` style API that locks once and runs both routing and handling inside the same critical section.

**💬 Non-blocking**
- 🟡 Warning: `GroupContext` prepares `group_members` statements in the constructor but does not create or verify the schema itself at [src/group-context.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-etmfppfm/repo/src/group-context.ts:36). This only works if `SessionStore.init()` has already run. That dependency should be documented, enforced, or moved into shared schema initialization.
- 🟡 Warning: `resolveMentions()` only captures `@` names made of word/CJK/Kana/Hangul characters and stops at spaces or punctuation at [src/group-context.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-etmfppfm/repo/src/group-context.ts:192). Display names containing spaces, hyphens, dots, or other common characters will not resolve.
- 🟡 Warning: I could not run the claimed verification in this checkout because dependencies are not installed: `vitest` and `typescript` are missing from `node_modules`, so `npm test` and `npx tsc --noEmit` fail before exercising the PR.

**✅ Highlights**
- The PR passes the relevance gate: Session routing, rate limiting, and group context are in scope for this bot gateway.
- The new tests cover many important behaviors: mention gating, rate-limit debounce, per-channel member isolation, rename handling, and context window budgeting.

yujiawei (approved):
# Code Review — PR #3 (cc-channel-octo)

**Scope:** `src/session-router.ts` (T8) + `src/group-context.ts` (T9), plus their new test files. +879/-4 over 4 files. Reviewed at head `047ea58`.

**Verdict: APPROVED.** Build is clean (`tsc --noEmit`, zero errors) and the full suite passes (72/72, including 17 new SessionRouter + 18 new GroupContext tests). All four blocking items raised in the previous rounds have been correctly addressed, and I could not find a remaining correctness or security defect that blocks merge. Remaining items below are P2 / hardening suggestions, not merge blockers.

> Both classes are still dead code (`gateway.ts` / `index.ts` are Phase-2 stubs); there is no live production impact today. They are reviewed now so the logic is correct before T7 wires them in.

---

## Previously-blocking items — verified fixed ✅

- **Mention gate no longer triggers on `@所有人`.** `isMentioned()` (`src/session-router.ts:80-87`) now gates only on `mention.uids` (includes robotId) or `mention.ais`; the `mention.all` trigger is removed, matching the documented contract in `octo/types.ts:64-83` (`all` = humans-only). Covered by the `REJECTS group message when only mention.all is set` test.
- **No more mis-resolution to the wrong user.** `resolveMentions()` (`src/group-context.ts:186-208`) does exact-match-after-punctuation-strip only; the aggressive progressive-prefix fallback is gone. The `@Alice` → `A` false-positive is now an explicit negative test.
- **Per-channel identity isolation.** Member maps are keyed by channel (`memberMapByChannel` / `nameToUidByChannel`, `src/group-context.ts:24-26`) and every lookup, including `resolveMentions(text, channelId)`, threads `channelId`. This matches the DB key `(group_id, uid)`. Cross-group name collisions are tested.
- **Full-pipeline serialization mechanism added.** `withSessionLock(key, fn)` (`src/session-router.ts:41-58`) is public so the T7 integrator can hold the per-key lock around the *entire* handler, not just routing. The gate promise only ever resolves (never rejects), so a throw inside `fn` cannot wedge the chain, and the `inboundQueues.get(key) === gate` guard avoids deleting a newer tail. Same-key-serial ordering is tested.

Also confirmed sound: token-bucket refill math and the rate-limit-before-non-text ordering (`checkRateLimit` at `src/session-router.ts:132-155`), the once-per-window notify debounce, stale-bucket eviction (`cleanStaleBuckets`, 5-min TTL), `replySafe` swallowing all outbound errors, `refreshMembers` recording `lastRefresh` only on success, the `learnMember` rename path that deletes the old reverse mapping only when it still points at the same uid, and the `buildContext` budget accounting (total output is provably ≤ `maxContextChars`).

---

## P2 — non-blocking (worth addressing, ideally before T7 integration)

1. **Reentrancy footgun in `withSessionLock` (`src/session-router.ts:41-63`).** `route()` internally calls `withSessionLock(key, ...)`. If the T7 integrator wraps the full handler in `withSessionLock(key, ...)` *and* that handler also calls `route(msg)` for the same key, the inner call awaits the outer gate, which only resolves after the handler returns → **self-deadlock**. The two entry points (`route` vs. `withSessionLock`) must not be nested for the same key. Please add a doc warning, or expose a single combined entry point, so the integrator can't fall into this.

2. **Unbounded per-channel growth in `GroupContext`.** `memberMapByChannel`, `nameToUidByChannel`, `lastRefresh`, and `messageCache` have no channel-level eviction. Bounded by the number of distinct channels the bot ever sees, so not attacker-amplifiable the way `tokenBuckets` was — but for a long-lived gateway in many groups it grows monotonically. Consider an LRU/TTL on idle channels, mirroring the `cleanStaleBuckets` treatment that `SessionRouter` already got.

3. **`resolveMentions` name coverage.** The token regex (`src/group-context.ts:194`) only matches `\w` plus the listed CJK/Hangul/Kana ranges, so display names containing spaces, hyphens, dots, accented Latin, or emoji will silently fail to resolve. If Octo display names are unrestricted, prefer resolving against `payload.mention.entities` (which carry exact offsets/uids) when present, falling back to name matching only as a heuristic.

4. **`buildContext` drops an over-budget single message entirely.** If the newest message alone exceeds the budget, the loop breaks on the first iteration and returns `''` rather than a truncated line (`src/group-context.ts:171-181`). Minor, but a single long message yields no context at all; consider truncating the last line to fill the budget.

---

## For the human reviewer (security-sensitive)

- **Bot-to-bot loop protection rests entirely on `botBlocklist` configuration.** In groups, only blocklisted senders are dropped (`src/session-router.ts:99-101`); any non-blocklisted bot that `@`-mentions this bot (or sends `@所有AI`) will be processed. The `GroupMember.robot` flag already exists in the API type (`octo/api.ts:184-191`) — consider dropping known robots structurally rather than relying on an operator-maintained blocklist, especially before multiple AI bots share a room.
- **`@所有AI` (`mention.ais`) means this bot speaks unsolicited to a broadcast.** Worth a conscious product confirmation that an AI broadcast should wake this adapter; the gate is correct per the documented contract, but it is the one path where the bot replies without being individually addressed.

Neither is a code defect in this PR — both are integration/policy points to confirm as T7 wires this up.

lml2468 (commented):
## cc-channel-octo PR #3 Re-review — 047ea5886ef7 (incremental from ca8b09613f1e)

| Source | Issue | Status |
|--------|-------|--------|
| yujiawei | Non-text messages bypass rate limiting | ✅ Rate limit check moved BEFORE non-text check |
| yujiawei | Rate limit rejection causes reply spam | ✅ `notified` flag debounce — one notification per window, resets on refill |
| Steve/Jerry | rename clobbers another user's nameToUid mapping | ✅ `nameMap.delete(existing)` guarded with `nameMap.get(existing) === uid` — only deletes if still owned by this uid |
| Steve/Jerry | Serial queue gap | ✅ Improved: `route()` now delegates to `withSessionLock()` — routing decision properly serialized |

### Analysis of withSessionLock design

`route()` now wraps `processMessage()` inside `withSessionLock()`. This is correct by itself. The public `withSessionLock()` API lets callers (index.ts in PR #7) extend serial scope to the full pipeline.

**Important usage constraint**: the caller must NOT nest `withSessionLock(key, () => route(msg))` — this would deadlock (inner lock waits for outer gate). The correct pattern is sequential: `route()` first (owns one lock), then `withSessionLock()` for the pipeline (owns another). A small gap between the two remains, but this is a pragmatic tradeoff. Recommend adding a JSDoc warning on `withSessionLock` about non-reentrancy.

### New tests

- Rename safety: two scenarios (same name takeover + rename doesn't clobber other user) ✓
- Rate limit debounce: subsequent rejections silent ✓
- Rate limit on non-text: image message gets rate-limited, not "暂不支持" ✓

### Verdict

APPROVED — all reviewer feedback addressed. Non-reentrant lock usage should be documented.

PR #4, opened 2026-06-04
feat: implement StreamRelay (T11)

lml2468:
## Summary

Implement `StreamRelay` class — throttled streaming output + typing heartbeat + message splitting.

## Changes

**src/stream-relay.ts** (+230 lines):
- `StreamRelay.deliver(channelId, channelType, chunks, apiUrl, botToken)` — consumes `AsyncIterable<string>` and delivers to Octo
- **Typing heartbeat**: 5s interval via `sendTyping`, fires immediately on start, cleaned up in `finally`
- **Throttled flush**: 800ms interval, accumulated (not incremental) text sent via stream API lifecycle (`streamStart` → `streamSend` → `streamEnd`)
- **Fallback**: when stream API throws, degrades to plain `sendMessage` with automatic splitting
- `splitMessage(text, maxChars)` — exported utility, splits at natural boundaries: paragraph (\\n\\n) > newline (\\n) > space > hard cut, 3500 char max per segment
- **Payload encoding**: base64-encoded JSON `{ type: 1, content }` for stream API

**src/__tests__/stream-relay.test.ts** (+348 lines):
- 8 `splitMessage` tests: single segment, exact boundary, paragraph/newline/space/hard split, empty string, reassembly
- 11 `StreamRelay` tests: stream lifecycle (start → send → end), typing indicator, fallback on stream failure, long text splitting, empty stream, timer cleanup on success/error, chunk accumulation, channel params

## Verification

```
npm test          # 56 tests passed (37 protocol + 19 stream-relay)
```

## Task

T11 from Phase 2. No dependencies on other Phase 2 tasks.

### DIFF
--- a/src/__tests__/stream-relay.test.ts
+++ b/src/__tests__/stream-relay.test.ts
@@ -339,4 +339,33 @@ describe("StreamRelay", () => {
     const segments = splitMessage("abc", 1);
     expect(segments).toEqual(["a", "b", "c"]);
   });
+
+  // ── P0 fix: source iterable exception cleans up timer + stream ───────
+
+  it("cleans up flushTimer and stream when source iterable throws", async () => {
+    // Create an iterable that yields one chunk then throws.
+    async function* throwingChunks(): AsyncIterable<string> {
+      yield "partial";
+      throw new Error("source exploded");
+    }
+
+    const chunks = throwingChunks();
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    const guarded = promise.catch(() => {});
+    await vi.runAllTimersAsync();
+    await guarded;
+
+    // Verify it rejected with the source error.
+    await expect(promise).rejects.toThrow("source exploded");
+
+    // streamEnd should have been called to close the started stream.
+    const endCalls = mockState.calls.filter((c) => c.fn === "streamEnd");
+    expect(endCalls.length).toBeGreaterThanOrEqual(1);
+
+    // Typing timer should be cleaned up (no more typing calls after error).
+    const countAfter = mockState.calls.filter((c) => c.fn === "sendTyping").length;
+    await vi.advanceTimersByTimeAsync(30_000);
+    const countLater = mockState.calls.filter((c) => c.fn === "sendTyping").length;
+    expect(countLater).toBe(countAfter);
+  });
 });
--- a/src/__tests__/stream-relay.test.ts
+++ b/src/__tests__/stream-relay.test.ts
@@ -78,16 +78,20 @@ interface ApiCall {
 const mockState = vi.hoisted(() => {
   const calls: ApiCall[] = [];
   let streamStartFail = false;
+  let streamSendFailAfter = -1; // Fail streamSend after N successful calls.
   let sendMessageFail = false;
   return {
     calls,
     get streamStartFail() { return streamStartFail; },
     set streamStartFail(v: boolean) { streamStartFail = v; },
+    get streamSendFailAfter() { return streamSendFailAfter; },
+    set streamSendFailAfter(v: number) { streamSendFailAfter = v; },
     get sendMessageFail() { return sendMessageFail; },
     set sendMessageFail(v: boolean) { sendMessageFail = v; },
     reset() {
       calls.length = 0;
       streamStartFail = false;
+      streamSendFailAfter = -1;
       sendMessageFail = false;
     },
   };
@@ -109,6 +113,12 @@ vi.mock("../octo/api.js", () => ({
   }),
   streamSend: vi.fn(async (params: Record<string, unknown>) => {
     mockState.calls.push({ fn: "streamSend", args: params });
+    if (mockState.streamSendFailAfter >= 0) {
+      const sendCount = mockState.calls.filter((c) => c.fn === "streamSend").length;
+      if (sendCount > mockState.streamSendFailAfter) {
+        throw new Error("streamSend failed");
+      }
+    }
   }),
   streamEnd: vi.fn(async (params: Record<string, unknown>) => {
     mockState.calls.push({ fn: "streamEnd", args: params });
@@ -295,5 +305,38 @@ describe("StreamRelay", () => {
     const payload = (startCalls[0].args as { payload: string }).payload;
     const decoded = JSON.parse(Buffer.from(payload, "base64").toString("utf-8")) as { content: string };
     expect(decoded.content).toBe("test data");
+  });
+
+  // ── Review fix: mid-stream failure calls streamEnd ─────────────────────
+
+  it("calls streamEnd on mid-stream failure before fallback", async () => {
+    // First streamSend succeeds (via streamStart), then streamSend fails on the next flush.
+    mockState.streamSendFailAfter = 0; // Fail from the 1st streamSend
+
+    // Yield two chunks: first triggers streamStart, second triggers a streamSend that fails.
+    const chunks = asyncChunks(["first", " second"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    // streamEnd should be called to close the dangling stream.
+    const endCalls = mockState.calls.filter((c) => c.fn === "streamEnd");
+    expect(endCalls.length).toBeGreaterThanOrEqual(1);
+
+    // Fallback sendMessage should deliver the content.
+    const sendCalls = mockState.calls.filter((c) => c.fn === "sendMessage");
+    expect(sendCalls.length).toBeGreaterThanOrEqual(1);
+  });
+
+  // ── Review fix: splitMessage maxChars guard ────────────────────────────
+
+  it("splitMessage throws on maxChars <= 0", async () => {
+    expect(() => splitMessage("hello", 0)).toThrow("maxChars must be >= 1");
+    expect(() => splitMessage("hello", -5)).toThrow("maxChars must be >= 1");
+  });
+
+  it("splitMessage works with maxChars = 1", () => {
+    const segments = splitMessage("abc", 1);
+    expect(segments).toEqual(["a", "b", "c"]);
   });
 });
--- a/src/__tests__/stream-relay.test.ts
+++ b/src/__tests__/stream-relay.test.ts
@@ -1,0 +1,299 @@
+/**
+ * Tests for StreamRelay — throttled flush, stream lifecycle, fallback, splitting.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { splitMessage } from "../stream-relay.js";
+
+// ─── splitMessage ───────────────────────────────────────────────────────────
+
+describe("splitMessage", () => {
+  it("returns single segment when text fits", () => {
+    const text = "hello world";
+    expect(splitMessage(text, 100)).toEqual(["hello world"]);
+  });
+
+  it("returns single segment when text equals maxChars exactly", () => {
+    const text = "a".repeat(3500);
+    expect(splitMessage(text)).toEqual([text]);
+  });
+
+  it("splits at paragraph boundary (\\n\\n)", () => {
+    const para1 = "a".repeat(1000);
+    const para2 = "b".repeat(1000);
+    const para3 = "c".repeat(1000);
+    const text = `${para1}\n\n${para2}\n\n${para3}`;
+    const segments = splitMessage(text, 2500);
+    expect(segments.length).toBeGreaterThanOrEqual(2);
+    expect(segments[0]).toContain(para1);
+    expect(segments.join("")).toBe(text);
+  });
+
+  it("splits at newline when no paragraph break", () => {
+    const line1 = "a".repeat(2000);
+    const line2 = "b".repeat(2000);
+    const text = `${line1}\n${line2}`;
+    const segments = splitMessage(text, 2500);
+    expect(segments.length).toBe(2);
+    expect(segments[0]).toBe(`${line1}\n`);
+    expect(segments[1]).toBe(line2);
+  });
+
+  it("splits at space when no newline", () => {
+    const word1 = "a".repeat(2000);
+    const word2 = "b".repeat(2000);
+    const text = `${word1} ${word2}`;
+    const segments = splitMessage(text, 2500);
+    expect(segments.length).toBe(2);
+    expect(segments[0]).toBe(`${word1} `);
+    expect(segments[1]).toBe(word2);
+  });
+
+  it("hard cuts when no natural boundary", () => {
+    const text = "a".repeat(7000);
+    const segments = splitMessage(text, 3500);
+    expect(segments.length).toBe(2);
+    expect(segments[0]).toBe("a".repeat(3500));
+    expect(segments[1]).toBe("a".repeat(3500));
+  });
+
+  it("handles empty string", () => {
+    expect(splitMessage("")).toEqual([""]);
+  });
+
+  it("reassembles to original for multi-segment split", () => {
+    const text = Array.from({ length: 20 }, (_, i) => `Paragraph ${i} content.`).join("\n\n");
+    const segments = splitMessage(text, 100);
+    expect(segments.join("")).toBe(text);
+  });
+});
+
+// ─── Shared mock state via vi.hoisted ───────────────────────────────────────
+
+interface ApiCall {
+  fn: string;
+  args: Record<string, unknown>;
+}
+
+const mockState = vi.hoisted(() => {
+  const calls: ApiCall[] = [];
+  let streamStartFail = false;
+  let sendMessageFail = false;
+  return {
+    calls,
+    get streamStartFail() { return streamStartFail; },
+    set streamStartFail(v: boolean) { streamStartFail = v; },
+    get sendMessageFail() { return sendMessageFail; },
+    set sendMessageFail(v: boolean) { sendMessageFail = v; },
+    reset() {
+      calls.length = 0;
+      streamStartFail = false;
+      sendMessageFail = false;
+    },
+  };
+});
+
+vi.mock("../octo/api.js", () => ({
+  sendTyping: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "sendTyping", args: params });
+  }),
+  sendMessage: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "sendMessage", args: params });
+    if (mockState.sendMessageFail) throw new Error("sendMessage failed");
+    return { message_id: "msg_1", client_msg_no: "c1", message_seq: 1 };
+  }),
+  streamStart: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "streamStart", args: params });
+    if (mockState.streamStartFail) throw new Error("stream API unavailable");
+    return "stream_001";
+  }),
+  streamSend: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "streamSend", args: params });
+  }),
+  streamEnd: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "streamEnd", args: params });
+  }),
+}));
+
+// ─── Helpers ────────────────────────────────────────────────────────────────
+
+async function* asyncChunks(items: string[]): AsyncIterable<string> {
+  for (const item of items) {
+    yield item;
+  }
+}
+
+// ─── StreamRelay ────────────────────────────────────────────────────────────
+
+// Import after mock setup so vi.mock hoisting takes effect.
+const { StreamRelay } = await import("../stream-relay.js");
+
+describe("StreamRelay", () => {
+  let relay: StreamRelay;
+
+  const CH_ID = "test_channel";
+  const CH_TYPE = 2; // Group
+  const API_URL = "https://api.test";
+  const BOT_TOKEN = "bf_test";
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+    mockState.reset();
+    relay = new StreamRelay();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("delivers short text via stream start + end", async () => {
+    const chunks = asyncChunks(["Hello, world!"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const streamCalls = mockState.calls.filter((c) => c.fn.startsWith("stream"));
+    expect(streamCalls.some((c) => c.fn === "streamStart")).toBe(true);
+    expect(streamCalls.some((c) => c.fn === "streamEnd")).toBe(true);
+  });
+
+  it("sends typing indicator at the start", async () => {
+    const chunks = asyncChunks(["Hi"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const typingCalls = mockState.calls.filter((c) => c.fn === "sendTyping");
+    expect(typingCalls.length).toBeGreaterThanOrEqual(1);
+  });
+
+  it("falls back to sendMessage when stream API fails", async () => {
+    mockState.streamStartFail = true;
+
+    const chunks = asyncChunks(["Fallback text"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const sendCalls = mockState.calls.filter((c) => c.fn === "sendMessage");
+    expect(sendCalls.length).toBe(1);
+    expect((sendCalls[0].args as { content: string }).content).toBe("Fallback text");
+  });
+
+  it("splits long fallback text into segments", async () => {
+    mockState.streamStartFail = true;
+
+    const longText = "word ".repeat(1500); // ~7500 chars
+    const chunks = asyncChunks([longText]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const sendCalls = mockState.calls.filter((c) => c.fn === "sendMessage");
+    expect(sendCalls.length).toBeGreaterThanOrEqual(2);
+    const reassembled = sendCalls.map((c) => (c.args as { content: string }).content).join("");
+    expect(reassembled).toBe(longText);
+  });
+
+  it("handles empty stream (no output)", async () => {
+    const chunks = asyncChunks([]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const nonTyping = mockState.calls.filter((c) => c.fn !== "sendTyping");
+    expect(nonTyping.length).toBe(0);
+  });
+
+  it("cleans up typing timer on completion", async () => {
+    const chunks = asyncChunks(["done"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const countAfter = mockState.calls.filter((c) => c.fn === "sendTyping").length;
+    await vi.advanceTimersByTimeAsync(30_000);
+    const countLater = mockState.calls.filter((c) => c.fn === "sendTyping").length;
+    expect(countLater).toBe(countAfter);
+  });
+
+  it("cleans up typing timer on error", async () => {
+    mockState.streamStartFail = true;
+    mockState.sendMessageFail = true;
+
+    const chunks = asyncChunks(["fail"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    // Attach a no-op catch to prevent Node's PromiseRejectionHandledWarning
+    // (the rejection is still observable via the original `promise` reference).
+    const guarded = promise.catch(() => {});
+    await vi.runAllTimersAsync();
+    await guarded;
+
+    // Verify the promise did reject.
+    await expect(promise).rejects.toThrow("sendMessage failed");
+
+    const countAfter = mockState.calls.filter((c) => c.fn === "sendTyping").length;
+    await vi.advanceTimersByTimeAsync(30_000);
+    const countLater = mockState.calls.filter((c) => c.fn === "sendTyping").length;
+    expect(countLater).toBe(countAfter);
+  });
+
+  it("accumulates chunks between flushes", async () => {
+    // Multiple chunks yielded synchronously: first chunk flushes immediately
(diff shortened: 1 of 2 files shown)


Jerry-Xin (changes requested):
Scope gate passed; the PR is relevant to `cc-channel-octo`, but the stream flush lifecycle has a blocking race condition.

**🔴 Blocking**
- 🔴 Critical: `StreamRelay` can start multiple Octo streams for one delivery when a timer-triggered flush overlaps with a loop-triggered flush. In [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:186), the timeout callback launches `flush()` without awaiting or otherwise serializing it. While that async `flush()` is still awaiting `streamStart()` at [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:158), `streamNo` remains `null` and `lastFlushTime` remains stale until [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:175). A later chunk can then hit [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:203) and call `flush()` again, causing a second `streamStart`. Only the final `streamNo` is ended at [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:249), so the first stream can be left open and users can see duplicate or inconsistent streamed output. Serialize all flushes through one in-flight promise/mutex and add a test with delayed `streamStart`/`streamSend` plus chunks arriving around the 800 ms boundary.

**💬 Non-blocking**
- 🟡 Warning: Successful deliveries always issue a final `streamSend` whenever `streamNo !== null`, even if the latest flush already sent the same accumulated content. See [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:228). This is not necessarily user-visible if the API treats sends as full-state replacement, but it adds a redundant request on every streamed response. Track the last flushed length/content to avoid no-op sends.
- 🟡 Warning: `streamEnd` failures are swallowed at [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:250). If Octo fails to close the stream, `deliver()` still resolves successfully, which can hide a broken lifecycle. Consider surfacing/logging this failure or defining why it is safe to ignore.
- 🔵 Suggestion: `splitMessage(text, maxChars)` can hang for invalid `maxChars <= 0` because `remaining` may never shrink. Since it is exported, guard `maxChars` and throw a clear error. See [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/stream-relay.ts:42).
- 🔵 Suggestion: The tests only use synchronous chunk production and immediately resolving stream mocks, so they miss the overlapping flush case. The helper at [src/__tests__/stream-relay.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/__tests__/stream-relay.test.ts:120) and mocks at [src/__tests__/stream-relay.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-cklr5jtx/repo/src/__tests__/stream-relay.test.ts:105) should include async latency coverage.

**✅ Highlights**
- The PR keeps the relay independent of the agent SDK and routes Octo operations through the existing API wrapper.
- Fallback message splitting preserves text reassembly and covers paragraph/newline/space/hard-cut cases.

I could not rerun tests locally because `vitest` and `tsc` were not installed in this checkout (`sh: vitest: command not found`, `sh: tsc: command not found`).

lml2468 (commented):
## cc-channel-octo PR #4 Review — e5063aafbcad

### Summary

StreamRelay (T11): async text stream → Octo delivery via stream API (start/send/end) with 800ms throttled flushes, typing heartbeat, graceful fallback to plain sendMessage with splitMessage segmentation. 280 lines implementation + 299 lines tests.

Well-designed: knows nothing about Claude SDK, clean stream protocol semantics (each flush sends full accumulated snapshot, not delta — correct for Octo's replace-mode streams).

### Findings

**P1 — Dangling stream on mid-stream failure** (`stream-relay.ts:~223`)

When `streamFailed` becomes true after a `streamSend` failure (stream already started, `streamNo !== null`), the fallback path returns without calling `streamEnd`. The Octo server is left with a dangling stream that never closes.

Fix:
```typescript
if (streamFailed) {
  if (streamNo !== null) {
    await streamEnd({ apiUrl, botToken, streamNo, channelId, channelType }).catch(() => {});
  }
  await this._deliverFallback(channelId, channelType, accumulated, apiUrl, botToken);
  return;
}
```

**P2 — `splitMessage("")` returns `[""]`** (`stream-relay.ts:41`)

`text.length <= maxChars` is true for empty string, so it returns `[""]`. The fallback path guards with `if (text.length === 0) return`, so this doesn't cause a bug in practice, but `splitMessage` as a public export should arguably return `[]` for empty input. Tests document current behavior — just noting the semantic inconsistency.

**P2 — `flushTimer` / `pendingFlush` potential interleave** (`stream-relay.ts`)

Timer-triggered `flush()` via `pendingFlush` could overlap with a direct `flush()` call from the for-await loop if a chunk arrives right as the timer fires. No data corruption (accumulated is append-only, stream protocol is snapshot-based), but two concurrent API calls is wasteful. Could guard with a `flushing` lock or clear `flushTimer` before direct flushes.

### Tests

Strong test suite:
- Fake timers for deterministic flush behavior
- Lifecycle ordering verification (start → send(s) → end)
- Fallback path + multi-segment splitting + reassembly invariant
- Typing heartbeat cleanup on both success and error paths
- Empty stream → no API calls (correct)
- Accumulated (snapshot) semantics explicitly tested — good documentation

### Verdict

CHANGES_REQUESTED — P1 dangling stream must be cleaned up on mid-stream failure.

yujiawei (approved):
# Code Review — PR #4 (cc-channel-octo)

## Verdict: APPROVED

The diff is well-structured, the happy path is correct, and it is covered by a thorough test suite (19 new `stream-relay` tests; 56 total pass, `tsc --noEmit` clean). `splitMessage` is solid — boundary priority (paragraph → newline → space → hard cut) is correct and every test verifies lossless reassembly. The typing heartbeat fires immediately, pings on a 5s interval, and is cleaned up in `finally` on both success and error paths.

I'm approving because none of the issues below are merge blockers: they are failure-path/edge-case bugs and quality items, and `StreamRelay` is a building block with no callers wired in yet (so it can't regress production today). I do recommend addressing the failure-path duplication and the cross-cutting endpoint question before this is integrated into the live delivery path.

---

## Recommendations (non-blocking)

### 1. Mid-stream failure leaks the stream and duplicates content
`src/stream-relay.ts:222-226`

When `streamStart` succeeds (so `streamNo` is set) but a later `streamSend` throws, `flush()` sets `streamFailed = true`. After the loop, the `streamFailed` branch calls `_deliverFallback(accumulated)` and returns **without calling `streamEnd`**:

```ts
if (streamFailed) {
  // Fallback: deliver entire accumulated text via plain messages.
  await this._deliverFallback(channelId, channelType, accumulated, apiUrl, botToken);
  return;
}
```

Two consequences when the failure happens after a successful start:
- The already-opened stream is never closed — `streamEnd` is skipped, leaving a dangling stream on the server.
- The user has already seen the streamed chunks up to the failure point, and `_deliverFallback` then re-sends the **entire** accumulated text as plain messages → duplicated content.

Note the asymmetry: the *other* fallback branch (`src/stream-relay.ts:240-244`) does the right thing — it calls `streamEnd(...).catch(() => {})` before falling back. Suggest applying the same pattern here: if a stream was started (`streamNo !== null`), call `streamEnd` before falling back, and consider only sending the *unsent remainder* rather than the full text to avoid duplication.

### 2. Possible concurrent `flush()` → double `streamStart`
`src/stream-relay.ts:150-208`

`flush()` reads and mutates shared `streamNo`/`accumulated`, and is invoked from two places: directly in the loop (`await flush()` when `elapsed >= FLUSH_INTERVAL_MS`) and from the scheduled timer (`pendingFlush = flush()`). With a real (network-delayed) async chunk source, the timer can fire and start an in-flight `flush()` while the loop subsequently calls `flush()` again. If the first hasn't yet completed `streamStart`, both observe `streamNo === null` and both issue `streamStart`, opening two streams.

This doesn't surface under the fake-timer tests (synchronous generators consume all chunks before timers run), but it's reachable in production. Consider a simple in-flight guard (e.g. an `isFlushing` flag, or always routing flushes through a single serialized promise chain).

### 3. Redundant final `streamSend`
`src/stream-relay.ts:228-246`

Because each flush sends the **full** accumulated text (cumulative, by design — confirmed by the tests), if the loop's last action was an immediate `await flush()` for the final chunk, the post-loop "flush any remaining buffered text" block re-sends the identical full payload. This is harmless if the server treats sends as replace-semantics, but it's an extra network round-trip every time. Consider tracking whether `accumulated` changed since the last successful flush and skipping the redundant send.

### 4. No cancellation path
`deliver(...)` doesn't accept an `AbortSignal`, though every underlying `api.ts` function supports one. For an agent-output relay, the user stopping generation mid-stream is a normal event; without a signal there's no way to abort an in-flight long delivery. Worth threading a signal through before integration.

### 5. Nit: typing ping on empty output
For an empty stream (no chunks), `deliver` still fires one `sendTyping` before discovering there's nothing to send. Minor UX wart — a "typing…" indicator with no follow-up message. Low priority.

---

## Cross-file verification item (please confirm before relying on the stream path)

This is outside the diff (it concerns the pre-existing `src/octo/api.ts`), but it determines whether `StreamRelay`'s primary path ever runs in production, so flagging it explicitly.

`api.ts` calls these stream endpoints:
- `POST {apiUrl}/bot/stream/start`
- `POST {apiUrl}/bot/stream/send`
- `POST {apiUrl}/bot/stream/end`

Note these omit the `/v1` prefix that every other bot call uses (`/v1/bot/sendMessage`, `/v1/bot/typing`, …). Checking the server side:
- The `/v1/bot/*` bot-API group registers **no** stream routes.
- The robot group registers `/v1/robots/:robot_id/:app_key/stream/start` and `/stream/end` — a different prefix, different auth, and **no `/stream/send`** at all.
- octo-lib (`config/stream.go`) only implements `IMStreamStart` and `IMStreamEnd` (proxying WuKongIM `/streammessage/start` and `/streammessage/end`); there is no stream "send"/append function.

If the deployment target is the standard octo-server, all three calls would 404 → `streamStart` throws → `StreamRelay` silently falls back to `sendMessage` for every delivery, and the entire throttled-stream feature would be dead code in practice. Before wiring this into the live path, please confirm the actual server contract (route paths, `/v1` prefix, whether an append/send endpoint exists, and the expected stream payload schema) and align `api.ts` accordingly. Graceful fallback means this won't crash — but it would mean the headline feature never activates.

---

## Summary

Clean, well-tested implementation of the building block. Approving. Before integration: (1) fix the mid-stream-failure dangling-stream + duplicate-content path, (2) guard against concurrent flushes, and (3) verify the stream endpoints against the real server so the stream path actually engages.

lml2468 (commented):
## cc-channel-octo PR #4 Re-review — 218d03a2ed4c (incremental from e5063aafbcad)

| 1 | P1 | `streamFailed` path doesn't call `streamEnd` — dangling stream | ✅ Fixed: `streamEnd` called before fallback (L233-235) |
| 2 | P2 | `splitMessage("")` returns `[""]` | Kept as-is (acceptable), but added `maxChars < 1` guard with throw |
| 3 | P2 | flushTimer/pendingFlush interleave | ✅ Fixed: `isFlushing` mutex guard added |

### Additional Improvements

- `lastSentLength` tracking prevents redundant final `streamSend` when accumulated text hasn't changed since last flush
- New test: "calls streamEnd on mid-stream failure before fallback" with `streamSendFailAfter` mock
- New tests: `splitMessage` throws on `maxChars <= 0`, works correctly with `maxChars = 1`
- `streamSendFailAfter` mock is well-designed — counts successful calls before triggering failure

### Verdict

APPROVED — all P1 issues resolved, P2 interleave addressed with mutex, new tests cover the fix paths.

Jerry-Xin (changes requested):
Scope check passed: this PR is relevant to `cc-channel-octo`, but there are blocking stream lifecycle correctness issues.

**🔴 Blocking**
- 🔴 Critical: `flush()` can mark unsent text as sent during an in-flight timer flush. At `src/stream-relay.ts:160`, the payload is encoded from the current `accumulated` value, but after awaiting `streamStart`/`streamSend`, `lastSentLength` is set from the later mutable `accumulated.length` at `src/stream-relay.ts:182`. If a scheduled flush is in flight and new chunks arrive meanwhile, those chunks are not in the payload but are counted as sent, so the final check at `src/stream-relay.ts:243` can skip them and silently drop output. Capture the sent length before the await and assign that captured value after success; also ensure a follow-up flush happens when `accumulated.length > lastSentLength`.

- 🔴 Critical: `_deliverViaStream` does not clean up scheduled flush timers when the source iterable throws. If `for await` fails after `scheduleFlush()` arms `flushTimer` at `src/stream-relay.ts:192`, execution skips the cleanup block at `src/stream-relay.ts:221`, while the outer `finally` only clears the typing timer at `src/stream-relay.ts:130`. That can leave a timer that sends partial content after `deliver()` has rejected, and can leave a started stream unclosed. Wrap the stream delivery internals in a `try/finally` that clears `flushTimer`, awaits/settles `pendingFlush` as appropriate, and closes any started stream on abort/error.

**💬 Non-blocking**
- 🟡 Warning: normal `streamEnd` failures are swallowed at `src/stream-relay.ts:263`. If Octo fails to close the stream, `deliver()` still resolves successfully and callers have no visibility into a stuck/incomplete stream. Consider logging or propagating this depending on the intended retry policy.
- 🟡 Warning: `splitMessage` only rejects `maxChars < 1` at `src/stream-relay.ts:43`. `NaN` would bypass that guard and can loop forever because slicing with `NaN` makes no progress. Use `Number.isFinite(maxChars)` and an integer/range check.
- 🟡 Warning: the tests cover happy-path fake timer behavior, but not slow in-flight stream API calls or throwing async iterables, which are the two paths above. Add targeted tests for both.

**✅ Highlights**
- The PR uses the existing Octo API wrapper layer instead of raw fetch calls.
- The fallback splitting behavior is covered with useful reassembly tests.

Local verification note: I attempted `npm test -- --runInBand`, but this checkout does not have `vitest` installed (`sh: vitest: command not found`).

yujiawei (approved):
# Code Review — PR #4 (cc-channel-octo)

## Verdict: APPROVED

Re-review of HEAD `218d03a2`, which addresses the four items raised in the prior review round. All four fixes are present, correct, and covered by tests. `tsc --noEmit` is clean and all 59 tests pass locally (22 stream-relay + 37 protocol). `StreamRelay` remains a self-contained building block with no production callers wired in yet, so it cannot regress live delivery today; the remaining items below are failure-path/edge-case quality notes, not merge blockers.

---

## Verified fixes

1. **Concurrent-flush race → fixed.** `flush()` now early-returns on `isFlushing` and sets the flag *synchronously* before the first `await` (`src/stream-relay.ts:156-159`). Because JS is single-threaded, a timer-triggered flush and a loop-triggered flush can no longer both observe `streamNo === null` and both call `streamStart`. The previously possible double-stream is closed.
2. **Dangling stream on mid-stream failure → fixed.** The `streamFailed` branch now calls `streamEnd` before falling back when a stream was already started (`src/stream-relay.ts:232-236`), so the Octo server no longer holds an orphaned stream.
3. **Redundant final `streamSend` → fixed.** `lastSentLength` is tracked and the post-loop send is gated on `accumulated.length > lastSentLength` (`src/stream-relay.ts:147, 243`). Length comparison is sound here because `accumulated` is append-only.
4. **`splitMessage(maxChars <= 0)` infinite loop → fixed.** Now throws on `maxChars < 1` (`src/stream-relay.ts:43-45`), with tests for `0`, `-5`, and `1`.

---

## Remaining (non-blocking) items

### 1. Mid-stream failure still duplicates already-streamed content
`src/stream-relay.ts:232-239`

The dangling-stream leak is fixed, but the content-duplication half of the original finding remains. If `streamStart` succeeds (user already sees streamed text) and a later `streamSend` throws, the fallback re-sends the **entire** `accumulated` text via `sendMessage`. The user then sees the streamed prefix *and* the full re-sent message. Consider sending only the unsent remainder (`accumulated.slice(lastSentLength)`) when a stream was already partially delivered. Failure-path only, so non-blocking.

### 2. No cancellation path
`deliver(...)` still does not accept an `AbortSignal`, even though every underlying `api.ts` function supports one. For an agent-output relay, a user stopping generation mid-stream is a normal event. Worth threading a signal through before this is wired into the live delivery path.

### 3. Concurrent flushes are silently dropped (acceptable, worth a comment)
When a flush is skipped due to `isFlushing`, that intermediate update is discarded rather than re-queued. This does not lose data — `accumulated` is append-only and the post-loop flush always sends the final state — but it does mean an occasional intermediate stream update can be skipped under load. A one-line comment documenting that the drop is intentional would help future readers.

### 4. Nit: typing ping on empty output
For a stream that yields no chunks, `deliver` still fires one `sendTyping` before discovering there is nothing to send, briefly showing a "typing…" indicator with no follow-up. Minor UX wart.

---

## Cross-file verification item (outside this diff — please confirm before relying on the stream path)

In `src/octo/api.ts` (added in an earlier task, not part of this PR), the stream endpoints use `/bot/stream/start`, `/bot/stream/send`, `/bot/stream/end`, whereas every other endpoint in the same file uses the `/v1/bot/...` prefix (`sendMessage`, `typing`, `register`, `messages/sync`, etc.). If the stream routes are also meant to be under `/v1`, every `streamStart` would 404 in production, `streamFailed` would always trip, and `StreamRelay`'s primary path would silently never run — every delivery would degrade to the `sendMessage` fallback. This is outside the PR diff, so it does not block merge of T11, but it should be verified against the Octo server routing before `StreamRelay` is integrated into the live path.

---

## Highlights

- `splitMessage` boundary priority (paragraph → newline → space → hard cut) is correct, and every test verifies lossless reassembly.
- Stream snapshot semantics (each flush sends the full accumulated text, not a delta) are explicitly tested and correct for Octo's replace-mode streams.
- Typing heartbeat fires immediately, pings on a 5s interval, and is cleaned up in `finally` on both success and error paths.
- The relay stays independent of the agent SDK and routes all Octo operations through the existing API wrapper.

yujiawei (changes requested):
# Code Review — PR #4 (cc-channel-octo)

## Verdict: CHANGES_REQUESTED

This supersedes my earlier Approve on the same HEAD (`218d03a2`). On a deeper pass I locally reproduced two blocking stream-lifecycle correctness bugs — one causes **silent data loss** (a tail chunk is permanently dropped from the delivered message), the other leaks a timer and a stream when the upstream source throws. Both fire only with realistic async latency, which is why the fake-timer test suite (59 passing, `tsc` clean) does not catch them. These must be fixed before merge.

The four fixes from the previous round are genuinely in place and correct (mutex guard, `streamEnd` before fallback, `lastSentLength` dedup, `splitMessage` guard). The problem is that the `lastSentLength` dedup introduced a subtler lost-update race.

---

## 🔴 Blocking

### P0 — Lost-update race drops the final chunk(s) (silent data loss)
`src/stream-relay.ts:155-190, 243`

`flush()` encodes its `payload` from `accumulated` at entry (line 160), but assigns `lastSentLength = accumulated.length` **after** the `await streamStart/streamSend` completes (line 182), reading the *then-current* `accumulated`. If new chunks are appended while a (timer-triggered) flush is in flight, those chunks are not in the sent payload yet are counted as sent. The post-loop guard `if (accumulated.length > lastSentLength && streamNo !== null)` (line 243) then evaluates false, and the trailing text is never sent or streamed.

Reproduced locally (slow `streamSend`, chunks `A`, `B`, then `C` arriving during the in-flight `AB` timer flush):
```
SENT PAYLOADS: ["A","AB"]   // "C" never sent
expected 'AB' to be 'ABC'
```
The user permanently loses the tail of the response.

**Fix:** capture the sent length *before* the await and assign that captured value on success:
```ts
isFlushing = true;
const snapshot = accumulated;            // capture
const payload = encodeStreamPayload(snapshot);
try {
  // ... streamStart / streamSend with payload ...
  lastSentLength = snapshot.length;       // assign the captured length, not accumulated.length
  lastFlushTime = Date.now();
}
```
Additionally ensure a follow-up flush actually runs when `accumulated.length > lastSentLength` after the loop (the current post-loop block does a single `streamSend`; with the capture fix that is sufficient, but please add a test that asserts the final payload equals the full text).

### P0 — Timer and stream leak when the source iterable throws
`src/stream-relay.ts:128-133, 192-270`

`_deliverViaStream` has no `try/finally` around the consume loop. If `for await (const chunk of chunks)` throws after `scheduleFlush()` has armed `flushTimer` (line 196), execution jumps past the cleanup block (lines 222-268). The outer `finally` in `deliver()` (line 130) only clears the *typing* timer, so:
- the armed flush timer still fires and issues a `streamSend` **after** `deliver()` has already rejected, and
- the started stream is never closed with `streamEnd` (dangling stream on the Octo server).

Reproduced locally (generator yields `X`, `Y`, then throws):
```
at reject:  ["start:X"]
after wait: ["start:X","send:XY"]   // send fires post-rejection; no streamEnd
```

**Fix:** wrap the delivery internals in `try/finally` that, on any exit path, clears `flushTimer`, settles/`await`s `pendingFlush`, and closes a started stream (`if (streamNo !== null) await streamEnd(...).catch(()=>{})`). A user stopping generation or an upstream error is a normal event for an agent relay, so this path needs deterministic cleanup.

---

## 🟡 Non-blocking

- **`splitMessage(NaN)` still loops forever.** The guard `maxChars < 1` (line 43) is `false` for `NaN`, so a `NaN` slips through and `slice(0, NaN)` makes no progress. Use `if (!Number.isInteger(maxChars) || maxChars < 1)` (or `Number.isFinite` + range check).
- **`streamEnd` failures are swallowed** on the happy path (line 264). If Octo fails to close the stream, `deliver()` still resolves successfully and callers have no visibility into a stuck stream. Consider logging, or document why it is safe to ignore.
- **Mid-stream failure still duplicates already-streamed content.** When `streamStart` succeeded and a later `streamSend` throws, the fallback re-sends the *entire* `accumulated` via `sendMessage`, so the user sees the streamed prefix plus a full re-send. Consider sending only `accumulated.slice(lastSentLength)`.
- **No `AbortSignal`.** `deliver(...)` accepts no cancellation signal even though every `api.ts` function supports one. Worth threading through before live integration (and it overlaps with the P0 cleanup path above).
- **Test gap.** The suite only exercises synchronous generators with immediately-resolving mocks, which is exactly why both P0s are invisible. Add tests with (a) a slow `streamSend` plus a chunk arriving during the in-flight flush, and (b) a throwing async iterable, asserting full-text delivery / no post-reject calls / stream closed.

---

## Cross-file note (outside this diff — verify before relying on the stream path)

In `src/octo/api.ts` (added in an earlier task, not part of this PR), the stream endpoints use `/bot/stream/start|send|end`, while every other endpoint uses the `/v1/bot/...` prefix. If the stream routes should also be under `/v1`, `streamStart` would 404 in production, `streamFailed` would always trip, and the stream path would silently never run (every delivery degrading to the `sendMessage` fallback). Does not block T11 since it is outside the diff, but please confirm against the Octo server routing before integrating `StreamRelay`.

---

## Highlights

- `splitMessage` boundary priority (paragraph → newline → space → hard cut) is correct, with lossless-reassembly tests.
- Stream snapshot semantics (full accumulated text per flush) are correct for Octo's replace-mode streams and explicitly tested.
- Typing heartbeat fires immediately, pings on 5s, and is cleaned up in `finally` on both paths.
- The relay stays independent of the agent SDK and routes all Octo operations through the existing API wrapper.

Jerry-Xin (approved):
The PR is in scope for `cc-channel-octo` and implements the existing StreamRelay layer with good cleanup behavior and focused test coverage.

💬 Non-blocking

🟡 Warning: Mid-stream fallback may produce duplicate user-visible output depending on Octo stream semantics. If `streamStart` has already created a visible partial message, the fallback path closes that stream and then sends the entire accumulated text again via `sendMessage`, so users could see both the partial stream and the full plain-text answer. See `src/stream-relay.ts:236` and `src/stream-relay.ts:243`, plus the final-send failure path at `src/stream-relay.ts:259`. This may be intended if failed streams are discarded server-side, but it should be confirmed or documented.

🔵 Suggestion: Typing failures are intentionally non-fatal, but `sendTyping(...).catch(() => {})` fully suppresses auth/API/config issues at `src/stream-relay.ts:123` and `src/stream-relay.ts:125`. A rate-limited debug log would make operational diagnosis easier without breaking delivery.

🔵 Suggestion: The mid-stream failure test currently checks that `streamEnd` and fallback happen, but not the intended user-visible outcome or duplication behavior. See `src/__tests__/stream-relay.test.ts:312`. Once Octo stream failure semantics are confirmed, add an assertion or comment locking in that decision.

✅ Highlights

- Stream lifecycle, fallback, timer cleanup, source-iterator errors, and splitting edge cases are covered well.
- `splitMessage` preserves reassembly and guards invalid `maxChars`.
- `finally` cleanup avoids leaking typing/flush timers and attempts to close dangling streams.

Verification run locally after `npm ci`: `npm test` passed 60 tests, `npm run type-check` passed, and `npm run lint` had only pre-existing warnings in `src/octo/socket.ts`.

yujiawei (approved):
# Code Review — PR #4 (cc-channel-octo)

## Verdict: APPROVED

This re-review of HEAD `6b63db7` confirms the two round-2 P0 blocking bugs are genuinely fixed. I reproduced both failure conditions with real timers and a latency-injected stream API (the path the fake-timer suite cannot exercise), and both now behave correctly. `tsc --noEmit` is clean and all 60 tests pass. `StreamRelay` is still a self-contained building block with no production callers wired in, so it cannot regress live delivery today. The remaining items below are non-blocking edge-case / quality notes.

---

## Verified P0 fixes (independently reproduced)

### 1. Lost-update race — tail chunk no longer dropped ✅
`src/stream-relay.ts:155-193, 247-266`

`flush()` now snapshots `snapshotLen = accumulated.length` **before** the `await` and assigns `lastSentLength = snapshotLen` on success (not the post-await `accumulated.length`). With a slow `streamStart`/`streamSend` and chunks arriving mid-flight, the post-loop guard `accumulated.length > lastSentLength` correctly fires the trailing send.

Reproduced with real timers (50ms stream calls, chunks `A`, then `B`/`C` arriving during the in-flight `start("A")` flush):
```
sent payloads: ["A","ABC"]   // "C" delivered; final payload is the full text
```
Previously this produced `["A","AB"]` and silently dropped `C`. Fixed.

### 2. Timer/stream leak on source-iterable throw ✅
`src/stream-relay.ts:205-290`

The consume loop and post-loop logic are now wrapped in `try/finally`. On any exit path the `finally` clears `flushTimer`, awaits `pendingFlush`, and closes a started stream via `streamEnd`. So when the upstream async iterable throws, no armed timer fires after rejection and the stream is closed.

Reproduced (generator yields `X`, `Y`, then throws): `deliver()` rejects with the source error, `streamEnd` is called, and **zero** API calls fire in the 1s after rejection (call count stable at 3 before/after). Fixed.

The four prior-round fixes also remain correct: `isFlushing` mutex (no double `streamStart`), `streamEnd` before fallback on mid-stream failure, `lastSentLength` dedup of the redundant final send, and the `splitMessage(maxChars < 1)` guard.

---

## Non-blocking items

### 1. `splitMessage(NaN)` still loops forever
`src/stream-relay.ts:43`

The guard `maxChars < 1` is `false` for `NaN`, so `NaN` slips through and `slice(0, NaN)` makes no progress — confirmed it loops indefinitely. Since `splitMessage` is an exported utility, harden the guard:
```ts
if (!Number.isInteger(maxChars) || maxChars < 1) {
  throw new Error(`splitMessage: maxChars must be a positive integer, got ${maxChars}`);
}
```
All current callers pass the safe default, so this is not reachable today — but the export invites future misuse.

### 2. Mid-stream failure still duplicates already-streamed content
`src/stream-relay.ts:236-245`

The dangling-stream leak is fixed (P1 from an earlier round), but the duplication half remains: if `streamStart` succeeded (user has already seen streamed text) and a later `streamSend` throws, the fallback re-sends the **entire** `accumulated` via `sendMessage`. The user then sees the streamed prefix plus a full re-send. Consider sending only the unsent remainder, `accumulated.slice(lastSentLength)`, when a stream was already partially delivered. Failure-path only.

### 3. `streamEnd` failures are swallowed on the happy path
`src/stream-relay.ts:270, 288`

If Octo fails to close the stream, `deliver()` still resolves successfully and callers have no visibility into a stuck/incomplete stream. Consider logging, or document why it's safe to ignore.

### 4. No `AbortSignal`
`deliver(...)` accepts no cancellation signal even though every underlying `api.ts` function supports one. For an agent-output relay, a user stopping generation mid-stream is a normal event. Worth threading a signal through before this is wired into the live delivery path.

### 5. Nit: typing ping on empty output
For a stream that yields no chunks, `deliver` still fires one `sendTyping` before discovering there's nothing to send — a brief "typing…" indicator with no follow-up message.

---

## Cross-file note (outside this diff — verify before relying on the stream path)

In `src/octo/api.ts` (added in an earlier task, not part of this PR), the stream endpoints use `/bot/stream/start|send|end`, while every other endpoint in the same file uses the `/v1/bot/...` prefix (`sendMessage`, `typing`, `register`, `messages/sync`, etc.). If the stream routes are also meant to be under `/v1`, every `streamStart` would 404 in production, `streamFailed` would always trip, and `StreamRelay`'s primary path would silently never run — every delivery degrading to the `sendMessage` fallback. This is outside the PR diff and does not block T11, but please confirm the route paths against the Octo server routing before integrating `StreamRelay` into the live path.

---

## Highlights

- `splitMessage` boundary priority (paragraph → newline → space → hard cut) is correct, with lossless-reassembly tests.
- Stream snapshot semantics (each flush sends the full accumulated text, not a delta) are correct for Octo's replace-mode streams and explicitly tested.
- Typing heartbeat fires immediately, pings on a 5s interval, and is cleaned up in `finally` on both success and error paths.
- The relay stays independent of the agent SDK and routes all Octo operations through the existing API wrapper.
- The round-2 fix added the right tests (mid-stream failure → `streamEnd`; source-throw cleanup), and the implementation correctly handles the snapshot-before-await ordering.

lml2468 (commented):
## cc-channel-octo PR #4 Re-review — 6b63db73f665 (incremental from 218d03a2ed4c)

### Previous Blockers Resolution

| Source | Issue | Status |
|--------|-------|--------|
| Steve/Jerry/yujiawei P0 | `lastSentLength` data loss — assigned after await | ✅ `const snapshotLen = accumulated.length` before await, `lastSentLength = snapshotLen` on success |
| Steve P1 | Source iterator exception — timer leak + dangling stream | ✅ Entire body wrapped in try/finally. Finally: clears timer, awaits pendingFlush, calls streamEnd if stream still open |

### Verification

- `snapshotLen` captured at line 159, used at line 184 — chunks arriving during await won't inflate lastSentLength ✓
- Double-close prevention: `streamNo = null` after each `streamEnd` call in normal/error paths, so `finally` only closes if not already closed ✓
- New test: source iterable throws → streamEnd called + typing timer cleaned + error propagated ✓

### Verdict

APPROVED — both P0 blockers resolved with proper cleanup and double-close prevention.

PR #6, opened 2026-06-04
feat: implement entry point orchestration (T12)

lml2468:
## Summary

Implements T12 — the entry point (`src/index.ts`) that wires up all Phase 2 modules into the complete message pipeline.

### Startup Sequence

```
loadConfig() → createAdapter(dbPath) → SessionStore.init()
→ cleanExpired() → GroupContext + StreamRelay
→ OctoGateway.start() (lock → register → WS → heartbeat)
→ SessionRouter(config, botId) → setMessageHandler()
```

### Message Pipeline (`handleMessage`)

1. Cache group messages for context (all messages, not just @bot)
2. `SessionRouter.route()` — filter self/blocklist, mention gate, rate limit
3. `SessionStore.getOrCreate()` + `appendUser()`
4. `GroupContext.refreshMembers()` + `buildContext()` (group chats only)
5. `buildHistoryPrefix()` from session store
6. `buildPrompt()` + `queryAgent()` via Claude Agent SDK
7. `teeChunks()` wrapper — collects full response text while streaming
8. `StreamRelay.deliver()` — stream output to Octo
9. `appendAssistant()` — store full response in session history
10. `resolveMentions()` — log resolved @mentions (group chats)

### Key Design Decisions

- **Generator tee pattern**: `queryAgent()` returns an `AsyncIterable<string>` that can only be consumed once. `teeChunks()` wraps it to simultaneously feed `StreamRelay.deliver()` and collect the full text for `appendAssistant()`.
- **Error recovery**: failed message processing sends a best-effort error reply; reply failures are swallowed to prevent crash cascades.
- **Startup logging**: prints config summary (no tokens), bot ID, expired session cleanup count.
- **Group context caching**: all group messages are cached for context, not just @bot messages.

### Dependencies

This PR merges T7-T11 branches to compile against all modules:
- T7: gateway.ts (PR #5)
- T8: session-router.ts (PR #3)
- T9: group-context.ts (PR #3)
- T10: agent-bridge.ts (PR #5)
- T11: stream-relay.ts (PR #4)

### Files changed (1 only, as scoped)

- `src/index.ts` (+161 lines)

### Verification

```
npm test          # 56 tests passed
```

### DIFF
--- a/src/__tests__/stream-relay.test.ts
+++ b/src/__tests__/stream-relay.test.ts
@@ -1,0 +1,299 @@
+/**
+ * Tests for StreamRelay — throttled flush, stream lifecycle, fallback, splitting.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { splitMessage } from "../stream-relay.js";
+
+// ─── splitMessage ───────────────────────────────────────────────────────────
+
+describe("splitMessage", () => {
+  it("returns single segment when text fits", () => {
+    const text = "hello world";
+    expect(splitMessage(text, 100)).toEqual(["hello world"]);
+  });
+
+  it("returns single segment when text equals maxChars exactly", () => {
+    const text = "a".repeat(3500);
+    expect(splitMessage(text)).toEqual([text]);
+  });
+
+  it("splits at paragraph boundary (\\n\\n)", () => {
+    const para1 = "a".repeat(1000);
+    const para2 = "b".repeat(1000);
+    const para3 = "c".repeat(1000);
+    const text = `${para1}\n\n${para2}\n\n${para3}`;
+    const segments = splitMessage(text, 2500);
+    expect(segments.length).toBeGreaterThanOrEqual(2);
+    expect(segments[0]).toContain(para1);
+    expect(segments.join("")).toBe(text);
+  });
+
+  it("splits at newline when no paragraph break", () => {
+    const line1 = "a".repeat(2000);
+    const line2 = "b".repeat(2000);
+    const text = `${line1}\n${line2}`;
+    const segments = splitMessage(text, 2500);
+    expect(segments.length).toBe(2);
+    expect(segments[0]).toBe(`${line1}\n`);
+    expect(segments[1]).toBe(line2);
+  });
+
+  it("splits at space when no newline", () => {
+    const word1 = "a".repeat(2000);
+    const word2 = "b".repeat(2000);
+    const text = `${word1} ${word2}`;
+    const segments = splitMessage(text, 2500);
+    expect(segments.length).toBe(2);
+    expect(segments[0]).toBe(`${word1} `);
+    expect(segments[1]).toBe(word2);
+  });
+
+  it("hard cuts when no natural boundary", () => {
+    const text = "a".repeat(7000);
+    const segments = splitMessage(text, 3500);
+    expect(segments.length).toBe(2);
+    expect(segments[0]).toBe("a".repeat(3500));
+    expect(segments[1]).toBe("a".repeat(3500));
+  });
+
+  it("handles empty string", () => {
+    expect(splitMessage("")).toEqual([""]);
+  });
+
+  it("reassembles to original for multi-segment split", () => {
+    const text = Array.from({ length: 20 }, (_, i) => `Paragraph ${i} content.`).join("\n\n");
+    const segments = splitMessage(text, 100);
+    expect(segments.join("")).toBe(text);
+  });
+});
+
+// ─── Shared mock state via vi.hoisted ───────────────────────────────────────
+
+interface ApiCall {
+  fn: string;
+  args: Record<string, unknown>;
+}
+
+const mockState = vi.hoisted(() => {
+  const calls: ApiCall[] = [];
+  let streamStartFail = false;
+  let sendMessageFail = false;
+  return {
+    calls,
+    get streamStartFail() { return streamStartFail; },
+    set streamStartFail(v: boolean) { streamStartFail = v; },
+    get sendMessageFail() { return sendMessageFail; },
+    set sendMessageFail(v: boolean) { sendMessageFail = v; },
+    reset() {
+      calls.length = 0;
+      streamStartFail = false;
+      sendMessageFail = false;
+    },
+  };
+});
+
+vi.mock("../octo/api.js", () => ({
+  sendTyping: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "sendTyping", args: params });
+  }),
+  sendMessage: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "sendMessage", args: params });
+    if (mockState.sendMessageFail) throw new Error("sendMessage failed");
+    return { message_id: "msg_1", client_msg_no: "c1", message_seq: 1 };
+  }),
+  streamStart: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "streamStart", args: params });
+    if (mockState.streamStartFail) throw new Error("stream API unavailable");
+    return "stream_001";
+  }),
+  streamSend: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "streamSend", args: params });
+  }),
+  streamEnd: vi.fn(async (params: Record<string, unknown>) => {
+    mockState.calls.push({ fn: "streamEnd", args: params });
+  }),
+}));
+
+// ─── Helpers ────────────────────────────────────────────────────────────────
+
+async function* asyncChunks(items: string[]): AsyncIterable<string> {
+  for (const item of items) {
+    yield item;
+  }
+}
+
+// ─── StreamRelay ────────────────────────────────────────────────────────────
+
+// Import after mock setup so vi.mock hoisting takes effect.
+const { StreamRelay } = await import("../stream-relay.js");
+
+describe("StreamRelay", () => {
+  let relay: StreamRelay;
+
+  const CH_ID = "test_channel";
+  const CH_TYPE = 2; // Group
+  const API_URL = "https://api.test";
+  const BOT_TOKEN = "bf_test";
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+    mockState.reset();
+    relay = new StreamRelay();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("delivers short text via stream start + end", async () => {
+    const chunks = asyncChunks(["Hello, world!"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const streamCalls = mockState.calls.filter((c) => c.fn.startsWith("stream"));
+    expect(streamCalls.some((c) => c.fn === "streamStart")).toBe(true);
+    expect(streamCalls.some((c) => c.fn === "streamEnd")).toBe(true);
+  });
+
+  it("sends typing indicator at the start", async () => {
+    const chunks = asyncChunks(["Hi"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const typingCalls = mockState.calls.filter((c) => c.fn === "sendTyping");
+    expect(typingCalls.length).toBeGreaterThanOrEqual(1);
+  });
+
+  it("falls back to sendMessage when stream API fails", async () => {
+    mockState.streamStartFail = true;
+
+    const chunks = asyncChunks(["Fallback text"]);
+    const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
+    await vi.runAllTimersAsync();
+    await promise;
+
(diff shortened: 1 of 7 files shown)


Jerry-Xin (changes requested):
Relevant PR, but the orchestration has blocking runtime issues that can break ordered conversation handling and streaming.

**🔴 Blocking**
- 🔴 Critical: Per-session concurrency is released before the actual message pipeline runs. `SessionRouter.route()` only queues through `processMessage()` and resolves the gate in `src/session-router.ts:41-45`; `src/index.ts:94-130` then runs session writes, history building, agent querying, and delivery outside that queue. Two same-session messages can run concurrently, producing out-of-order assistant history and responses built without the prior assistant turn. Keep the session gate held for the whole processing callback, or return an explicit release guard used after delivery/storage finishes.

- 🔴 Critical: Stream API paths appear to omit the `/v1` prefix used by every other bot endpoint. `src/octo/api.ts:355`, `src/octo/api.ts:380`, and `src/octo/api.ts:401` call `/bot/stream/*`, while send, typing, heartbeat, register, group, sync, and user info all use `/v1/bot/*`. If the server follows the same versioned route convention, streaming will always fail and silently fall back to plain messages. Fix the routes or add server-backed proof/tests showing these unversioned paths are intentional.

**💬 Non-blocking**
- 🟡 Warning: The current message is duplicated in the prompt. `src/index.ts:112-117` appends the user message before calling `buildHistoryPrefix()`, then passes the same content again as `[Current message]`. In group chats, `src/index.ts:83-91` also caches the same message before context is built, so the latest user text may appear in group context, history, and current message. Build history/context from prior messages, then persist the current user turn.

- 🟡 Warning: The entry point has no tests covering the full pipeline ordering, error reply path, prompt composition, or same-session concurrent messages. The existing stream relay tests are useful, but they would not catch the queue boundary bug above.

**✅ Highlights**
- Startup wiring is readable and keeps tokens out of config logs.
- `teeChunks()` correctly handles the single-consumer async iterable while preserving the full response for persistence.
- Verified locally after `npm ci`: `npm run type-check`, `npm test`, and `npm run lint` all completed with no errors. Lint reports warnings only.

lml2468 (commented):
## cc-channel-octo PR #6 Review — c5c5235d300b

### Summary

Full orchestration PR (T12): entry point (`index.ts`), gateway (`gateway.ts`), agent bridge (`agent-bridge.ts`), plus bundled copies of session-router (T8), group-context (T9), and stream-relay (T11+tests).

New code is well-structured — clean startup sequence, tee pattern for streaming + history capture, graceful shutdown with lock file. But there are blocking issues.

### Critical

**P0 — User message duplicated in prompt** (`index.ts:120-123`)

```typescript
store.appendUser(sessionKey, userContent);                    // L120: stores msg in DB
const historyPrefix = store.buildHistoryPrefix(...);           // L121: retrieves it (including L120)
const prompt = buildPrompt(historyPrefix, contextStr, userContent); // L123: adds it AGAIN
```

`buildHistoryPrefix` returns the just-appended message as part of history, then `buildPrompt` adds `[Current message]\n<message>` again. The LLM sees the user's message twice — token waste + model confusion. Fix: either exclude the current message from history (append AFTER building prefix), or remove the `[Current message]` section from `buildPrompt`.

### P1 — Unfixed findings from PRs #3 and #4

This PR bundles identical copies of session-router.ts, group-context.ts, and stream-relay.ts from PRs #3/#4 without incorporating the CR feedback:

| Source PR | Issue | Status |
|---|---|---|
| #3 | `isMentioned` treats `mention.all` as bot-relevant (should only check `ais`) | ❌ unfixed |
| #3 | `tokenBuckets` unbounded memory growth | ❌ unfixed |
| #3 | Non-text messages reply to system events | ❌ unfixed |
| #4 | `streamFailed` path doesn't call `streamEnd` (dangling stream) | ❌ unfixed |

These were flagged by me, Steve, and Jerry. They should be fixed before merging in any PR.

### P1 — New findings

**`fromName` extraction is type-unsafe** (`index.ts:85`)

```typescript
const fromName = (msg.payload as Record<string, unknown>).from_name as string | undefined;
```

`MessagePayload` doesn't define `from_name`. This bypasses TypeScript's type system via double cast. If the field location changes, this silently breaks. Should either extend the type definition or use a proper accessor.

**Mention resolution results unused** (`index.ts:136-139`)

`groupContext.resolveMentions(fullResponse)` computes UIDs but only logs them — never passes them to any API call. Dead code that gives the impression mentions are working.

**`permissionMode` / `settingSources` cast as `any`** (`agent-bridge.ts:46,50`)

The SDK options use `as any` / `as any[]` to bypass type checking. If config contains an invalid value, it passes through silently. Should validate or use SDK-defined type unions.

### P2 observations

- Gateway WS socket construction is duplicated between `registerAndConnect` and `attemptTokenRefresh` — extract a factory.
- `attemptTokenRefresh` can recurse via `onError` → `attemptTokenRefresh` after the cooldown window. The 60s cooldown mitigates tight loops but doesn't eliminate them.
- Lock file write is non-atomic (TOCTOU between `existsSync` and `writeFileSync`). Acceptable for a bot but worth noting.

### Verdict

CHANGES_REQUESTED — P0 prompt duplication must be fixed. P1s from PRs #3/#4 should not be merged in bundled form while still unfixed.

yujiawei (changes requested):
# Code Review — PR #6 (cc-channel-octo)

Thanks for the thorough write-up — the orchestration shape is clean and the tee/streaming design is well documented. `tsc --noEmit` is clean and all 56 tests pass locally.

However, this PR is the entry point that wires untrusted chat input into a tool-capable agent, and review surfaced several blocking issues: a mention-gate bug that makes the bot answer broadcasts not directed at it, a concurrency gap where the "per-session serialization" does not actually serialize the agent pipeline (corrupting stored history), and a stream-failure path that duplicates output and leaks the stream. Requesting changes on those; security-posture notes and nits follow.

> Note: the diff touches 7 files (gateway/session-router/group-context/agent-bridge/stream-relay + index + a test), not "1 file only" as the description states. The review covers all changed files.

---

## P0 — Bot answers `@所有人` (humans-only) broadcasts

**`src/session-router.ts:67-74`**

```ts
private isMentioned(msg: BotMessage): boolean {
  const mention = msg.payload.mention;
  if (!mention) return false;
  if (mention.uids?.includes(this.robotId)) return true;
  if (mention.all) return true;   // <-- humans-only signal
  if (mention.ais) return true;
  return false;
}
```

`mention.all` is documented in `src/octo/types.ts:64-73` as a **humans-only** signal: the server double-writes `all=1` on outbound for legacy clients, and "Adapter treats `all=1` as a humans-only signal (NOT ais)". This bot is an AI, so `@所有人` is *not* addressed to it. The mention gate at line 91 is the only group filter, so every `@所有人` broadcast in every group the bot is in fires a full SDK query + reply — unsolicited spam and token cost on a routine operator action. Because the server keeps double-writing `all=1` post-split, this is a live path, not dead code.

**Fix:** drop the `if (mention.all) return true;` branch. Only a direct uid mention or `mention.ais` should trigger an AI bot. Do **not** add `mention.humans` either — it is also humans-only.

---

## P1 — Per-session "serialization" doesn't cover the agent pipeline → history corruption

**`src/session-router.ts:32-50` + `src/index.ts:95-136`**

`SessionRouter` advertises "concurrency control" and uses `inboundQueues` to chain per-session work, but `route()` releases the gate in its `finally` immediately after `processMessage()` returns — i.e. after only the *routing decision* (filter / mention / rate-limit):

```ts
try {
  await prev;
  return await this.processMessage(msg, key);   // routing decision only
} finally {
  resolveGate();                                 // gate released here
}
```

The expensive work — `queryAgent()`, `streamRelay.deliver()`, `appendUser()`, `appendAssistant()` — all runs in `handleMessage` **after** `route()` has returned and the gate is already released (`index.ts:95` onward). Two messages from the same session that arrive close together therefore run their agent pipelines **concurrently**. Concretely, the interleaving of `appendUser` (line 113) and `appendAssistant` (line 135) across two in-flight turns can store messages out of order — e.g. `user1, user2, assistant2, assistant1` — permanently corrupting `buildHistoryPrefix` for that session.

**Fix:** if per-session serialization is the intent (the docstring says so), the gate must wrap the whole pipeline, not just the routing decision. Either move the heavy processing inside the queued critical section, or have `index.ts` await a per-session chained promise around `handleMessage`'s body.

---

## P1 — Partial stream failure duplicates the response and leaks the stream

**`src/stream-relay.ts:193-224`**

Stream payloads are full snapshots (confirmed by the test "sends accumulated text (not incremental) in each stream flush"). If `streamStart` succeeds (a snapshot is shown to the user) and a *later* in-loop `streamSend` throws, `flush()` sets `streamFailed = true` while `streamNo` stays non-null. The loop keeps accumulating, then the final block runs:

```ts
if (streamFailed) {
  await this._deliverFallback(channelId, channelType, accumulated, apiUrl, botToken);
  return;
}
```

`_deliverFallback` re-sends the **entire** text via plain `sendMessage`, so the user sees the streamed prefix **plus** a full second copy. And there is no `streamEnd` on this path (the `streamEnd` at ~line 250 is after the early `return`), so the already-started stream is left open — a leak. Note the *other* failure path (final-send failure, ~lines 240-245) deliberately calls `streamEnd` before falling back, which makes this omission look unintentional.

**Fix:** only fall back to a fresh plain message when nothing has been streamed yet (`streamNo === null`). If a stream was already started, best-effort `streamEnd` it and avoid re-sending content already delivered (track sent length and send only the remainder).

---

## P1 — Security posture: untrusted chat input drives a `bypassPermissions` agent with file + web tools

**`src/agent-bridge.ts:42-55`, `src/index.ts:84-130`** (defaults in `config.ts:48-52`, forwarded here)

This PR is where external, attacker-influenced input is connected end-to-end to a tool-capable agent:

- Default `permissionMode: 'bypassPermissions'` → `allowDangerouslySkipPermissions: true` (`agent-bridge.ts:49-53`), with `allowedTools` = `Read, Glob, Grep, WebFetch, WebSearch` running in `config.cwd`.
- Any group member who `@`-mentions the bot (or any DM) reaches `queryAgent` with a prompt built **verbatim** from chat content (`index.ts:112-117`), and the agent's output is streamed back to the requester.

Three concrete consequences, all confirmed reachable:

1. **File-read exfiltration** — a user can ask the bot to `Read` files under `cwd` (configs, source, secrets) and have the contents streamed into chat.
2. **SSRF** — `WebFetch` has no host allowlist; a user can target `169.254.169.254`, loopback admin ports, or the internal `apiUrl` host and exfiltrate the response.
3. **Prompt injection** — `index.ts:84-92` caches **all** group messages (not just @bot), and `buildContext` concatenates them as `${fromName}：${content}` with no escaping/fencing into the prompt (`group-context.ts:136-144`, `agent-bridge.ts:24-26`). The structural headers (`[Current message]`, `[Group context]`) are plain text any member can type, so a member can forge headers / inject instructions that persist across the 100-message window for other users' turns.

I'm flagging this as P1 rather than a hard P0 because some of it stems from defaults outside the diff — but this is the PR that assembles the untrusted-input → privileged-agent path, so it should not merge without an explicit decision. **At minimum:** confirm whether `bypassPermissions` + `WebFetch` are intended for an externally-triggered bot; if so, add a tool/permission gate or `cwd` sandbox, an outbound URL allowlist (block RFC1918 / link-local / loopback), and fence untrusted chat content as data (random per-invocation delimiter + a system-prompt instruction that fenced text is reference only).

---

## P2 / nits (non-blocking)

- **Redundant final stream send** (`stream-relay.ts:200-246`): `accumulated` is never cleared after a flush, so when an in-loop or timer flush already ran, the final block re-encodes and re-sends the same snapshot. Harmless only if the Octo stream API treats each `streamSend` as full-content *replace* (not documented in `api.ts`/`types.ts`); under append semantics it duplicates. Track `lastSentLength` and skip the final send when nothing new accumulated.
- **Double-flush race** (`stream-relay.ts:182-209`): the timer-scheduled `flush()` is stored in `pendingFlush` but a direct `await flush()` on the next chunk can run concurrently with it; both read/write `accumulated`/`lastFlushTime`/`streamNo` with no mutex. Low-probability given snapshot idempotency, but worth a single in-flight-flush guard.
- **No tests for the mention gate** (`session-router.ts:67-74`): nothing in `src/__tests__` exercises `isMentioned`, which is exactly why the `all`-vs-`humans` bug slipped through. Add a matrix: `uids` includes robotId → mentioned; `all=1` → NOT mentioned; `humans=1` → NOT mentioned; `ais=1` → mentioned; no mention → not.
- **Stale `nameToUid` on rename** (`group-context.ts:70-105`): when a member's display name changes, `memberMap` is overwritten but the old `name → uid` entry in `nameToUid` is never deleted, leaving a stale reverse mapping that `resolveMentions` can hit.
- **Bot-loop guard is opt-in only** (`session-router.ts:76-93`): self/blocklist filtering relies on an empty-by-default `botBlocklist`; a noisy peer bot emitting `@所有AI` can repeatedly trigger expensive runs (bounded only by the 5/min token bucket). Consider a human-sender-or-explicit-allowlist policy for group triggers. (A true A↔B loop is not reachable — the bot's own replies carry no mention payload.)

---

## Verdict

Requesting changes. The blocking items are the `mention.all` gate (P0), the concurrency/history-corruption gap (P1), and the stream partial-failure duplicate + leak (P1); the security-posture item (P1) needs at least an explicit acknowledgement before this entry point ships. The P2s/nits can be folded in or deferred.

lml2468:
Superseded by PR #7 which includes all T12 code + PR#5/PR#6 review fixes (security hardening, concurrency control, prompt dedup, stream fixes).

PR #10, opened 2026-06-04
test: add config, gateway, and routeAndHandle concurrency tests (T13)

lml2468:
### T13-a: config.test.ts (33 tests)
- `loadConfig` three-level priority (env > file > defaults)
- Required field validation (botToken / apiUrl throw on missing)
- Invalid JSON error includes file path
- `_`-prefix key filtering (e.g. `_comment`)
- Full `CC_OCTO_*` env override coverage (all 14 variables)
- `parseIntStrict`: rejects hex (`0xff`), negative, zero, scientific notation (`1e3`), float (`3.14`), non-numeric
- `parseCsv`: trims whitespace, filters empty segments, single item
- Missing config file fallback

### T13-b: gateway.test.ts (11 tests)
- Process lock: acquire/release lifecycle, stale PID detection, live PID rejection
- Bot registration + WKSocket factory verification
- Message handler self-filter
- Token refresh 60s cooldown (blocks within window, allows after)
- Heartbeat 3x consecutive failure → reconnect trigger
- Heartbeat failure count reset on success

### T13-c: session-router.test.ts extended (6 new tests)
- `routeAndHandle`: same-session FIFO serial execution
- `routeAndHandle`: different-session parallel execution
- Max-1 concurrent handler per session key
- Non-processable messages skip handler
- Processable messages invoke handler
- 10-message burst: FIFO + max-1 concurrent assertion

### Verification
- `npm test` → 145 tests all passing (was 95)
- No source files modified

### DIFF
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -1,0 +1,352 @@
+/**
+ * Config tests.
+ *
+ * Coverage:
+ *  - loadConfig three-level priority (env > file > defaults)
+ *  - parseCsv / parseIntStrict boundary cases
+ *  - Required field validation (botToken / apiUrl)
+ *  - Invalid JSON error message includes file path
+ *  - _-prefix key filtering
+ *  - Full CC_OCTO_* env override coverage
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { join } from 'node:path';
+import { mkdtempSync, writeFileSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { loadConfig } from '../config.js';
+
+// ─── Helpers ────────────────────────────────────────────────────────────────
+
+let tmpDir: string;
+const savedEnv: Record<string, string | undefined> = {};
+const CC_VARS = [
+  'CC_OCTO_BOT_TOKEN', 'CC_OCTO_API_URL', 'CC_OCTO_CWD', 'CC_OCTO_DATA_DIR',
+  'CC_OCTO_SDK_MODEL', 'CC_OCTO_SDK_ALLOWED_TOOLS', 'CC_OCTO_SDK_PERMISSION_MODE',
+  'CC_OCTO_SDK_MAX_TURNS', 'CC_OCTO_SDK_SYSTEM_PROMPT', 'CC_OCTO_SDK_SETTING_SOURCES',
+  'CC_OCTO_RATE_LIMIT_MAX_PER_MINUTE', 'CC_OCTO_CONTEXT_MAX_CHARS',
+  'CC_OCTO_CONTEXT_HISTORY_LIMIT', 'CC_OCTO_BOT_BLOCKLIST',
+];
+
+function setup() {
+  tmpDir = mkdtempSync(join(tmpdir(), 'cc-octo-config-test-'));
+  for (const v of CC_VARS) {
+    savedEnv[v] = process.env[v];
+    delete process.env[v];
+  }
+}
+
+function teardown() {
+  for (const v of CC_VARS) {
+    if (savedEnv[v] !== undefined) {
+      process.env[v] = savedEnv[v];
+    } else {
+      delete process.env[v];
+    }
+  }
+  rmSync(tmpDir, { recursive: true, force: true });
+}
+
+function writeConfig(obj: Record<string, unknown>, filename = 'config.json'): string {
+  const path = join(tmpDir, filename);
+  writeFileSync(path, JSON.stringify(obj));
+  return path;
+}
+
+// ─── 1. Defaults ────────────────────────────────────────────────────────────
+
+describe('loadConfig defaults', () => {
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it('returns correct defaults when only required fields are provided via file', () => {
+    const path = writeConfig({ botToken: 'bf_test', apiUrl: 'https://api.test' });
+    const cfg = loadConfig(path);
+
+    expect(cfg.botToken).toBe('bf_test');
+    expect(cfg.apiUrl).toBe('https://api.test');
+    expect(cfg.cwd).toBe(process.cwd());
+    expect(cfg.dataDir).toBe('./data');
+    expect(cfg.sdk.allowedTools).toEqual(['Read', 'Write', 'Edit', 'Bash', 'Glob', 'Grep', 'WebFetch', 'WebSearch']);
+    expect(cfg.sdk.permissionMode).toBe('bypassPermissions');
+    expect(cfg.sdk.settingSources).toEqual(['user']);
+    expect(cfg.rateLimit.maxPerMinute).toBe(5);
+    expect(cfg.context.maxContextChars).toBe(6000);
+    expect(cfg.context.historyLimit).toBe(40);
+    expect(cfg.botBlocklist).toBeUndefined();
+  });
+});
+
+// ─── 2. Three-Level Priority ────────────────────────────────────────────────
+
+describe('three-level priority: env > file > defaults', () => {
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it('file overrides defaults', () => {
+    const path = writeConfig({
+      botToken: 'bf_file',
+      apiUrl: 'https://file.test',
+      dataDir: '/custom/data',
+      rateLimit: { maxPerMinute: 10 },
+    });
+    const cfg = loadConfig(path);
+    expect(cfg.dataDir).toBe('/custom/data');
+    expect(cfg.rateLimit.maxPerMinute).toBe(10);
+  });
+
+  it('env overrides file', () => {
+    const path = writeConfig({
+      botToken: 'bf_file',
+      apiUrl: 'https://file.test',
+      dataDir: '/from-file',
+    });
+    process.env.CC_OCTO_DATA_DIR = '/from-env';
+    const cfg = loadConfig(path);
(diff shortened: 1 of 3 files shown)


Jerry-Xin (approved):
The PR is in scope and adds relevant regression coverage for existing config, gateway, and session routing behavior. No blocking issues found.

💬 Non-blocking
- 🟡 Warning: [src/__tests__/gateway.test.ts:283](src/__tests__/gateway.test.ts:283) only checks that reconnect happened after advancing through three heartbeat failures. It does not assert that reconnect did *not* happen after the first or second failure, so an implementation with a too-low threshold could still pass. Add intermediate `registerBot` call-count assertions after failure 1 and failure 2.
- 🟡 Warning: [src/__tests__/gateway.test.ts:297](src/__tests__/gateway.test.ts:297) does not fully prove heartbeat failure count reset on success. A missing reset would still pass this test unless another post-success failure is exercised. Add one or two failures after the successful heartbeat and assert no reconnect occurs too early.
- 🔵 Suggestion: [src/__tests__/gateway.test.ts:84](src/__tests__/gateway.test.ts:84) and other `gw.start()` tests accumulate `process.once('SIGINT'/'SIGTERM')` handlers through `OctoGateway.start()`. Consider stubbing `process.once` in this test file or adding cleanup support to avoid process-global listener leakage as the suite grows.
- 🔵 Suggestion: [src/__tests__/config.test.ts:292](src/__tests__/config.test.ts:292) says “rejects empty string” but asserts that an empty env var is ignored. Rename the test to match the expected behavior.

✅ Highlights
- Good coverage of config precedence, required fields, JSON parse errors, and env overrides.
- The new `routeAndHandle` tests cover the important same-session serialization behavior and cross-session parallelism.
- Gateway tests use mocked network/socket boundaries while keeping lock-file behavior realistic.

Verification note: I could not rerun tests locally because dev dependencies are not installed in this checkout (`vitest`, `tsc`, and `eslint` were not found).

yujiawei (approved):
# Code Review — PR #10 (cc-channel-octo)

**Verdict: APPROVED** — test-only PR, no source changes, all gates green. Findings below are non-blocking (P2 / nit) test-quality and coverage suggestions.

## Summary

This PR adds 825 lines across three test files and modifies **no source** (`git diff main...c507e53 --name-only` → only `config.test.ts`, `gateway.test.ts`, `session-router.test.ts`). I verified locally at the head SHA:

- `npm test` → **145 passed** (was 95; +50 new) ✅
- `tsc --noEmit` → clean ✅
- `eslint src/` → 0 errors (13 warnings, all pre-existing in untouched `socket.ts`) ✅
- **Mock/source contract fidelity: clean.** Every `vi.mock()` return shape matches the real signatures — `registerBot` mock matches `BotRegisterResp` (`api.ts:156-180`), the `WKSocket` mock exposes exactly the methods the gateway calls (`connect`, `disconnectAndWait`), and the `sendMessage` mock matches the param object `replySafe` passes. No test asserts a call shape the source never makes.

The new tests are genuinely useful: the `routeAndHandle` concurrency suite (FIFO ordering + `maxConcurrent === 1` per session key, parallel across keys, 10-message burst) meaningfully exercises the per-session lock, and the config priority / `parseIntStrict` / `parseCsv` coverage is thorough. Because this PR was routed `security_sensitive`, I reviewed the gateway (token/lock handling) and router (blocklist + mention gate) coverage extra carefully; the core authorization boundary is well covered. Nothing blocks merge.

## Findings (all non-blocking)

### P2 — worth a follow-up

**1. PID-`1` "live process holds the lock" test is uid-dependent and likely tests the wrong branch.** `gateway.test.ts:117-125` writes PID `1` and asserts `start()` rejects with `/Another instance is running/`. But `acquireLock` (`gateway.ts:71-86`) has **two branches that throw a byte-identical message**: the live-process path (`process.kill(pid,0)` succeeds → throw, L73-75) and the EPERM path (L83-85). As a non-root user `process.kill(1, 0)` throws **EPERM** (verified on this runner, uid≠0), so the test passes through the EPERM branch — *not* the "signal-0 succeeds" branch its name advertises — and as root it would hit the other branch instead. The regex can't tell them apart. The reject-on-live-lock behavior *is* still asserted, so this isn't a pass-when-should-fail bug, but the genuine live-process detection path may have no deterministic coverage. Suggest splitting into two tests that stub `process.kill` to (a) return normally and (b) throw `{code:'EPERM'}`, and/or using distinct messages so assertions can distinguish the branches.

**2. No test for `process.kill` throwing a non-ESRCH/non-EPERM error → silent lock-steal path.** `acquireLock`'s catch (`gateway.ts:76-90`) handles only `ESRCH`, the `"Another instance"` re-throw, and `EPERM`; **any other error code falls through with no `else` and execution continues to `writeFileSync` at L90, overwriting (stealing) the lock**. For a single-instance gateway a silent steal means a double-running bot / double token refresh. The stale test uses `999999999` (returns `ESRCH` on Linux), so this fall-through is unexercised. Consider a test that mocks `process.kill` to throw e.g. `{code:'EINVAL'}` to pin down the intended behavior (and consider whether the source should fail-closed here — but that's a source concern, out of scope for this test PR).

**3. Canonical three-state `mention.humans` rejection is untested.** `types.ts:82-83` documents `humans=1` as the **server-authoritative** "@所有人" signal (the form that arrives in production post three-state split), with legacy `all=1` double-written for old clients. `isMentioned` (`session-router.ts:98-105`) correctly ignores `humans` (returns false → bot stays silent), but the suite only asserts the rejection for the **legacy** `{ all: 1 }` (`session-router.test.ts:124-130`). The production-bound field guarding this authorization boundary has no test — a regression adding `if (mention.humans) return true` would not be caught. Add a `{ humans: 1 }` → `null` case.

**4. `ChannelType.CommunityTopic` authorization surface is never exercised.** `isGroupLike` (`session-router.ts:94-96`) gates **both** the blocklist drop (L117) and the mention gate (L122) for `CommunityTopic` (=5), but every router test uses only `DM` or `Group`. A refactor dropping `CommunityTopic` from `isGroupLike` would silently let the bot answer un-mentioned/blocklisted senders in community topics with the suite still green. Add at least one `CommunityTopic` mention-gate + blocklist case.

**5. Serial-queue test passes regardless of locking.** `session-router.test.ts:241-257` ("processes same session key sequentially") pushes to `order` inside a `.then()` with no async delay in the work, so FIFO holds even if the lock were broken. The newer `routeAndHandle` suite (L270+) fixes this with real `setTimeout` inside the handler and is the genuine assertion; the older test gives false confidence and could be tightened or considered superseded.

**6. Reconnect assertion is loose.** `gateway.test.ts:275-295` asserts `registerBot` call count `toBeGreaterThan(initialRegCalls)` after 3 heartbeat failures — it confirms reconnect *fired* but can't detect *over*-firing (e.g. a regression triggering multiple refreshes). `toBe(initialRegCalls + 1)` would be stricter.

### nit — optional polish

- **`config.test.ts:292` test name is misleading.** Named `rejects empty string` and placed in the `parseIntStrict via env` block (whose siblings all `toThrow`), but the body asserts the **opposite**: empty string is falsy → `if (env.CC_OCTO_SDK_MAX_TURNS)` short-circuits (`config.ts:144`) → no override, no throw. The assertion is correct behavior; only the name lies. Rename to "empty env var is ignored (no override)", and optionally add a direct `parseIntStrict('')` test since that throw path is never exercised.
- **Stale-PID `999999999`** (`gateway.test.ts:104-115`) relies on the kernel returning `ESRCH` (not `EINVAL`) for an out-of-range PID. Robust on Linux CI; a forked-then-exited child PID would be more portable.
- **`'different session keys run in parallel'`** (`session-router.test.ts:295-319`) is the one assertion gated on wall-clock (real 10ms `setTimeout`) rather than promise ordering. The loose `> 1` bound makes it safe in practice; noted only as the single timing-sensitive case.
- **`WKSocket` mock over-provisions** `disconnect`/`updateCredentials` (`gateway.test.ts:25-33`) — methods the gateway never calls. Harmless, minor noise.
- **`releaseLock` only-deletes-own-PID guard** (`gateway.ts:97`) is untested — no case writes a foreign PID then asserts the file survives `stop()`.
- **Lock-file `0o600` mode** (`gateway.ts:90`) is never asserted (`statSync(...).mode`); dropping the mode arg would go unnoticed.
- **"token never logged"** has no negative assertion. Source doesn't log `im_token`/`botToken` today; a `console.log/error` spy asserting no arg contains the token would guard against a future credential-leak regression.
- **`attemptTokenRefresh` re-entrancy guard** (`isRefreshing`, `gateway.ts:155`) isn't isolated — the cooldown (also tested) overlaps it, so the tests can't tell which suppressed a concurrent refresh. Low risk since `lastRefreshTime` is set synchronously before the first `await`.
- **Gateway `start()` registers SIGINT/SIGTERM listeners** (`gateway.ts:231-240`) that aren't removed between tests; uses `process.once` so impact is bounded, but repeated `start()` across cases accumulates listeners.

## Items a human may want to glance at (security-sensitive routing)

None are introduced by this PR (it adds tests only), but flagged per the `security_sensitive` classification as the source paths least covered by the new tests: the lock-file fall-through-and-steal branch (finding #2), and the `mention.humans` / `CommunityTopic` authorization gates (#3, #4). All are source behaviors that are currently correct; the note is purely about test coverage of the trust boundary.

## Conclusion

No P0/P1 issues. The PR does exactly what it claims (config/gateway/concurrency test supplement, 95→145 tests), all checks pass, and mocks faithfully mirror source. **Approving.** The P2 items above are good follow-up candidates for a future test pass but do not block this merge.

PR #11, opened 2026-06-04
test: add end-to-end smoke tests (T18)

lml2468:
## Summary

8 end-to-end smoke tests covering the complete message pipeline with mocked external dependencies and real internal components.

### Mock Strategy
- **Mocked**: Claude Agent SDK (`queryAgent`), Octo REST API (sendMessage, streamStart/Send/End, sendTyping, getGroupMembers)
- **Real**: SessionStore (`:memory:` SQLite), SessionRouter, GroupContext, StreamRelay, buildPrompt

### Test Cases

| # | Test | Verifies |
|---|------|----------|
| 1 | DM text happy path | Stream output + history storage (user + assistant) |
| 2 | Group @mention triggers | Agent called, history stored with group session key |
| 3 | Group non-@mention silent | No agent call, message cached in group context |
| 4 | Non-text DM unsupported | Replies `暂不支持此类消息`, no agent call |
| 5 | Rate limit rejection | Beyond maxPerMinute rejected, debounced reply |
| 6 | Multi-turn history | Second prompt includes first turn's history |
| 7 | Bot self-message filtered | Router filters own messages |
| 8 | Group context dedup | Current message only in `[Current message]`, not `[Group context]` |

### Pipeline Simulation
`handleMessage` is not exported from `index.ts`, so the test recreates the pipeline via a `simulateMessage()` helper that mirrors the real logic.

### Files changed (1)
- `src/__tests__/e2e.test.ts` (+396 lines)

### Verification
```
npm test          # 103 tests passed (95 existing + 8 new e2e)
```

### DIFF
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -1,0 +1,396 @@
+/**
+ * End-to-end smoke tests — simulate the complete message pipeline
+ * without real WS connections or Claude API calls.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// --- Mocks (hoisted before imports) ---
+
+vi.mock('../octo/api.js', () => ({
+  sendMessage: vi.fn().mockResolvedValue(undefined),
+  sendTyping: vi.fn().mockResolvedValue(undefined),
+  streamStart: vi.fn().mockResolvedValue('stream-001'),
+  streamSend: vi.fn().mockResolvedValue(undefined),
+  streamEnd: vi.fn().mockResolvedValue(undefined),
+  getGroupMembers: vi.fn().mockResolvedValue([]),
+  sendHeartbeat: vi.fn().mockResolvedValue(undefined),
+  registerBot: vi.fn().mockResolvedValue({
+    robot_id: 'bot-001',
+    im_token: 'token',
+    ws_url: 'ws://localhost',
+    api_url: 'https://test.example.com',
+    owner_uid: 'owner',
+    owner_channel_id: 'ch-owner',
+  }),
+  generateClientMsgNo: vi.fn().mockReturnValue('client-msg-001'),
+  fetchUserInfo: vi.fn().mockResolvedValue(null),
+}));
+
+vi.mock('../agent-bridge.js', async (importOriginal) => {
+  const original = await importOriginal<typeof import('../agent-bridge.js')>();
+  return {
+    ...original,
+    buildPrompt: original.buildPrompt,
+    queryAgent: vi.fn(),
+  };
+});
+
+// --- Imports (after mocks) ---
+
+import { SessionStore } from '../session-store.js';
+import { SessionRouter } from '../session-router.js';
+import { GroupContext } from '../group-context.js';
+import { StreamRelay } from '../stream-relay.js';
+import { createAdapter, type DbAdapter } from '../db-adapter.js';
+import { buildPrompt, queryAgent } from '../agent-bridge.js';
+import {
+  sendMessage,
+  sendTyping,
+  streamStart,
+  streamSend,
+  streamEnd,
+} from '../octo/api.js';
+import { ChannelType, MessageType } from '../octo/types.js';
+import type { BotMessage } from '../octo/types.js';
+import type { Config } from '../config.js';
+
+// --- Constants ---
+
+const BOT_ID = 'bot-001';
+const USER_UID = 'user-001';
+const GROUP_CHANNEL = 'group-ch-001';
+
+// --- Helpers ---
+
+function makeConfig(overrides?: Partial<Config>): Config {
+  return {
+    botToken: 'test-token',
+    apiUrl: 'https://test.example.com',
+    cwd: '/tmp/test-project',
+    dataDir: '/tmp/data',
+    sdk: {
+      allowedTools: ['Read', 'Glob', 'Grep'],
+      permissionMode: 'bypassPermissions',
+      settingSources: ['user'],
+    },
+    rateLimit: { maxPerMinute: 5 },
+    context: { maxContextChars: 6000, historyLimit: 40 },
+    botBlocklist: [],
+    ...overrides,
+  };
+}
+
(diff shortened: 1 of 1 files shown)


Jerry-Xin (approved):
The PR is relevant to this repository and adds useful smoke coverage for the Octo message pipeline; I found no blocking issues.

💬 Non-blocking

- 🟡 Warning: [src/__tests__/e2e.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-3j3_29by/repo/src/__tests__/e2e.test.ts:132) duplicates the private handler logic instead of exercising the production handler in [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-3j3_29by/repo/src/index.ts:70). This still has value, but future changes to the real handler can diverge from the simulated helper without these tests failing. Consider exporting the handler behind a testable module boundary.

- 🔵 Suggestion: [src/__tests__/e2e.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-3j3_29by/repo/src/__tests__/e2e.test.ts:49) imports `sendTyping` and `streamSend` but does not assert on them. Either remove the unused imports or add assertions if they are part of the intended smoke coverage.

✅ Highlights

- The PR covers the main route outcomes: direct messages, group mention gating, ignored group chatter, unsupported payloads, rate limiting, history accumulation, self-message filtering, and group-context deduplication.
- The tests use real storage, routing, context, prompt construction, and stream relay components, which gives better integration confidence than isolated unit tests.

I could not run the test file locally because dependencies are not installed in this checkout; the test runner binary was missing.

yujiawei (approved):
# Code Review — PR #11 (cc-channel-octo)

**Verdict: APPROVED** — test-only PR, no blocking issues. Two optional follow-ups noted below.

## Summary

This PR adds `src/__tests__/e2e.test.ts` (396 lines, 8 smoke tests) covering the message pipeline with mocked external dependencies (Claude Agent SDK via `queryAgent`, Octo REST API) and real internal components (`SessionStore` on `:memory:` SQLite, `SessionRouter`, `GroupContext`, `StreamRelay`, real `buildPrompt`).

Verification on the head commit (`50eb2ab`):
- ✅ Full suite: **103/103 tests pass** (8 new + 95 existing)
- ✅ `tsc --noEmit`: clean
- ✅ `eslint`: 0 errors (2 warnings, see nit #1)

## Strengths

- **Good seam choice.** Mocking only `queryAgent` and the Octo `api.js` boundary while exercising the real router/context/relay/store gives genuine integration coverage — `buildPrompt` is deliberately kept real (`agent-bridge.js` mock preserves `original.buildPrompt`), so prompt assembly is validated end-to-end.
- **Meaningful behavioral coverage.** The 8 cases hit the routing decisions that matter: DM happy path, group @mention gate (positive + negative), non-text rejection, rate-limit debounce, multi-turn history accumulation, bot self-message filtering, and the group-context dedup invariant (current message must not double-appear). Tests #4/#5/#7 correctly rely on the real `SessionRouter` behavior rather than re-stubbing it.
- **Stream lifecycle is asserted through the real relay.** `StreamRelay` is instantiated for real; `streamStart`/`streamEnd` calls confirm the streaming path actually runs.

## Findings (none blocking)

### P2 — Error-handling path is not exercised
The test helper `simulateMessage()` (test L132–210) mirrors the real `handleMessage()` (`src/index.ts:70–166`) but **omits the `try/catch` + error-reply branch** (`index.ts:139–153`). No test configures `queryAgent` to reject or `streamRelay.deliver` to fail, so the user-facing fallback (`'An error occurred while processing your message. Please try again.'`, `index.ts:148`) has zero coverage. This is a missing-coverage gap, not a correctness defect — but it's the most valuable case to add next.

*Suggested follow-up:* one test that makes `queryAgent` throw/reject and asserts `sendMessage` is called with the error reply.

### P2 — `simulateMessage()` duplicates pipeline logic instead of importing it
Because `handleMessage` is not exported from `index.ts`, the test re-implements ~78 lines of pipeline orchestration. I diffed the copy against the original: the ordering (refreshMembers → buildContext → pushMessage → buildHistoryPrefix → appendUser → buildPrompt → deliver → appendAssistant) matches faithfully, so the tests are not currently asserting anything false. The risk is **drift**: future changes to `handleMessage` won't be caught unless the copy is updated in lockstep. Consider exporting `handleMessage` and calling it directly so the test binds to the real pipeline.

### P2 — A few assertions could be tightened
- Rate-limit test (L334–337) counts reply occurrences but doesn't assert `channelId`/`channelType` via `expect.objectContaining`.
- DM happy-path (L249) uses `toContain` on a prompt substring; an exact-structure check would catch malformed/duplicated prompt assembly.

These are assertion-strength nits; the behaviors they target are covered elsewhere.

### nit — Unused imports
`sendTyping` (L49) and `streamSend` (L51) are imported but never used (the 2 eslint warnings). Drop them.

### nit — Non-deterministic `message_id`
`makeDmMsg`/`makeGroupMsg` build `message_id` from a millisecond clock (L86, L104), which can collide within a millisecond, while the rate-limit test overrides with sequential ids (L325). Verified harmless — no tested code path keys on `message_id` (routing uses `from_uid`/`channel_id`, `session-router.ts:83–88`) — but uniform sequential ids would be cleaner.

## Notes on non-issues (checked and cleared)

- **No timer leak.** `StreamRelay.deliver()` clears its `setInterval` typing heartbeat in a `finally` (`stream-relay.ts:130–132`) and clears/awaits the `setTimeout` flush in `_deliverViaStream`'s `finally` (`stream-relay.ts:231–248`) on every path. Since `deliver()` is awaited inside `simulateMessage()` (which each test awaits), no real timers survive a test — consistent with the clean run.
- **No DB resource leak.** `GroupContext` shares the same `:memory:` adapter that `afterEach` closes (test L237), and a fresh adapter + `GroupContext` is built in `beforeEach`, so no state crosses tests.

## Conclusion

Solid, well-targeted smoke-test suite that meaningfully raises confidence in the routing/context/streaming pipeline. All findings are P2/nit and do not block merge. Approving; the error-path test is worth a quick follow-up.

PR #12, opened 2026-06-05
docs: add ARCHITECTURE.md

lml2468:
## T16: ARCHITECTURE.md

Full system design documentation for cc-channel-octo. Written from actual source code analysis (12 source files, ~1200 lines).

### Contents

- **Design Goals** — independent process, minimal surface, secure by default, protocol fidelity
- **System Overview** — ASCII architecture diagram showing all modules and their relationships
- **Module Reference** — three-layer breakdown:
  - L0 Protocol: WuKongIM binary protocol, REST API, types
  - L1 Infrastructure: config, db-adapter, session-store, gateway
  - L2 Application: session-router, group-context, agent-bridge, stream-relay, index.ts
- **Data Flow** — step-by-step DM and group message processing paths
- **Concurrency Model** — per-session promise-chain locks, FIFO ordering, parallel session isolation
- **Security Model** — threat model, 7-layer mitigation table, permission mode rationale, attack surface reduction guide
- **Persistence** — SQLite schema, WAL mode, TTL, cascade deletes
- **Configuration Resolution** — three-level priority chain with validation
- **Streaming Output** — two-tier delivery (stream API primary, plain message fallback), typing heartbeat, flush throttling
- **Error Handling** — per-component recovery patterns
- **Relationship to openclaw-channel-octo** — fork strategy and sync procedure
- **Future Directions** — media, multi-bot, branching, webhook mode

### Verification

- `tsc --noEmit`: zero errors
- `npm test`: 103/103 passed
- Only file changed: `ARCHITECTURE.md` (+224 lines)
- All internal links consistent with existing README.md and CONTRIBUTING.md

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,224 @@
+# Architecture
+
+This document describes the internal architecture of cc-channel-octo: a standalone Node.js gateway that bridges [Claude Code](https://claude.ai/code) (via the [Claude Agent SDK](https://www.npmjs.com/package/@anthropic-ai/claude-agent-sdk)) to [Octo](https://github.com/nicco-io/octo) instant messaging.
+
+## Design Goals
+
+1. **Independent process.** No dependency on OpenClaw or any orchestration layer. One binary, one config file, `npm start`.
+2. **Minimal surface area.** 12 source files, ~1200 lines. Every module has a single responsibility and a clear boundary.
+3. **Secure by default.** Headless automation requires `bypassPermissions`, so security is enforced through tool whitelisting (`allowedTools`), working directory isolation (`cwd`), and setting source restriction (`settingSources`).
+4. **Protocol fidelity.** The Octo layer implements the WuKongIM binary protocol directly — DH key exchange, AES-CBC encryption, binary framing — rather than depending on a high-level SDK that may lag behind protocol changes.
+
+## System Overview
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│                     cc-channel-octo                         │
+│                                                             │
+│  ┌──────────┐   ┌────────────────┐   ┌──────────────────┐  │
+│  │  config   │──▶│    gateway      │──▶│  session-router  │  │
+│  └──────────┘   │  (WS lifecycle)  │   │  (routing +      │  │
+│                 │                  │   │   concurrency +   │  │
+│                 └────────┬─────────┘   │   rate limiting)  │  │
+│                          │             └────────┬──────────┘  │
+│                          │ BotMessage           │             │
+│                          ▼                      ▼             │
+│  ┌──────────────┐   ┌──────────┐   ┌──────────────────────┐ │
+│  │ group-context │   │ session- │   │    agent-bridge      │ │
+│  │ (member cache │   │  store   │   │  (Claude Agent SDK)  │ │
+│  │  + context    │   │ (SQLite) │   └──────────┬───────────┘ │
+│  │  window)      │   └──────────┘              │             │
+│  └──────────────┘                    AsyncIterable<string>   │
+│                                                │             │
+│                                     ┌──────────▼───────────┐ │
+│                                     │    stream-relay      │ │
+│                                     │  (typing + throttle  │ │
+│                                     │   + split + fallback)│ │
+│                                     └──────────────────────┘ │
+└─────────────────────────────────────────────────────────────┘
+         ▲                                      │
+         │ WuKongIM binary protocol             │ Octo REST API
+         │ (WebSocket)                          │ (stream/send)
+         ▼                                      ▼
+┌─────────────────────────────────────────────────────────────┐
+│                      Octo Server                            │
+└─────────────────────────────────────────────────────────────┘
+```
+
+## Module Reference
+
+The system is organized into three layers and a shared protocol package.
+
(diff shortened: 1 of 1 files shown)


Jerry-Xin (approved):
The new architecture document is relevant to this repository and broadly matches the checked source; I found only non-blocking documentation accuracy issues.

💬 Non-blocking

🟡 Warning — [ARCHITECTURE.md:217] references `CONTRIBUTING.md`, but that file is not present in the repository. The sync-procedure link will be dead for readers.

🟡 Warning — [ARCHITECTURE.md:185] says README has the full configuration reference table, but [README.md:21] only points users to `config.example.json` and `src/config.ts`. Update the target or add the table.

🔵 Suggestion — [ARCHITECTURE.md:7] says “One binary,” but the project is a Node package started via `node dist/index.js` / `npm start`, not a distributed binary. “One process” or “one package” would be more precise.

✅ Highlights

The module breakdown aligns well with the implementation, especially the session lock in [src/session-router.ts:65], persistence schema in [src/session-store.ts:28], and stream fallback behavior in [src/stream-relay.ts:250].

Verification caveat: I could not run tests or type-check locally because `vitest` and `tsc` are not installed in this checkout.

yujiawei (approved):
# Code Review — PR #12 (cc-channel-octo)

**Scope:** docs-only — adds `ARCHITECTURE.md` (+224 lines), no source changes. This PR was flagged security-sensitive, so I fact-checked every claim in the document against the source at the head SHA, with particular attention to the Security Model section.

**Verdict: APPROVED.** The document is accurate and high-quality. Every security claim checks out against the code, and the remaining issues are factual imprecisions, not blockers.

## Security Model — all claims verified ✅

I traced each row of the Security Model and Mitigations tables to source:

- Default `allowedTools` = the 8 listed tools — `src/config.ts:49`, passed through unmodified at `src/agent-bridge.ts:83`.
- `permissionMode: 'bypassPermissions'` default + rationale — `src/config.ts:50`; `allowDangerouslySkipPermissions` gated on it at `src/agent-bridge.ts:88`.
- `settingSources: ['user']` (excludes `project`) — `src/config.ts:51`, applied at `src/agent-bridge.ts:87`. This is the correct mitigation against a malicious `.claude/settings.json` in `cwd`, as the doc claims.
- Hardcoded system prompt rejecting credential/secret reads and exfiltration — `DEFAULT_SYSTEM_PROMPT` at `src/agent-bridge.ts:16-23`, applied at `:82`.
- Rate limit token bucket, default 5/min — `src/session-router.ts:16-29,150-173`; `src/config.ts:54`.
- Process lock `data/gateway.lock` via `kill(pid, 0)` — `src/gateway.ts:31,72`.
- Attack-surface subset table is internally coherent (Read-only/No-shell/No-network are valid subsets of the 8 tools).

One genuinely good design detail the doc highlights and the code backs up: `routeAndHandle` holds a single per-session lock across **both** the route decision and the handler execution (`src/session-router.ts:65-76`), closing the TOCTOU gap. Confirmed.

## Findings

### P2 — Line count is off by ~2.5×
`Design Goals #2` states *"12 source files, ~1200 lines."* The file count is correct, but the actual non-test source is **2,978 lines** (`socket.ts` alone is 740). "Minimal surface area" is a security selling point, so an accurate figure matters here. Suggest "~3,000 lines" (or "~1,700 excluding the forked protocol layer", if the intent was to count only project-authored code — worth stating explicitly which).

### P2 — Broken link to `CONTRIBUTING.md`
Line ~217 links to `./CONTRIBUTING.md` for the protocol sync procedure, but that file does not exist in the repo. Either add the file or inline the procedure / drop the link. (The `./README.md` link is fine — it exists.)

### Nits (non-blocking)
- **`parseOctoJson` scope** — doc says "16+ digit numeric IDs"; the regex only rewrites the `message_id` field (`src/octo/api.ts:37-43`), not all numeric IDs. The 16+ threshold itself is exact. Suggest "the `message_id` field" for precision.
- **Routing pipeline list is non-exhaustive** — between the mention gate and rate limiting there is a system-event filter (`if (msg.payload.event) return null;`, `src/session-router.ts:127`) that drops join/leave events. The documented order is correct; the list just omits this stage.
- **Bot blocklist default** — doc says "Empty by default"; the field is actually left `undefined` (`src/config.ts` defaults omit it), which is behaviorally equivalent (`?? false` at `src/session-router.ts:91`). Fine as-is, just technically "unset" not "empty list".
- **PR description: "103/103 tests passed"** — a static scan of the 4 spec files counts 95 top-level `it`/`test` cases with no table-driven `.each`. Not a doc issue, but worth a maintainer recount in case the number is stale.

## Verified accurate (sample)
Protocol layer (curve25519 DH, AES-128-CBC, salt-from-CONNACK IV, 1-byte nibble header, 7-bit varint framing, packet-type enum, ±25% jitter backoff, 64-bit big-endian IDs), persistence (WAL, `foreign_keys=ON`, 5s busy_timeout, 7-day TTL, cascade delete, hourly member refresh), config (3-level resolution, strict integer parsing, `_`-prefix stripping, exactly 14 env-overridable fields), and streaming (full-accumulated-text flushes, base64, 800ms throttle, `isFlushing` mutex, split priority, 3500-char segments, 5s typing heartbeat) all match source exactly.

The two P2 items are easy follow-ups and do not block merge. Nice, well-grounded documentation.

PR #13, opened 2026-06-05
fix(security): replace Math.random DH seed with crypto.randomBytes + fix spread stack overflow (Q4+Q5)

lml2468:
### Q4: DH key seed — crypto.randomBytes(32) instead of Math.random()

`generateDeviceID()` used `Math.random()` to generate the seed for the Curve25519 DH key pair. `Math.random()` is not a cryptographically secure PRNG — its internal state can be predicted from observed outputs. Since the device ID (derived from the same PRNG) is transmitted in cleartext in the CONNECT packet, an attacker observing WS traffic could potentially recover the PRNG state and derive the DH private key, breaking all payload encryption.

**Fix:** Use `crypto.randomBytes(32)` from Node.js `node:crypto` for the DH seed. `generateDeviceID()` is still used for the device ID string (non-security-critical).

### Q5: Spread/fromCharCode stack overflow on large payloads

Six sites in socket.ts used `...spread` or `String.fromCharCode(...array)` patterns that hit the V8 maximum call stack size (~10K arguments) on large payloads:

| Site | Before | After |
|------|--------|-------|
| `Encoder.writeBytes` | `this.w.push(...b)` | for loop |
| `Encoder.writeString` | `this.w.push(...arr)` | for loop |
| `stringToUint` | `unescape(encodeURIComponent())` | `TextEncoder` |
| `uintToString` | `String.fromCharCode(...array)` | `TextDecoder` |
| `aesDecrypt` | `String.fromCharCode(...Array.from(data))` | `Buffer.toString('binary')` |
| `handleRawData` | `tempBuffer.push(...Array.from(data))` | for loop |

### Tests

3 new regression tests in `protocol.test.ts`:
- `writeBytes` with 100K bytes
- `writeString/readString` round-trip with 60K UTF-8 bytes (20K CJK chars)
- `writeString/readString` round-trip with 65K ASCII bytes

### Verification

- `tsc --noEmit`: ✅ clean
- `eslint src/octo/socket.ts`: 0 errors (13 pre-existing warnings)
- `vitest run`: 153 tests pass (40 protocol tests = 37 existing + 3 new)

**Scope:** Only `src/octo/socket.ts` and `src/__tests__/octo/protocol.test.ts` modified.

### DIFF
--- a/src/__tests__/octo/protocol.test.ts
+++ b/src/__tests__/octo/protocol.test.ts
@@ -569,3 +569,35 @@ describe("Edge cases", () => {
     expect(remaining.length).toBe(0);
   });
 });
+
+// ─── 6. Large-Payload Safety (Q5 regression) ───────────────────────────────
+
+describe("Large-payload safety (>64K, no stack overflow)", () => {
+  it("Encoder.writeBytes handles >64K bytes without stack overflow", () => {
+    const enc = new Encoder();
+    const largeArray = new Array(100_000).fill(0x42);
+    enc.writeBytes(largeArray);
+    const result = enc.toUint8Array();
+    expect(result.length).toBe(100_000);
+    expect(result[0]).toBe(0x42);
+    expect(result[99_999]).toBe(0x42);
+  });
+
+  it("Encoder.writeString / Decoder.readString round-trip for large UTF-8 string", () => {
+    const enc = new Encoder();
+    // 20K CJK chars = ~60K UTF-8 bytes, fits in Int16 length prefix (max 65535)
+    const longStr = "中".repeat(20_000);
+    enc.writeString(longStr);
+    const dec = new Decoder(enc.toUint8Array());
+    expect(dec.readString()).toBe(longStr);
+  });
+
+  it("Encoder.writeString / Decoder.readString round-trip for >64K ASCII", () => {
+    const enc = new Encoder();
+    // 65000 ASCII chars = 65000 bytes, still fits in Int16 (max 65535)
+    const longStr = "A".repeat(65_000);
+    enc.writeString(longStr);
+    const dec = new Decoder(enc.toUint8Array());
+    expect(dec.readString()).toBe(longStr);
+  });
+});
--- a/src/octo/socket.ts
+++ b/src/octo/socket.ts
@@ -8,6 +8,7 @@ import { generateKeyPair, sharedKey } from "curve25519-js";
 import { Buffer } from "buffer";
 import CryptoJS from "crypto-js";
 import { Md5 } from "md5-typescript";
+import { randomBytes } from "node:crypto";
 import type { BotMessage, MessagePayload } from "./types.js";
 
 // ─── WuKongIM Binary Protocol Constants ─────────────────────────────────────
@@ -31,7 +32,7 @@ const PROTO_VERSION = 4;
 export class Encoder {
   private w: number[] = [];
   writeByte(b: number) { this.w.push(b & 0xff); }
-  writeBytes(b: number[]) { this.w.push(...b); }
+  writeBytes(b: number[]) { for (let i = 0; i < b.length; i++) this.w[this.w.length] = b[i]; }
   writeInt16(b: number) { this.w.push((b >> 8) & 0xff, b & 0xff); }
   writeInt32(b: number) { this.w.push((b >> 24) & 0xff, (b >> 16) & 0xff, (b >> 8) & 0xff, b & 0xff); }
   writeInt64(n: bigint) {
@@ -44,7 +45,7 @@ export class Encoder {
     if (s && s.length > 0) {
       const arr = stringToUint(s);
       this.writeInt16(arr.length);
-      this.w.push(...arr);
+      for (let i = 0; i < arr.length; i++) this.w[this.w.length] = arr[i];
     } else {
       this.writeInt16(0);
     }
@@ -121,15 +122,11 @@ export class Decoder {
 }
 
 function stringToUint(str: string): number[] {
-  const encoded = unescape(encodeURIComponent(str));
-  const arr: number[] = [];
-  for (let i = 0; i < encoded.length; i++) arr.push(encoded.charCodeAt(i));
-  return arr;
+  return Array.from(new TextEncoder().encode(str));
 }
 
 function uintToString(array: number[]): string {
-  const encoded = String.fromCharCode(...array);
-  return decodeURIComponent(escape(encoded));
+  return new TextDecoder().decode(new Uint8Array(array));
 }
 
 function encodeVariableLength(len: number): number[] {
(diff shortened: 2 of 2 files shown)


Jerry-Xin (approved):
This PR is in scope for `cc-channel-octo`: it directly fixes the Octo socket protocol’s DH seed generation and large-payload protocol encoding paths.

💬 Non-blocking

🟡 Warning — [src/octo/socket.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-5q12_ypz/repo/src/octo/socket.ts:128): `TextDecoder()` defaults to replacement-mode decoding, so malformed UTF-8 in protocol strings is now silently converted to `U+FFFD`. The previous `decodeURIComponent(...)` path would throw. If invalid protocol UTF-8 should reject the packet, use `new TextDecoder("utf-8", { fatal: true })`.

🔵 Suggestion — [src/octo/socket.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-5q12_ypz/repo/src/octo/socket.ts:360): the `randomBytes(32)` change is the main security fix, but there is no regression test around DH seed generation. A small mocked WebSocket/crypto test would help prevent accidental reintroduction of `Math.random()` for key material.

🔵 Suggestion — [src/__tests__/octo/protocol.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-5q12_ypz/repo/src/__tests__/octo/protocol.test.ts:595): the test title says `>64K ASCII`, while `65_000` is below 64 KiB (`65_536`). Since the protocol length prefix cannot encode strings above `65_535` bytes anyway, consider naming this “near 64K ASCII” to avoid confusion.

✅ Highlights

The security fix correctly uses Node CSPRNG output for Curve25519 seed material, while keeping `generateDeviceID()` limited to non-secret device ID generation.

The spread removal addresses the stack-overflow class without changing the protocol byte layout for valid inputs.

`git diff --check` is clean. I could not run tests or type-check locally because dependencies are not installed in this checkout (`vitest` and `tsc` were not found).

lml2468 (commented):
## QA Review: APPROVED ✅

Q4 + Q5 fixes verified — all 6 spread/fromCharCode sites eliminated, DH seed switched to crypto.randomBytes(32). Checked head branch source (737 lines) — zero residual `String.fromCharCode(...`, `push(...spread)`, or `Math.random()` seed patterns.

### Verified per site:

1. **`Encoder.writeBytes`** — `push(...b)` → index assignment loop ✅
2. **`Encoder.writeString`** — same pattern ✅
3. **`stringToUint`** — deprecated `unescape(encodeURIComponent())` → `TextEncoder` ✅
4. **`uintToString`** — `String.fromCharCode(...array)` → `TextDecoder` ✅
5. **`aesDecrypt`** — `String.fromCharCode(...Array.from(data))` → `Buffer.toString("binary")` ✅ (latin1 equivalent, correct for base64 ciphertext)
6. **`handleRawData`** — `tempBuffer.push(...Array.from(data))` → for loop ✅
7. **DH seed** — `Uint8Array.from(stringToUint(generateDeviceID()))` → `randomBytes(32)` ✅

Tests: 3 regression tests cover >64K writeBytes, 60K UTF-8 round-trip, 65K ASCII round-trip.

### 💬 Non-blocking notes

- `writeBytes` uses `this.w[this.w.length] = b[i]` while `handleRawData` uses `this.tempBuffer.push(data[i])` — both correct, minor style inconsistency.
- `stringToUint` creates an intermediate `Array.from(Uint8Array)` copy — could return `Uint8Array` directly and adapt `Encoder` to accept it, but that is a broader refactor, not this PR scope.

✅ Scope clean (socket.ts + protocol.test.ts only). No unrelated changes.

— 郭守卫 QA

yujiawei (approved):
# Code Review — PR #13 (cc-channel-octo)

**Verdict: APPROVED.** Both fixes are correct, minimal, well-scoped, and backed by regression tests. No P0/P1 issues. I independently verified the security claims and the byte-level equivalence of every refactored site.

## Scope verified
- 2 files changed: `src/octo/socket.ts` (+8/-11), `src/__tests__/octo/protocol.test.ts` (+32).
- `tsc --noEmit`: clean. `eslint src/octo/socket.ts`: 0 errors (13 pre-existing warnings, all unrelated `_unused`/`any`). Full test suite: 156 pass; `protocol.test.ts`: 40 pass (37 existing + 3 new).

## Q4 — CSPRNG for the DH seed ✅

`socket.ts:360` now seeds the Curve25519 key pair with `randomBytes(32)` from `node:crypto` instead of deriving it from `generateDeviceID()` (which is `Math.random()`-based).

This is the right fix and the threat model in the description is sound: the device ID is sent in cleartext in the CONNECT packet, and when the DH seed was derived from the same `Math.random()` stream, an observer could in principle recover PRNG state and reconstruct the private key, defeating payload encryption. Decoupling the seed from any transmitted value closes that.

Verified:
- `curve25519-js` `generateKeyPair` requires **exactly** a 32-byte array (`if (seed.length !== 32) throw`). `randomBytes(32)` satisfies this; a `Buffer` is a `Uint8Array`, so `checkArrayTypes` passes. Confirmed at runtime — key pair generates with 32-byte public/private outputs.
- `generateDeviceID()` is correctly retained only for the non-security-critical device-ID string (`socket.ts:365`). With the seed now independent, the device ID's predictability is no longer security-relevant.

**For human verification (security-sensitive):** please confirm there is no out-of-band requirement that the DH seed be reproducible/derivable from the device ID (none is apparent in this codebase, and the WuKongIM handshake derives the AES key purely from the DH shared secret + server salt, so a random seed is correct). Behavior is unchanged on the wire — only the seed's entropy source changed.

## Q5 — Spread / fromCharCode stack-overflow fixes ✅

All six sites avoid passing large arrays as variadic arguments (the ~10K-arg V8 call-stack limit). I checked each for behavioral equivalence, not just compilation:

- `Encoder.writeBytes` / `writeString` — `this.w.push(...x)` → index-assignment loop. Appends in identical order; no masking was applied before or after, so byte output is unchanged.
- `stringToUint` — `unescape(encodeURIComponent())` → `TextEncoder().encode()`. Verified byte-identical across ASCII, CJK, astral-plane emoji, mixed, empty, and Latin-1 inputs.
- `uintToString` — `String.fromCharCode(...)` → `TextDecoder().decode()`. Round-trips identically for the same input set.
- `aesDecrypt` — `String.fromCharCode(...Array.from(data))` → `Buffer.from(data).toString("binary")`. Verified identical for bytes 0–255 (latin1/binary maps each byte 1:1).
- `handleRawData` — `tempBuffer.push(...Array.from(data))` → per-element loop. Same result, no stack risk.

The 3 new tests (100K-byte `writeBytes`, 60K-byte CJK round-trip, 65K ASCII round-trip) directly exercise the previously-overflowing paths and pass.

## Non-blocking notes (P2 / informational)

1. **Int16 length cap is unchanged (pre-existing).** `writeString` writes `arr.length` via `writeInt16`, so any UTF-8 payload >65535 bytes silently truncates the length prefix (e.g. 70000 → 4464) and corrupts the frame. The new tests deliberately stay at/below 65535, which is correct for this protocol, but the encoder still does not guard against an over-length string. Out of scope for this PR; worth a follow-up assertion (`if (arr.length > 0xffff) throw`) if oversized payloads are ever possible.
2. **`this.w[this.w.length] = x` idiom** is functionally equivalent to `push` and avoids the spread; fine to keep. A plain `this.w.push(b[i])` would read more conventionally with no measurable cost, but this is purely stylistic.
3. `tempBuffer` remains a `number[]` accumulated byte-by-byte (pre-existing design). Not introduced here; no action needed.

## Conclusion

Correct, surgical security hardening with good test coverage. Approving.

PR #14, opened 2026-06-05
fix(Q2): add default 30s timeout to postJson — prevent session deadlock

lml2468:
### Problem

`postJson` passed the caller's `signal` directly to `fetch` without a fallback. All callers (`sendMessage`, `sendTyping`, `sendHeartbeat`, `stream*`) pass no signal. A hanging Octo API (network partition, DNS timeout, unresponsive server) blocks the session's promise chain indefinitely → all subsequent messages for that session queue forever.

### Fix

```typescript
const effectiveSignal = signal ?? AbortSignal.timeout(DEFAULT_TIMEOUT_MS);
```

One line in `postJson` ensures every call gets a 30s timeout (`DEFAULT_TIMEOUT_MS`, already defined but unused) unless the caller provides their own signal. Caller-provided signals are respected unchanged.

Also applied the same fix to `getChannelMessages` (dead code, but consistent).

### Impact

- `sendMessage`: now protected ✅
- `sendTyping`: now protected ✅  
- `sendHeartbeat`: now protected ✅
- `streamStart/Send/End`: now protected ✅
- `registerBot`: already passes signal from caller → unchanged ✅
- `getGroupMembers`: already has `AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` → unchanged ✅
- `fetchUserInfo`: already has `AbortSignal.timeout(5000)` → unchanged ✅

### Tests

8 new tests in `api-timeout.test.ts`:
1. Default timeout applied when no signal provided
2. Caller-provided signal respected
3. Abort error when timeout fires
4. Correct headers including Authorization
5. URL trailing slash normalization
6. Error response with status and body
7. Empty response returns undefined
8. int64 message_id JSON protection

### Verification

- `tsc --noEmit`: zero errors
- `vitest run`: 161/161 passed (153 existing + 8 new)
- Only `src/octo/api.ts` changed (+3 lines / -2 lines)

### DIFF
--- a/src/__tests__/api-timeout.test.ts
+++ b/src/__tests__/api-timeout.test.ts
@@ -1,0 +1,164 @@
+/**
+ * Tests for postJson default timeout behavior (Q2 fix).
+ *
+ * Verifies that postJson applies a 30s default timeout when no signal is provided,
+ * and respects caller-provided signals when explicitly passed.
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { postJson } from "../octo/api.js";
+
+// Mock the global fetch before any postJson calls.
+const mockFetch = vi.fn();
+vi.stubGlobal("fetch", mockFetch);
+
+describe("postJson default timeout", () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+  });
+
+  it("applies default timeout when no signal is provided", async () => {
+    mockFetch.mockResolvedValue({
+      ok: true,
+      text: () => Promise.resolve('{"result": "ok"}'),
+    });
+
+    await postJson("https://api.example.com", "test-token", "/v1/bot/test", {
+      key: "value",
+    });
+
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+    const options = mockFetch.mock.calls[0][1] as RequestInit;
+    // After Q2 fix, postJson should always pass a signal (30s default)
+    expect(options.signal).toBeDefined();
+    expect(options.signal).toBeInstanceOf(AbortSignal);
+    expect((options.signal as AbortSignal).aborted).toBe(false);
+  });
+
+  it("uses caller-provided signal when explicitly passed", async () => {
+    mockFetch.mockResolvedValue({
+      ok: true,
+      text: () => Promise.resolve('{"result": "ok"}'),
+    });
+
+    const controller = new AbortController();
+    await postJson(
+      "https://api.example.com",
+      "test-token",
+      "/v1/bot/test",
+      { key: "value" },
+      controller.signal,
+    );
+
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+    const options = mockFetch.mock.calls[0][1] as RequestInit;
+    expect(options.signal).toBe(controller.signal);
+  });
+
+  it("rejects with abort error when timeout fires", async () => {
+    const shortSignal = AbortSignal.timeout(1);
+
+    mockFetch.mockImplementation(
+      (_url: string, opts: RequestInit) =>
+        new Promise((resolve, reject) => {
+          const timer = setTimeout(
+            () => resolve({ ok: true, text: () => "" }),
+            60_000,
+          );
+          opts?.signal?.addEventListener("abort", () => {
+            clearTimeout(timer);
+            reject(new DOMException("The operation was aborted.", "AbortError"));
+          });
+        }),
+    );
+
+    await expect(
+      postJson(
+        "https://api.example.com",
+        "test-token",
+        "/v1/bot/test",
+        {},
+        shortSignal,
+      ),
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for `cc-channel-octo` and correctly bounds Octo API `fetch` calls to prevent indefinite per-session queue blocking.

💬 Non-blocking
- 🔵 Suggestion `src/__tests__/api-timeout.test.ts:20`: The default-timeout test confirms a signal is passed, but not that `AbortSignal.timeout()` receives `30_000`. Consider spying on `AbortSignal.timeout` to assert the exact default duration. This is test precision only, not a merge blocker.

✅ Highlights
- `src/octo/api.ts:53`: Centralizing the fallback timeout in `postJson` covers `sendMessage`, `sendTyping`, `sendHeartbeat`, and stream calls without changing caller behavior.
- `src/octo/api.ts:53`: Caller-provided signals are preserved via `signal ?? ...`, which keeps explicit cancellation semantics intact.
- `src/octo/api.ts:269`: Applying the same default to `getChannelMessages` is consistent with the rest of the API layer.
- Verification passed locally: `npm run type-check` and `npm test` both succeeded, with `161` tests passing.

yujiawei (approved):
# Code Review — PR #14 (cc-channel-octo)

## Verdict: APPROVED

A clean, minimal, and well-targeted fix for a real availability bug. Verified locally on head `bb313f7`: `tsc --noEmit` passes with zero errors and `vitest run` is green at 161/161 (153 existing + 8 new).

## Summary

`postJson` previously forwarded the caller's optional `signal` straight to `fetch` with no fallback. Every production caller (`sendMessage`, `sendTyping`, `sendHeartbeat`, `streamStart/Send/End`) passes **no** signal, so a hung Octo endpoint (network partition, DNS stall, unresponsive server) would leave the `fetch` promise pending forever and stall the session's promise chain. The one-line `signal ?? AbortSignal.timeout(DEFAULT_TIMEOUT_MS)` closes this correctly.

## Verification of claims

- ✅ **All non-signal callers now protected.** Confirmed each call site passes no signal: `sendMessage` (`stream-relay.ts:124`, `index.ts:143`, `session-router.ts:188`), `sendHeartbeat` (`gateway.ts:202`), and the three `stream*` functions (`api.ts:353/378/399`, which don't even expose a `signal` param). All now inherit the 30s default. ✅
- ✅ **Caller-provided signals respected unchanged.** `registerBot` forwards `params.signal`; the `?? ` fallback only kicks in when it's `undefined`, so an explicit signal is passed through untouched. Test `uses caller-provided signal when explicitly passed` asserts this. ✅
- ✅ **`getChannelMessages` consistency fix** (`api.ts:269`) matches the existing `getGroupMembers`/`fetchUserInfo` timeout pattern. Dead code today, but harmless and consistent. ✅
- ✅ **`DEFAULT_TIMEOUT_MS` was already declared** (`api.ts:16`) and is now actually used in `postJson` — no new constant, no magic number. ✅
- ✅ **Test suite is meaningful**, not just smoke: it covers the default-applied path, the caller-override path, abort-on-timeout, headers, URL normalization, error propagation, empty-body handling, and int64 `message_id` protection. ✅

## Non-blocking notes (nits / future hardening)

1. **`AbortSignal.timeout` does not bound response-body reads.** The timeout aborts connection + waiting for response headers, but once headers arrive the subsequent `await response.text()` (`api.ts:69`) is not covered by the same signal. A server that returns headers and then stalls the body stream could still hang. This is the same limitation in the pre-existing `getGroupMembers`/`fetchUserInfo` code, so it's not a regression — just worth noting for a future "true total deadline" pass if body-stall is a realistic failure mode for Octo.

2. **A caller-provided signal fully opts out of the default timeout.** Today no caller passes a long-lived cancellation signal, so this is purely latent. But if a future caller passes an `AbortController` signal used only for cancellation (one that may never fire on its own), it silently loses the 30s safety net. If you want belt-and-suspenders, `AbortSignal.any([signal, AbortSignal.timeout(DEFAULT_TIMEOUT_MS)])` (Node ≥20.3 / available on Node 22, which this repo targets) would combine both. Not required for merge.

3. **Test nit:** in `rejects with abort error when timeout fires`, the mock's success branch resolves `{ ok: true, text: () => "" }` where `text` returns a string rather than a Promise — fine because that branch is never taken (the 60s timer is cleared by the abort), but it's slightly inconsistent with the other mocks. Cosmetic only.

None of the above block merge. The fix is correct, scoped to the smallest viable change, and adequately tested.

lml2468 (commented):
QA Review: APPROVED. Core fix verified — effectiveSignal = signal ?? AbortSignal.timeout(DEFAULT_TIMEOUT_MS), all callers protected. 8 new tests also serve as first direct postJson tests (int64 protection, headers, URL normalization). Scope clean. — guoshouwei

PR #20, opened 2026-06-05
fix: P1 batch — context persistence, surrogate safety, empty reply, segment resilience (Q15, Q16, Q21, Q22, Q23)

lml2468:
## Summary

Five P1 fixes in one PR, covering stream relay reliability and group context persistence.

### Q15 — Group message context persistence (`group-context.ts`)
- New `group_messages` SQLite table with index on `(channel_id, id DESC)`
- `pushMessage()` now persists to DB alongside in-memory cache
- DB trimmed to 2× window size (200 rows per channel) on each write
- New `loadMessagesFromDb(channelId)` method restores messages from DB
- Proper snake_case → camelCase mapping for DB rows

### Q16 — Load persisted data on startup (`group-context.ts` + `index.ts`)
- New `loadAllFromDb()` method queries all distinct group_ids and loads both members and messages
- Called once in `index.ts` during initialization, immediately after GroupContext creation
- Activates the previously-dead `loadMembersFromDb()` method

### Q21 — splitMessage surrogate pair safety (`stream-relay.ts`)
- Hard cut now checks if the character at `splitAt - 1` is a high surrogate (0xD800–0xDBFF)
- If so, backs up one code unit to keep the surrogate pair intact
- Prevents invalid Unicode in message segments

### Q22 — Empty reply feedback (`index.ts`)
- When agent produces no output, sends `[No response generated. Please try rephrasing your question.]`
- Previously, users saw typing indicator flash then nothing

### Q23 — Segment failure resilience (`stream-relay.ts`)
- Per-segment try/catch in `deliver()` — one `sendMessage` failure logs the error and continues
- Previously, first segment failure dropped all remaining segments

## Tests Added
- `splitMessage`: surrogate pair hard cut test
- `StreamRelay`: segment failure continuation test
- `GroupContext`: message persistence, loadMessagesFromDb restoration, loadAllFromDb full restore

## Verification
- `npx tsc --noEmit` — zero errors
- `npx vitest run` — 188/188 tests pass (9 test files)

## Files Changed
- `src/stream-relay.ts` — Q21 + Q23
- `src/group-context.ts` — Q15 + Q16
- `src/index.ts` — Q16 + Q22
- `src/__tests__/stream-relay.test.ts` — Q21 + Q23 tests
- `src/__tests__/group-context.test.ts` — Q15 + Q16 tests

### DIFF
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -215,4 +215,58 @@ describe('GroupContext', () => {
     expect(name2).toBe('Alice');
     expect(fetchUserInfo).toHaveBeenCalledTimes(1);
   });
+
+  // --- Q15: pushMessage persists to SQLite ---
+
+  it('pushMessage persists messages to group_messages table', () => {
+    ctx.pushMessage('ch1', 'u1', 'Alice', 'hello world', 1000);
+    ctx.pushMessage('ch1', 'u2', 'Bob', 'hi there', 2000);
+
+    const rows = adapter.prepare(
+      'SELECT from_uid, from_name, content, timestamp FROM group_messages WHERE channel_id = ? ORDER BY id',
+    ).all('ch1') as Array<{ from_uid: string; from_name: string; content: string; timestamp: number }>;
+    expect(rows.length).toBe(2);
+    expect(rows[0].from_uid).toBe('u1');
+    expect(rows[0].content).toBe('hello world');
+    expect(rows[1].from_uid).toBe('u2');
+  });
+
+  it('loadMessagesFromDb restores messages after fresh GroupContext creation', () => {
+    // Push messages with first context
+    ctx.pushMessage('ch1', 'u1', 'Alice', 'msg1', 1000);
+    ctx.pushMessage('ch1', 'u2', 'Bob', 'msg2', 2000);
+
+    // Create a fresh GroupContext with same adapter
+    const ctx2 = new GroupContext(adapter, 6000);
+    ctx2.loadMessagesFromDb('ch1');
+
+    // Should restore messages and produce context
+    const context = ctx2.buildContext('ch1');
+    expect(context).toContain('Alice：msg1');
+    expect(context).toContain('Bob：msg2');
+  });
+
+  // --- Q16: loadAllFromDb ---
+
+  it('loadAllFromDb loads members and messages for all groups', () => {
+    // Setup: push data to two different groups
+    ctx.pushMessage('ch1', 'u1', 'Alice', 'hello', 1000);
+    ctx.pushMessage('ch2', 'u2', 'Bob', 'world', 2000);
+    ctx.learnMember('ch1', 'u1', 'Alice');
+    ctx.learnMember('ch2', 'u2', 'Bob');
+
+    // Create a fresh context and loadAllFromDb
+    const ctx2 = new GroupContext(adapter, 6000);
+    ctx2.loadAllFromDb();
+
+    // Members should be restored
+    expect(ctx2.getName('u1', 'ch1')).toBe('Alice');
+    expect(ctx2.getName('u2', 'ch2')).toBe('Bob');
+
+    // Messages should be restored
+    const context1 = ctx2.buildContext('ch1');
+    expect(context1).toContain('Alice：hello');
+    const context2 = ctx2.buildContext('ch2');
+    expect(context2).toContain('Bob：world');
+  });
 });
--- a/src/__tests__/stream-relay.test.ts
+++ b/src/__tests__/stream-relay.test.ts
@@ -76,6 +76,19 @@ describe("splitMessage", () => {
     const segments = splitMessage("abc", 1);
     expect(segments).toEqual(["a", "b", "c"]);
   });
+
+  it("does not split surrogate pairs on hard cut", () => {
+    // 😀 = U+1F600 = 😀 (2 code units)
+    const emoji = "😀";
+    // Fill to force a hard cut right at a surrogate pair boundary
+    const filler = "a".repeat(9); // 9 + 2 = 11 code units, maxChars=10 would cut inside emoji
+    const text = filler + emoji;  // 11 code units
+    const segments = splitMessage(text, 10);
+    // Should NOT split the surrogate pair — back up to 9
+    expect(segments[0]).toBe(filler);
+    expect(segments[1]).toBe(emoji);
+    expect(segments.join("")).toBe(text);
+  });
 });
 
 // ─── Shared mock state via vi.hoisted ───────────────────────────────────────
@@ -206,7 +219,8 @@ describe("StreamRelay", () => {
     await vi.runAllTimersAsync();
     await guarded;
 
-    await expect(promise).rejects.toThrow("sendMessage failed");
+    // Q23: sendMessage failures are swallowed per-segment; deliver() should not reject.
+    await expect(promise).resolves.toBeUndefined();
 
     const countAfter = mockState.calls.filter((c) => c.fn === "sendTyping").length;
     await vi.advanceTimersByTimeAsync(30_000);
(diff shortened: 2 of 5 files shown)


Jerry-Xin (changes requested):
The PR is in scope for `cc-channel-octo`, but one edge-case regression in `splitMessage` can hang indefinitely.

**🔴 Blocking**

- 🔴 Critical: `splitMessage()` can enter an infinite loop when `maxChars === 1` and the next code point is a surrogate pair. In [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-i4czh9fr/repo/src/stream-relay.ts:79), `splitAt` backs up from `1` to `0`; then [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-i4czh9fr/repo/src/stream-relay.ts:85) pushes an empty segment and [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-i4czh9fr/repo/src/stream-relay.ts:86) leaves `remaining` unchanged. Repro: `splitMessage("😀", 1)`. Since `maxChars = 1` is explicitly allowed and tested for ASCII, the surrogate fix must guarantee forward progress, for example by special-casing `splitAt === 0` or rejecting impossible surrogate-safe limits.

**💬 Non-blocking**

- 🟡 Warning: `loadAllFromDb()` discovers channels only from `group_members`, so persisted rows in `group_messages` are not restored if the member row is missing or was not written successfully. See [src/group-context.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-i4czh9fr/repo/src/group-context.ts:292). Consider unioning distinct IDs from both `group_members.group_id` and `group_messages.channel_id`.

- 🟡 Warning: `StreamRelay.deliver()` now swallows every `sendMessage` failure, and [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-i4czh9fr/repo/src/index.ts:136) still stores the full assistant response even if zero segments were delivered. That can leave the user with no visible response while history advances. Consider tracking successful segment count and surfacing an error if all sends fail.

- 🔵 Suggestion: The new empty-reply behavior in [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-i4czh9fr/repo/src/index.ts:138) is not covered by the E2E pipeline test helper, which still omits that branch. Add a test where `queryAgent` yields no chunks and assert the fallback message is sent.

**✅ Highlights**

- ✅ The new DB row mapping in `loadMessagesFromDb()` correctly converts `snake_case` columns to the in-memory `GroupMessage` shape.
- ✅ Per-segment send continuation is covered by a focused test.
- ✅ The PR is relevant to this repository’s stream relay and group context persistence behavior.

Verification note: I attempted `npm run type-check` and `npm test`, but this checkout does not have dependencies installed locally (`tsc` and `vitest` were not found).

lml2468:
**Review: APPROVED** ✅ (齐静春)

1 commit, 5 files, +195/-9. tsc clean, 188/188 tests pass.

**Q15 — Message persistence**: `group_messages` table with `CREATE IF NOT EXISTS` (migration-safe). `pushMessage` writes to DB + trims to 2x window. Index on `(channel_id, id DESC)` supports the `ORDER BY id DESC LIMIT` query.

**Q16 — loadAllFromDb**: Queries distinct `group_id` from `group_members`, loads both members and messages per group. `loadMessagesFromDb` correctly skips if live data exists (`existing.length > 0`). Called once at startup in `index.ts`.

**Q21 — Surrogate pair safety**: Checks `charCodeAt(splitAt - 1)` for high surrogate range (`0xD800-0xDBFF`), backs up one code unit. Verified with REPL: filler + emoji case, no-surrogate case, low-surrogate-at-boundary case — all correct.

**Q22 — Empty reply**: `[No response generated. Please try rephrasing your question.]` sent when `fullResponse` is empty. Positioned correctly after the stream relay `deliver()` completes.

**Q23 — Segment resilience**: Per-segment `try/catch` in `deliver()`. Failed segment logged + skipped, remaining segments continue. Test verifies all 3 segments attempted even when segment 2 throws.

One performance nit (non-blocking): `deleteOldMessages` runs a subquery on every `pushMessage`. Could be deferred to periodic cleanup instead. Not urgent — 100-message windows keep it bounded.

yujiawei (approved):
# Code Review — PR #20 (cc-channel-octo)

## Summary

Five focused P1 fixes (Q15, Q16, Q21, Q22, Q23) across stream relay reliability and group-context persistence. I read the full surrounding source (not just the diff), built it, and ran the suite locally:

- `npx tsc --noEmit` → **0 errors**
- `npx vitest run` → **193/193 pass** (10 files)
- `eslint` on changed files → **clean**

All five fixes are correct for production usage. No P0/P1 blockers found. A few low-priority hardening notes below.

## Verification by item

| Item | Verdict | Notes |
|---|---|---|
| Q15 — message persistence | ✅ | New `group_messages` table + index; `pushMessage` persists and trims to `2× window` (200/channel). DB-side cap (200) ≥ in-memory cap (100), consistent. |
| Q16 — load on startup | ✅ | `loadAllFromDb()` wired once in `index.ts:49`, before gateway start; restores members + messages. `loadMessagesFromDb` loads `LIMIT 100` matching the live window cap. |
| Q21 — surrogate safety | ✅ | Hard-cut path backs up one code unit when `charCodeAt(splitAt-1)` is a high surrogate (0xD800–0xDBFF). Correct: in the loop `remaining.length > maxChars` always holds, so the low half exists at `[maxChars]`. Natural-boundary splits (`\n\n`/`\n`/space) are BMP and can't bisect a pair. |
| Q22 — empty reply feedback | ✅ | When the agent yields nothing, `index.ts:138-147` sends a user-visible placeholder. No duplication with `deliver()` (which sends nothing on empty). |
| Q23 — segment resilience | ✅ | Per-segment `try/catch` in `deliver()` logs and continues; the updated test asserting `resolves.toBeUndefined()` matches. |

## Findings (all P2 / non-blocking)

**P2-1 — Q23: total-send-failure is now silent, and history records an undelivered reply.**
`stream-relay.ts:137-140` now swallows every `sendMessage` error. Previously the first failure propagated to `handleMessage`'s catch (`index.ts:149`), which sent an "An error occurred…" reply. Now if *all* segments fail, the user sees nothing — only a console log. Separately, `collected`/`fullResponse` is still recorded via `store.appendAssistant` (`index.ts:135-137`) even when no segment was delivered, so session history can claim a reply the user never received. The partial-success goal is sound; consider tracking whether ≥1 segment succeeded and, if zero succeeded, surfacing a failure to the user / skipping the history append. Suggestion, not a blocker.

**P2-2 — Q16: group discovery keys off `group_members` only.**
`loadAllFromDb` enumerates `SELECT DISTINCT group_id FROM group_members` (`group-context.ts:292-293`). A channel that has rows in `group_messages` but no `group_members` row (e.g. a message whose `from_uid` is empty, so `learnMember` early-returns at line 125 but `insertMessage` still runs) won't have its messages restored. In practice every `pushMessage` also calls `learnMember`, so this is an edge case. More robust would be to union both tables (`SELECT DISTINCT channel_id FROM group_messages UNION SELECT DISTINCT group_id FROM group_members`).

**P2-3 — Q21: latent infinite loop only at `maxChars === 1` with a leading surrogate pair.**
If `maxChars` were `1` and a segment began with a surrogate pair, the back-up sets `splitAt = 0`, pushing an empty segment and never shrinking `remaining`. Not reachable in production (`deliver` always uses the default `MAX_SEGMENT_CHARS = 3500`), but a one-line guard (`if (splitAt < 1) splitAt = 1;` after the back-up, or `Math.max(1, …)`) would make `splitMessage` safe for any caller-supplied `maxChars`.

## Nits

- Q22 placeholder `'[No response generated. Please try rephrasing your question.]'` is hardcoded English while the surrounding UX uses CJK (`fromName：content`). Consider i18n/config if the deployment is Chinese-facing.
- Q15: `insertMessage` + `deleteOldMessages` run as two separate statements per push (not wrapped in a transaction). Harmless given better-sqlite3's synchronous model; noting only for completeness.

## Conclusion

Correct, well-tested, and cleanly scoped. The findings are non-blocking quality/robustness suggestions that can be addressed in a follow-up. Approving.

PR #22, opened 2026-06-05
style: clarify onPacket flag semantics per packet type (Q39, Q42)

lml2468:
## Summary

Two P2 readability items.

### Q39 — onPacket hasServerVersion/noPersist same-bit semantic confusion (`socket.ts`)

The WuKongIM header byte flags' bit 0 has **different semantics per packet type**:
- **CONNACK**: bit 0 = `hasServerVersion` (server includes version byte in body)
- **RECV**: bit 0 = `noPersist` (message should not be persisted)

Previously, both were extracted at the top of `onPacket()` as separate variables reading the same bit, which was misleading. Now each flag is extracted inside its own `case` branch with proper naming and a comment block explaining the per-packet-type semantics.

Also fixed the misleading inline comment at the `encodeConnectPacket` header byte.

### Q42 — pendingFlush readability (no changes needed)

The old pendingFlush/isFlushing timing complexity was entirely removed in Q1 (PR#15). The current `deliver()` method is a clean 30-line accumulate→split→send flow with no timing logic. Nothing to document or simplify.

## Verification
- `npx tsc --noEmit` — zero errors
- `npx vitest run` — 237/237 tests pass

### DIFF
--- a/src/octo/socket.ts
+++ b/src/octo/socket.ts
@@ -180,7 +180,7 @@ function encodeConnectPacket(opts: {
   const bodyBytes = Array.from(body.toUint8Array());
 
   const frame = new Encoder();
-  // header: packetType << 4 | flags (noPersist bit0 = hasServerVersion for CONNACK)
+  // header: packetType << 4 | flags
   frame.writeByte((PacketType.CONNECT << 4) | 0);
   frame.writeBytes(encodeVariableLength(bodyBytes.length));
   frame.writeBytes(bodyBytes);
@@ -572,9 +572,6 @@ export class WKSocket extends EventEmitter {
   private onPacket(data: Uint8Array): void {
     const firstByte = data[0];
     const packetType = firstByte >> 4;
-    const hasServerVersion = (firstByte & 0x01) > 0;
-    const noPersist = (firstByte & 0x01) > 0;
-    const reddot = ((firstByte >> 1) & 0x01) > 0;
 
     // Skip the header and variable-length bytes to get body
     const dec = new Decoder(data);
@@ -583,13 +580,24 @@ export class WKSocket extends EventEmitter {
       dec.readVariableLength(); // remaining length
     }
 
+    // WuKongIM header byte layout: [packetType:4][flags:4]
+    // Bit 0 of flags has DIFFERENT semantics per packet type:
+    //   CONNACK: bit 0 = hasServerVersion (server includes version byte in body)
+    //   RECV:    bit 0 = noPersist (message should not be persisted)
+    // Bit 1 of flags:
+    //   RECV:    bit 1 = reddot (show unread badge)
     switch (packetType) {
-      case PacketType.CONNACK:
+      case PacketType.CONNACK: {
+        const hasServerVersion = (firstByte & 0x01) > 0;
         this.onConnack(dec, hasServerVersion);
         break;
-      case PacketType.RECV:
-        this.onRecv(dec, noPersist, reddot);
+      }
+      case PacketType.RECV: {
+        const _noPersist = (firstByte & 0x01) > 0;
+        const _reddot = ((firstByte >> 1) & 0x01) > 0;
+        this.onRecv(dec, _noPersist, _reddot);
         break;
+      }
       case PacketType.DISCONNECT:
         this.onDisconnect(dec);
         break;


yujiawei (approved):
# Code Review — PR #22 (cc-channel-octo)

## Verdict: APPROVED

This is a focused, behavior-preserving readability refactor. The single changed file (`src/octo/socket.ts`, +15/-7) clarifies the per-packet-type semantics of the WuKongIM header flag byte without altering any runtime behavior. I read the full `onPacket` / `onConnack` / `onRecv` context at the review SHA; the change is correct and an improvement.

## What changed

1. Removed the top-of-`onPacket` extraction where `hasServerVersion` and `noPersist` both read `firstByte & 0x01` from the same bit — genuinely confusing, since the two names suggested different bits.
2. Each flag is now extracted inside its own `case` branch (`CONNACK` → `hasServerVersion`, `RECV` → `_noPersist` / `_reddot`).
3. Added a comment block documenting that bit 0 of the flags nibble has different meaning per packet type.
4. Removed a misleading inline comment on the `encodeConnectPacket` header byte (flags are always `0` for the outbound CONNECT, so the CONNACK note didn't belong there).

## Correctness check

- **Equivalent behavior.** Bit math is unchanged: `hasServerVersion = (firstByte & 0x01) > 0`, `_noPersist = (firstByte & 0x01) > 0`, `_reddot = ((firstByte >> 1) & 0x01) > 0`. The CONNACK branch still passes the version flag to `onConnack`, which still gates the `serverVersion` read on it. The RECV branch still forwards both flags to `onRecv`. ✅
- **Comment accuracy.** The documented mapping (CONNACK bit0 = hasServerVersion, RECV bit0 = noPersist, RECV bit1 = reddot) matches actual usage: `onConnack(dec, hasServerVersion)` reads the version byte only when set. ✅
- **Scope discipline.** `case` bodies are now wrapped in block scopes (`{ ... }`) so the per-branch `const` declarations don't leak across cases — correct and necessary given the lexical bindings. ✅
- **No reachability regression.** `firstByte` is still read once at the top and remains in scope inside the new blocks. ✅

## Notes (non-blocking)

- `_noPersist` and `_reddot` remain unused inside `onRecv` (underscore-prefixed, with an `eslint-disable` for the unused-vars rule). Keeping them documents the wire format and preserves the method signature, which is a reasonable trade-off. If these flags are never expected to be consumed, a future cleanup could drop them from the `onRecv` signature entirely — but that's out of scope here and not required for this PR.

## Verification

Author reports `npx tsc --noEmit` clean and `npx vitest run` 237/237 passing. The diff is consistent with those claims — no type-affecting or logic-affecting changes that would plausibly break either. No further blocking items.

Jerry-Xin (approved):
The PR is relevant to `cc-channel-octo`: it clarifies WuKongIM packet flag semantics in the socket protocol implementation, with no observed behavior change.

💬 Non-blocking

🔵 Suggestion: In [src/octo/socket.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qr2_8zpt/repo/src/octo/socket.ts:596), consider naming the local variables `noPersist` and `reddot` instead of `_noPersist` and `_reddot`. The underscore convention usually signals unused variables, but these locals are used when calling `onRecv()`. The existing `_` names make sense on the ignored `onRecv` parameters at line 652.

✅ Highlights

🔵 The protocol comment at [src/octo/socket.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-qr2_8zpt/repo/src/octo/socket.ts:583) correctly documents that flag bit 0 has packet-specific meaning for `CONNACK` versus `RECV`.

🔵 Moving `hasServerVersion`, `noPersist`, and `reddot` extraction into their relevant `switch` branches improves readability without changing parsing behavior.

Verification note: I attempted `npm run type-check`, `npm run lint`, and `npm test`, but this checkout does not have dev dependencies installed, so `tsc`, `eslint`, and `vitest` were not available locally.

lml2468 (commented):
QA Review: APPROVED ✅

Q39: Flag extraction moved from shared top-of-function to per-case branches with correct naming. Comment block clearly documents bit 0 semantic difference (hasServerVersion for CONNACK, noPersist for RECV). Misleading CONNECT comment also cleaned up. Q42: Correct judgment — pendingFlush deleted in Q1, nothing to do.

Scope clean: socket.ts only, +15/-7, pure readability. — guoshouwei QA

PR #23, opened 2026-06-05
fix(P2): ws listener cleanup, heartbeat during refresh, dependency risk docs (Q30+Q33+Q35)

lml2468:
### Q30: WS listener cleanup on disconnect (socket.ts)

`disconnect()` and `disconnectAndWait()` now call `ws.removeAllListeners()` before `ws.close()`. Previously, event listeners (open/message/close/error) remained on old WS instances until GC. Stale guards prevented message processing (security was fine), but listeners accumulated in reconnect loops — a memory leak.

### Q33: Heartbeat timer during token refresh (gateway.ts)

`attemptTokenRefresh()` now calls `stopHeartbeat()` at the start and `startHeartbeat()` after successful reconnect. Previously the API heartbeat timer continued firing during refresh, sending requests that would fail and increment `heartbeatFailCount`, potentially triggering a second overlapping refresh cycle.

### Q35: Known Dependency Risks documentation (ARCHITECTURE.md)

New section documenting maintenance status and replacement paths for three cryptographic dependencies:

| Package | Last Published | Risk |
|---------|---------------|------|
| `crypto-js@4.2.0` | 2023-10-24 | CVE-2023-46233 (PBKDF2, not used here). 2+ years unmaintained. |
| `md5-typescript@1.0.5` | 2018-03-03 | 8+ years unmaintained. |
| `curve25519-js@0.0.4` | 2019-08-02 | 7+ years unmaintained. |

Assessment: production risk is low (simple operations, no exploitable CVEs for our usage). Risk is supply chain. Replacement paths documented.

### Verification

- `tsc --noEmit`: ✅ clean
- `vitest run`: 237 tests pass
- Scope: 3 files, +20/-1 lines

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -166,6 +166,20 @@ The WuKongIM protocol mandates specific cryptographic choices that cc-channel-oc
 
 These are protocol-level constraints, not implementation choices. Changing them requires WuKongIM server-side changes.
 
+### Known Dependency Risks
+
+Three cryptographic dependencies have low maintenance activity:
+
+| Package | Version | Last Published | Risk | Replacement Path |
+|---------|---------|---------------|------|------------------|
+| `crypto-js` | 4.2.0 | 2023-10-24 | CVE-2023-46233 (PBKDF2, not used here). No releases in 2+ years. | Node.js built-in `crypto.createCipheriv` / `crypto.createDecipheriv` for AES-128-CBC. Drop-in replacement, ~20 lines. |
+| `md5-typescript` | 1.0.5 | 2018-03-03 | No known CVEs. Unmaintained (8+ years). | Node.js built-in `crypto.createHash('md5')`. One-line replacement. |
+| `curve25519-js` | 0.0.4 | 2019-08-02 | No known CVEs. Unmaintained (7+ years). Pure JS implementation. | `@noble/curves` (actively maintained, audited). Requires API adaptation. |
+
+**Current assessment:** Production runtime risk is **low** — these libraries perform simple, well-understood operations (AES-CBC, MD5 hash, X25519 DH). No known exploitable vulnerabilities affect our usage patterns. The risk is supply chain (abandoned packages receiving no security patches) rather than functional.
+
+**Replacement plan:** Tracked as tech debt. Replace when (a) a CVE affects our usage, (b) we diverge from upstream `openclaw-channel-octo` anyway, or (c) a dependency audit policy requires actively maintained packages.
+
 ## Persistence
 
 All persistent state lives in a single SQLite database (`data/cc-octo.db`):
--- a/src/gateway.ts
+++ b/src/gateway.ts
@@ -201,6 +201,9 @@ export class OctoGateway {
     try {
       console.log('Attempting token refresh...');
 
+      // Q33: Stop heartbeat during refresh to avoid empty pings
+      this.stopHeartbeat();
+
       if (this.socket) {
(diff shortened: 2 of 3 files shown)


Jerry-Xin (changes requested):
Relevance gate passed: this PR modifies in-scope WebSocket/gateway lifecycle code and project architecture docs, but one refresh failure path is a blocking availability regression.

🔴 Blocking

🔴 Critical — `src/gateway.ts:204-228`: `attemptTokenRefresh()` now stops the API heartbeat before refresh, but only restarts it after `registerBot()` and reconnect succeed. If `registerBot({ forceRefresh: true })` fails transiently, the old socket has already been disconnected, `this.socket` is set to `null`, and the heartbeat timer remains stopped forever. That removes the only automatic retry path, leaving the gateway stuck offline until process restart or an external trigger. Restart the heartbeat in the failure path, or schedule an explicit retry/backoff, and add a test where refresh registration rejects.

💬 Non-blocking

🟡 Warning — `src/octo/socket.ts:285`: `disconnect()` removes all listeners before calling `close()`, including the `error` listener. If `ws` emits an asynchronous `error` during/after close, Node can treat it as an unhandled error. Safer patterns are to close first and remove listeners from the `close` callback, or keep a no-op `error` listener until the socket is fully closed.

🟡 Warning — test coverage is missing for both new behaviors: failed refresh after `stopHeartbeat()`, and listener cleanup on `disconnect()` / `disconnectAndWait()`. The existing gateway tests cover successful refresh and heartbeat-triggered refresh, but not the new failure mode.

✅ Highlights

The PR is correctly scoped to the described Q30/Q33/Q35 work. `disconnectAndWait()` preserves its own close listener until resolution, so the listener cleanup there does not break the wait path.

Verification note: I attempted `npm test -- --runInBand`, but the workspace does not have `vitest` installed (`sh: vitest: command not found`). `git diff --check` passed.

lml2468 (commented):
## Q30+Q33+Q35 Review (by 李飞飞)

🔴 **Blocking — agree with existing review**

**Q33 refresh failure leaves heartbeat permanently stopped**

`gateway.ts:204-225`: `stopHeartbeat()` runs before `registerBot()`, but `startHeartbeat()` only runs on the success path. If `registerBot({ forceRefresh: true })` throws, the heartbeat stays dead — no automatic reconnection, gateway stuck offline.

Fix: add `this.startHeartbeat()` in the `catch` block (or move it to `finally`). The heartbeat needs to restart even on failure so it can detect the problem and trigger another refresh attempt after cooldown.

```typescript
} catch (err) {
  console.error('Token refresh failed:', String(err));
  this.startHeartbeat(); // Restart heartbeat so it can retry
} finally {
  this.isRefreshing = false;
}
```

💬 **Non-blocking**

1. `socket.ts:285` `removeAllListeners()` before `close()` — agree with existing review's concern. If `ws` emits an async `error` during close, it becomes unhandled. Safe minimum: keep a no-op error listener `ws.on('error', () => {})` after `removeAllListeners()`, before `close()`. Low probability but defensive.

2. Q30 `disconnectAndWait` correctly preserves its own `close` listener by adding it AFTER `removeAllListeners()` (L310). Good — the wait path is not broken.

✅ **Highlights**

- Q35 dependency risk table is well-structured — covers CVE status, maintenance activity, and concrete replacement paths.
- Q33 concept is correct (heartbeat空跑 during refresh is unnecessary), just needs the failure path fix.
- Q30 `removeAllListeners` in both `disconnect` and `disconnectAndWait` is thorough.

PR #24, opened 2026-06-05
fix(P2): Q34 apiUrl SSRF protection + Q37 encodeVariableLength(0) + Q38 aesIV salt warning

lml2468:
### Q34 — apiUrl SSRF protection (config.ts)
New `isAllowedApiUrl()` in `loadConfig`:
- `https://` — always allowed
- `http://` — only localhost / 127.0.0.1 / [::1]
- `file://`, `ftp://`, malformed URLs — rejected

Prevents SSRF if apiUrl is accidentally or maliciously set to an internal service address.

### Q37 — encodeVariableLength(0) (socket.ts)
Special-case `len === 0` to return `[0]` instead of empty array. Per MQTT-style variable-length encoding, length 0 must be a single `0x00` byte. Not triggered in practice but completes the protocol implementation.

### Q38 — aesIV salt length warning (socket.ts)
`console.warn` when CONNACK salt is shorter than 16 bytes. AES-128-CBC requires a 16-byte IV; shorter salt relies on CryptoJS implicit zero-padding. Helps diagnose server-side issues.

### Tests
10 new tests:
- Q34: accept https, localhost, 127.0.0.1, [::1]; reject arbitrary http, file://, malformed
- Q37: variable-length decode roundtrip (0, 1, 127, 128)
- Q38: placeholder (runtime path requires Q29 WKSocket tests)

### Verification
- `tsc --noEmit`: zero errors
- `vitest run`: 247/247 passed (237 + 10 new)
- Files: config.ts / socket.ts / q34-q37-q38.test.ts (+169 lines)

### DIFF
--- a/src/__tests__/q34-q37-q38.test.ts
+++ b/src/__tests__/q34-q37-q38.test.ts
@@ -1,0 +1,139 @@
+/**
+ * Tests for Q34 (apiUrl SSRF protection), Q37 (encodeVariableLength(0)),
+ * Q38 (aesIV salt length warning).
+ */
+
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { writeFileSync, mkdtempSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+// ─── Q34: apiUrl SSRF protection ───────────────────────────────────────────
+
+describe("apiUrl SSRF protection (Q34)", () => {
+  const savedEnv: Record<string, string | undefined> = {};
+
+  function writeConfig(overrides: Record<string, unknown> = {}): string {
+    const dir = mkdtempSync(join(tmpdir(), "q34-"));
+    const cfgPath = join(dir, "config.json");
+    writeFileSync(
+      cfgPath,
+      JSON.stringify({ botToken: "bf_test", apiUrl: "https://safe.example.com", ...overrides }),
+    );
+    return cfgPath;
+  }
+
+  afterEach(() => {
+    for (const key of Object.keys(savedEnv)) {
+      if (savedEnv[key] === undefined) {
+        delete process.env[key];
+      } else {
+        process.env[key] = savedEnv[key];
+      }
+    }
+    vi.resetModules();
+  });
+
+  it("accepts https:// URLs", async () => {
+    const { loadConfig } = await import("../config.js");
+    const path = writeConfig({ apiUrl: "https://api.octo.example.com" });
+    const cfg = loadConfig(path);
+    expect(cfg.apiUrl).toBe("https://api.octo.example.com");
+  });
+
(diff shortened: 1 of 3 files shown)


Jerry-Xin (approved):
The PR is relevant to this repository: it modifies config validation, WuKongIM socket framing/crypto diagnostics, and related tests for existing project functionality.

🔴 Blocking: None.

💬 Non-blocking:
- 🟡 Warning `src/__tests__/q34-q37-q38.test.ts:87`: the Q37 test does not exercise the changed production `encodeVariableLength()` path. It only decodes a manually provided `[0x00]`, so reverting `src/octo/socket.ts:133` would still leave this test green.
- 🟡 Warning `src/__tests__/q34-q37-q38.test.ts:123`: the Q38 test is effectively a no-op placeholder. It spies on `console.warn`, restores it, and asserts `true`; it does not verify the warning string or the CONNACK salt-length path.
- 🟡 Warning `src/__tests__/q34-q37-q38.test.ts:13`: the Q34 tests rely on ambient `process.env`. Since `loadConfig()` gives `CC_OCTO_API_URL` precedence over file config, these tests can become flaky or stop testing the intended URL if that env var is set in the runner.
- 🔵 Suggestion `src/config.ts:204`: allowing every `https://` URL is consistent with the PR description, but it is not complete SSRF protection against internal HTTPS hosts or DNS rebinding. If the intended boundary is “no internal targets,” this should eventually validate resolved addresses, not just scheme.

✅ Highlights:
- `src/config.ts:201` cleanly rejects malformed and non-HTTP(S) URLs during config loading.
- `src/octo/socket.ts:133` correctly fixes zero-length variable-length encoding to emit `0x00`.
- `src/octo/socket.ts:623` adds useful diagnostics for short CONNACK salts without changing connection behavior.

Verification note: I attempted targeted tests and lint, but this checkout does not have dependencies installed: `vitest` and `eslint` were not found.

lml2468 (commented):
**Security Review — LGTM**

## Q34: SSRF protection ✅

`isAllowedApiUrl()` is correct:
- `https:` → allowed
- `http:` → only localhost / 127.0.0.1 / [::1]
- Other protocols (file://, ftp://) → rejected
- Malformed URLs → rejected via try/catch on `new URL()`
- Called in `loadConfig()` after merge, blocks startup if unsafe

Error message includes the offending URL for debugging — good.

## Q37: encodeVariableLength(0) ✅

`if (len === 0) return [0]` — correct. Previously the `while (len > 0)` loop returned `[]` for 0, which would produce a frameless packet. The WuKongIM variable-length encoding of 0 is a single 0x00 byte.

## Q38: aesIV salt warning ✅

`console.warn` when `salt.length < 16` — warning level is appropriate (don't break the connection, but surface the issue). Message accurately describes CryptoJS zero-padding behavior.

PR #26, opened 2026-06-05
fix: restore heartbeat on token refresh failure (Q33 catch path)

lml2468:
Follow-up fix for PR#23 Q33: `attemptTokenRefresh()` calls `stopHeartbeat()` at the start, but the catch block did not restart it. If `registerBot` throws, the gateway permanently loses self-healing. Now both success and failure paths call `startHeartbeat()`.

1 file, +1 line.

### DIFF
--- a/src/gateway.ts
+++ b/src/gateway.ts
@@ -230,6 +230,7 @@ export class OctoGateway {
       this.startHeartbeat(); // Q33: Restart API heartbeat after successful refresh
     } catch (err) {
       console.error('Token refresh failed:', String(err));
+      this.startHeartbeat(); // Q33: Restore heartbeat on failure for self-healing
     } finally {
       this.isRefreshing = false;
     }


Jerry-Xin (changes requested):
The PR is in scope for `cc-channel-octo`, but the new failure-path heartbeat restart needs a shutdown guard.

🔴 Blocking

🔴 Critical — [src/gateway.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-plcr5id8/repo/src/gateway.ts:233): `attemptTokenRefresh()` can outlive `stop()`. If a refresh is in flight, `stop()` sets `_draining`, stops heartbeat, disconnects, and releases the lock, but it does not await or cancel the refresh. If `registerBot()` then rejects, this new catch-path `this.startHeartbeat()` creates a fresh interval after the gateway has been stopped. That can keep the process alive and continue sending heartbeats after shutdown, potentially alongside a new instance that acquired the released lock. Gate heartbeat restarts on a stopped/draining state, or add an explicit stopped flag/cancellation path. The same guard should cover the success restart at line 230 for consistency.

💬 Non-blocking

🟡 Warning — There is no direct test for the new failure path. Add a regression test where refresh starts, `registerBot()` rejects, and heartbeat resumes only when the gateway is still running. Also add the shutdown race case above so `stop()` remains terminal.

✅ Highlights

The normal running-state behavior is directionally correct: restoring the API heartbeat after a failed forced registration preserves the existing self-healing loop.

Test note: I attempted to run the gateway test file, but local dependencies are not installed here (`vitest: command not found`).

PR #28, opened 2026-06-05
feat(G15,G16,G17): add fetchBotGroups, getGroupInfo, searchSpaceMembers API

lml2468:
## G15 + G16 + G17: API 补全

Three new API functions aligned with openclaw-channel-octo upstream, using the existing `getJson` helper.

### G15 — fetchBotGroups
`GET /v1/bot/groups` — returns the list of groups the bot has joined. Graceful fallback (empty array) on failure.

### G16 — getGroupInfo  
`GET /v1/bot/groups/:groupNo` — returns group metadata (name, owner_uid, member_count).

### G17 — searchSpaceMembers
`GET /v1/bot/space/members?keyword=&space_id=&limit=` — fuzzy search Space members. Returns `robot` field (enables future G14/G23 bot detection).

### Tests
10 new tests: success/failure/URL-correctness for each API.

### Verification
- `tsc --noEmit`: zero errors  
- `vitest run`: 276/276 passed (266 + 10 new)
- Files: api.ts (+78) / g15-g16-g17-api.test.ts (+215)

### DIFF
--- a/src/__tests__/g15-g16-g17-api.test.ts
+++ b/src/__tests__/g15-g16-g17-api.test.ts
@@ -1,0 +1,215 @@
+/**
+ * Tests for G15 (fetchBotGroups), G16 (getGroupInfo), G17 (searchSpaceMembers).
+ */
+
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+const mockFetch = vi.fn();
+vi.stubGlobal("fetch", mockFetch);
+
+import {
+  fetchBotGroups,
+  getGroupInfo,
+  searchSpaceMembers,
+} from "../octo/api.js";
+
+describe("fetchBotGroups (G15)", () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+  });
+
+  it("returns group list on success", async () => {
+    mockFetch.mockResolvedValue({
+      ok: true,
+      text: () =>
+        Promise.resolve(
+          JSON.stringify([
+            { group_no: "g1", name: "Group One" },
+            { group_no: "g2", name: "Group Two" },
+          ]),
+        ),
+    });
+
+    const groups = await fetchBotGroups({
+      apiUrl: "https://api.test",
+      botToken: "tok",
+    });
+    expect(groups).toHaveLength(2);
+    expect(groups[0].group_no).toBe("g1");
+    expect(groups[1].name).toBe("Group Two");
+  });
+
+  it("returns empty array on API failure", async () => {
+    mockFetch.mockResolvedValue({
+      ok: false,
+      status: 500,
+      statusText: "Internal Server Error",
+      text: () => Promise.resolve("server error"),
+    });
+
+    const groups = await fetchBotGroups({
+      apiUrl: "https://api.test",
+      botToken: "tok",
+    });
+    expect(groups).toEqual([]);
+  });
+
+  it("calls correct URL", async () => {
+    mockFetch.mockResolvedValue({
+      ok: true,
+      text: () => Promise.resolve("[]"),
+    });
+
+    await fetchBotGroups({
+      apiUrl: "https://api.test",
+      botToken: "tok",
+    });
+
+    const url = mockFetch.mock.calls[0][0] as string;
+    expect(url).toBe("https://api.test/v1/bot/groups");
+  });
+});
+
+describe("getGroupInfo (G16)", () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+  });
+
+  it("returns group info on success", async () => {
+    mockFetch.mockResolvedValue({
+      ok: true,
+      text: () =>
+        Promise.resolve(
+          JSON.stringify({
+            group_no: "g1",
+            name: "My Group",
+            owner_uid: "owner123",
+            member_count: 42,
+          }),
+        ),
+    });
+
+    const info = await getGroupInfo({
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
The PR is relevant to `cc-channel-octo`: it extends the existing Octo API wrapper with bot group and Space member APIs, and the implementation generally matches the project’s current `getJson` helper pattern.

💬 Non-blocking

🟡 Warning — `groupNo` is interpolated as a raw path segment in `getGroupInfo`, so values containing `/`, `?`, or `#` could produce an unintended request path. Consider `encodeURIComponent(params.groupNo)` for path safety. Existing `getGroupMembers` has the same pattern, so this is not a PR-specific blocker, but the new API repeats it. See `src/octo/api.ts:313-317`.

🔵 Suggestion — `searchSpaceMembers` omits `limit` when it is `0` because it checks truthiness. If `0` should be allowed or explicitly sent to the API, use `params.limit !== undefined` instead. See `src/octo/api.ts:337-340`.

🔵 Suggestion — The query parameter test checks only substring presence. A small additional test with spaces or non-ASCII keyword values would lock in the `URLSearchParams` encoding behavior. See `src/__tests__/g15-g16-g17-api.test.ts:164-182`.

✅ Highlights

The new APIs use the existing `getJson` helper, preserving authorization, timeout behavior, error body propagation, and JSON parsing semantics.

`fetchBotGroups` implements the described graceful fallback behavior and validates the response is an array before returning it.

The tests cover success, failure, and URL construction for all three new functions.

Verification note: I attempted to run the targeted tests and type-check locally, but this checkout does not have dev dependencies installed: `vitest: command not found` and `tsc: command not found`.

yujiawei (approved):
# Code Review — PR #28 (cc-channel-octo)

**Verdict: APPROVED** — no blocking issues. Three small, well-tested GET helpers that faithfully mirror the `openclaw-channel-octo` upstream signatures and endpoint paths. `tsc --noEmit` clean, 276/276 tests pass. Suggestions below are non-blocking polish (P2/nit).

## ✅ Verified

- **Endpoint paths** match upstream exactly: `GET /v1/bot/groups`, `GET /v1/bot/groups/:groupNo`, `GET /v1/bot/space/members`.
- **Function signatures & return shapes** align with upstream `fetchBotGroups` / `getGroupInfo` / `searchSpaceMembers`.
- **Error-handling contracts** are intentionally consistent with upstream: `fetchBotGroups` degrades gracefully (logs + returns `[]`); `getGroupInfo` and `searchSpaceMembers` throw on failure.
- **Security**: `URLSearchParams.set()` percent-encodes all query params (keyword/space_id/limit), so no query-string injection. `botToken` is sent only as a header, never logged or placed in the URL; error messages include response body but not the token. `apiUrl` is bot/instance config, not request-time user input — no new SSRF surface introduced.
- **Tests**: 10 tests cover success, one error branch each, and URL/query construction (including trailing-slash normalization and the no-params case).

## Suggestions (non-blocking)

### P2 — `searchSpaceMembers` lacks the `Array.isArray` guard that `fetchBotGroups` has
`getJson` returns `{} as T` on an empty 200 body (`if (!text) return {} as T;`). `fetchBotGroups` defends against this with `return Array.isArray(data) ? data : []`, but `searchSpaceMembers` returns the `getJson` result directly. If the server ever replies `200` with an empty body, the caller receives `{}` falsely typed as `SpaceMember[]`, and a consumer doing `.map()`/`.length` would misbehave.

This is latent (no consumer exists in this PR; a search-with-no-matches conventionally returns `[]`, not an empty body) and it actually mirrors upstream behavior — so it's not a blocker. But since `fetchBotGroups` already guards in this same PR, mirroring it would make the two list helpers consistent:

```ts
const data = await getJson<SpaceMember[]>(params.apiUrl, params.botToken, `/v1/bot/space/members${qs ? `?${qs}` : ""}`);
return Array.isArray(data) ? data : [];
```

Keep the throw-vs-return-`[]` difference between the two helpers as-is — that's intentional upstream parity.

### P2 — No empty-body test for any of the three helpers
The 10 tests cover happy path, one HTTP-error branch each, and URL construction — but none mocks a `200` with an empty/whitespace body, which is exactly the `getJson` `if (!text) return {} as T` branch. An empty-body test for each helper (asserting `fetchBotGroups -> []`, and `searchSpaceMembers -> []` after the guard above) would lock in the array contract and would have surfaced the guard asymmetry.

### nit — `limit=0` is silently dropped
`if (params.limit) query.set("limit", String(params.limit))` treats `0` as falsy, so an explicit `limit: 0` is never forwarded. Almost certainly never a meaningful caller intent for a member search, and the truthy guard is the idiomatic pattern; flagging only for awareness. If `0` should ever be forwarded, use `if (params.limit !== undefined)`.

### nit — `getGroupInfo` does not `encodeURIComponent(groupNo)`
The path segment `/v1/bot/groups/${params.groupNo}` is interpolated raw. `groupNo` is a server-issued opaque ID (not end-user input), and upstream `getGroupInfo` also doesn't encode it, so this is upstream-parity, not a regression. Optional defense-in-depth: wrap with `encodeURIComponent` to align with the `mention_pref`/`threads` helpers that do encode. Safe to defer.

## Summary
Clean, focused PR with good test coverage and correct upstream alignment. None of the above blocks merge — the array-guard consistency item (P2) is the most worthwhile follow-up. Approving.

PR #29, opened 2026-06-05
feat(security): G14+G18+G19+G20+G23

lml2468:
Bot loop prevention, owner identity, per-user rate limit, robot flags. 5 files +121/-3. 266 tests pass.

### DIFF
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -278,3 +278,44 @@ describe('GroupContext', () => {
     expect(context2).toContain('Bob：world');
   });
 });
+
+// ─── G23: robot flag tracking ───────────────────────────────────────────────────────────────
+
+describe('G23: robot flag', () => {
+  let adapter: DbAdapter;
+  let ctx: GroupContext;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    adapter = createTestAdapter();
+    ctx = new GroupContext(adapter, 6000);
+  });
+
+  it('isRobot returns undefined for unknown channel/uid', () => {
+    expect(ctx.isRobot('ch-unknown', 'u-unknown')).toBeUndefined();
+  });
+
+  it('refreshMembers populates robot flags from GroupMember.robot field', async () => {
+    (getGroupMembers as ReturnType<typeof vi.fn>).mockResolvedValueOnce([
+      { uid: 'alice', name: 'Alice', robot: 0 },
+      { uid: 'helper_bot', name: 'Helper', robot: 1 },
+      { uid: 'no-flag', name: 'NoFlag' }, // robot undefined — should NOT be stored
+    ]);
+    await ctx.refreshMembers('ch1', 'https://api', 'token');
+
+    expect(ctx.isRobot('ch1', 'alice')).toBe(false);
+    expect(ctx.isRobot('ch1', 'helper_bot')).toBe(true);
+    expect(ctx.isRobot('ch1', 'no-flag')).toBeUndefined();
+  });
+
+  it('robot flags are scoped per channel — no cross-channel leakage', async () => {
+    (getGroupMembers as ReturnType<typeof vi.fn>)
+      .mockResolvedValueOnce([{ uid: 'shared', name: 'A', robot: 1 }])
+      .mockResolvedValueOnce([{ uid: 'shared', name: 'A', robot: 0 }]);
+    await ctx.refreshMembers('ch1', 'https://api', 'token');
+    // Force second refresh by clearing lastRefresh state — use a different channel
+    await ctx.refreshMembers('ch2', 'https://api', 'token');
+    expect(ctx.isRobot('ch1', 'shared')).toBe(true);
+    expect(ctx.isRobot('ch2', 'shared')).toBe(false);
+  });
+});
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -470,3 +470,159 @@ describe('Message length limit (Q10)', () => {
     expect(result!.shouldProcess).toBe(false);
   });
 });
+
+// ─── G14: bot-to-bot DM loop prevention ────────────────────────────────────────────
+
+describe('G14: bot-to-bot DM loop prevention', () => {
+  it('drops DM from a uid ending in _bot', async () => {
+    const router = new SessionRouter(makeConfig(), ROBOT_ID);
+    const result = await router.route(
+      makeMsg({
+        channel_type: ChannelType.DM,
+        from_uid: 'random_bot',
+        payload: { type: MessageType.Text, content: 'hi' },
+      }),
+    );
+    expect(result).toBeNull();
+  });
+
+  it('drops DM from the bot itself (knownBotUids includes self)', async () => {
+    const router = new SessionRouter(makeConfig(), ROBOT_ID);
+    const result = await router.route(
+      makeMsg({
+        channel_type: ChannelType.DM,
+        from_uid: ROBOT_ID,
+        payload: { type: MessageType.Text, content: 'hi' },
+      }),
+    );
+    expect(result).toBeNull();
+  });
+
+  it('drops DM from a registered known bot uid', async () => {
+    const router = new SessionRouter(makeConfig(), ROBOT_ID);
+    router.registerKnownBot('peer-bot-uid');
+    const result = await router.route(
+      makeMsg({
+        channel_type: ChannelType.DM,
+        from_uid: 'peer-bot-uid',
+        payload: { type: MessageType.Text, content: 'hi' },
+      }),
+    );
+    expect(result).toBeNull();
+  });
+
+  it('allows DM from a bot in allowedBotUids whitelist', async () => {
+    const router = new SessionRouter(
+      makeConfig({ allowedBotUids: ['trusted_bot'] }),
+      ROBOT_ID,
+    );
+    const result = await router.route(
+      makeMsg({
+        channel_type: ChannelType.DM,
+        from_uid: 'trusted_bot',
+        payload: { type: MessageType.Text, content: 'hi' },
+      }),
+    );
+    expect(result).not.toBeNull();
+    expect(result!.shouldProcess).toBe(true);
+  });
+
+  it('allows DM from a regular human user (no _bot suffix)', async () => {
+    const router = new SessionRouter(makeConfig(), ROBOT_ID);
+    const result = await router.route(
+      makeMsg({
+        channel_type: ChannelType.DM,
+        from_uid: 'alice123',
+        payload: { type: MessageType.Text, content: 'hi' },
+      }),
+    );
+    expect(result).not.toBeNull();
+    expect(result!.shouldProcess).toBe(true);
+  });
+
+  it('does NOT drop group messages from _bot uid (mention gate handles those)', async () => {
+    const router = new SessionRouter(makeConfig(), ROBOT_ID);
+    // Without @mention, group msg from bot would be dropped by mention gate, not G14.
+    // With @mention, it should pass.
+    const result = await router.route(
+      makeMsg({
+        channel_type: ChannelType.Group,
+        from_uid: 'someone_bot',
+        payload: {
+          type: MessageType.Text,
+          content: 'hi',
+          mention: { uids: [ROBOT_ID] },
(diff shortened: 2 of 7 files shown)


Jerry-Xin (approved):
Summary: Project relevance gate passed; this PR modifies core cc-channel-octo security/routing behavior and is in scope. I found no blocking bugs.

💬 Non-blocking

🟡 Warning — `src/session-router.ts:187`: the rate-limit checks consume global and per-session tokens before checking the new per-user bucket. Once a user exceeds the per-user quota, further cross-channel messages can still drain the global bucket and per-session buckets for each touched channel. Consider checking the per-user bucket before consuming broader/shared buckets, or making checks transactional.

🟡 Warning — Test coverage is thin for the new behavior. I do not see direct tests for `_bot` DM rejection/allowlisting (`src/session-router.ts:124`, `src/session-router.ts:130`), cross-channel per-user limiting (`src/session-router.ts:247`), `CC_OCTO_ALLOWED_BOT_UIDS` config parsing (`src/config.ts:206`), `ownerUid` propagation (`src/gateway.ts:186`, `src/index.ts:66`), or `GroupContext.isRobot()` (`src/group-context.ts:263`). The implementation looks reasonable, but these are security-adjacent paths and should be pinned down.

🔵 Suggestion — `src/group-context.ts:169`: robot flags are memory-only and only updated when `m.robot !== undefined`. That is fine for the stated “future routing integrations” use, but if this becomes an enforcement signal later, stale/missing values and restart behavior should be designed explicitly.

✅ Highlights

The DM bot-loop gate is conservative and configurable via `allowedBotUids`.

Owner identity is captured without changing runtime behavior yet, which keeps the permission-model groundwork low risk.

Group mention semantics remain aligned with the existing `uids` / `ais` behavior and still ignore legacy `all`.

I attempted to run tests, but the local checkout does not have `vitest` installed: `sh: vitest: command not found`.

Jerry-Xin (approved):
Summary: Project relevance gate passed; this PR modifies core cc-channel-octo security/routing behavior and is in scope. I found no blocking bugs.

💬 Non-blocking

🟡 Warning — `src/session-router.ts:187`: the rate-limit checks consume global and per-session tokens before checking the new per-user bucket. Once a user exceeds the per-user quota, further cross-channel messages can still drain the global bucket and per-session buckets for each touched channel. Consider checking the per-user bucket before consuming broader/shared buckets, or making checks transactional.

🟡 Warning — Test coverage is thin for the new behavior. I do not see direct tests for `_bot` DM rejection/allowlisting (`src/session-router.ts:124`, `src/session-router.ts:130`), cross-channel per-user limiting (`src/session-router.ts:247`), `CC_OCTO_ALLOWED_BOT_UIDS` config parsing (`src/config.ts:206`), `ownerUid` propagation (`src/gateway.ts:186`, `src/index.ts:66`), or `GroupContext.isRobot()` (`src/group-context.ts:263`). The implementation looks reasonable, but these are security-adjacent paths and should be pinned down.

🔵 Suggestion — `src/group-context.ts:169`: robot flags are memory-only and only updated when `m.robot !== undefined`. That is fine for the stated “future routing integrations” use, but if this becomes an enforcement signal later, stale/missing values and restart behavior should be designed explicitly.

✅ Highlights

The DM bot-loop gate is conservative and configurable via `allowedBotUids`.

Owner identity is captured without changing runtime behavior yet, which keeps the permission-model groundwork low risk.

Group mention semantics remain aligned with the existing `uids` / `ais` behavior and still ignore legacy `all`.

I attempted to run tests, but the local checkout does not have `vitest` installed: `sh: vitest: command not found`.

yujiawei (changes requested):
# Code Review — PR #29 (cc-channel-octo)

**Scope:** `feat(security): G14+G18+G19+G20+G23` — 5 files, +121/-3. Reviewed at head `b058b29`. Typecheck clean; all 266 existing tests pass. This PR is security-sensitive (rate-limiting + bot loop prevention), so it gets extra scrutiny.

Net: the intent is good and the gates fail **closed** (no rate-limit bypass, no security hole that lets traffic through). But G20 introduces a confirmed functional regression in its own reply-debounce, and a security-sensitive change to the rate limiter ships with **zero** automated tests for any of the new behavior. Requesting changes on those two before merge.

---

## P1 — must fix before merge

### 1. G20 breaks the rate-limit reply debounce → "请稍后再试" reply amplification

`src/session-router.ts:187-199`. The rejection branch is:

```ts
if (
  !this.checkGlobalRateLimit() ||
  !this.checkRateLimit(key) ||
  !this.checkUserRateLimit(msg.from_uid)
) {
  const bucket = this.tokenBuckets.get(key); // per-SESSION bucket only
  if (bucket && !bucket.notified) {
    bucket.notified = true;
    await this.replySafe(msg, '请稍后再试');
  }
  ...
}
```

The debounce flag lives on the **per-session** bucket, but `checkRateLimit(key)` (line 217) always runs *before* the new per-user check (`||` short-circuit), and it unconditionally resets the flag while the session bucket still has tokens:

```ts
// session-router.ts:232-235
if (bucket.tokens >= 1) {
  bucket.notified = false;
}
```

So when a user is throttled by the **new per-user limit** (`checkUserRateLimit`, the cross-channel case G20 exists to handle) while the per-session bucket is fresh, every incoming message re-creates/refreshes the session bucket, clears `notified`, and sends another `请稍后再试`. The debounce — whose stated purpose is "only notify once per rate-limit window to avoid reply spam" — never engages on this path.

- Concretely (default `maxPerMinute=5`, `config.ts:63`): a uid exhausts its per-user bucket across group A, then messages group B. Group B's session bucket is fresh, so the bot replies `请稍后再试` on roughly every message until that session bucket also drains (~5 replies/fresh-session), capped in aggregate by the global 10× bucket (~50 replies/min). That's a self-inflicted reply-spam / mild self-DoS, and it's the exact behavior the debounce was added to prevent.
- The existing debounce test (`session-router.test.ts:200-216`) uses `maxPerMinute:1` + DM, so the session bucket hits 0 immediately and line 233 never re-fires — which is why the regression is invisible to CI.

**Fix direction:** record the "notified" state against the bucket(s) that actually caused the rejection (or a dedicated per-key notify timestamp), and stop clearing `notified` on every message that still has tokens — clear it on window rollover instead. Add a regression test with `maxPerMinute>=2` exercising the per-user-only rejection path.

### 2. Security-sensitive PR with zero tests for the new logic

`git diff` for this PR touches only the 5 source files — no test file was added or extended. None of the new symbols appear anywhere under `src/__tests__/`:

- `looksLikeBot` / `allowedBotUids` (G14 DM bot-drop + whitelist bypass)
- `checkUserRateLimit` / `userBuckets` (G20 per-user limit)
- `isRobot` / `robotFlags` (G23)
- `registerKnownBot`

The G14 bot-drop (`session-router.ts:158-164`, silent `return null`) and the G20 per-user throttle are precisely the **silent-failure** paths where a regression — failing open (loop not prevented) or failing closed (legit DMs dropped) — would be invisible. For a security-sensitive change this is the gap most worth closing. Please add unit tests covering at minimum:

1. DM from a `*_bot` uid is dropped; a uid in `allowedBotUids` bypasses the drop.
2. Per-user bucket throttles the same uid across distinct session keys (the G20 cross-channel scenario).
3. `GroupContext.isRobot` returns the expected tri-state.

---

## P2 — should fix / follow up (non-blocking)

### 3. Rate-limit checks consume tokens before the decision → global budget drained by rejected traffic

`src/session-router.ts:187-191`. Checks run global → session → user, and each decrements its bucket *on the pass path* before the next runs (`globalBucket.tokens -= 1` L276, session `tokens -= 1` L238). So a message ultimately rejected by the per-user check has already burned a **global** token (and a session token). Everything fails closed — no bypass — but a single user repeatedly hitting *their own* per-user limit still drains the **shared** global budget, which degrades capacity for unrelated users (a real, if low-impact, cross-user fairness flaw). Suggest a two-phase peek-then-commit: evaluate all three limits without consuming, decrement only when all pass.

### 4. Bucket cleanup is unreachable under global saturation; per-user buckets have no independent sweep

`cleanStaleBuckets()` is only invoked from inside `checkRateLimit()` (`session-router.ts:218`). When `checkGlobalRateLimit()` returns false (line 188), the `||` short-circuit means `checkRateLimit` — and thus the only sweeper of both `tokenBuckets` and the new `userBuckets` — never runs. During sustained global saturation (a flood, exactly when the maps grow fastest) cleanup is skipped on every message, so the maps accumulate per-key/per-uid entries until traffic subsides and a message finally passes the global check. The growth is transient and self-correcting rather than a permanent leak, but the cleanup design assumes `checkRateLimit` is always reached — which both global saturation and idle periods violate. Consider running cleanup unconditionally (or on a timer) rather than coupling it to the per-session check.

### 5. G18 / G23 / `registerKnownBot` add dead "for future" scaffolding

Confirmed unconsumed in `src/`:

- `SessionRouter.ownerUid` (G18): assigned at `session-router.ts:50` from `gateway.ownerUid`, never read.
- `GroupContext.isRobot` + `robotFlags` (G23): `isRobot` (`group-context.ts:263`) has no callers; `robotFlags` is written at `group-context.ts:170-173` and read only inside the unused `isRobot`, so the whole write/read pair is unreachable.
- `registerKnownBot` (`session-router.ts:55`): defined, never called.

(For the record, `knownBotUids` itself **is** live — consumed at `session-router.ts:125` — and the `gateway.ts` `ownerUid` getter is real plumbing; only the consumer side dead-ends.) Dead-but-harmless, but it's untested surface area. Either wire it into a consumer in this PR or split it out until the permission model / 免@ gate that uses it actually lands.

### 6. G14 `_bot` suffix heuristic is the only DM-side bot signal

`looksLikeBot` (`session-router.ts:124-128`) relies on `knownBotUids` + a `uid.endsWith('_bot')` suffix for DM senders. The new server-authoritative `robot` flag (`GroupMember.robot` → `GroupContext.isRobot`) is structurally **unavailable** for DM senders (it's group-member-scoped only), so the suffix is genuinely the best DM signal available today. The false-positive concern (a human picking a `*_bot` uid) is not substantiated in the repo — nothing shows octo uids are user-chosen — so this is acceptable as a documented heuristic. Suggestions if you want to harden later: extend `fetchUserInfo` to return a robot flag and prefer it over the suffix; and log dropped DMs instead of a fully silent `return null` so any real false positive is observable.

---

## Items to manually verify (human reviewer)

- The intended product behavior under the G20 cross-channel throttle: is repeated `请稍后再试` (finding #1) acceptable, or must it be one-per-window as the debounce comment claims? This drives whether #1 is truly blocking.
- Whether octo uids can ever be self-chosen (affects #6's false-positive risk on the silent DM drop).

## Summary

| # | Issue | Severity |
|---|-------|----------|
| 1 | G20 breaks reply debounce → `请稍后再试` amplification | P1 |
| 2 | Zero tests for new security-sensitive logic | P1 |
| 3 | Global token consumed by per-user-rejected traffic | P2 |
| 4 | Bucket cleanup unreachable under global saturation | P2 |
| 5 | Dead G18/G23/`registerKnownBot` scaffolding | P2 |
| 6 | `_bot` suffix is sole DM bot signal | P2 |

Requesting changes for #1 and #2. The remaining items are non-blocking follow-ups.

lml2468 (commented):
QA Re-review: APPROVED ✅

P1.1 (peek-then-consume) verified — checkAllRateLimits refills all three buckets first, returns the blocking bucket without consuming on failure, consumes from all three on success. Per-bucket notified flag means a user blocked by per-user limit gets debounced on the user bucket, not the per-session bucket. No more reflection DoS — verified by the new "at most 2 replies for 10 blocked messages" test.

P1.2 (zero tests) verified — 12 new tests, all green:
- G14: 6 tests (suffix / self / known-bot / whitelist / human / group-mention)
- G18: 1 test (constructor accepts ownerUid)
- G20: 2 tests (cross-channel block + bounded debounce replies)
- G23: 3 tests (unknown undefined / robot=1/0 stored / per-channel isolation)

Verified locally on head branch: 288/288 pass, tsc clean, eslint 0 errors.

Minor non-blocking notes:
- looksLikeBot self-check is dead (line 117 returns null for self first) but harmless
- ownerUid is stored but not yet used — fine, G18 is "store for future permission model"
- _ownerUid in gateway uses leading underscore for "unused-but-stored" — eslint passes

Ready to merge. — guoshouwei QA

PR #30, opened 2026-06-05
feat(G9,G10,G21): Space isolation, history segmentation, streamOn filter

lml2468:
G9 Space isolation + G10 history segmentation + G21 streamOn filter. 8 new tests. 274/274 pass.

### DIFF
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -161,8 +161,8 @@ async function simulateMessage(
     }
 
     const userContent = msg.payload.content ?? '';
-    const historyPrefix = store.buildHistoryPrefix(sessionKey, config.context.historyLimit);
-    store.appendUser(sessionKey, userContent);
+    const historyPrefix = store.buildSegmentedHistoryPrefix(sessionKey, config.context.historyLimit);
+    store.appendUser(sessionKey, userContent, msg.message_seq);
 
     const rawChunks = queryAgent(userContent, historyPrefix, contextStr, config);
 
@@ -184,7 +184,8 @@ async function simulateMessage(
 
     const fullResponse = collected.join('');
     if (fullResponse) {
-      store.appendAssistant(sessionKey, fullResponse);
+      store.appendAssistant(sessionKey, fullResponse, msg.message_seq);
+      store.setLastBotReplySeq(sessionKey, msg.message_seq);
     }
     } catch (err) {
       console.error('simulateMessage error:', err);
@@ -205,6 +206,7 @@ async function simulateMessage(
   if (
     !wasProcessed &&
     isGroup &&
+    !msg.streamOn &&
     msg.payload.type === MessageType.Text &&
     msg.payload.content
   ) {
@@ -443,5 +445,56 @@ describe('E2E smoke tests', () => {
     const history = store.buildHistoryPrefix(USER_UID, 40);
     expect(history).toContain('[user]: empty response');
     expect(history).not.toContain('[assistant]');
+  });
+
+  // --- PR#30 review: G21 streamOn must not pollute group context ---
+
+  it('streamOn=true group message is NOT cached in group context (G21 fix)', async () => {
+    const baseMsg = makeGroupMsg('streaming partial update');
+    const streamingMsg: BotMessage = { ...baseMsg, streamOn: true };
+
+    await simulateMessage(streamingMsg, config, store, router, groupContext, streamRelay);
+
+    // The streaming update must NOT appear in the group context window.
+    const context = groupContext.buildContext(GROUP_CHANNEL);
+    expect(context).not.toContain('streaming partial update');
+  });
+
+  it('streamOn=false group message IS cached in group context (G21 baseline)', async () => {
+    const finalMsg = makeGroupMsg('final message');
+    await simulateMessage(finalMsg, config, store, router, groupContext, streamRelay);
+
+    const context = groupContext.buildContext(GROUP_CHANNEL);
+    expect(context).toContain('final message');
+  });
+
+  // --- PR#30 review: G10 segmentation actually runs end-to-end ---
+
+  it('handleMessage records lastBotReplySeq so next turn segments history (G10 fix)', async () => {
+    mockQueryYield('first answer');
+    await simulateMessage(
+      { ...makeDmMsg('first question'), message_seq: 100 },
+      config, store, router, groupContext, streamRelay,
+    );
+
+    // After the first turn, lastBotReplySeq should be set to 100.
+    expect(store.getLastBotReplySeq(USER_UID)).toBe(100);
+
+    // Second turn: user sends a follow-up with seq=101 — it must be labeled [new].
+    mockQueryYield('second answer');
+    await simulateMessage(
+      { ...makeDmMsg('follow-up question'), message_seq: 101 },
+      config, store, router, groupContext, streamRelay,
+    );
+
+    // Inspect history segmentation as it was built for the second turn.
+    // Set lastBotReplySeq back to 100 to simulate the state the second turn saw.
+    store.setLastBotReplySeq(USER_UID, 100);
+    const segHistory = store.buildSegmentedHistoryPrefix(USER_UID, 40);
+    expect(segHistory).toContain('[answered history]');
+    expect(segHistory).toContain('first question');
+    expect(segHistory).toContain('first answer');
+    expect(segHistory).toContain('[new messages]');
+    expect(segHistory).toContain('follow-up question');
   });
 });
--- a/src/__tests__/g9-g10-g21.test.ts
+++ b/src/__tests__/g9-g10-g21.test.ts
@@ -107,11 +107,12 @@ describe("SessionStore history segmentation (G10)", () => {
     store.init();
 
     store.getOrCreate("test-session", "ch1", 2);
-    store.appendUser("test-session", "first question");
-    store.appendAssistant("test-session", "first answer");
-    store.appendUser("test-session", "second question");
-    store.appendUser("test-session", "follow up");
+    store.appendUser("test-session", "first question", 1);
+    store.appendAssistant("test-session", "first answer", 1);
+    store.appendUser("test-session", "second question", 2);
+    store.appendUser("test-session", "follow up", 3);
 
+    // Bot last replied to message_seq 1; messages with seq > 1 are [new messages].
     store.setLastBotReplySeq("test-session", 1);
 
     const history = store.buildSegmentedHistoryPrefix("test-session", 40);
@@ -147,8 +148,8 @@ describe("SessionStore history segmentation (G10)", () => {
     store.init();
 
     store.getOrCreate("test-session", "ch1", 2);
-    store.appendUser("test-session", "msg1");
-    store.appendUser("test-session", "msg2");
+    store.appendUser("test-session", "msg1", 1);
+    store.appendUser("test-session", "msg2", 2);
     store.setLastBotReplySeq("test-session", 0);
 
     const history = store.buildSegmentedHistoryPrefix("test-session", 40);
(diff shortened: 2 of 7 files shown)


Jerry-Xin (commented):
The PR is in scope for `cc-channel-octo`, but the advertised G10 history segmentation is not integrated and G21 is incomplete for group context caching.

🔴 Blocking

- 🔴 Critical: G10 is effectively dead code in the real pipeline. `src/index.ts:138` still calls `store.buildHistoryPrefix(...)`, and no production path calls `buildSegmentedHistoryPrefix(...)`. Also, after the bot replies, `src/index.ts:160` only appends the assistant response and never records `msg.message_seq` with `setLastBotReplySeq(...)`. As a result, the new segmentation behavior is only exercised by unit tests and will never affect actual agent prompts.

- 🔴 Critical: `lastBotReplySeq` is not actually used to segment by sequence. `src/session-store.ts:177` reads the value, but `src/session-store.ts:183-198` ignores it and simply splits at the last assistant message. The `messages` table also does not store `message_seq`, so the implementation cannot honor the documented behavior at `src/session-store.ts:167-170` (“Messages before lastBotReplySeq...” / “messages after...”). Tests such as `src/__tests__/g9-g10-g21.test.ts:115` pass with an arbitrary sequence value because the sequence is not involved in the result.

- 🔴 Critical: G21’s `streamOn` filter does not fully skip streaming updates for groups. `src/session-router.ts:150-153` prevents handler execution, but `src/index.ts:189-197` then caches every non-processed group text message into `groupContext` without checking `msg.streamOn`. Partial streaming updates can therefore still enter future context and influence later prompts.

💬 Non-blocking

- 🟡 Warning: `src/session-router.ts:107-112` treats any DM `from_uid` starting with `s` and containing `_` as a space-scoped UID. If ordinary user IDs can match that shape, this changes session keys unexpectedly. Consider validating against the full space DM pattern, preferably using `channel_id` consistency, before extracting a space ID.

✅ Highlights

- G9’s direction is relevant and the router tests cover basic space separation.
- Passing `setting.streamOn` from `WKSocket` into `BotMessage` is a reasonable place to expose the WuKongIM flag.

Jerry-Xin (changes requested):
The PR is in scope for `cc-channel-octo`, but the advertised G10 history segmentation is not integrated and G21 is incomplete for group context caching.

🔴 Blocking

- 🔴 Critical: G10 is effectively dead code in the real pipeline. `src/index.ts:138` still calls `store.buildHistoryPrefix(...)`, and no production path calls `buildSegmentedHistoryPrefix(...)`. Also, after the bot replies, `src/index.ts:160` only appends the assistant response and never records `msg.message_seq` with `setLastBotReplySeq(...)`. As a result, the new segmentation behavior is only exercised by unit tests and will never affect actual agent prompts.

- 🔴 Critical: `lastBotReplySeq` is not actually used to segment by sequence. `src/session-store.ts:177` reads the value, but `src/session-store.ts:183-198` ignores it and simply splits at the last assistant message. The `messages` table also does not store `message_seq`, so the implementation cannot honor the documented behavior at `src/session-store.ts:167-170` (“Messages before lastBotReplySeq...” / “messages after...”). Tests such as `src/__tests__/g9-g10-g21.test.ts:115` pass with an arbitrary sequence value because the sequence is not involved in the result.

- 🔴 Critical: G21’s `streamOn` filter does not fully skip streaming updates for groups. `src/session-router.ts:150-153` prevents handler execution, but `src/index.ts:189-197` then caches every non-processed group text message into `groupContext` without checking `msg.streamOn`. Partial streaming updates can therefore still enter future context and influence later prompts.

💬 Non-blocking

- 🟡 Warning: `src/session-router.ts:107-112` treats any DM `from_uid` starting with `s` and containing `_` as a space-scoped UID. If ordinary user IDs can match that shape, this changes session keys unexpectedly. Consider validating against the full space DM pattern, preferably using `channel_id` consistency, before extracting a space ID.

✅ Highlights

- G9’s direction is relevant and the router tests cover basic space separation.
- Passing `setting.streamOn` from `WKSocket` into `BotMessage` is a reasonable place to expose the WuKongIM flag.

yujiawei (changes requested):
# Code Review — PR #30 (cc-channel-octo)

## Summary

This PR ships three independent features. Two are correct and fully wired; one is non-functional dead code despite being advertised as delivered.

| Feature | Status | Verdict |
|---|---|---|
| G9 — Space isolation in session key | Wired, correct | ✅ |
| G21 — `streamOn` filter | Wired, correct, protocol-accurate | ✅ |
| G10 — History segmentation | **Not wired — no production caller** | ❌ P1 |

Build (`tsc --noEmit`) and the full suite (274/274) pass locally.

---

## P1 — G10 history segmentation is never invoked in production (blocking)

The new segmentation API in `src/session-store.ts` is dead code. It has no production caller — only the test file exercises it:

- `setLastBotReplySeq()` (`session-store.ts:158`), `getLastBotReplySeq()` (`:163`), and `buildSegmentedHistoryPrefix()` (`:172`) are referenced only in `src/__tests__/g9-g10-g21.test.ts` and their own definitions. A tree-wide grep finds zero other call sites. `getLastBotReplySeq()` has no caller at all, not even in tests.
- The live message handler still uses the **old** flat builder: `src/index.ts:138` → `store.buildHistoryPrefix(sessionKey, config.context.historyLimit)`. Nothing in `index.ts` calls `setLastBotReplySeq()` or `buildSegmentedHistoryPrefix()`.

Consequence: the `lastBotReplySeq` map is always empty in production, so even if `buildSegmentedHistoryPrefix()` were called, it would always hit the `lastReplySeq === undefined` branch (`session-store.ts:177-181`) and fall back to flat history. The feature labeled "G10 history segmentation" in the PR title/commit does nothing at runtime, and the passing tests give false confidence that it is live.

**Required to unblock:** either wire G10 into the handler — call `setLastBotReplySeq()` after the bot replies (around `index.ts:160`) and swap `buildHistoryPrefix` → `buildSegmentedHistoryPrefix` at `index.ts:138` — or remove the unused API from this PR and drop G10 from the scope/title so the changelog stays accurate.

## P2 — `buildSegmentedHistoryPrefix` ignores the numeric seq it is named after

Even once wired, the implementation does not use the `lastBotReplySeq` value to find the boundary. After the `=== undefined` check, the numeric value is never read again; the split point is computed purely from the **last `assistant` message** by role (`session-store.ts:183-190`), then sliced at that index (`:197-198`).

This is structurally unavoidable as written: `MessageRow` (`session-store.ts:23-26`) only carries `role` and `content` — there is no `seq`/`message_seq` column on the rows to match against. The docstring "Messages before lastBotReplySeq are labeled [answered history]" (`:167-171`) therefore describes behavior the code does not implement. If a session ever has multiple assistant turns, or produces assistant output after the tracked reply seq, the role-derived boundary will diverge from the intended seq boundary. Tracking a seq and then splitting by role is a smell worth resolving before this ships.

---

## Verified correct ✅

### G9 — Space isolation
- `sessionKey()` embeds the spaceId for DMs (`session-router.ts:90-96`) and is invoked in the routing path (`:76` in `routeAndHandle`, `:86` in `route`); production wires it via `index.ts:110` → `result.sessionKey` → `store.getOrCreate(...)`.
- `extractSpaceId()` (`:103-127`) parses `s{spaceId}_{peer}` from `from_uid`, then falls back to `channel_id`, then to `""`. Returning `""` for group-like channels is intentional (groups isolate via `channel_id`).

### G21 — streamOn filter
- Parsed from the in-body setting byte: `parseSettingByte` returns `streamOn: ((v >> 1) & 0x01) > 0` (`socket.ts:214-219`), read at `socket.ts:661-662`, attached to `BotMessage` at `socket.ts:708`, typed at `types.ts:25`, and filtered at `session-router.ts:153` (`if (msg.streamOn) return null`).
- Bit positions match the WuKongIM wire protocol exactly: `SettingReceiptEnabled = 1<<7`, `SettingTopic = 1<<3`, `SettingStream = 1<<1` (`socket.ts:216-218`). No collision with the RECV frame-header reddot bit at `socket.ts:599` — that is a separate byte. The topic-conditional payload read (`socket.ts:674`) correctly mirrors the protocol so the setting byte is not misaligned.

---

## Recommendation

`CHANGES_REQUESTED` — G9 and G21 are good to merge. G10 should be either wired into the runtime (and its seq-vs-role boundary logic reconciled) or removed from this PR's scope before merge, so the delivered behavior matches the advertised feature set.

lml2468 (commented):
QA Re-review: APPROVED ✅

Three P1 fixes verified on head branch (37b281a). Local: tsc clean, 291/291 pass.

**P1.1 G10 dead code → wired up**
- index.ts:138 now calls buildSegmentedHistoryPrefix (was the old buildHistoryPrefix)
- index.ts:163 calls setLastBotReplySeq(msg.message_seq) on successful assistant reply
- e2e simulator mirrors the exact same logic — keeps tests honest

**P1.2 lastBotReplySeq actually drives segmentation now**
- messages table: ALTER TABLE adds message_seq column with PRAGMA-guarded migration for pre-G10 DBs
- appendUser/appendAssistant accept optional messageSeq; persisted to DB
- buildSegmentedHistoryPrefix splits strictly on `r.message_seq > lastReplySeq` — no more "last assistant position" cheat
- NULL message_seq (legacy rows / non-tracked assistant) attaches to the active side via seenNew flag — sensible
- When no new messages since last reply, segmentation labels are omitted (avoids confusing empty [new messages] section)

**P1.3 streamOn filter at both layers**
- session-router.ts: `if (msg.streamOn) return null;` defensive gate
- index.ts:194: `!msg.streamOn` check before pushMessage to groupContext
- socket.ts: parses streamOn from settingByte bit 1 onto BotMessage
- e2e test verifies streaming partial NOT cached + final IS cached

**Test scope**: 27 new tests (20 G9/G10/G21 + 4 segmentation + 3 e2e). Coverage:
- segment by real seq ✅
- no-new-messages omits labels ✅
- NULL seq follows active side ✅
- DB persists message_seq ✅
- streamOn=true skipped, false processed ✅
- e2e roundtrip: 1st turn records seq, 2nd turn segments ✅

### 💬 Non-blocking notes

- G9 spaceId extraction uses lastIndexOf("_") on from_uid — works for canonical `s{spaceId}_{peerId}`. UIDs with underscores in the peerId portion (uncommon but possible) would split at the wrong place. Edge case, document or revisit if hit.
- Migration check uses try/catch + console.warn — silent on failure. Probably fine for forward-only schema, but ALTER TABLE failure on a real DB would mean message_seq column missing → all subsequent inserts throw. Consider failing loud on migration errors.
- The setting bit 1 mapping (streamOn) lives in socket.ts now — worth a one-line ARCHITECTURE.md note since multiple modules now depend on this bit semantics.

Ready to merge.

— guoshouwei QA

lml2468 (commented):
Security Review LGTM. P1.1 segmented prefix wired in production + e2e. P1.2 message_seq column + real seq-based split + backward-compat migration + 4 new tests. P1.3 double-gate streamOn (router + cache). Non-blocking note: extractSpaceId lastIndexOf could collide if Octo peerIds ever contain underscores (current UUIDs safe). Approving.

lml2468 (commented):
## ✅ APPROVED by 李飞飞 (re-review of commit 37b281a)

GitHub 限制：lml2468 是 PR 推送凭证，无法用 approve state，只能用 comment——这条消息等同于 APPROVED。

三个 P1 全修：

**P1.1 G10 wire-up** ✅
- index.ts:147 已切到 `buildSegmentedHistoryPrefix`
- index.ts:175 成功回复后调 `setLastBotReplySeq(sessionKey, msg.message_seq)`
- e2e simulator 同步改

**P1.2 真按 message_seq 切分** ✅
- 加 `message_seq INTEGER` 列 + ALTER TABLE 迁移兼容旧 DB
- `appendUser/Assistant` 接受 `messageSeq?: number`
- `buildSegmentedHistoryPrefix` 严格按 `seq > lastReplySeq` 切，不再用 last-assistant 偷懒
- NULL seq（assistant reply / legacy 行）跟随 `seenNew` 状态——边界正确
- 没新消息时不显示分段标签，避免假阳性

**P1.3 streamOn 缓存泄漏** ✅
- index.ts:209 `!msg.streamOn` 守卫在 pushMessage 之前
- e2e simulator 同步改

**测试** ✅ 7 个新测试：
- 4 segmentation 单测（真按 seq / 无新消息无标签 / NULL seq 跟随 / DB 持久化）
- 3 e2e（streamOn=true 不缓存 / streamOn=false 缓存 / segmentation 端到端 roundtrip）
- 总测试 282→291 全过

**细节验证**：assistant reply 用 user 的 msg.message_seq 作为其 seq——同一 turn 的 user+assistant 始终在分区同侧，正确。

PR#30 可以合入。

PR #31, opened 2026-06-05
feat(G3,G8,G12,G13): reply context + read receipt + mention-free + @botname strip

lml2468:
## Summary

Implements 4 feature GAPs in a single commit.

### G3: Reply/quoted message context
- In `handleMessage`, extract `msg.payload.reply` and prepend a `[Quoted message from <name>]: <content>` prefix to the user content sent to the LLM.
- Original `userContent` is still stored in history unchanged.

### G8: Read receipt
- Added `sendReadReceipt` to `src/octo/api.ts` (POST `/v1/bot/readReceipt`).
- In `handleMessage`, after `streamRelay.deliver()` completes, fire-and-forget a read receipt for the processed message.

### G12: Mention-free mode
- Added `mentionFreeGroups?: string[]` to `Config` interface, `PartialConfig`, `mergeConfig`, and `applyEnv` (env var: `CC_OCTO_MENTION_FREE_GROUPS`).
- In `SessionRouter.processMessage`, the group mention gate now allows through messages from groups listed in `mentionFreeGroups` even without an @mention. System events, rate limits, and non-text checks still apply.

### G13: Strip @botname from group messages
- Added `cleanContent?: string` to `RouteResult`.
- In `processMessage`, after all checks pass, strip the leading @botname from group messages using entities-based removal (precise offset/length) with regex fallback. Empty result falls back to original content.
- In `handleMessage`, use `result.cleanContent ?? userContent` as the LLM input.

## Tests
- New `src/__tests__/g3-g8-g12-g13.test.ts` with 9 unit tests covering all 4 features.
- Updated `src/__tests__/e2e.test.ts`: added `sendReadReceipt` to mock, updated `simulateMessage` to mirror `index.ts` (G3/G8/G13), added G3 quote context test and G8 read receipt test.

## Verification
- `npx tsc --noEmit`: zero errors
- `npx eslint src/`: no new warnings (one pre-existing any in q13-q14-q17.test.ts)
- `npx vitest run`: 277/277 tests pass across 18 test files

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ Three-level priority: **environment variables** > **config.json** > **defaults**
 | `context.maxContextChars` | `CC_OCTO_CONTEXT_MAX_CHARS` | `6000` | Max characters of group context injected into prompts |
 | `context.historyLimit` | `CC_OCTO_CONTEXT_HISTORY_LIMIT` | `40` | Max messages in session history window |
 | `botBlocklist` | `CC_OCTO_BOT_BLOCKLIST` | `[]` | Comma-separated bot UIDs to ignore in DMs (prevents bot loops) |
+| `mentionFreeGroups` | `CC_OCTO_MENTION_FREE_GROUPS` | `[]` | Comma-separated group channel IDs where the bot responds to every text message without requiring an `@bot` mention (G12). |
 
 ## Security Model
 
--- a/config.example.json
+++ b/config.example.json
@@ -21,5 +21,8 @@
     "historyLimit": 40
   },
 
-  "botBlocklist": []
+  "botBlocklist": [],
+
+  "_comment_mention_free": "G12: group_ids listed here process every text message without requiring an @bot mention. Use group channel IDs as strings. Env override: CC_OCTO_MENTION_FREE_GROUPS (csv).",
+  "mentionFreeGroups": []
 }
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -26,6 +26,7 @@ const CC_VARS = [
   'CC_OCTO_SDK_MAX_TURNS', 'CC_OCTO_SDK_SYSTEM_PROMPT', 'CC_OCTO_SDK_SETTING_SOURCES',
   'CC_OCTO_RATE_LIMIT_MAX_PER_MINUTE', 'CC_OCTO_CONTEXT_MAX_CHARS',
   'CC_OCTO_CONTEXT_HISTORY_LIMIT', 'CC_OCTO_BOT_BLOCKLIST',
+  'CC_OCTO_MENTION_FREE_GROUPS', 'CC_OCTO_MAX_RESPONSE_CHARS',
 ];
 
 function setup() {
@@ -250,6 +251,12 @@ describe('CC_OCTO_* env override coverage', () => {
     const path = writeConfig({ botToken: 'bf_t', apiUrl: 'https://a' });
     process.env.CC_OCTO_BOT_BLOCKLIST = 'bot-a, bot-b, bot-c';
     expect(loadConfig(path).botBlocklist).toEqual(['bot-a', 'bot-b', 'bot-c']);
+  });
+
+  it('CC_OCTO_MENTION_FREE_GROUPS (csv) overrides config (G12)', () => {
+    const path = writeConfig({ botToken: 'bf_t', apiUrl: 'https://a' });
+    process.env.CC_OCTO_MENTION_FREE_GROUPS = 'group-1, group-2,group-3';
+    expect(loadConfig(path).mentionFreeGroups).toEqual(['group-1', 'group-2', 'group-3']);
   });
 });
 
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -23,7 +23,7 @@ vi.mock('../octo/api.js', () => ({
   }),
   generateClientMsgNo: vi.fn().mockReturnValue('client-msg-001'),
   fetchUserInfo: vi.fn().mockResolvedValue(null),
-}));;
+}));
 
 vi.mock('../agent-bridge.js', async (importOriginal) => {
   const original = await importOriginal<typeof import('../agent-bridge.js')>();
@@ -164,15 +164,24 @@ async function simulateMessage(
 
     const userContent = msg.payload.content ?? '';
 
-    // G3: Extract quoted/replied message content for LLM context
+    // G3: Extract quoted/replied message content for LLM context (truncated to 4KB).
     let quotePrefix = '';
     const replyData = msg.payload?.reply;
     if (replyData) {
-      const replyPayload = replyData.payload;
-      const replyContent = replyPayload?.content ?? '';
+      const replyPayload = replyData?.payload;
+      const rawReplyContent = replyPayload?.content ?? '';
       const replyFrom = replyData.from_name ?? replyData.from_uid ?? 'unknown';
-      if (replyContent) {
-        quotePrefix = `[Quoted message from ${replyFrom}]: ${replyContent}\n---\n`;
+      if (rawReplyContent) {
+        const QUOTE_MAX_BYTES = 4_096;
+        let truncated = rawReplyContent;
+        if (Buffer.byteLength(rawReplyContent, 'utf-8') > QUOTE_MAX_BYTES) {
+          truncated = rawReplyContent.slice(0, QUOTE_MAX_BYTES);
+          while (Buffer.byteLength(truncated, 'utf-8') > QUOTE_MAX_BYTES) {
+            truncated = truncated.slice(0, -1);
+          }
+          truncated += '…[truncated]';
+        }
+        quotePrefix = `[Quoted message from ${replyFrom}]: ${truncated}\n---\n`;
       }
     }
     const userContentForLLM = quotePrefix + (result.cleanContent ?? userContent);
@@ -547,6 +556,34 @@ describe('E2E smoke tests', () => {
     expect(userMsg).toContain('[Quoted message from Alice]');
     expect(userMsg).toContain('the original code');
     expect(userMsg).toContain('what does this mean');
+  });
+
+  // P1 regression: oversized reply payload is truncated before being prepended
+  // to user content. Without this guard, a small current message could carry
+  // an unbounded payload.reply.payload.content into queryAgent, bypassing the
+  // 32KB content guard in session-router.
+  it('G3 P1: truncates oversized quoted reply to ~4KB', async () => {
+    const hugeReply = 'X'.repeat(50_000);
+    const msg = makeDmMsg('tell me about this', {
+      payload: {
+        type: MessageType.Text,
+        content: 'tell me about this',
+        reply: {
+          from_uid: 'other-user',
+          from_name: 'Alice',
+          payload: { type: MessageType.Text, content: hugeReply },
+        },
+      },
+    });
+    await simulateMessage(msg, config, store, router, groupContext, streamRelay);
+
+    expect(queryAgent).toHaveBeenCalledTimes(1);
+    const userMsg = (queryAgent as ReturnType<typeof vi.fn>).mock.calls[0][0] as string;
+    // Truncation marker present.
+    expect(userMsg).toContain('[truncated]');
+    // Total prompt size is bounded (4KB cap + small framing overhead).
+    expect(userMsg.length).toBeLessThan(5_000);
+    expect(userMsg).toContain('tell me about this');
   });
 
   // --- G8: Read receipt ---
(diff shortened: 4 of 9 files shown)


Jerry-Xin (changes requested):
Summary: The PR is relevant to `cc-channel-octo`, but the new mention-stripping logic can corrupt valid group messages and should be fixed before merge.

**🔴 Blocking**
- 🔴 Critical: `src/session-router.ts:180` strips any leading `@...` token from every group message when entity-based stripping did not run. This is not limited to the bot mention. In mention-free groups, a normal message like `@alice please check this` will be sent to the agent as `please check this`; in mentioned groups, `@alice ... @bot ...` can also lose the wrong mention. This violates the G13 intent of stripping `@botname` only. The fallback should only remove a mention known to be the bot, or avoid stripping when identity cannot be proven.
- 🔴 Critical: `src/index.ts:141`-`150` adds quoted reply content to the agent prompt without applying the existing 32 KB guard from `src/session-router.ts:160`-`164`. A small current message can therefore carry a very large `payload.reply.payload.content` into `queryAgent`, bypassing the message-size protection and risking excessive prompt size or resource usage. The quoted prefix should be bounded, truncated, or included in the same size validation as the final prompt sent to the agent.

**💬 Non-blocking**
- 🟡 Warning: `README.md:84`-`99` and `config.example.json:24` do not document `mentionFreeGroups` / `CC_OCTO_MENTION_FREE_GROUPS`, so operators have no discoverable way to enable G12 from the project docs.
- 🟡 Warning: `src/__tests__/config.test.ts:27`-`31` does not include `CC_OCTO_MENTION_FREE_GROUPS` in `CC_VARS`, and there is no config test for the new env override. This can leave env pollution between tests and misses coverage for the new setting.

**✅ Highlights**
- The entity-based mention stripping path uses UTF-16 offsets consistently with the declared type.
- The read receipt API is isolated behind `sendReadReceipt`, matching the existing API helper pattern.

Validation note: I attempted `npm run type-check`, `npm test -- --runInBand`, and `npm run lint`, but this checkout has no installed dev binaries (`tsc`, `vitest`, and `eslint` were not found).

yujiawei (changes requested):
# Code Review — PR #31 (cc-channel-octo)

## Verdict: CHANGES_REQUESTED

Solid, well-tested PR — four cleanly-scoped features, `tsc --noEmit` clean, and 277/277 tests pass. One blocking issue: a feature interaction between **G12** and **G13** that silently corrupts user input sent to the LLM. The fix is small. The rest below are non-blocking suggestions.

---

## P1 — Blocking

### G13 + G12 interaction: regex fallback strips a non-bot leading @mention

`src/session-router.ts:181-182`

```ts
// Fallback: regex strip leading @word
if (cleanContent === content) {
  cleanContent = content.replace(/^@\S+\s*/, '').trim();
}
```

The G13 stripping block runs for **every** group-like message that reaches `shouldProcess` — and with G12 (mention-free groups) added in this same PR, that now includes messages where the bot was **not** mentioned at all. In that case the entities-based check (`e.uid === this.robotId && e.offset === 0`, lines 173-178) correctly finds nothing, `cleanContent` stays equal to `content`, and the regex fallback fires — stripping whatever `@token` happens to be first.

Concrete failure (mention-free group):

- User sends `@张三 你看下这个` (addressing a teammate, not the bot).
- No bot entity → fallback runs → LLM receives `你看下这个`.
- The bot loses the "addressed to 张三" context and may answer a message that wasn't directed at it.

The same misfire also exists in a normal group when the bot mention isn't first, e.g. `@alice @bot help` → the offset-0 entity check fails for the bot, fallback strips `@alice` instead of `@bot`.

This is a silent input-corruption bug, and the scenario it bites is exactly the one G12 enables. Suggested fix — skip the fallback when the bot was not mentioned (i.e. in the mention-free path), or only strip a leading @token verified to be the bot:

```ts
// Only fall back when the bot is actually the leading mention.
if (cleanContent === content && this.isMentioned(msg)) {
  cleanContent = content.replace(/^@\S+\s*/, '').trim();
}
```

(There is also no test covering this case — the existing G13 regex test uses `@bot ...` where the bot *is* first, so the bug is invisible to the suite. Please add a mention-free-group test asserting a leading non-bot @mention is preserved.)

---

## P2 — Non-blocking (recommend addressing)

### 1. G3 quote content bypasses the `MAX_CONTENT_BYTES` cap
`src/index.ts:137-150`, vs `src/session-router.ts:161-164`

The 32 KB length gate validates only `msg.payload.content`. The G3 quote prefix is built from `msg.payload.reply.payload.content` and prepended to `userContentForLLM` without any size check, so combined input to `queryAgent` can reach ~2× the cap. The quoted message is itself a previously-sent (bounded) message, so this is a soft-limit relaxation rather than unbounded amplification — but it does violate the documented contract. Consider validating combined size, or truncating the quoted content, before calling `queryAgent`.

### 2. G3 quote context not persisted in history (asymmetry undocumented)
`src/index.ts:150-153`

`userContentForLLM` (quote + cleaned content) goes to the LLM, but `store.appendUser(sessionKey, userContent)` stores only the raw original. Storing the raw message is a reasonable choice (avoids prefix duplication on replay), but on later turns the reply relationship is gone from history. At minimum add a one-line comment explaining the intent.

### 3. Inconsistent optional chaining on reply payload
`src/index.ts:143`

`const replyPayload = replyData.payload;` — `ReplyPayload.payload` is optional (`src/octo/types.ts:61`), and the next line already uses `replyPayload?.content`. No runtime bug, but make line 143 `replyData?.payload` for consistency.

### 4. Read receipt skipped on the error path
`src/index.ts:171-180`

The G8 read receipt sits inside the `try` block after `streamRelay.deliver()`. If delivery (or anything before it) throws, no receipt is sent and the `catch` doesn't send one either. Decide the intended semantics ("read" = "successfully processed" vs "received") and either move it to a `finally` or document the choice.

### 5. Missing test plumbing for `CC_OCTO_MENTION_FREE_GROUPS`
`src/__tests__/config.test.ts`

The new env var is not in the `CC_VARS` save/restore list (test-isolation risk) and has no parsing test, unlike its sibling `CC_OCTO_BOT_BLOCKLIST`. Add both.

---

## Nits

- `src/session-router.ts:177` vs `:182` — entity path uses `trimStart()`, regex path uses `trim()`. Pick one for consistency.
- `src/__tests__/e2e.test.ts:26` — double semicolon `}));;`.

---

## What's good

- Clean role separation maintained (`userContentForLLM` → user role, history/context → system) — G3 doesn't regress the Q3 fix.
- G12 correctly preserves all downstream gates (blocklist, system-event, rate-limit, non-text, length) for mention-free groups; the ordering is right and tested.
- G8 endpoint/payload shape is consistent with the rest of `octo/api.ts`, fire-and-forget with error logging is appropriate for a non-critical signal.
- The rejected-finding checks confirmed: non-text quote handling and empty-`channel_id` coalescing are correct/unreachable, not bugs.

Once the G13 fallback is guarded (and ideally covered by a test), this is good to merge.

lml2468 (commented):
**Security Review — LGTM**

## G13 fix ✅
`cleanContent === content && this.isMentioned(msg)` properly gates the regex fallback. In mention-free groups, if bot isn't mentioned, no stripping happens — "@alice please check" stays intact. The test for this is correct.

Documented edge case (`@alice @bot help` where bot not at offset 0 → regex strips `@alice `): accepted. Bot was explicitly @'d so the user knew they were addressing the bot; the security concern (mention-free group leaking user @ another user) is closed.

## G3 fix ✅
4KB byte-safe truncation with `…[truncated]` marker. The shrink loop correctly handles multi-byte UTF-8 (CJK chars don't get half-truncated). Total user prompt bounded ~36KB max (32KB content + 4KB quote + framing) — respects the original 32KB user content guard.

`rawReplyContent ?? ''` safely handles missing nested fields. Storing raw `userContent` (without quote prefix) for replay avoids duplication on subsequent turns — good call.

## Tests ✅
- e2e: integration test for 50KB reply → <5KB prompt
- regression test for mention-free + non-bot @mention
- multi-mention with bot not at offset 0 edge case documented
- byte-safe truncation marker check

Non-blocking note: the regex fallback's "strip first @token when bot is mentioned" heuristic could be tightened later by checking if ANY entity matches the bot and stripping only that range. Current behavior is safe but slightly lossy in multi-mention scenarios. Not a blocker.

yujiawei (approved):
# Code Review — PR #31 (cc-channel-octo)

## Verdict: APPROVED

This PR implements 4 well-scoped feature gaps (G3/G8/G12/G13) cleanly and with strong test coverage. I verified the head SHA locally (`b3abbe0`), built it, and ran the full suite: **293/293 tests pass**, `tsc --noEmit` is clean, and `eslint src/` reports only the single **pre-existing** `any` warning in `q13-q14-q17.test.ts` (untouched by this PR). No P0/P1 blockers found. The items below are P2/nits — none of them block merge.

## What was verified (cross-file)

- **G3 (reply quote context)** — `index.ts:139-171`. Quote body is byte-truncated to a 4 KB budget (UTF-8 safe: char slice then byte-trim loop) and prepended only to the LLM input; `store.appendUser` persists the raw content (`index.ts:174`), so replay isn't polluted with the prefix. Good separation. The corresponding `ReplyPayload` type exists in `types.ts:60-64`, so `from_name`/`payload.content` access is type-safe.
- **G8 (read receipt)** — `api.ts:210-223` follows the existing `postJson` helper contract; the call site (`index.ts:192-201`) is fire-and-forget with a `.catch()`, so no unhandled rejection and no impact on the response path. The endpoint `/v1/bot/readReceipt` matches the upstream surface that the fork header notes was removed, so re-adding it on the same path is consistent.
- **G12 (mention-free groups)** — `session-router.ts:130-137`. The mention gate now lets configured groups through without an `@`, while system-event, rate-limit, non-text, and 32 KB content checks all still run *after* the gate. Config plumbing (`config.ts`) mirrors the existing `botBlocklist` pattern across `Config`, `PartialConfig`, `mergeConfig`, and `applyEnv` (`CC_OCTO_MENTION_FREE_GROUPS`).
- **G13 (@botname strip)** — `session-router.ts:167-196`. Entities-based path uses precise offset/length (UTF-16 units, matching `String.substring`), and the regex fallback is now correctly gated behind `isMentioned(msg)`, so a leading `@teammate` in a mention-free group is no longer stripped. The empty-result guard preserves the original content.
- **Test mirror integrity** — `e2e.test.ts` `simulateMessage` faithfully mirrors the production `handleMessage` (quote prefix, `cleanContent ?? userContent`, read-receipt block), so the e2e assertions actually exercise equivalent logic rather than a divergent copy.

## Findings (all P2 / non-blocking)

### P2-1 — G13 regex fallback strips the *wrong* mention when the bot isn't first
`session-router.ts:188-193`. When entities are present but the bot is not at offset 0 (e.g. `"@alice @bot help"` with bot entity at offset 7), Path 1 finds nothing (`offset === 0` required) and Path 2 runs because the bot *is* mentioned — the regex then strips the leading `@alice`, yielding `"@bot help"`. This both removes the wrong addressee and fails G13's own goal (the `@bot` token survives). The test at `g3-g8-g12-g13.test.ts:347-377` documents this as an accepted limitation, which is honest, but it produces corrupted LLM input in a realistic case.

Suggested fix: when `mention.entities` is present, treat it as authoritative and skip the regex fallback entirely (entities are precise; if they don't show the bot leading, there is nothing safe to strip):
```ts
if (cleanContent === content && !mention?.entities && this.isMentioned(msg)) {
  cleanContent = content.replace(/^@\S+\s*/, '').trimStart();
}
```

### P2-2 — Stale header comment now contradicts the file
`api.ts:1-4` still lists `read receipts` under "Removed:", directly contradicting the new `sendReadReceipt` added in this same PR (`api.ts:208-223`). Drop "read receipts" from that line to avoid future confusion.

### P2-3 — Read-receipt timing is post-generation (minor UX)
`index.ts:192` fires the receipt only *after* `streamRelay.deliver()` completes. For a slow agent response the "read" signal is delayed by the full generation time. If the intent is "we saw your message," sending it before/at the start of processing would feel more responsive. Non-blocking — current placement is defensible as "fully handled."

### P2-4 — Regex fallback over-strips space-less CJK mentions (cosmetic, self-healing)
`session-router.ts:192`. `^@\S+\s*` is greedy on non-whitespace, so a space-less mention like `@bot你好` matches the whole token and empties the result, which the `if (!cleanContent) cleanContent = content` guard then reverts to the raw string. Net effect: no corruption, but no cleaning either in that case. The entities path covers this correctly when entities are provided, so impact is limited.

## Summary
Solid, focused, well-tested change. The features behave correctly on the common paths and the P1 feedback from the prior round (regex guard + quote truncation) was addressed in `b3abbe0`. Approving; please consider P2-1 and P2-2 as quick follow-ups.

lml2468 (commented):
## ✅ APPROVED by 李飞飞

GitHub 限制：lml2468 是 PR 推送凭证，无法用 approve state，只能用 comment——这条消息等同于 APPROVED。

第二次 re-review 确认（针对 commit b3abbe0）：

**G13** ✅ `cleanContent === content && this.isMentioned(msg)` 双重 gate 正确。mention-free 群 + bot 未被 @ → 不剥任何 @mention（@alice 等保留）。已验证测试覆盖 P1 回归场景。

**G3** ✅ 4KB byte-safe truncation（while loop trim 单字符直到字节数达标）+ `…[truncated]` 标记。50KB reply 测试验证总 prompt < 5KB，不绕过 32KB content gate。store.appendUser 用 raw content 不带 prefix——replay 时不重复，正确。

**Tests** ✅ 4 P1 regression + 1 e2e integration + 1 config env，全过。

PR#31 可以合入。

PR #33, opened 2026-06-05
feat(G1,G2,G4,G11,G22): inbound resolution + file inlining + history backfill

lml2468:
## Summary

Second-batch GAP fixes for inbound message handling.

| GAP | Description | Implementation |
|-----|-------------|----------------|
| **G1** | Media message handling | New `src/inbound.ts` `resolveContent()` covers all 10 `MessageType` values (Text/Image/GIF/Voice/Video/File/Location/Card/MultipleForward/RichText) — replaces the previous '暂不支持' rejection |
| **G2** | Text-file inlining | `tryResolveFile()` HEAD-aware download with 20KB inline cap and 5MB hard cap; .py/.ts/.js/.json/.md/.csv (25+ extensions) get inlined into the prompt; oversize text files stream to `/tmp/cc-channel-octo/inbound-files/` |
| **G4** | API history backfill | New `getChannelMessages()` over POST `/v1/bot/messages/sync` with base64 payload decoding; `index.ts` cold-start backfill for empty group sessions, one-shot per session via in-memory Set |
| **G11** | Enriched history format | `HistoricalMessage` type carries `from_uid/from_name/type/timestamp`; `resolveHistoricalMessagePlaceholder()` maps non-text types to compact replay placeholders; `seedHistoryFromApi()` populates SessionStore so subsequent turns hit cache |
| **G22** | Voice context | Voice payload now reaches the agent with `[语音消息] <url>` marker; full STT integration tracked separately |

## RichText (type=14) details

- Adds `MessageType.RichText = 14` plus `RichTextBlock` / block-type constants to `octo/types.ts`
- `resolveRichTextContent()` honors server-authoritative top-level `plain` when present, falls back to assembling text from content blocks
- Image blocks collected into `mediaUrls` (sanitized for non-string url defense)
- Legacy string-content compat preserved

## Session router changes

- `session-router.ts` no longer auto-replies '暂不支持此类消息' for non-text payloads — all types flow to `handleMessage` where `resolveContent` produces LLM-friendly text
- G13 `cleanContent` (@bot stripping) now scoped to `Text` payloads only; non-text returns `cleanContent: undefined` so the pipeline uses `resolveContent.text`
- Size limit (32KB) still enforced for Text payloads (media URLs are bounded by their own size)

## Files

| File | Lines | Change |
|------|-------|--------|
| `src/inbound.ts` | +487 (new) | `resolveContent`, `tryResolveFile`, RichText / MultipleForward expansion, file inlining |
| `src/octo/types.ts` | +41 | RichText enum + block types + ForwardMessage shapes |
| `src/octo/api.ts` | +83 | `getChannelMessages` + `HistoricalMessage` type |
| `src/index.ts` | +149/-19 | resolveContent integration, G2 file inline path, G4 backfill, context cache for non-text |
| `src/session-router.ts` | +25/-43 | Drop non-text rejection, scope cleanContent to Text |
| `src/__tests__/inbound.test.ts` | +345 (new) | 37 tests covering all MessageType + RichText / MultipleForward parsing |
| `src/__tests__/g4-history-api.test.ts` | +221 (new) | 8 tests covering API parsing, base64 decoding, error paths |
| `src/__tests__/g2-file-inline.test.ts` | +183 (new) | 7 tests covering inline path, temp-file path, size caps, errors |
| `src/__tests__/e2e.test.ts` | +12/-13 | Updated 'unsupported notice' test → 'G1 routes non-text to agent' |
| `src/__tests__/session-router.test.ts` | +9/-2 | Same update at the unit level |

## Verification

- `tsc --noEmit`: ✅ clean
- `eslint src/`: ✅ 0 new warnings on changed files (7 pre-existing in other people's files)
- `vitest run`: 372/372 pass (+52 new tests)
- Scope: only inbound parsing + index.ts handler — does **not** touch outbound (mention/sendMedia/RichText send) which belongs to 豆豆's G5+G6+G7+G24 PR

## Out of scope (intentional)

- History string-format change (sender/timestamp tags per row) — needs a schema migration coordinated with PR#30's `message_seq` column; will follow as a focused G11 phase-2 PR to avoid churn here
- Outbound mention construction (G7) — 豆豆's PR
- COS upload / sendMediaMessage (G5+G24) — 豆豆's PR
- RichText *send* (G6) — 豆豆's PR

## Coordination with 豆豆

Confirmed boundaries:
- 飞飞 owns inbound parsing (resolveContent, mention payload **reading**)
- 豆豆 owns outbound construction (sendMediaMessage, mention payload **writing**, RichText send)
- No file-level overlap expected

### DIFF
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -44,6 +44,7 @@ import { GroupContext } from '../group-context.js';
 import { StreamRelay } from '../stream-relay.js';
 import { createAdapter, type DbAdapter } from '../db-adapter.js';
 import { queryAgent } from '../agent-bridge.js';
+import { resolveContent } from '../inbound.js';
 import {
   sendMessage,
   sendReadReceipt,
@@ -164,6 +165,10 @@ async function simulateMessage(
 
     const userContent = msg.payload.content ?? '';
 
+    // G1: resolve non-text payloads through inbound.resolveContent.
+    const resolved = resolveContent(msg.payload, config.apiUrl);
+    const bodyText = result.cleanContent ?? resolved.text;
+
     // G3: Extract quoted/replied message content for LLM context (truncated to 4KB).
     let quotePrefix = '';
     const replyData = msg.payload?.reply;
@@ -184,7 +189,7 @@ async function simulateMessage(
         quotePrefix = `[Quoted message from ${replyFrom}]: ${truncated}\n---\n`;
       }
     }
-    const userContentForLLM = quotePrefix + (result.cleanContent ?? userContent);
+    const userContentForLLM = quotePrefix + bodyText;
 
     const historyPrefix = store.buildSegmentedHistoryPrefix(sessionKey, config.context.historyLimit);
     store.appendUser(sessionKey, userContent, msg.message_seq);
@@ -343,19 +348,23 @@ describe('E2E smoke tests', () => {
     expect(ctx).toContain('TestUser：Just chatting');
   });
 
-  // --- 4. Non-text message gets "unsupported" reply ---
+  // --- 4. Non-text message: G1 routes it through to the agent ---
 
-  it('Non-text DM message: replies with unsupported notice', async () => {
+  it('G1: non-text DM message is resolved and sent to the agent (not rejected)', async () => {
     const msg = makeDmMsg('', {
-      payload: { type: MessageType.Image, url: 'https://example.com/img.png' },
+      payload: { type: MessageType.Image, url: 'file/preview/img.png' },
     });
     await simulateMessage(msg, config, store, router, groupContext, streamRelay);
 
-    // queryAgent should NOT be called
-    expect(queryAgent).not.toHaveBeenCalled();
+    // queryAgent IS called — the image is resolved as "[图片] <url>" and the
+    // agent gets a chance to respond.
+    expect(queryAgent).toHaveBeenCalled();
+    const userMsg = (queryAgent as ReturnType<typeof vi.fn>).mock.calls[0][0] as string;
+    expect(userMsg).toContain('[图片]');
+    expect(userMsg).toContain('img.png');
 
-    // sendMessage called with unsupported notice
-    expect(sendMessage).toHaveBeenCalledWith(
+    // No “不支持” notice.
+    expect(sendMessage).not.toHaveBeenCalledWith(
       expect.objectContaining({
         content: '暂不支持此类消息，请发送文字',
       }),
--- a/src/__tests__/g2-file-inline.test.ts
+++ b/src/__tests__/g2-file-inline.test.ts
@@ -1,0 +1,183 @@
+/**
+ * G2 (file inlining) + G4 (history backfill) integration tests.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+describe('G2: file inlining via tryResolveFile', () => {
+  let originalFetch: typeof globalThis.fetch;
+
+  beforeEach(() => {
+    originalFetch = globalThis.fetch;
+  });
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch;
+  });
+
+  it('inlines small text file content', async () => {
+    const { tryResolveFile } = await import('../inbound.js');
+
+    const fileContent = 'def hello():\n    print("world")\n';
+    const encoder = new TextEncoder();
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(encoder.encode(fileContent));
+        controller.close();
+      },
+    });
+
+    globalThis.fetch = vi.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      body: stream,
+    } as unknown as Response);
+
+    const result = await tryResolveFile({
+      url: 'https://api.example.com/file/hello.py',
+      botToken: 'tok',
+      filename: 'hello.py',
+    });
+
+    expect(result).toHaveProperty('inlined');
+    if ('inlined' in result) {
+      expect(result.inlined).toBe(fileContent);
+    }
+  });
+
+  it('rejects non-text extensions with size description', async () => {
+    const { tryResolveFile } = await import('../inbound.js');
+
+    const result = await tryResolveFile({
+      url: 'https://api.example.com/file/photo.jpg',
+      botToken: 'tok',
+      filename: 'photo.jpg',
+      knownSize: 5_000,
+    });
+
+    expect(result).toHaveProperty('description');
+    if ('description' in result) {
+      expect(result.description).toContain('photo.jpg');
+      expect(result.description).toContain('4.9KB');
+    }
+  });
+
+  it('rejects files exceeding hard cap by knownSize', async () => {
+    const { tryResolveFile } = await import('../inbound.js');
+
+    const result = await tryResolveFile({
+      url: 'https://api.example.com/file/big.md',
+      botToken: 'tok',
+      filename: 'big.md',
+      knownSize: 10 * 1024 * 1024, // 10 MB > 5MB cap
+    });
+
+    expect(result).toHaveProperty('description');
+    if ('description' in result) {
+      expect(result.description).toContain('超过下载上限');
+    }
+  });
+
+  it('returns description on HTTP error', async () => {
+    const { tryResolveFile } = await import('../inbound.js');
+
+    globalThis.fetch = vi.fn().mockResolvedValue({
+      ok: false,
+      status: 404,
+      body: null,
+    } as unknown as Response);
+
+    const result = await tryResolveFile({
+      url: 'https://api.example.com/file/missing.md',
+      botToken: 'tok',
+      filename: 'missing.md',
+    });
+
+    expect(result).toHaveProperty('description');
+    if ('description' in result) {
+      expect(result.description).toContain('HTTP 404');
+    }
+  });
+
+  it('returns description on network error', async () => {
+    const { tryResolveFile } = await import('../inbound.js');
+
+    globalThis.fetch = vi.fn().mockRejectedValue(new Error('ECONNRESET'));
+
+    const result = await tryResolveFile({
+      url: 'https://api.example.com/file/foo.md',
+      botToken: 'tok',
+      filename: 'foo.md',
+    });
+
+    expect(result).toHaveProperty('description');
+    if ('description' in result) {
+      expect(result.description).toContain('网络错误');
+    }
+  });
+
+  it('extracts extension from URL when filename has none', async () => {
+    const { tryResolveFile } = await import('../inbound.js');
+
+    globalThis.fetch = vi.fn().mockResolvedValue({
+      ok: true,
+      status: 200,
+      body: new ReadableStream({
+        start(c) {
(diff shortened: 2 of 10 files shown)


yujiawei (changes requested):
# Code Review — PR #33 (cc-channel-octo)

**Scope:** `feat(G1,G2,G4,G11,G22)` — inbound message resolution + file inlining + history backfill. 10 files, single commit `de67a60`. Reviewed against `main` (merge-base `86b3092`).

**Verdict: CHANGES_REQUESTED** — one P0 (credential exfiltration / SSRF) plus several P1 correctness/security issues must be addressed before merge.

**Build/test status (verified locally):** `tsc --noEmit` clean ✅, `vitest run` 372/372 pass ✅, eslint clean on changed files ✅. The functional happy paths are well tested; the issues below are mostly adversarial/edge paths the test suite does not cover.

---

## P0 — Blocking

### 1. SSRF + bot-token exfiltration via attacker-controlled file URL
`src/inbound.ts:75` (`buildMediaUrl`) + `src/inbound.ts:400-403` (`tryResolveFile`)

```ts
export function buildMediaUrl(relUrl?: string, apiUrl?: string): string | undefined {
  if (!relUrl) return undefined;
  if (relUrl.startsWith('http://') || relUrl.startsWith('https://')) return relUrl; // passthrough, no host check
  ...
}

// tryResolveFile
const resp = await fetch(url, {
  headers: { Authorization: `Bearer ${botToken}` },
  signal: AbortSignal.timeout(FILE_DOWNLOAD_TIMEOUT_MS),
});
```

`payload.url` is fully client-controlled (`MessagePayload.url`, populated from the inbound socket frame in `octo/socket.ts`). `buildMediaUrl` returns any absolute `http(s)` URL **unchanged**, and `tryResolveFile` then issues a `fetch` to that URL **with the bot's secret `Authorization: Bearer <botToken>` header attached**. A group member can send a `File` message whose `url` is `https://attacker.example/collect` and the bot will deliver its credential to the attacker's server.

This is especially clear-cut because the codebase already takes SSRF seriously elsewhere: `src/config.ts:233-263` rejects a non-allowlisted `apiUrl` ("SSRF protection"), and `config.ts:81` warns that the config file contains the `botToken` secret. The new download path bypasses that same protection.

**Required fix:** before attaching the bot token, validate that the resolved URL's origin matches the configured `apiUrl` host (or a dedicated storage-host allowlist). Only send the `Authorization` header to trusted origins; for any other host either refuse the download or fetch without credentials. The same `buildMediaUrl` passthrough also feeds Image/GIF/Voice/Video/RichText URLs that are handed to the LLM/agent — those should be constrained too.

---

## P1 — Should fix before merge

### 2. Prompt injection via inlined file contents
`src/index.ts:160-161`

```ts
if ('inlined' in fileResult) {
  bodyText = `[文件: ${filename}]\n\n--- 文件内容 ---\n${fileResult.inlined}\n--- 文件结束 ---`;
```

The raw file body (up to 20KB, attacker-supplied) is concatenated into `bodyText` → `userContentForLLM` → `queryAgent(...)` (`index.ts:208,245`) with no escaping or delimiter hardening. A `.md`/`.py`/etc. file can carry instructions that override the conversation. There is a soft mitigation (`SECURITY_PROMPT_PREFIX` in `agent-bridge.ts:28`), but that is a probabilistic defense, and the agent has Bash + filesystem tools, so blast radius is high. Consider clearer untrusted-content fencing and noting this is *file content from an untrusted user* in the prompt.

### 3. Dangerous extensions preserved in temp file path handed to the agent
`src/inbound.ts:445-446`

```ts
const safeName = filename.replace(/[^a-zA-Z0-9._-]/g, '_') || 'file';
const tempPath = join(TEMP_DIR, `${randomUUID()}-${safeName}`);
```

Path traversal is correctly neutralized (slashes become `_`), but executable extensions (`.sh`, `.py`, `.js`) survive, and the full temp path is then surfaced to the agent (`index.ts:163`). Combined with #2 and the agent's Bash tool, a user-uploaded script plus a "please run this" message is a realistic RCE-by-social-engineering chain. Recommend forcing a non-executable suffix (e.g. append `.txt`) on temp files, and/or instructing the agent never to execute user-uploaded files.

### 4. G4 backfill stores every historical message as `user` role
`src/index.ts:348-374` (`seedHistoryFromApi`)

All API-backfilled messages — **including the bot's own past replies** — are persisted via `store.appendUser(...)`. `buildSegmentedHistoryPrefix` then renders them as `[user]:` (`session-store.ts`). The comment acknowledges this ("we don't know which side … all historical messages are stored as user role") and `currentUserUid` is accepted but `void`-ed. The result is that on group cold-start the model sees its own prior answers labelled as user turns, which corrupts the structural role separation the rest of the pipeline (Q3) is built around. Either resolve sender→role using the bot's `robot_id` (available from registration), or don't backfill until role attribution is implemented.

### 5. G4 backfill likely double-inserts the current message
`src/index.ts:213-241`

The cold-start backfill (`getChannelMessages`, `pull_mode:1`, `start_message_seq:0`, limit 20) runs immediately before `store.appendUser(sessionKey, userContent, msg.message_seq)` at line 241. If the sync endpoint returns the just-received message (very likely, since it's the newest), it is seeded once in `seedHistoryFromApi` and appended again at line 241 — duplicating the current turn in history. Please confirm against the server's `messages/sync` semantics and de-dupe by `message_seq` if so.

### 6. RichText with empty `plain` and empty blocks sends an empty body to the agent
`src/inbound.ts:152-154` + `301-306`

```ts
const text = topPlain.trim() !== '' ? topPlain : buildRichTextPlain(blocks);
return { text, mediaUrls };          // text can be ''
...
case MessageType.RichText: {
  const rt = resolveRichTextContent(...);
  return { text: rt.text, ... };     // no fallback
}
```

`resolveInnerMessageText` (forward children) correctly uses `rt.text || '[图文消息]'` (line 185), but the top-level `resolveContent` RichText case does not. An empty/whitespace RichText therefore yields `bodyText = ''`, and the agent receives only the quote prefix (or nothing). Add the same `|| '[图文消息]'` fallback.

### 7. Location coordinate fields read with unguarded `as number`
`src/inbound.ts:275-281` (fields not declared on `MessagePayload`, `octo/types.ts:68-83`)

```ts
const lat = (payload.latitude ?? payload.lat) as number | undefined;
const lng = (payload.longitude ?? payload.lng ?? payload.lon) as number | undefined;
return { text: lat != null && lng != null ? `[位置信息: ${lat},${lng}]` : '[位置信息]' };
```

The `!= null` guard does not exclude objects/arrays, so malformed coordinates render as `[位置信息: [object Object],[object Object]]` into the prompt. The neighbouring Card case (lines 283-284) correctly uses `typeof === 'string'` guards — apply the same `typeof === 'number'` check here, and declare these fields on the type rather than casting.

### 8. Unvalidated `as unknown as` casts on MultipleForward/RichText payloads
`src/inbound.ts:293-296`, `301-302`

`resolveMultipleForwardText` does `const msgs = payload?.msgs ?? []` with no array check; if `msgs` arrives as a string, `for (const m of msgs)` iterates characters and emits a garbled transcript into the prompt. Add a structural guard (`Array.isArray`) before iterating, and validate each child `m.payload` before reading `.type`/`.from_uid`.

---

## P2 / Nits (non-blocking)

- **Binary-as-UTF-8 inlining** (`inbound.ts:438-439`): `buf.toString('utf-8')` on a `.py`/`.json` file containing binary bytes yields mojibake (U+FFFD) injected into the prompt. Consider validating UTF-8 (reject on replacement chars) before inlining.
- **`backfilledSessions` grows unbounded** (`index.ts:340-341`): module-global `Set` never pruned; a long-lived process accumulates one entry per session. Bound it (LRU) or move the sentinel into the store with TTL.
- **Blocking network I/O under the session lock** (`index.ts:112-239`): file download (up to 30s) and G4 backfill run inside `routeAndHandle`, serializing the session. Acceptable for now but worth noting for latency.
- **Session router no longer size-checks non-text** (`session-router.ts:244-253`): File size is only enforced downstream in `tryResolveFile` (`knownSize > MAX_FILE_DOWNLOAD_BYTES`, which `NaN`/negative bypass — though the streaming byte counter at `inbound.ts:424` still caps it). Defense-in-depth only.
- **MultipleForward recursion has no depth cap** (`inbound.ts:193-215`): server-bounded in practice, but a `maxDepth` guard is cheap insurance.
- **Dead `description` field on the `tempPath` branch** (`inbound.ts:471`): the cast-spread `description` is never read by the caller (`index.ts:162`). Return just `{ tempPath }`.
- **`seedHistoryFromApi` misleading comment + unused param** (`index.ts:348-374`): the "heuristic" comment describes logic that isn't implemented; `currentUserUid` is `void`-ed. Tie to fix #4 or simplify the comment.
- **G4 cold-start segmentation** (`index.ts:213-234`): after seeding, `lastBotReplySeq` is undefined so the first turn falls back to flat history (no `[answered]/[new]` labels). Minor, by design.

---

## Security-sensitive sign-off note for a human reviewer

This PR was correctly flagged `security_sensitive`. The **must-verify** item before merge is finding #1 (token exfiltration to attacker-controlled hosts) — it is a direct secret-leak with a low-privilege trigger (any group member sending a file). Findings #2/#3 together form a plausible RCE-via-agent chain and warrant a human decision on the agent's tool permissions (Bash in default `allowedTools`, `config.ts:61`).

lml2468:
**Review: APPROVED with two non-blocking issues** ✅ (齐静春)

10 files, +1559/-62, 372/372 tests pass, tsc clean. inbound.ts well-structured: clean separation between sync rendering (resolveContent) and async file inlining (tryResolveFile). RichText/MultipleForward expansion mirrors upstream correctly.

### What's clearly right
- G1 inbound.ts: All 10 MessageType cases handled. RichText plain field preference + blocks fallback matches upstream. MultipleForward nested expansion preserves transcript structure.
- G2 file inlining: Streaming read with double size guard (20KB inline / 5MB temp). Stale temp file cleanup (1h TTL) is good housekeeping.
- G4 backfill: backfilledSessions Set marker set BEFORE await — race-safe. Cold-start only, falls through on failure.
- G13 fix retained: Router still only strips @bot when explicitly mentioned (the PR#31 P1 fix).
- Router refactor: cleanContent is optional, non-text payloads fall through to resolveContent in index.ts.

### Issue 1 (non-blocking): seedHistoryFromApi stores bot's own past replies as [user]
index.ts:348-374 — seedHistoryFromApi takes currentUserUid but immediately discards it with 'void currentUserUid' and appends EVERY historical message as user role. Bot's previous replies in the same group get stored as [user]: rows, next turn's buildSegmentedHistoryPrefix shows them with the [user] label. LLM sees its own past replies attributed to the user — confusing context.

Fix: pass robotId to seedHistoryFromApi and route bot's messages through appendAssistant instead.

### Issue 2 (non-blocking): inlined file content persisted in session history
index.ts:175 — userContent = msg.payload.content ?? bodyText for File messages, bodyText contains the full inlined file (up to 20KB). This goes into appendUser. With historyLimit=40, a user spamming files could grow the system prompt to ~800KB per turn.

Fix: when File payload is inlined, persist only the metadata ([文件: filename]) in appendUser, keep the full inlined content for the current turn's userContentForLLM only.

### Nit
inbound.ts:88 guessMime is marked unused with eslint-disable — fine for now but if G22 STT integration isn't coming soon, delete it.

Both issues are tail risks, not blockers. PR can land; suggest tracking as follow-up.

PR #34, opened 2026-06-05
feat(G24,G5,G6,G7): media + RichText + @mention send capabilities

lml2468:
Third-batch GAP. G24 API foundations. G5 media upload pipeline. G6 RichText combined send. G7 mention resolution. 386/386 tests pass.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -210,9 +210,40 @@ All 14 config fields are overridable via environment variables. See [README.md](
 
 ## Output Delivery
 
-The stream relay delivers agent output via plain `sendMessage` with intelligent text splitting. Split priority: paragraph break (`\n\n`) > newline (`\n`) > space > hard cut. Maximum segment size: 3500 characters.
+### Text
+
+The stream relay delivers agent text via plain `sendMessage` with intelligent text splitting. Split priority: paragraph break (`\n\n`) > newline (`\n`) > space > hard cut (surrogate-pair safe). Maximum segment size: 3500 characters. Per-segment try/catch ensures one failed segment does not drop the rest.
 
 A typing indicator heartbeat runs at 5-second intervals so the user sees activity during long agent runs.
+
+### @mentions (G7)
+
+Before each `sendMessage` call, the relay runs `resolveMentions()` from `mention-utils.ts`:
+- **Structured form** `@[uid:displayName]` (preferred, generated from system prompt) — converted to `@displayName` with precise `MentionEntity` payload.
+- **Plain form** `@displayName` — resolved via optional `memberMap` (displayName → uid). Longest-prefix match handles names containing spaces.
+- **`@all` / `@所有人`** — detected and surfaced as `mentionAll: true`.
+
+Resolved `mentionUids`, `mentionEntities`, and `mentionAll` are attached to the outbound payload so the IM server can route notifications correctly.
+
+### Media (G5/G24)
+
+`media-upload.ts` provides `uploadAndSendMedia()` for one-shot media delivery:
+1. Resolve input (`data:` / `file://` / `http(s)://`) into a file body, size, content-type, and dimensions (for images).
+2. Call `getUploadCredentials` to obtain short-lived COS STS credentials.
+3. Stream-upload via `cos-nodejs-sdk-v5` `putObject`; resolve CDN URL (re-encoded for non-ASCII keys).
+4. Call `sendMediaMessage` (Image with `width`/`height`/`name`/`size`, File-like with `name`/`size` only).
+
+500 MB max file size. Temp files cleaned in `finally`. PNG/JPEG/GIF/WebP dimension parsing reads only header bytes (no image decoding).
+
+### RichText (G6)
+
+`sendRichTextCombined()` packs text and `![alt](url)` markdown image references into a single `type=14` (RichText) message:
+- Parses markdown image refs, uploads each in parallel to COS.
+- Assembles interleaved `text` / `image` blocks.
+- Successful images become `{ type: "image", url, width, height }` blocks.
+- Failed images fall back inline as `[alt]` text inside the surrounding text block.
+- A `plain` field with `[图片]` placeholders is sent alongside for legacy-client compat (server recomputes from blocks authoritatively).
+- Returns `richText: false` when no images were uploaded successfully — caller falls back to plain text.
 
 ## Error Handling
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "better-sqlite3": "^12.10.0",
+        "cos-nodejs-sdk-v5": "^2.15.4",
         "crypto-js": "^4.2.0",
         "curve25519-js": "^0.0.4",
         "md5-typescript": "^1.0.5",
@@ -1926,6 +1927,24 @@
       "dev": true,
       "license": "Python-2.0"
     },
+    "node_modules/asn1": {
+      "version": "0.2.6",
+      "resolved": "https://registry.npmjs.org/asn1/-/asn1-0.2.6.tgz",
+      "integrity": "sha512-ix/FxPn0MDjeyJ7i/yoHGFt/EX6LyNbxSEhPPXODPL+KB0VPk86UYfL0lMdy+KCnv+fmvIzySwaK5COwqVbWTQ==",
+      "license": "MIT",
+      "dependencies": {
+        "safer-buffer": "~2.1.0"
+      }
+    },
+    "node_modules/assert-plus": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
+      "integrity": "sha512-NfJ4UzBCcQGLDlQq7nHxH+tv3kyZ0hHQqF5BO6J7tNJeP5do1llPr8dZ8zHonfhAu0PHAdMkSo+8o0wxg9lZWw==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=0.8"
+      }
+    },
     "node_modules/assertion-error": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/assertion-error/-/assertion-error-2.0.1.tgz",
@@ -1935,6 +1954,36 @@
       "engines": {
         "node": ">=12"
       }
+    },
+    "node_modules/asynckit": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
+      "integrity": "sha512-Oei9OH4tRh0YqU3GxhX79dM/mwVgvbZJaSNaRk+bshkj0S5cfHcgYakreBjrHwatXKbz+IoIdYLxrKim2MjW0Q==",
+      "license": "MIT"
+    },
+    "node_modules/atomically": {
+      "version": "1.7.0",
+      "resolved": "https://registry.npmjs.org/atomically/-/atomically-1.7.0.tgz",
+      "integrity": "sha512-Xcz9l0z7y9yQ9rdDaxlmaI4uJHf/T8g9hOEzJcsEqX2SjCj4J20uK7+ldkDHMbpJDK76wF7xEIgxc/vSlsfw5w==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=10.12.0"
+      }
+    },
+    "node_modules/aws-sign2": {
+      "version": "0.7.0",
+      "resolved": "https://registry.npmjs.org/aws-sign2/-/aws-sign2-0.7.0.tgz",
+      "integrity": "sha512-08kcGqnYf/YmjoRhfxyu+CLxBjUtHLXLXX/vUfx9l2LYzG3c1m61nrpyFUZI6zeS+Li/wWMMidD9KgrqtGq3mA==",
+      "license": "Apache-2.0",
+      "engines": {
+        "node": "*"
+      }
+    },
+    "node_modules/aws4": {
+      "version": "1.13.2",
+      "resolved": "https://registry.npmjs.org/aws4/-/aws4-1.13.2.tgz",
+      "integrity": "sha512-lHe62zvbTB5eEABUVi/AwVh0ZKY9rMMDhmm+eeyuuUQbQ3+J+fONVQOZyj+DdrvD4BY33uYniyRJ4UJIaSKAfw==",
+      "license": "MIT"
     },
     "node_modules/balanced-match": {
       "version": "4.0.4",
@@ -1965,6 +2014,15 @@
         }
       ],
       "license": "MIT"
+    },
+    "node_modules/bcrypt-pbkdf": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/bcrypt-pbkdf/-/bcrypt-pbkdf-1.0.2.tgz",
+      "integrity": "sha512-qeFIXtP4MSoi6NLqO12WfqARWWuCKi2Rn/9hJLEmtB5yTNr9DqFWkJRCf2qShWzPeAMRnOgCrq0sg/KLv5ES9w==",
+      "license": "BSD-3-Clause",
+      "dependencies": {
+        "tweetnacl": "^0.14.3"
+      }
     },
     "node_modules/better-sqlite3": {
       "version": "12.10.0",
@@ -2127,6 +2185,12 @@
         "node": ">=6"
       }
     },
+    "node_modules/caseless": {
+      "version": "0.12.0",
+      "resolved": "https://registry.npmjs.org/caseless/-/caseless-0.12.0.tgz",
+      "integrity": "sha512-4tYFyifaFfGacoiObjJegolkwSU4xQNGbVgUiNYVUxbQ2x2lUsFvY4hVgVzGiIe6WLOPqycWXA40l+PWsxthUw==",
+      "license": "Apache-2.0"
+    },
     "node_modules/chai": {
       "version": "5.3.3",
       "resolved": "https://registry.npmjs.org/chai/-/chai-5.3.3.tgz",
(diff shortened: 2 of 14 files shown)


lml2468 (commented):
**Security Review — CHANGES_REQUESTED**

Excellent breadth and care on the COS pipeline, mention resolution, and partial-failure handling. But two P0 vulnerabilities in `resolveMedia()` need to be fixed before merge — the rest is solid.

## 🔴 P0.1 — Local File Disclosure via `file://` in agent output

`media-upload.ts resolveMedia()` handles `file://` URLs:
```ts
if (mediaUrl.startsWith("file://")) {
  const filePath = decodeURIComponent(mediaUrl.slice(7));
  const st = statSync(filePath);  // ⚠️ no path validation
  return { fileBody: createReadStream(filePath), ... };
}
```

The agent runs with `bypassPermissions` and has full host read access. If an LLM emits (or is prompt-injected into emitting) `file:///etc/passwd` or `file:///home/user/.ssh/id_rsa` in a RichText image ref or media URL, the bot will read that file and upload it to COS — publicly visible in the chat.

This is a **direct file-exfiltration channel from the agent host to any chat room**.

**Fix options:**
1. **Reject `file://` entirely.** Outbound media should come from explicit COS uploads, not arbitrary filesystem paths.
2. **Whitelist path prefix.** Restrict to `config.cwd` or a dedicated `mediaDir`, validate via `path.resolve()` + prefix check + `path.relative()` not starting with `..`.

Same vulnerability in `uploadImageForRichText()` via `resolveMedia(ref.url)` — RichText markdown `alt` would exfil.

## 🔴 P0.2 — SSRF via `http(s)://` URL in agent output

`downloadToTempFile()` fetches any URL the agent emits:
```ts
const head = await fetch(url, { method: "HEAD", ... });
const resp = await fetch(url, ...);
```

No protocol/host validation. An agent emitting `http://169.254.169.254/latest/meta-data/iam/security-credentials/` or `http://localhost:6379/` would have the bot fetch internal infrastructure responses and upload them to COS.

Symmetrical to Q34 apiUrl SSRF protection but on the outbound side.

**Fix:**
1. Reject loopback / link-local / private RFC1918 ranges (`127.0.0.0/8`, `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`, `169.254.0.0/16`, `::1`, `fc00::/7`) before fetching.
2. Resolve hostname — if it resolves to a private IP, reject.
3. Handle DNS rebinding (validate IP of actual connection or pin DNS).

`ipaddr.js` or a simple check over the resolved IP is enough.

## P1 — `getUploadCredentials` error path

```ts
throw new Error(`Octo API /v1/bot/upload/credentials failed (${response.status}): ${text || response.statusText}`);
```

Response body `text` is included verbatim in error. If server returns a 401 body echoing the request Authorization header for debugging, this leaks botToken into logs. Cap or sanitize body (truncate to 200 chars, strip header-looking patterns).

Same class as Q11. Lower severity here because body would need to specifically echo the token.

## P1 — `mentionAll` regex precision

`/(?:^|(?<=\s))@(?:all|所有人)(?=\s|[^\w]|$)/i`

Lookahead matches a lot of legitimate trailing chars. Combined with leading boundary, false positives rare but possible. Not security — just precision.

## What's solid ✅

- `sanitizeFilename` correctly strips path separators + rejects traversal segments, bounded 255 chars
- Temp file paths use `randomUUID()` prefix — unpredictable even if name sanitization bypassed
- Always under `UPLOAD_TEMP_DIR` via `path.join`
- COS uses short-lived STS tokens
- 500MB hard cap on all three input paths
- HEAD-first size probe avoids downloading huge bodies before rejecting
- temp file cleanup in `finally`
- `uploadFileToCOS` error extraction safely handles unknown error shapes
- `mention-utils` longest-prefix-match correctly handles displayName collisions
- structured `@[uid:name]` conversion uses incremental cursor, not indexOf rescan
- partial RichText failure gracefully degrades — failed images become `[alt]` text

P0 fixes are scoped to `media-upload.ts resolveMedia()` and `downloadToTempFile()` — tight changes. Tests for rejection paths welcome.

yujiawei (changes requested):
# Code Review — PR #34 (cc-channel-octo)

**Scope reviewed:** full diff vs `main` (14 files, +2805/-15) at head `c0fbc3e`. Verified locally on a clean checkout: `tsc --noEmit` passes, `vitest run` → **386/386 green**. The new modules are well-structured and the test files are thorough for the happy paths.

The verdict below hinges on a distinction the PR makes easy to miss: **most of the new surface (G5 media + G6 RichText) is not wired into the runtime**, so its defects are latent. But **G7 @mention resolution *is* wired** into `stream-relay.deliver()`, and it ships two reproducible functional bugs that reach real users.

---

## Blocking (must fix before merge)

### P1-1 — A structured `@[uid:name]` token can be split across two outbound segments, dropping the mention and leaking raw `@[…]` text to users
`src/stream-relay.ts:146-150` · `src/mention-utils.ts:54-67`

`deliver()` splits the accumulated reply **first** (`splitMessage(accumulated)`), then runs `resolveMentions(segment)` on each segment independently. `splitMessage` is token-agnostic — it cuts on `\n\n` > `\n` > space > hard-cut and has no awareness of mention tokens. Because the structured token's name charset (`[^\]\n]+`) permits spaces, a token like `@[u123:Bob Smith]` is cut at its internal space by the space tier well before any 3500-char hard cut.

Reproduced:
```
input : "...help @[u123:Bob Smith] please review"  (forced split @40 chars)
seg[0]: "...help @[u123:Bob "   → parsed mentions: []   (no closing ] → match fails)
seg[1]: "Smith] please review"  → parsed mentions: []
```
Result: the uid is **silently dropped** (no notification routed) and the user literally sees `@[u123:Bob` / `Smith]` in the message. This path runs unconditionally in production (it does not depend on `memberMap`).

**Fix direction:** resolve mentions on the *full* accumulated text once, computing entities against the whole string, then split while recomputing/preserving offsets — or, at minimum, make `splitMessage` mention-aware so it never cuts inside an `@[…]` token. Add a test that forces a token across a split boundary.

### P1-2 — `@all` / `@所有人` detection false-positives when immediately followed by a CJK character → spurious group-wide ping
`src/mention-utils.ts:242`

```js
const mentionAll = /(?:^|(?<=\s))@(?:all|所有人)(?=\s|[^\w]|$)/i.test(finalContent);
```
The trailing lookahead `[^\w]` treats CJK as a valid terminator (JS `\w` is ASCII-only). So an `@all`/`@所有人` embedded inside CJK prose is read as an everyone-ping. Reproduced:
```
"@all大家好"        → mentionAll = true
"我看了@all相关配置"  → mentionAll = true   (no ping intended)
"请@所有人都看一下"    → mentionAll = true
```
In a Chinese-language IM this is a realistic mis-fire that notifies the entire group and cannot be undone. (English is correctly guarded: `@allocation`, `@allow` → false.)

**Fix direction:** require a non-name terminator on the trailing side — e.g. allow only whitespace / end-of-string / ASCII punctuation after the token, excluding letters, digits **and** CJK/Kana/Hangul ranges (mirror the negative set already used for the leading boundary). Add CJK-adjacent cases to the test matrix.

---

## Non-blocking — latent defects in unwired G5/G6 code (fix before wiring)

The following are real bugs, but `uploadAndSendMedia()` and `sendRichTextCombined()` have **no production caller** (only `__tests__`), so production blast radius today is zero. They must be fixed before G5/G6 are wired, and several would become P1 at that point.

- **Non-base64 `data:` URI corrupts the payload.** `src/media-upload.ts:298` — regex `/^data:([^;,]+)?(?:;base64)?,(.*)$/` makes `;base64` optional but always does `Buffer.from(match[2], "base64")`. `data:text/plain,Hello` decodes to garbage. Branch on whether `;base64` was actually present; use `decodeURIComponent` otherwise. Tests only cover the `;base64` form.
- **Whitespace-only text block fails the entire RichText send.** `src/media-upload.ts:580,596` — `if (segment)` / `if (trailing)` push a `{type:"text", text:"   "}` block for whitespace between adjacent images (`a   b`). octo-lib `richtext.go` (`ValidateRichTextBlocks`) rejects any `TrimSpace(text)==""` block with `ErrRichTextTextEmpty`, 4xx-ing the whole message. Guard with `.trim()`.
- **HTTP download has no streaming size cap.** `src/media-upload.ts:256-275` — HEAD `content-length || 0` passes the gate when the header is absent/chunked/lying; the GET body is then `pipeline`'d to `/tmp` with no byte counter, and the 500MB cap is only checked post-hoc via `statSync`. Count bytes during streaming and abort+unlink past the cap.
- **No SSRF guard / `file://` arbitrary-read.** `src/media-upload.ts` `resolveMedia` — `mediaUrl` is agent/LLM-influenced. `http(s)` fetch has no host/IP allowlist (internal metadata endpoints reachable) and redirects are followed; `file://` reads any local path and exfiltrates it to a public CDN. Add an allowlist / block private ranges, and gate or remove the `file://` branch for untrusted input.
- **Dimension parsers trust the MIME label, not magic bytes.** `src/media-upload.ts:79-114` — a mislabeled file yields bogus width/height (defaulted to 800×600 in the RichText path). Low impact but worth a signature check.

## Non-blocking — wiring / docs / tests

- **G5/G6 orchestrators are dead code but ARCHITECTURE.md describes them in the present tense** as live delivery paths. Either wire them or add an explicit "implemented, not yet invoked by the relay" status note. The same applies to the G7 section (see next item).
- **G7 plain-`@name` fallback is never exercised in production.** `deliver()` gained a 7th positional `memberMap` param (`src/stream-relay.ts:117`) but the sole caller `src/index.ts:190` passes only 6 args, so `buildEntitiesFromFallback` / `tryLongestMemberMatch` (~half of `mention-utils.ts`) are dead at runtime. `GroupContext` already maintains the name→uid map but exposes no public accessor. Wire it through and add an end-to-end test.
- **Cross-system offset unit is assumed, not verified.** `MentionEntity.offset` is documented as UTF-16 code units; the implementation is correct-by-construction (all offsets derive from `match.index`/`string.length`), but there is no test with an astral-plane char before a mention and no contract test against the server. Add one to lock the contract.
- **Over-mocked tests:** the COS key re-encode (`media-upload.ts:226-230`, the whole reason that branch is non-trivial) is only tested with ASCII keys, so `encodeURIComponent` is a no-op and a regression wouldn't be caught — add a CJK-key assertion. The no-`cdnBaseUrl` `Location` fallback is never happy-path tested. RichText `plain` is asserted only via `toContain('[图片]')`.

## Nits
- `getUploadCredentials` validates `bucket/region/key/credentials` but not `startTime`/`expiredTime`, which then flow as `undefined` into the COS signer (`src/octo/api.ts:280-296`).
- Temp file can leak if `statSync` throws a non-size error between download and return (`src/media-upload.ts:344`).
- Markdown image regex `([^)\s]+)` breaks on URLs containing `)` (`src/media-upload.ts:439`).
- A real member literally named `all` / `所有人` can never be mentioned via the fallback path (unconditionally skipped).

---

## Summary

Solid, well-tested foundations. **Requesting changes on the two production-reachable G7 mention bugs (P1-1, P1-2)** — both are user-visible, both have trivial fixes. The large P2 cluster lives in unwired G5/G6 code and can be addressed before those paths are connected, but the `data:`-URI corruption, whitespace-block rejection, and SSRF/`file://` items should be cleared before any wiring lands.

Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository, but several shipped paths are either not wired into production or leave security/resource-control gaps in untrusted media fetching.

**🔴 Blocking**
- 🔴 Critical: RichText/media delivery is implemented but never used by the actual outbound path. `handleMessage()` still calls only `streamRelay.deliver(...)` at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/index.ts:257), and `StreamRelay` still only calls `sendMessage()` at [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/stream-relay.ts:152). `uploadAndSendMedia()` and `sendRichTextCombined()` are not imported or invoked anywhere outside tests, so the advertised G5/G6 send capabilities do not work in the running bot.

- 🔴 Critical: Plain mention resolution is also not wired into production. `StreamRelay.deliver()` accepts `memberMap` at [src/stream-relay.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/stream-relay.ts:117), but the only production call omits it at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/index.ts:257). Also, `GroupContext.buildContext()` renders names without identifiers at [src/group-context.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/group-context.ts:223), so structured mention syntax cannot be reliably produced either.

- 🔴 Critical: SSRF protection can be bypassed through redirects. `assertPublicUrl()` validates only the original URL at [src/media-upload.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/media-upload.ts:342), but both `fetch()` calls follow redirects by default at [src/media-upload.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/media-upload.ts:349) and [src/media-upload.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/media-upload.ts:355). A public URL can redirect to loopback or metadata endpoints after validation. Use `redirect: "manual"` and validate each `Location`, or implement a redirect-following helper that re-runs the public-address checks on every hop.

- 🔴 Critical: The HTTP download size limit is enforced only before and after download, not during streaming. If `content-length` is missing or false, `pipeline(nodeStream, ws)` at [src/media-upload.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/media-upload.ts:364) can write an arbitrary-size response to disk before the `statSync()` check at [src/media-upload.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/media-upload.ts:431). Count bytes while streaming and abort/unlink as soon as the 500 MB limit is crossed.

- 🔴 Critical: The new production dependency introduces known critical vulnerabilities. Adding `cos-nodejs-sdk-v5` at [package.json](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/package.json:20) pulls deprecated `request` and vulnerable transitive packages; `npm audit --omit=dev` reports 10 production vulnerabilities, including 3 critical. This needs a safer SDK/version strategy or documented mitigation before merging.

**💬 Non-blocking**
- 🟡 Warning: `data:` URLs without `;base64` are still decoded as base64 at [src/media-upload.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-w7520rxr/repo/src/media-upload.ts:392), despite the parser accepting non-base64 data URIs. Decode percent-encoded/plain data separately or reject non-base64 explicitly.

**✅ Highlights**
- Type-check passes: `npm run type-check`.
- Tests pass locally after installing dependencies: `451 passed`.
- The structured mention offset conversion avoids same-name rebinding issues, and the upload credential error sanitization is a good defensive addition.

lml2468 (commented):
**Security Review — LGTM** ✅

Both P0s + P1 are fixed correctly with comprehensive test coverage.

## P0.1 file:// ✅
`resolveMedia()` throws unconditionally on `file://` scheme. Test verifies fetchMock + cosPutObjectMock never called. Both `uploadAndSendMedia` and `uploadImageForRichText` (RichText path) go through the same `resolveMedia` chokepoint — no bypass path exists.

## P0.2 SSRF ✅
`assertPublicUrl()` runs at the very top of `downloadToTempFile()`, BEFORE HEAD/GET — verified by `expect(fetchMock).not.toHaveBeenCalled()` in every rejection test. Coverage is thorough:
- IPv4: 127/8, 10/8, 172.16/12, 192.168/16, 169.254/16 (AWS/GCP metadata), 100.64/10 CGN, 0/8
- IPv6: ::1, ::, fc00::/7 (fc/fd startsWith), fe80::/10 (fe8/fe9/fea/feb), ::ffff: v4-mapped (recursive check)
- IP literals rejected without DNS round-trip; hostnames go through `dns.lookup({ all: true })` and reject if ANY resolved address is private

`startsWith("fc")/("fd")` correctly captures `fc00::/7` for compressed forms. `fe8/fe9/fea/feb` correctly captures `fe80::/10`. DNS rebinding caveat is explicitly noted in the comment — acceptable given short-TTL mitigation and the attacker complexity required.

Test list covers 9 distinct SSRF targets (each private range × IPv4/IPv6). 

## P1 credential error log ✅
```ts
const sanitized = text
  .slice(0, 200)
  .replace(/Bearer\s+\S+/gi, "Bearer ***")
  .replace(/(authorization"?\s*[:=]\s*"?)[^"\s,}]+/gi, "$1***");
```
Truncates to 200 chars + strips `Bearer <token>` and `authorization: <value>` patterns (handles JSON-quoted and bare forms). Defense-in-depth even though the body would have to specifically echo the header.

## Minor nits (non-blocking)
1. `::ffff:` v4-mapped path is implemented but not directly tested. Could add `'http://[::ffff:127.0.0.1]/'` as a test case to lock down the regex.
2. The DNS rebinding window remains (validate→connect TOCTOU). Acceptable per the documented residual risk; can be hardened later with HTTP agent that pins the resolved IP if Octo ever runs in a more hostile environment.

Approving. Clean rebase, scope unchanged from v1 + 3 surgical security fixes.

yujiawei (changes requested):
# Code Review — PR #34 (cc-channel-octo)

Independent review of the media upload, RichText combined-send, and @mention features. The implementation is generally high quality — clear module boundaries, thorough unit tests (451 passing), clean `tsc --noEmit`, careful int64/idempotency handling in `api.ts`, and good defense-in-depth instincts (filename sanitization, content-disposition encoding, temp-file cleanup in `finally`). I verified the `RichTextBlock.type` change from numeric `1`/`2` to string `"text"`/`"image"` against the `octo-lib` wire contract (`common/richtext.go`: `RichTextBlockText = "text"`) — this is a **correct latent-bug fix**: `inbound.ts` was previously comparing server-sent string types against numeric constants and never matching.

That said, I'm requesting changes for two reasons: a security control added in this very PR is bypassable, and the headline send capabilities are not actually reachable in the live message flow.

## P1 — Blocking

### 1. SSRF defense is bypassable via HTTP redirect
`src/media-upload.ts:342` (`assertPublicUrl`) validates the **initial** URL, but the subsequent `fetch()` calls at `:349` (HEAD) and `:355` (GET) use the default `redirect: "follow"`. Node's global `fetch` follows 3xx redirects to any host, so a publicly-routable URL that 302-redirects to `http://169.254.169.254/...` (cloud metadata) or an internal host defeats the check entirely. This is the exact threat the PR's security commit ("SSRF defense") sets out to block.

Fix: pass `redirect: "manual"` and re-run `assertPublicUrl` on each `Location` hop (or resolve+pin the IP and connect by IP). The `DNS rebinding` caveat is already documented in the comment; the redirect gap should get the same treatment.

### 2. Media / RichText / v1-mention paths are not wired into production
- `uploadAndSendMedia()` (`media-upload.ts:452`) and `sendRichTextCombined()` (`media-upload.ts:613`) have **zero callers** outside `__tests__`, and are not re-exported from `index.ts` (package `main`). The "media + RichText send capabilities" in the PR title are unreachable in the running adapter.
- `StreamRelay.deliver()` gained a `memberMap` param (`stream-relay.ts:117`), but the only production caller at `index.ts:257` passes 6 of 7 args — `memberMap` is always `undefined`. As a result the v1 plain-`@name` fallback (`mention-utils.ts:229`, `buildEntitiesFromFallback`) is dead in production; only v2 structured `@[uid:displayName]` resolves. `groupContext.getMemberMap()` is also `private`, so the map can't be threaded through without a small accessor.

If this is intentional staging ("G24 foundations", wire-up in a follow-up), please say so in the PR description and gate the unreachable code behind that plan — otherwise it lands latent security/correctness bugs (below) that will silently activate when a later PR connects them. The v1 fallback gap is the most likely surprise: plain `@name` from the LLM will silently produce no mention.

### 3. `data:` URI decoding is incorrect
`src/media-upload.ts:388-392`. Two bugs:
- The payload is **always** decoded as base64 (`Buffer.from(match[2], "base64")`) regardless of whether the `;base64` marker is present. A raw data URI such as `data:text/plain,Hello` is silently corrupted (`Buffer.from("Hello","base64")` → garbage bytes).
- The regex `/^data:([^;,]+)?(?:;base64)?,(.*)$/` does not match data URIs with extra parameters, e.g. `data:text/plain;charset=utf-8;base64,...` returns **no match** → "Invalid data URI format".

Fix: detect the `;base64` token explicitly and branch (base64 vs `decodeURIComponent`), and allow interim `;param` segments in the pattern. (Currently masked only because the code is unreachable — see #2.)

## P2 — Should fix

### 4. No incremental download size cap
`downloadToTempFile()` checks `Content-Length` on the HEAD response only (`media-upload.ts:351`); the GET stream is written to disk with no running-byte limit, and the `MAX_UPLOAD_SIZE` `statSync` check (`:431`) runs only after the full file is on disk. A server that omits/lies on `Content-Length` can write an arbitrarily large file to `/tmp` (up to the 300s timeout) → local disk exhaustion. Enforce the cap incrementally during `pipeline()` (count bytes in a passthrough and destroy on overflow).

### 5. Mention entity offsets are UTF-16 — verify against server
`convertStructuredMentions` (`mention-utils.ts:103`) and `buildEntitiesFromFallback` (`:188`) compute `offset`/`length` in JS UTF-16 code units, matching `types.ts:30` and the inbound `session-router.ts` convention (internally consistent, existing tests green). But `octo-lib` is Go (byte/rune-indexed strings). If the server expects byte or rune offsets, mentions following any astral-plane char (emoji) — and potentially multi-byte CJK — would be misplaced. Please confirm the server's expected unit; if it isn't UTF-16, this needs a conversion layer. Not necessarily introduced by this PR, but the new outbound path makes it load-bearing.

### 6. Markdown image URLs containing `)` are truncated
`MARKDOWN_IMAGE_RE` (`media-upload.ts:526`) uses `[^)\s]+` for the URL, so `a.jpg)` captures only `http://x/path(1`. Edge case, but worth a balanced-paren handling or a documented "percent-encode parens" contract.

### 7. Partial credential sanitization
`api.ts:279-282` strips `Bearer \S+` and `authorization[:=]...` from error text — good defensive move, but it's best-effort post-hoc scrubbing of a body that may echo headers in other shapes. Consider not surfacing raw response bodies for the credentials endpoint at all, or whitelisting known error fields.

## Summary
The foundations are well-written and tested, and the RichText constant fix is correct. Blocking on: the SSRF control being redirect-bypassable (#1), the advertised send paths not being reachable + v1 mention fallback being dead (#2), and the `data:` decoding bugs (#3). #4–#7 are non-blocking but should be addressed before this code goes live. Most fixes are small and localized.

PR #39, opened 2026-06-05
fix(D1): protocol DoS + correctness — socket tempBuffer cap, base64 cap, system prompt cap, SDK null guard

lml2468:
## D1 — Wave 2 batch (4 P0/P1 fixes from Phase 6 audit)

Originally proposed five fixes; one (onPacket stale guard) dropped after re-verification — see below.

### S6 (P0-1) — socket.ts tempBuffer OOM
Two-layer defense:
1. `MAX_TEMP_BUFFER_BYTES = 1 MiB` cap in `handleRawData`. If exceeded → drop buffer + close ws (reconnect logic recovers).
2. `MAX_VARLEN_BYTES = 4` inside `unpackOne` — matches MQTT spec. Stream of 0x80 continuation bytes can never exceed 4 bytes legally.

### S7 (P0-2) — getChannelMessages base64 + count caps
1. Per-message payload base64 string capped at 256 KiB (≈ 192 KiB decoded). Oversized payload skipped + warn log.
2. Result array capped at `limit` (default 20) even if server returns more.

### P1-3 — buildSystemPrompt 100 KiB hard cap
`MAX_SYSTEM_PROMPT_CHARS = 100 KiB`. Beyond it: history truncated tail-preserving (most recent kept), group context capped at 20 KiB share (oldest dropped). Security prefix + custom prompt never truncated.

### P1-4 — queryAgent SDK null safety
`message.message?.content ?? []` guards against malformed SDK output.

### Dropped (P1-6 onPacket stale guard)
After rechecking the code, all packet handlers (`onPacket`, `onConnack`, `onRecv`, `onDisconnect`) are synchronous — no awaits between the `ws.on('message')` entry guard and the writes to `this.connected` / `this.aesKey` etc. The existing entry guard is sufficient. My audit concern was theoretical.

### Tests (13 new)
- 10 unit tests in `d1-protocol-dos.test.ts`: buildSystemPrompt under-cap pass-through, history truncation, custom-prompt preservation, group budget, exhausted-budget; queryAgent null content / missing message field; getChannelMessages oversized payload drop / count cap / normal decode
- 3 socket integration tests in `octo/d1-socket-dos.test.ts`: 2 MiB partial-packet flood triggers buffer cap + close, 5-byte varlen triggers cap + close, normal PONG doesn't close

### Verification
- `tsc --noEmit`: clean
- `npm run lint` (eslint --max-warnings 0): clean
- `npx vitest run`: 473/473 (460 baseline + 13 new)
- Pre-commit hook (lint-staged + tsc) ran on commit

### Coordination notes
- **Conflict potential with PR#37 (S3)**: both touch `agent-bridge.ts buildSystemPrompt`. My change is purely additive at the END (cap + helper function). PR#37's edits are inside SECURITY_PROMPT_PREFIX / SECTION_MARKER_RE / buildSystemPrompt body (sanitize groupContext). Should rebase cleanly. If PR#37 lands first I'll rebase D1; if D1 lands first PR#37 rebases.
- C1 (飞飞) blocked on D1's buildSystemPrompt cap so D1 lands first per Wave 2 plan.

### DIFF
--- a/src/__tests__/d1-protocol-dos.test.ts
+++ b/src/__tests__/d1-protocol-dos.test.ts
@@ -1,0 +1,222 @@
+/**
+ * D1 (Wave 2) tests:
+ *  - S6 socket.ts tempBuffer OOM guard + variable-length 4-byte cap
+ *  - S7 api.ts getChannelMessages base64 payload size + array length caps
+ *  - P1-3 agent-bridge.ts buildSystemPrompt 100 KiB cap with truncation
+ *  - P1-4 agent-bridge.ts SDK output null-safety guard
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { buildSystemPrompt } from '../agent-bridge.js';
+
+// ─── P1-3: buildSystemPrompt length cap ───────────────────────────────────────
+
+describe('buildSystemPrompt MAX_SYSTEM_PROMPT_CHARS (D1/P1-3)', () => {
+  it('returns assembled prompt unchanged when under cap', () => {
+    const out = buildSystemPrompt('history line', 'group chat', 'custom prompt');
+    expect(out.length).toBeLessThan(2_000);
+    expect(out).toContain('custom prompt');
+    expect(out).toContain('[Group context]\ngroup chat');
+    expect(out).toContain('[Conversation history]\nhistory line');
+  });
+
+  it('caps total length at 100 KiB when history is huge', () => {
+    // 40 turns × 4 KiB each = 160 KiB worth of history.
+    const turns: string[] = [];
+    for (let i = 0; i < 40; i++) {
+      turns.push(`[user]: turn ${i} ${'x'.repeat(4_000)}`);
+    }
+    const huge = turns.join('\n');
+    const out = buildSystemPrompt(huge, '', '');
+    // 100 KiB hard ceiling plus a small slack for headers/labels.
+    expect(out.length).toBeLessThanOrEqual(110 * 1024);
+    expect(out).toContain('[older turns dropped]');
+    // Tail must be preserved (most recent turn = turn 39).
+    expect(out).toContain('turn 39');
+    // Oldest turn should be dropped.
+    expect(out).not.toContain('turn 0 ');
+  });
+
+  it('preserves security prefix and custom prompt verbatim when truncating', () => {
+    const huge = '[user]: ' + 'a'.repeat(120 * 1024);
+    const out = buildSystemPrompt(huge, '', 'OPERATOR_DIRECTIVE_42');
+    // The non-truncatable sections must survive.
+    expect(out).toContain('OPERATOR_DIRECTIVE_42');
+    expect(out).toContain('coding assistant'); // security prefix excerpt
+  });
+
+  it('caps group context separately when total exceeds the budget', () => {
+    // Push total over 100 KiB so the cap kicks in.
+    const giantGroup = 'msg\n'.repeat(8_000);  // ~32 KB — group is big
+    const giantHistory = '[user]: x'.repeat(20_000); // ~200 KB — forces truncation
+    const out = buildSystemPrompt(giantHistory, giantGroup, '');
+    expect(out.length).toBeLessThanOrEqual(110 * 1024);
+    // Group context budget is 20 KiB — should still appear, but drop oldest.
+    expect(out).toContain('[Group context]');
+    expect(out).toContain('[older messages dropped]');
+  });
+
+  it('drops history entirely if security + custom + group already saturate budget', () => {
+    const giantCustom = 'OP_'.repeat(40_000); // 120 KB custom prompt
+    const out = buildSystemPrompt('history line', 'group chat', giantCustom);
+    // Custom prompt is non-truncatable; we still produce a result.
+    expect(out).toContain(giantCustom);
+    expect(out.length).toBeGreaterThan(115 * 1024); // custom dominates
+    // The trailing [Conversation history] section (the appended one, not the
+    // literal mention inside the security prefix) should be truncated to a
+    // tiny tail since budget is exhausted.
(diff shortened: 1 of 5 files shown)


Jerry-Xin (changes requested):
Scope gate passed: this PR is relevant to `cc-channel-octo`, but one DoS fix is incomplete enough to block merge.

**🔴 Blocking**
- 🔴 Critical: `src/octo/socket.ts:514-520` enforces `MAX_TEMP_BUFFER_BYTES` only after copying the entire inbound WebSocket chunk into `this.tempBuffer`. A malicious server can still send one large binary frame and force the client to push every byte into a JS number array before the cap runs, which can consume far more memory than the intended 1 MiB limit. Check `this.tempBuffer.length + data.length` before the append loop, drop/close immediately when it would exceed the cap, and avoid copying the oversized chunk.

**💬 Non-Blocking**
- 🟡 Warning: `src/agent-bridge.ts:124-199` describes a hard `100 KiB` prompt cap, but the implementation can still exceed it when non-truncated sections, especially `customPrompt`, are large. `src/agent-bridge.ts:186` also forces at least 1 KiB of history budget even if the fixed sections already exceed the cap. If this is intentionally best-effort for trusted config, document it as such; otherwise validate or cap configured prompt text.
- 🟡 Warning: `src/octo/api.ts:596-615` caps message count and payload decoding after `postJson()` has already read and parsed the full response body. That still leaves oversized sync responses able to consume memory in `response.text()` / `JSON.parse()` before these caps apply. This is less severe than the socket issue, but worth addressing if the threat model includes a malicious or buggy server.
- 🔵 Suggestion: New comments and tests include non-English fragments in otherwise English OSS code paths. For consistency, keep new code comments and test descriptions fully English.

**✅ Highlights**
- The variable-length integer guard in `src/octo/socket.ts:573-591` correctly rejects malformed encodings beyond four bytes.
- The SDK message null guard in `src/agent-bridge.ts:268-279` is a low-risk correctness improvement.
- The new tests cover the intended happy paths and malformed input cases well.

Verification note: I attempted a focused Vitest run, but this checkout is missing the local `vitest` package, so the config failed to load with `ERR_MODULE_NOT_FOUND`.

lml2468 (commented):
QA Review: APPROVED ✅

Local verification on head branch (60e0f0c): tsc clean, eslint zero warnings, 473/473 pass.
CI also green (W0 working — type-check/lint/test/coverage all SUCCESS).

### S6 socket OOM — verified
Two-layer defense correct: handleRawData 1 MiB cap + unpackOne 4-byte varlen cap. Edge case noted: handleRawData check is post-loop so a 5 MiB single ws.send could briefly grow tempBuffer to ~5 MiB before cap fires. Not exploitable (Node ws maxPayload upstream cap), minor not blocking.

### S7 base64 + count caps — verified
Length checked BEFORE Buffer.from prevents 75 MB sync allocation. params.limit cap on array slice correct. Minor: console.warn per oversized could log-spam if attacker sends many — not blocking.

### P1-3 buildSystemPrompt cap — verified
Fast path under cap, truncateSystemPrompt preserves securityPrefix + customPrompt unconditionally, groupContext 20 KiB share drop-oldest, history remaining budget tail-preserve. Math.max(1024, ...) graceful fallback when custom dominates (test case 5).

### P1-4 SDK null safety — verified
message.message?.content ?? [] guards both missing and undefined paths.

### Dropped P1-6 — judgment correct
Re-verified all packet handlers are sync no-await. Original ws.on stale guard is sufficient. Self-correcting a theoretical concern is better than hard-fixing a non-issue — good QA discipline.

### Test quality
13 new tests: 5 buildSystemPrompt boundary cases, 2 SDK null cases, 3 getChannelMessages cases, 3 socket integration with same WebSocket mock as wksocket-packet.test.ts. The 2 MiB partial-packet test correctly asserts on outcome (close called) not specific path since both caps result in close.

### Coordination
PR#37 (S3) just merged on main. D1 description says "if PR#37 lands first I will rebase". Rebase needed before merge — both touched agent-bridge.ts buildSystemPrompt area (S3 modified body, D1 appended cap logic). Should rebase cleanly per齐分析.

Ready to merge after rebase. — guoshouwei QA

Jerry-Xin (changes requested):
The PR is in scope for `cc-channel-octo`, but the socket DoS fix still allows an oversized inbound frame to grow memory well past the intended cap before enforcement.

**🔴 Blocking**
- 🔴 Critical: [src/octo/socket.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-2k52fg66/repo/src/octo/socket.ts:514) appends the entire incoming `data` frame into `tempBuffer` before checking `MAX_TEMP_BUFFER_BYTES` at line 520. A malicious peer can send one very large WebSocket frame and force a byte-by-byte copy into an unbounded `number[]`, potentially exhausting memory before the close path runs. The cap needs to be enforced before or during append, for example by checking `this.tempBuffer.length + data.length > MAX_TEMP_BUFFER_BYTES` before copying and closing without appending. Add a regression test that verifies an oversized single frame is rejected without growing `tempBuffer` beyond the cap.

**💬 Non-blocking**
- 🟡 Warning: [src/agent-bridge.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-2k52fg66/repo/src/agent-bridge.ts:176) does not implement a strict 100 KiB cap despite the PR contract. The group budget is added independently of remaining space, and line 202 forces at least 1 KiB of history even when the budget is already exhausted. The tests also allow 110 KiB. If the intended contract is a hard cap for all truncatable sections, compute group and history budgets from the remaining space after non-truncatable sections.
- 🔵 Suggestion: Several new comments include non-English labels. For a top-tier OSS repository, keep code comments and test descriptions consistently English.

**✅ Highlights**
- The SDK null guard is narrowly scoped and improves robustness.
- The historical message payload cap and returned-message count cap are directionally correct and covered by focused tests.

PR #41, opened 2026-06-05
fix(C1): RichText crash + budgets + G4 payload merge + rejection cache guard

lml2468:
C1 五项修复 — 三个是我 PR#33 引入的真回归。P1.1 userContent typeof guard (RichText crash). P1.3 MultipleForward 3/50/8KiB caps. P1.4 RichText 50/20/32KiB caps. P1.5 G4 backfill 合并 decoded payload. P2.5 RouteResult.rejectionReason 抑制 group context 缓存。27 新测试，517/517 全绿。

### DIFF
--- a/src/__tests__/c1-rejection-userguard.test.ts
+++ b/src/__tests__/c1-rejection-userguard.test.ts
@@ -1,0 +1,161 @@
+/**
+ * C1 (Stage 6) — RouteResult.rejectionReason + userContent type guard.
+ *
+ * P1.1: index.ts userContent fallback now uses typeof === 'string' guard so
+ *       RichText/File payloads with array/object content don't crash SQLite.
+ * P2.5: SessionRouter attaches rejectionReason on rate-limited/oversized so
+ *       index.ts can suppress group context caching for those messages.
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { SessionRouter } from '../session-router.js';
+import { ChannelType, MessageType } from '../octo/types.js';
+import type { BotMessage } from '../octo/types.js';
+import type { Config } from '../config.js';
+
+// Mock sendMessage so replySafe doesn't hit the network
+vi.mock('../octo/api.js', async () => {
+  const actual = await vi.importActual<typeof import('../octo/api.js')>('../octo/api.js');
+  return {
+    ...actual,
+    sendMessage: vi.fn().mockResolvedValue(undefined),
+  };
+});
+
+const ROBOT_ID = 'bot-xyz';
+const OWNER_UID = 'owner-1';
+
+function makeConfig(overrides: Partial<Config> = {}): Config {
+  return {
+    botToken: 'tok',
+    apiUrl: 'https://test',
+    cwd: '/tmp',
+    dataDir: '/tmp/data',
+    sdk: { allowedTools: [], permissionMode: 'bypassPermissions', settingSources: ['user'] },
+    rateLimit: { maxPerMinute: 5 },
+    context: { maxContextChars: 6000, historyLimit: 40 },
+    maxResponseChars: 524_288,
+    ...overrides,
+  };
+}
+
+function makeMsg(opts: Partial<BotMessage> = {}): BotMessage {
+  return {
+    message_id: '1',
+    message_seq: 1,
+    from_uid: 'user-alice',
+    channel_id: 'dm-1',
+    channel_type: ChannelType.DM,
+    timestamp: Date.now(),
+    payload: { type: MessageType.Text, content: 'hi' },
+    ...opts,
+  };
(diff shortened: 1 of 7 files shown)


Jerry-Xin (approved):
This PR is in scope for `cc-channel-octo` and addresses real inbound-message routing, parsing, history, and group-context behavior. I found no blocking issues.

💬 Non-blocking

- 🟡 Warning — `src/inbound.ts:113`: the RichText block cap still walks and allocates all valid blocks before slicing because it does `filter(...).slice(...)`. For adversarially large arrays, prefer slicing first or manually collecting until `RICH_TEXT_MAX_BLOCKS`.
- 🟡 Warning — `src/inbound.ts:245`: `MultipleForward` caps `msgs`, but `users` is still fully mapped. A huge `users` array can still burn CPU/memory even when only 50 messages are rendered.
- 🔵 Suggestion — `src/inbound.ts:124`: `truncateByBytes` says the result is at most `maxBytes`, but it appends the marker after truncating, so output can exceed the cap by marker length. Either reserve marker bytes or adjust the comment.

✅ Highlights

- `src/index.ts:195` correctly prevents non-string RichText/File content from reaching SQLite history writes.
- `src/session-router.ts:123` returning `RouteResult | null` is a clean way to let callers suppress side effects for active rejections.
- `src/octo/api.ts:650` preserves top-level sync fields while backfilling missing data from decoded payloads.
- Added coverage targets the main regressions around RichText budgets, forwarded-message caps, history payload merge, and group cache suppression.

Test note: I could not execute the test suite in this checkout because dependencies are not installed; `vitest` was not found.

lml2468:
**Review: APPROVED with two non-blocking findings** ✅ (齐静春)

7 files, +743/-18, 517/517 pass, tsc clean, eslint zero warnings.

Five-item collapse well-organized. Three fixes (P1.1 / P1.3 / P1.5) are self-audited regressions from your own PR#33/G4 — proactive root-cause closure beats waiting for reviewer.

### Design verification ✅

- **P1.1 typeof guard**: RichText payload.content is Array, previous `?? historyRecord` only fired on null/undefined → arrays crashed SQLite binding. Explicit typecheck covers array/object/number/null/undefined uniformly. 6 tests cover all variants incl. `content: ''` (empty string preserved, not fallback'd).
- **P1.3 MultipleForward 3-axis**: depth=3 / msgs=50 / output=8KiB. Hop-counted _depth from top-level=0. Truncation markers preserve user-facing signal. 100-deep stack-safety test confirms no overflow.
- **P1.4 RichText 3-axis**: blocks=50 / mediaUrls=20 / output=32KiB. Matches MAX_CONTENT_BYTES so RichText can't bypass Text gate via type-14. Strictly tighter than D1's 100KiB system-prompt cap — layered defense.
- **P1.5 G4 top-level + payload merge**: 'strict superset' semantic (top-level wins when non-empty, else fall to decoded payload) is right. Previously every Text history row had content=undefined → G4 backfill no-op. content !== '' guard preserves explicit empty.
- **P2.5 → P1 upgrade**: rejectionReason via RouteResult + SUPPRESS_GROUP_CACHE set covers actively-blocked without affecting silent-drop (legitimate chatter). e2e regression test proves flooder can't inject into [Group context] for next turn — precise attack closed.

### Finding 1 (P2 perf, non-blocking): truncateByBytes worst-case O(n²) on CJK

src/inbound.ts:135-145 — `input.slice(0, maxBytes)` is UTF-16 code units, not bytes. For CJK (3 bytes/char) keeps up to 3×maxBytes bytes, then shrinks 1 char per iter with O(n) byteLength. For 32KB target + 96KB CJK input: ~22K iter × ~60KB avg ≈ ~1.3GB work (1-2s blocking).

Bounded in practice (RichText 32KB cap + D1 socket cap), but low-effort fix:
```typescript
const buf = Buffer.from(input, 'utf-8');
if (buf.length <= maxBytes) return { text: input, truncated: false };
let end = maxBytes;
while (end > 0 && (buf[end] & 0xc0) === 0x80) end--; // walk back to UTF-8 boundary
return { text: buf.subarray(0, end).toString('utf-8') + marker, truncated: true };
```
O(1) backoff, no quadratic scan.

### Finding 2 (nit): dead enum members in RejectionReason

session-router.ts:26-32 declares 'blocked_bot' | 'self_message' | 'system_event' | 'global_rate_limited' | 'not_mentioned' but only 'rate_limited' and 'oversized' are ever set (G20 collapsed three rate-limit paths into one). Either set them at the return-null sites for clean classification, or trim the type. Keeping for forward-compat is also defensible.

### Cross-PR check
Independent of PR#38 (S1) and PR#40 (S2) — different files. Merged against current main cleanly.

LGTM, ready to land. Both findings are non-blocking follow-up material.

PR #44, opened 2026-06-05
test(Q1): coverage gaps + tautology removal + migration fail-loud

lml2468:
## Stage 6 Wave 2 Q1 — test coverage holes from QA audit

Three audit findings (P0-1, P0-2, P0-3) plus one bonus tautology cleanup (Q38).

PR#34 token sanitize regression test (originally Q1 item 4) was found during preparation to already exist in api.test.ts:237-285 (added in PR#34 itself). Audit miss — scope reduced to 3 items + 1 bonus.

## Q1-1: JPEG and WebP parseImageDimensions tests (+8 tests)

`media-upload.ts:175-200` supports PNG/JPEG/GIF/WebP. Pre-Q1 the suite covered only PNG and GIF; **JPEG SOF scan + WebP VP8 RIFF parse (incl. 0x3FFF mask) had zero coverage**. parseImageDimensions failure silently falls back to 800x600 in `uploadImageForRichText` — every JPEG/WebP could have wrong dimensions and we would never know.

`media-upload.test.ts` adds:
- JPEG SOF0 marker (baseline + jpg-alias mime variant)
- JPEG SOF2 marker (progressive)
- JPEG with no SOF marker → null
- WebP VP8 dimensions from RIFF header
- WebP 0x3FFF mask verification (high bits must drop)
- WebP VP8L (lossless) chunk → null (current parser only handles VP8)

## Q1-2: SQLite G10 migration tests + console.warn → throw (+3 tests + 5 prod lines)

`session-store.ts:90-105` ALTER TABLE message_seq migration was wrapped in try/catch + console.warn. The audit found this silently hides real failures: every subsequent INSERT/SELECT against message_seq would fail with cryptic SQL errors far from root cause.

**Prod fix**: changed `console.warn` → `throw` with clear context (`G10 message_seq migration failed — database is in an unknown state`). Operator sees the failure at startup, not 50 messages in.

(Same fix齐静春 also independently proposed in PR#30 review note + 李飞飞 P1.D — three reviewers cross-validated this fail-loud guarantee.)

`session-store.test.ts` adds:
- migrates populated v0.1.0 schema (no message_seq column) cleanly
- buildSegmentedHistoryPrefix handles pre-migration NULL message_seq rows
- throws clear error when migration cannot run (fail-loud guarantee)

## Q1-3: Delete tautology tests (+2 real coverage, -3 tautology/weak tests)

`pr33-followup.test.ts:191-216` `handleMessage wiring (smoke)` block had two weak tests:
- `expect(true).toBe(true)` tautology (would pass even if index.ts was deleted)
- `two different strings are not equal` weak contract assertion

Both were placeholder tests added to document a contract that is already directly tested by the `LLM no longer sees its own replies` and `classifies bot messages as assistant` tests earlier in the file. **Block deleted** — net coverage unchanged, false-confidence eliminated.

## Q1 bonus: Q38 aesIV salt warning tautology (+2 real tests, -1 tautology)

`q34-q37-q38.test.ts:121-139` `aesIV salt length warning` had a similar `expect(true).toBe(true)` placeholder noting `runtime testing deferred to Q29`. Q29 (`wksocket-packet.test.ts`) has been implemented since.

**Moved the real test into `wksocket-packet.test.ts`** using the existing WKSocket + ws-mock infrastructure:
- `CONNACK with salt shorter than 16 bytes emits console.warn (Q38)`
- `CONNACK with salt >= 16 bytes does NOT warn` (negative case)

`q34-q37-q38.test.ts` Q38 block replaced with a pointer comment.

## Verification

```
npm run type-check  # clean
npm run lint        # 0 warnings, 0 errors (W0 gate)
npm test            # 665 → 674 (+13 added, -4 deleted)
bash .husky/pre-commit  # fires + passes (verified during commit)
```

## Net coverage impact

| Area | Branches gained |
|------|------------------|
| JPEG SOF0/SOF2 scan + jpg-alias | 0% → covered |
| WebP VP8 RIFF + 0x3FFF mask | 0% → covered |
| SQLite migration on populated v0.1.0 schema | 0% → covered |
| Migration fail-loud assertion | new contract |
| onConnack short-salt warn path | 0% → covered |

13 false-confidence-free tests added, 4 weak tests removed = net higher signal density.

## Audit method (固化进 MEMORY)

This Q1 PR closes Stage 6 audit P0-1/P0-2/P0-3 + Q38 placeholder cleanup. The audit method used:

1. **Triple grep matrix** to scope (not just `*.test.ts` vs `*.ts` file pairing):
   - `grep -rln '<keyword>' src/__tests__/` — by behavior keyword
   - `grep -rln '<funcName>' src/__tests__/` — by target function
   - `git log --all --grep='<PR#|topic>'` — by commit-message intent

2. **Tautology hunt** — `grep -rn 'expect(true).toBe(true)' src/__tests__/` found 2 occurrences; both addressed.

3. **PR#34 token sanitize discovery** — Stage 6 audit originally listed this as zero regression test but triple-grep on a final pass found `api.test.ts:237-285` already covered it (added in PR#34 itself; QA audit only looked at new test files vs new source files). Lesson: always check `git log --grep`/test bodies, not just new file pairing.

### DIFF
--- a/src/__tests__/media-upload.test.ts
+++ b/src/__tests__/media-upload.test.ts
@@ -132,6 +132,107 @@ describe('parseImageDimensions', () => {
     const buf = Buffer.alloc(100);
     expect(parseImageDimensions(buf, 'image/tiff')).toBeNull();
   });
+
+  // Q1-1: JPEG SOF marker scan. Production scans for 0xFFC0/0xFFC2 markers,
+  // reads height at offset+5 and width at offset+7 (both 2 bytes BE).
+  // Pre-Q1 this code path was entirely untested; a parse failure silently
+  // falls back to 800x600 defaults in uploadImageForRichText.
+  it('parses JPEG dimensions via SOF0 marker', () => {
+    // Minimal JPEG: SOI (FFD8) + APP0 segment + SOF0 segment.
+    const buf = Buffer.alloc(40);
+    // SOI
+    buf[0] = 0xFF; buf[1] = 0xD8;
+    // APP0 marker at offset 2
+    buf[2] = 0xFF; buf[3] = 0xE0;
+    // APP0 segment length: 16 bytes (includes the length field itself)
+    buf.writeUInt16BE(16, 4);
+    // SOF0 marker at offset 20 (2 marker + 2 + 16 APP0 segment = offset 20)
+    buf[20] = 0xFF; buf[21] = 0xC0;
+    // SOF0 segment length: 17 bytes
+    buf.writeUInt16BE(17, 22);
+    // SOF0 data layout: precision (1) + height (2 BE) + width (2 BE) + ...
+    buf[24] = 8;                  // precision (offset+4 from marker start)
+    buf.writeUInt16BE(720, 25);  // height at offset 25 (= marker+5)
+    buf.writeUInt16BE(1280, 27); // width at offset 27 (= marker+7)
+    const result = parseImageDimensions(buf, 'image/jpeg');
+    expect(result).toEqual({ width: 1280, height: 720 });
+  });
+
+  it('parses JPEG dimensions via SOF2 marker (progressive)', () => {
+    // Same shape but using SOF2 (0xFFC2) instead of SOF0.
+    const buf = Buffer.alloc(40);
+    buf[0] = 0xFF; buf[1] = 0xD8;
+    buf[2] = 0xFF; buf[3] = 0xE0;
+    buf.writeUInt16BE(16, 4);
+    buf[20] = 0xFF; buf[21] = 0xC2; // SOF2
+    buf.writeUInt16BE(17, 22);
+    buf[24] = 8;
+    buf.writeUInt16BE(480, 25);  // height
+    buf.writeUInt16BE(640, 27);  // width
+    const result = parseImageDimensions(buf, 'image/jpeg');
+    expect(result).toEqual({ width: 640, height: 480 });
+  });
+
+  it('accepts image/jpg mime alias for JPEG (some uploads use .jpg → image/jpg)', () => {
+    const buf = Buffer.alloc(40);
+    buf[0] = 0xFF; buf[1] = 0xD8;
+    buf[2] = 0xFF; buf[3] = 0xE0;
+    buf.writeUInt16BE(16, 4);
+    buf[20] = 0xFF; buf[21] = 0xC0;
+    buf.writeUInt16BE(17, 22);
+    buf[24] = 8;
+    buf.writeUInt16BE(100, 25);
+    buf.writeUInt16BE(200, 27);
+    const result = parseImageDimensions(buf, 'image/jpg');
+    expect(result).toEqual({ width: 200, height: 100 });
+  });
+
+  it('returns null for JPEG with no SOF marker (truncated after SOI)', () => {
+    // SOI only, no SOF marker reachable before end of buffer
+    const buf = Buffer.from([0xFF, 0xD8, 0xFF, 0xE0, 0x00, 0x10]);
+    expect(parseImageDimensions(buf, 'image/jpeg')).toBeNull();
+  });
+
+  // Q1-1: WebP VP8 RIFF header. Production checks `VP8 ` (with trailing
+  // space) at offset 12-16, then reads width/height as 2-byte LE with
+  // 0x3FFF mask. The mask is a VP8 quirk; wrong mask = wrong dimensions.
+  it('parses WebP VP8 dimensions from RIFF header', () => {
+    const buf = Buffer.alloc(40);
+    buf.write('RIFF', 0);
+    buf.writeUInt32LE(32, 4);   // file size minus 8
+    buf.write('WEBP', 8);
+    buf.write('VP8 ', 12);      // VP8 chunk fourcc (with trailing space!)
+    buf.writeUInt32LE(20, 16);  // VP8 chunk size
+    // VP8 bitstream: 6-byte tag/keyframe (offsets 20-25) we don't validate,
+    // then width at offset 26 (2 LE), height at offset 28 (2 LE), both masked 0x3FFF.
+    buf.writeUInt16LE(640, 26);
+    buf.writeUInt16LE(480, 28);
+    const result = parseImageDimensions(buf, 'image/webp');
+    expect(result).toEqual({ width: 640, height: 480 });
+  });
+
+  it('applies 0x3FFF mask to WebP VP8 dimensions (high bits = scaling, not dim)', () => {
+    const buf = Buffer.alloc(40);
+    buf.write('RIFF', 0);
+    buf.writeUInt32LE(32, 4);
+    buf.write('WEBP', 8);
+    buf.write('VP8 ', 12);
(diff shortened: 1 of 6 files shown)


yujiawei (approved):
# Code Review — PR #44 (cc-channel-octo)

**Verdict: APPROVED** — no blocking issues. This is a well-scoped, high-signal test PR plus one small, correct production hardening. All claims in the description were independently verified against the source.

## Summary

- `+319 / -47` across 6 files; 5 test files + one 10-line prod change in `session-store.ts`.
- Net test count `665 → 674` confirmed (`vitest run`: **674 passed, 36 files**). The 5 touched test files: **72 passed**.
- `npm run type-check` clean, `npm run lint` clean (0 warnings under `--max-warnings 0`).

## 1. Verification

| Item | Status | Evidence |
|------|--------|----------|
| JPEG SOF0/SOF2 + `image/jpg` alias tests | ✅ correct | `parseImageDimensions` scans `0xC0`/`0xC2`, reads height at `offset+5`, width at `offset+7` (`src/media-upload.ts:86-99`). Tests construct byte-accurate buffers matching this; the `image/jpg` alias is genuinely supported (`mime === "image/jpg"`, line 86). |
| WebP VP8 RIFF + `0x3FFF` mask | ✅ correct | Prod reads width/height as LE at offsets 26/28 with `& 0x3FFF` (`src/media-upload.ts:106-112`). The mask test sets high bits `0xC000` and asserts they drop — `640\|0xC000 = 0xC280`, `& 0x3FFF = 0x280 = 640`. Math checks out. |
| WebP "missing VP8 chunk" → null | ✅ correct | `VP8L` ≠ `"VP8 "` so the `toString("ascii",12,16)` guard fails → null. |
| JPEG no-SOF → null | ✅ correct | 6-byte buffer makes the `offset < buf.length - 8` loop body unreachable → returns null. |
| SQLite migration on populated v0.1.0 schema | ✅ correct | `init()` runs `CREATE TABLE IF NOT EXISTS` (no-op on the pre-seeded table), `PRAGMA table_info` finds no `message_seq`, then `ALTER TABLE … ADD COLUMN` (`src/session-store.ts:100-107`). Test confirms pre-rows survive with NULL seq and history is preserved. |
| Segmented history handles NULL `message_seq` | ✅ correct | `buildSegmentedHistoryPrefix` routes NULL-seq rows by the `seenNew` flag (`src/session-store.ts:225-228`); the test exercises the legacy-NULL path and the `[new messages]` label. |
| Migration fail-loud throw | ✅ correct | `console.warn` → `throw new Error(... G10 message_seq migration failed ...)` (`src/session-store.ts:108-111`). Test shadows `adapter.exec` to throw on the ALTER and asserts `toThrow(/G10 message_seq migration failed/)`. |
| Q38 salt-warn moved to real WKSocket test | ✅ correct | Prod warns when `!salt \|\| salt.length < 16` (`src/octo/socket.ts:670-674`). New test drives a real CONNACK with an 8-byte salt via `doHandshake` and asserts the warn; negative case (20-byte `TEST_SALT`) asserts no warn. |
| Tautology deletions are safe | ✅ correct | The two deleted placeholders in `pr33-followup.test.ts` documented the botId-classification contract, which is **already** directly tested in the same file: `classifies bot messages as assistant…` (line 51) and `LLM no longer sees its own replies…` (line 76). No coverage lost. The deleted `import('../index.js')` smoke test added nothing beyond what `type-check` already guarantees. |

## 2. Issues

**None (P0/P1).** Nothing blocks merge.

## 3. Security note (PR was flagged security-sensitive)

The only behavioral change is `session-store.ts` converting a silent `console.warn` into a startup-time `throw`. This is the correct **fail-loud** posture for a security-sensitive migration path: a corrupt/partially-migrated DB now aborts at boot (`store.init()` is called exactly once, `src/index.ts:48`) instead of degrading silently and surfacing cryptic SQL errors later. No new attack surface, no credential/token handling touched, no swallowed exceptions introduced. The salt-length warning behavior (`socket.ts:670`) is unchanged — this PR only adds a real test for it. Nothing here needs additional manual security verification beyond confirming the startup-abort behavior is acceptable operationally (it is: an unknown-state DB should not serve traffic).

## 4. Nits / suggestions (non-blocking)

- `media-upload.test.ts:103` — test name "returns null for WebP with **missing VP8 chunk marker**" actually exercises a **VP8L (lossless)** chunk. The inline comment clarifies it, but the title could read "VP8L lossless chunk → null (parser only handles lossy VP8)" for precision.
- `session-store.test.ts:366-392` — the fail-loud test's lead comment walks through two rejected approaches before landing on the `exec` shadow. The final mechanism (override `adapter.exec` to throw on the ALTER) is clean and correct; trimming the exploratory prose would make intent clearer to future readers.
- `session-store.ts:108-111` — minor: the thrown error chains `String(err)` rather than using `{ cause: err }`. The string form is fine for an operator-facing startup message; `cause` would preserve the original stack if anything ever programmatically inspects it. Optional.

Overall: clean, byte-accurate tests that close real coverage gaps (JPEG/WebP dimension parsing, SQLite migration on legacy schemas), legitimate removal of false-confidence placeholders, and a sound fail-loud hardening. Approving.

lml2468:
**Review: APPROVED** (齐静春)

6 files, +319/-47, 674/674 pass, tsc clean, eslint zero warnings. W0 pre-commit hook + CI both gating.

**SQLite migration throw fix verified end-to-end** (this was my PR#30 review concern + 飞飞 P1.D + 守卫 audit triple-flag):

Reverse-verified by temporarily reverting throw back to console.warn and re-running session-store tests — the new "throws with clear error" test correctly fails (1/11 fail), proving the test really probes the fix. Restored. The "fail loud at startup" contract is now enforced.

Migration tests are real, not mocks-of-mocks:
- "populated v0.1.0 schema migrates without throwing" uses raw adapter.exec to construct the pre-G10 schema with rows, then calls store.init() — exactly the upgrade path real operators hit.
- "NULL message_seq rows handled gracefully" validates that buildSegmentedHistoryPrefix works on the mixed (legacy NULL + new seq) state — directly exercises the legacy-row path I added in PR#30 P1.2 fix.
- "simulated migration failure" shadows adapter.exec to throw on the ALTER TABLE — clever way to force the catch arm without faking better-sqlite3.

JPEG/WebP tests construct minimal real binary payloads (SOI + APP0 + SOF0/SOF2 with precise height/width offsets matching production buf.readUInt16BE(offset + 5/7)). VP8 0x3FFF mask test sets high bits explicitly and confirms they're masked out — exactly the VP8 quirk that would silently produce wrong dimensions if the mask were dropped. VP8L → null is also good negative coverage.

Q38 aesIV salt warning moved to wksocket-packet.test.ts with real WS mock + doHandshake — the "CONNACK with salt >= 16 bytes does NOT warn" negative case is what makes this a real test instead of a tautology. **This deletes the Q34+Q37+Q38 placeholder I wrote when I couldn't drive WKSocket.onConnack directly — 守卫's real-WS approach is the right answer. Glad to see it killed.**

Audit three-step grep matrix固化 also valuable — PR#34 token sanitize already covered miss is the kind of cross-PR audit gap that recurs.

LGTM, ready to land. No blockers, no findings.

PR #36, opened 2026-06-05
ci(W0): GitHub Actions + husky pre-commit + coverage + strict tsc

lml2468:
## Stage 6 Wave 0 — CI/Gate infrastructure (root cause fix)

Stage 6 audit identified **zero automated quality gates on PRs** as the root cause that let stage 5 land 13 P0 issues mechanical checks would have caught. This PR fixes the infrastructure gap. All subsequent stage-6 fix PRs (Wave 1 S1-S5 security, Wave 2 S6-S12 DoS+correctness) will land under enforced gates.

### 1. `.github/workflows/ci.yml` — four CI jobs on push/PR

- **type-check** — `tsc --noEmit`
- **lint** — `eslint --max-warnings 0`
- **test** — `vitest run` (460 tests)
- **coverage** — `vitest run --coverage` with artifact upload (14d retention); runs after the first three pass

### 2. `.husky/pre-commit` — local gate before any commit

- `lint-staged` runs ESLint `--max-warnings 0` on staged `.ts` files
- `tsc --noEmit` catches type regressions
- Full test suite deferred to CI (keep pre-commit fast)
- Auto-installed by `npm install` via `prepare` script

### 3. `@vitest/coverage-v8` + `vitest.config.ts`

- v8 coverage provider, reports text/html/lcov to `coverage/`
- No hard threshold yet; baseline measured this run:
  - **81.33%** statements / **81.82%** branches / **90.19%** functions
- Lowest coverage: `index.ts` (14%, mostly `main()` bootstrap), `octo/api.ts` (70%), `media-upload.ts` (82%) — will be improved in the Q1 PR

### 4. `husky` + `lint-staged` devDependencies

`lint-staged` config in `package.json` runs ESLint on `src/**/*.ts`.

### 5. `tsconfig.json` — enabled strict flags

- `noUnusedLocals: true`
- `noUnusedParameters: true`

One existing dead-code site (`SessionRouter.ownerUid`) was exposed via a public `getOwnerUid()` getter since it's intentional state for the future G18 permission model.

### 6. `g9-g10-g21.test.ts` — closed 6 ESLint warnings

Replaced 5 `as any` Config casts with a typed `makeTestConfig()` helper plus removed unused `beforeEach` import. ESLint now zero warnings, which `--max-warnings 0` enforces.

### 7. `README.md` — Quality gates subsection

Documents the three enforcement layers and how to run them locally.

## Verification

```bash
npm run type-check    # tsc clean, strict noUnusedLocals/Params enforced
[...]

PR #45, opened 2026-06-05
fix(C2): output + media pipeline defects

lml2468:
C2 batch: 5 fixes reverse-verified. P0-1 splitMessage breaks @[uid:name] / P1-3 RichText concurrency / P1-4 chunked transfer bypass / P1-5 SVG XSS / P1-6 truncation surrogate. 672/672 tests.

### DIFF
--- a/src/__tests__/media-upload.test.ts
+++ b/src/__tests__/media-upload.test.ts
@@ -529,4 +529,153 @@ describe('sendRichTextCombined', () => {
     expect(result.imageCount).toBe(0);
     expect(result.failedMedia).toHaveLength(1);
   });
+
+  // ─── C2 P1-3: bounded parallelism + per-message image cap ──────────────
+  //
+  // Without bounds, LLM emitting `![](data:...)` 100× would spawn 100 parallel
+  // COS uploads (fd / memory exhaustion). Fix: cap image count to
+  // RICHTEXT_MAX_IMAGES=20, run uploads with worker pool of
+  // RICHTEXT_MAX_CONCURRENT_UPLOADS=4.
+  it('P1-3: caps RichText image count to RICHTEXT_MAX_IMAGES (20)', async () => {
+    // Build 25 markdown images. Cap should drop refs >=20 into failedMedia.
+    // Mock credentials + send for each upload that proceeds.
+    const png = Buffer.alloc(33);
+    png[0] = 0x89; png[1] = 0x50; png[2] = 0x4E; png[3] = 0x47;
+    png[4] = 0x0D; png[5] = 0x0A; png[6] = 0x1A; png[7] = 0x0A;
+    png.writeUInt32BE(13, 8);
+    png.write('IHDR', 12);
+    png.writeUInt32BE(8, 16);
+    png.writeUInt32BE(8, 20);
+    const dataUri = `data:image/png;base64,${png.toString('base64')}`;
+
+    // Provide getUploadCredentials responses for 20 uploads (cap), plus 1 final
+    // sendRichTextMessage. Total expected fetch calls = 21.
+    for (let i = 0; i < 20; i++) {
+      fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+        bucket: 'b', region: 'r', key: `path/img${i}.png`,
+        credentials: { tmpSecretId: 'i', tmpSecretKey: 'k', sessionToken: 't' },
+        startTime: 1, expiredTime: 4600,
+        cdnBaseUrl: 'https://cdn.example.com',
+      }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+    }
+    cosPutObjectMock.mockImplementation((_params, cb) => {
+      cb(null, { Location: 'b.cos.r.myqcloud.com/path/img.png' });
+    });
+    fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+      message_id: 'rt1', client_msg_no: 'c1', message_seq: 1,
+    }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+
+    const text = Array.from({ length: 25 }, (_, i) => `![img${i}](${dataUri})`).join(' ');
+    const result = await sendRichTextCombined({
+      apiUrl: 'https://test.example.com',
+      botToken: 'bf_test',
+      channelId: 'ch1',
+      channelType: ChannelType.Group,
+      text,
+    });
+
+    // 5 images dropped to failedMedia with "cap exceeded" message.
+    expect(result.failedMedia.length).toBe(5);
+    for (const fail of result.failedMedia) {
+      expect(fail.error).toMatch(/cap exceeded/);
+    }
+    // 20 uploads attempted, all succeeded.
+    expect(cosPutObjectMock).toHaveBeenCalledTimes(20);
+    expect(result.imageCount).toBe(20);
+  });
+
+  // ─── C2 P1-4: chunked transfer bypasses HEAD size check ───────────────
+  //
+  // Without the stream-side size limiter, a server returning chunked encoding
+  // (no Content-Length) lets the download write to /tmp until disk fills. Fix:
+  // Transform stream counter rejects past MAX_UPLOAD_SIZE.
+  it('P1-4: chunked transfer with no Content-Length is bounded by stream size limit', async () => {
+    // Simulate chunked transfer: HEAD response has no content-length header,
+    // GET body emits 600MB worth of data.
+    fetchMock.mockResolvedValueOnce(new Response(null, {
+      status: 200,
+      // No 'content-length' header — typical of chunked transfer.
+    }));
+
+    // Build a ReadableStream that emits 1MB chunks until 600MB (exceeds 500MB cap)
+    const CHUNK_SIZE = 1024 * 1024;
+    const TOTAL_CHUNKS = 600;
+    let emitted = 0;
+    const stream = new ReadableStream({
+      pull(controller) {
+        if (emitted >= TOTAL_CHUNKS) { controller.close(); return; }
+        controller.enqueue(new Uint8Array(CHUNK_SIZE));
+        emitted++;
+      },
+    });
+    fetchMock.mockResolvedValueOnce(new Response(stream, {
+      status: 200,
+      headers: { 'Content-Type': 'application/octet-stream' },
+    }));
+
+    await expect(uploadAndSendMedia({
+      apiUrl: 'https://test.example.com',
+      botToken: 'bf_test',
+      channelId: 'ch1',
+      channelType: ChannelType.Group,
+      mediaUrl: 'https://example.com/chunked-payload.bin',
+    })).rejects.toThrow(/exceeded MAX_UPLOAD_SIZE/);
+  });
+
+  // ─── C2 P1-5: SVG XSS — SVG must NOT be sent as inline Image ───────────
+  //
+  // SVG can embed <script>/<foreignObject> which may execute if an IM client
+  // renders the image inline. Fix: route SVG to File type (forces download).
+  it('P1-5: SVG MIME type is sent as File not Image', async () => {
+    fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+      bucket: 'b', region: 'r', key: 'path/x.svg',
+      credentials: { tmpSecretId: 'i', tmpSecretKey: 'k', sessionToken: 't' },
+      startTime: 1, expiredTime: 4600,
+      cdnBaseUrl: 'https://cdn.example.com',
+    }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+    cosPutObjectMock.mockImplementation((_params, cb) => {
+      cb(null, { Location: 'b.cos.r.myqcloud.com/path/x.svg' });
+    });
+    fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+      message_id: 'm1', client_msg_no: 'c1', message_seq: 1,
+    }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+
+    const svgPayload = '<svg xmlns="http://www.w3.org/2000/svg"><script>alert(1)</script></svg>';
+    const dataUri = `data:image/svg+xml;base64,${Buffer.from(svgPayload).toString('base64')}`;
+
+    await uploadAndSendMedia({
+      apiUrl: 'https://test.example.com',
+      botToken: 'bf_test',
+      channelId: 'ch1',
+      channelType: ChannelType.Group,
+      mediaUrl: dataUri,
+    });
+
+    // Verify SVG was sent as File (type 8), NOT Image (type 2).
+    const sendCall = fetchMock.mock.calls[1];
+    const sendBody = JSON.parse(sendCall[1].body as string);
+    expect(sendBody.payload.type).toBe(8); // MessageType.File
+    expect(sendBody.payload.width).toBeUndefined();
+    expect(sendBody.payload.height).toBeUndefined();
+  });
+
+  it('P1-5: SVG in RichText markdown is rejected (not uploaded as image block)', async () => {
+    // RichText image blocks render inline too — SVG must be rejected via the
+    // isSafeInlineImage gate so failedMedia records it and [alt] falls back.
+    const svgPayload = '<svg xmlns="http://www.w3.org/2000/svg"><script>alert(1)</script></svg>';
+    const dataUri = `data:image/svg+xml;base64,${Buffer.from(svgPayload).toString('base64')}`;
+
+    const result = await sendRichTextCombined({
+      apiUrl: 'https://test.example.com',
+      botToken: 'bf_test',
+      channelId: 'ch1',
+      channelType: ChannelType.Group,
+      text: `header ![svg](${dataUri}) footer`,
+    });
+
+    expect(result.richText).toBe(false); // no images uploaded → caller falls back to plain
+    expect(result.failedMedia).toHaveLength(1);
+    expect(result.failedMedia[0].error).toMatch(/Not a safe inline image|svg/i);
+    expect(cosPutObjectMock).not.toHaveBeenCalled();
+  });
 });
--- a/src/__tests__/media-upload.test.ts
+++ b/src/__tests__/media-upload.test.ts
@@ -4,6 +4,13 @@
  * DNS isolation: downloadToTempFile calls assertPublicUrl which does a DNS
  * lookup for non-IP hosts. Test URLs use fictitious `example.com` hostnames —
  * without DNS mock these hit the real resolver and fail in CI (ENOTFOUND).
+ *
+ * Mocking contract (PR#45 review nit from 王大锤): the `node:dns/promises`
+ * mock below is REQUIRED for any test that drives a code path through
+ * `downloadToTempFile` (anything resolving an http(s):// mediaUrl). If you
+ * add a new test using a new hostname, extend the mock's hostname branches
+ * — do NOT rely on the real resolver. Removing this mock makes the suite
+ * green locally and red in CI, the classic "works on my machine" antipattern.
  */
 
 import { describe, it, expect, vi, beforeEach } from 'vitest';
--- a/src/__tests__/media-upload.test.ts
+++ b/src/__tests__/media-upload.test.ts
@@ -678,4 +678,54 @@ describe('sendRichTextCombined', () => {
     expect(result.failedMedia[0].error).toMatch(/Not a safe inline image|svg/i);
     expect(cosPutObjectMock).not.toHaveBeenCalled();
   });
+
+  it('P1-5 hardening: SVG with charset/case-variant params still rejected (PR#45 review nit)', async () => {
+    // PR#45 review-loop probe (陈皮皮): isSafeInlineImage must strip MIME
+    // params and lowercase. "image/svg+xml; charset=utf-8" and uppercase
+    // variants must NOT bypass the SVG gate. Without param-strip + lowercase,
+    // a data URI like `data:image/svg+xml; charset=utf-8;base64,...` would
+    // pass startsWith("image/") and === check would miss → SVG routed to
+    // Image, XSS open.
+    const svgPayload = '<svg xmlns="http://www.w3.org/2000/svg"><script>alert(1)</script></svg>';
+    const variants = [
+      `data:image/svg+xml; charset=utf-8;base64,${Buffer.from(svgPayload).toString('base64')}`,
+      `data:IMAGE/SVG+XML;base64,${Buffer.from(svgPayload).toString('base64')}`,
+      `data:Image/Svg+Xml;base64,${Buffer.from(svgPayload).toString('base64')}`,
+    ];
+    for (const dataUri of variants) {
+      // Setup: getUploadCredentials + COS putObject must NOT be triggered
+      // for SVG via Image path — set up mocks so a bypass would actually
+      // get to sendMediaMessage and we could inspect payload.type.
+      fetchMock.mockReset();
+      cosPutObjectMock.mockReset();
+      fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+        bucket: 'b', region: 'r', key: 'path/x.svg',
+        credentials: { tmpSecretId: 'i', tmpSecretKey: 'k', sessionToken: 't' },
+        startTime: 1, expiredTime: 4600,
+      }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+      cosPutObjectMock.mockImplementation((_params, cb) => {
+        cb(null, { Location: 'b.cos.r.myqcloud.com/path/x.svg' });
+      });
+      fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+        message_id: 'm1', client_msg_no: 'c1', message_seq: 1,
+      }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+
+      await uploadAndSendMedia({
+        apiUrl: 'https://test.example.com',
+        botToken: 'bf_test',
+        channelId: 'ch1',
+        channelType: ChannelType.Group,
+        mediaUrl: dataUri,
+      });
+
+      // SVG variant MUST route to File (type=8), not Image (type=2).
+      const sendBody = JSON.parse(fetchMock.mock.calls[1][1].body as string);
+      expect(sendBody.payload.type).toBe(8);
+      // And COS upload MUST set Content-Disposition: attachment so the CDN
+      // serves SVG as download instead of inline render.
+      const putCall = cosPutObjectMock.mock.calls[0][0] as { ContentDisposition?: string };
+      expect(putCall.ContentDisposition).toBeDefined();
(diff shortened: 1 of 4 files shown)


dolphinsboy (approved):
✅ APPROVED (fadfea8e3c4d)

C2 五个修复全部正确，无阻塞项。672 tests pass。

P0-1 splitMessage @mention 保护：protectedRanges 设计正确，adjustSplitForProtectedRanges 回退到 range.start，病态超长 mention 有 hard-cut 而不 infinite loop，mentionAll consumed flag 避免跨段 spam。手工 trace 验证 offset rebase 正确。

P1-3 RichText 并发 + 数量 cap：RICHTEXT_MAX_IMAGES=20 预先推 failedMedia，RICHTEXT_MAX_CONCURRENT_UPLOADS=4 bounded pool，JS 单线程 await 边界 workCursor++ 无 race。

P1-4 chunked transfer 绕过：Transform sizeLimiter pipeline 插入正确，覆盖无 content-length / lying server 场景，cleanup unlink tempPath 路径完整。

P1-5 SVG XSS：isSafeInlineImage() 拒绝 image/svg+xml，两处应用（msgType 选择 + RichText 拒绝），threat model 有注释。

P1-6 代理对截断：cutAt-1 high surrogate [0xD800,0xDBFF] 检测 cutAt--，与 splitMessage 同 pattern，正确。

yujiawei (changes requested):
# Code Review — PR #45 (cc-channel-octo)

**Verdict: CHANGES_REQUESTED** — 4 of the 5 fixes (P0-1, P1-3, P1-4, P1-6) are correct, well-reasoned, and backed by genuine tests. The remaining fix, **P1-5 (SVG XSS)**, only mitigates the in-app inline-render path and leaves a residual serving-layer XSS open. Since closing SVG XSS is the stated goal of this change, that residual should be addressed before merge.

I traced each fix against the full post-change files plus the `resolveMentions`/`uploadFileToCOS` contracts they depend on, and ran a boundary/fuzz analysis of the split logic.

---

## P1-5 — SVG XSS: residual vector (blocking)

The message-type fix is good but does not close the actual execution path.

After `isSafeInlineImage()` routes an SVG to `MessageType.File`, the object is **still uploaded to COS with `Content-Type: image/svg+xml` and no `Content-Disposition` header**:

- `uploadFileToCOS` (`src/media-upload.ts:218-226`) sets a disposition only for `video/`·`audio/` (`inline`) and for `!image/` (`attachment`). `image/svg+xml` starts with `image/`, so it falls through and **gets no disposition at all**.
- `ensureTextCharset` (`src/media-upload.ts:75-80`) only touches `text/*`, so the `ContentType` reaches COS unchanged as `image/svg+xml`.

`MessageType.File` only changes how the IM client renders the *bubble* (download chip vs inline ``). It does nothing to how the CDN serves the bytes. A user who opens the file's CDN URL directly (top-level navigation, or any client that previews File attachments in a webview) gets a document served as `image/svg+xml` with no attachment hint → the browser parses it as an SVG document and executes embedded `<script>`/`<foreignObject>` in the CDN origin. The attacker delivery path the ticket describes (ask the LLM to post a crafted SVG) is unchanged for the direct-URL case.

This is a strict improvement over the prior behavior (which inline-rendered SVG as `Image`), so it is not a regression — but it does not fully close the vector the PR set out to fix.

**Suggested fix:** in `uploadFileToCOS`, for `image/svg+xml` force `Content-Disposition: attachment` and/or rewrite the stored `ContentType` to `application/octet-stream` (or `text/plain`). A one-line branch alongside the existing disposition logic covers it.

**Test gap that hid this** (`src/__tests__/media-upload.test.ts:106-137`): the P1-5 tests assert only `payload.type === 8` and that width/height are undefined. Neither asserts the COS object's `ContentDisposition`/`ContentType`. A test pinning `attachment` (or a non-active content type) for `image/svg+xml` would have caught the serving-layer gap — please add it with the fix.

A related correctness wart (non-blocking, **not** exploitable): `isSafeInlineImage` keys off the *claimed* MIME, so SVG bytes mislabeled as `image/png` (`data:image/png;base64,<svg>`) pass the gate and route to `Image`. Because COS then serves them as `image/png`, no browser parses them as SVG, so no script runs — it renders as a broken image, not an XSS. Worth noting, not worth blocking.

---

## P0-1 — split must not break `@[uid:name]` mentions: correct ✅

`deliver()` now resolves mentions once on the full text, derives `protectedRanges` from the resolved entity offsets, splits with those ranges, then re-bases entities per segment (`src/stream-relay.ts:200-252`). `splitMessage`'s `adjustSplitForProtectedRanges` keeps any candidate cut from landing strictly inside a range.

Verified:
- Because `protectedRanges == entity ranges` and the splitter never cuts inside a range, every entity stays wholly within one segment — the per-segment filter `offset >= segStart && offset+length <= segEnd` (`src/stream-relay.ts:226-228`) cannot silently drop an entity. Confirmed with a large fuzz over random spaces/newlines/spaced names and emoji-adjacent ranges: zero drops, `segments.join("") === input` in every case.
- Per-segment offset rebasing (`e.offset - segStart`, with `segStart` accumulating `segment.length`) yields exact segment-local offsets; the segment-local slice equals the original `@name` substring.
- No infinite-loop/no-progress risk: `splitAt` stays `> 0` and advances every iteration.
- Strict improvement over the old per-segment `resolveMentions`, which broke structured mentions split mid-name.

Minor notes (non-blocking):
- `mentionAll` is hoisted to **segment 0**, not to the segment whose visible text contains the `@all`/`@所有人` token (`src/stream-relay.ts:221-234`). It still fires exactly once (the anti-spam goal is met) and no text is lost; only the visible-token/flag pairing can differ when `@all` falls in a later segment. Consider attaching it to the segment that actually contains the token, or documenting the hoist.
- Acknowledged pathological case (`src/stream-relay.ts:133-138`): a single mention longer than `maxChars=3500` is hard-cut and its entity is then dropped from both segments. Unreachable for real `@[uid:name]` and already called out in the comment — flagging only for completeness.
- The added test covers a single mention; a two-mention-in-different-segments case would strengthen the per-segment rebasing coverage.

## P1-3 — image cap + bounded concurrency: correct ✅

`RICHTEXT_MAX_IMAGES=20` cap with overflow refs pre-failed to `[alt]`, first 20 processed by a 4-worker pool over a shared `workCursor` (`src/media-upload.ts:598-636`).

Verified: `const slot = work[workCursor++]` is synchronous (only the subsequent `await` suspends), so no work item is skipped or duplicated; `slice(0,20)`/`slice(20)` form an exact complementary partition; overflow indices never enter the `uploaded` map so block assembly renders them as `[alt]`; `imageCount` reflects only successes; all-fail path returns `richText:false`. Unbounded fanout is eliminated. (Note: the cap bounds image *count*, not bytes — worst case ~4×500MB transient memory; acceptable for the stated goal.)

## P1-4 — chunked-transfer size bypass: correct ✅

The `Transform` size limiter (`src/media-upload.ts:313-330`) enforces `MAX_UPLOAD_SIZE` mid-stream under backpressure, so a no-`Content-Length` body can't fill the disk. On overflow the chunk is rejected, `pipeline` destroys the chain (cancelling the fetch body, no fd leak), and the partial temp file is unlinked. The HEAD check and the post-download `statSync` backstop remain; all three boundaries use a consistent `>` comparison. Defended against lying `Content-Length` and redirect bypass. (Transient: up to ~MAX bytes hit disk before abort — fine, but combined with the RichText concurrency of 4 there's no global `/tmp` budget; minor.)

## P1-6 — truncation surrogate guard: correct ✅

`src/stream-relay.ts:188-191` backs off one code unit when the cut lands on a high surrogate, mirroring `splitMessage`'s hard-cut guard. No orphan high surrogate; an orphan *low* surrogate is impossible (a slice can only end in a low surrogate when its high partner is included). A mid-`@[uid:name]` truncation degrades gracefully to literal text (the structured pattern fails to match), and that path is pre-existing — P1-6 shifts the cut by at most one unit and does not worsen it.

---

## Tests

Genuine, not tautological: P0-1 asserts on real sent content + segment-local entity offsets; P1-3 drives 25 refs → 5 failed / 20 uploads; P1-4 runs a real 600MB chunked stream through the limiter. Two follow-ups:
- The `uploadAndSendMedia` tests for P1-4/P1-5 hit `assertPublicUrl` (real DNS for `example.com`) before `fetchMock` intercepts — they'll be flaky/hang in offline CI. Consider stubbing the URL-policy check.
- Add the SVG serving-layer assertion described under P1-5.

672/672 passing is a good sign; the gaps above are about what the suite doesn't yet assert, not failures.

## Summary

Solid, well-documented batch. P0-1/P1-3/P1-4/P1-6 are ready. Please pair the P1-5 message-type change with a serving-layer mitigation (force `attachment` disposition / non-active content type for `image/svg+xml`) plus a test, and this is good to go.

lml2468 (commented):
**Verdict: CHANGES_REQUESTED** — agree with Yujiawei's blocking P1-5 finding. 4 of 5 fixes solid; P1-5 SVG XSS has residual serving-layer vector that's the **same "修了一半" pattern** we固化'd from PR#38 (cross-parser canonical-equivalent forms — in-app render parser fixed but CDN-served SVG parser by the browser still active).

Independent verification matrix:

### ✅ Reverse-verified working

| Fix | My revert test | Result |
|-----|----------|--------|
| P0-1 protectedRanges | Removed `protectedRanges` arg → P0-1 test fails (broken mention) → restored → 31/31 pass | ✅ test really probes bug |
| P1-6 truncation surrogate | Removed `cutAt--` → P1-6 test fails (orphan high surrogate) → restored → 31/31 pass | ✅ test really probes bug |

Both reverse-verifications confirm tests aren't tautological. Also ran independent edge probes (entity at exact boundary / adjacent entities / pathological entity > maxChars / paragraph break inside range / paragraph break pulled back to range.start): all 15/16 pass, 1 was my own off-by-one assertion error not an impl issue.

### 🔴 P1-5 SVG residual — same root cause as PR#38 "修了一半" pattern

Yujiawei's analysis is correct. The fix changes `msgType` from `Image` to `File`, but `uploadFileToCOS` still serves the bytes with `Content-Type: image/svg+xml` and **no Content-Disposition**:

- The IM bubble UI is fixed (no inline render in chat)
- **But the CDN URL directly opened in any browser (top-level navigation, webview preview, copy-link-paste)** = browser parses as SVG document → embedded `<script>`/`<foreignObject>` executes in CDN origin

This is **exactly** the "fix one parser, leave the other parser in the data flow untouched" pattern we固化'd as MEMORY lessons #28-29 (PR#38 cross-host header / encoded traversal / `%2F` server-side). The browser is another parser; fixing in-app render but not CDN serving is the same shape.

**Required fix per Yujiawei's suggestion**: in `uploadFileToCOS`, for `image/svg+xml`:
1. Force `Content-Disposition: attachment` (browser downloads instead of renders), AND
2. Optionally rewrite stored `ContentType` to `application/octet-stream` or `text/plain` (defense-in-depth — even if Content-Disposition is stripped by CDN, the MIME is inert)

Plus test pinning either `ContentDisposition === 'attachment'` or `ContentType !== 'image/svg+xml'` for SVG uploads. This would have caught the residual vector — fits our固化 "byte-safe truncation必测 N×K boundary"-style rule: **security fix tests must assert on the actual security invariant, not just adjacent behavior**.

### 🟡 Additional concerns I'd flag (non-blocking, fold into same revision)

**1. Case-sensitive MIME match bypass** (Yujiawei mentioned a related wart, here's another):

```ts
contentType === "image/svg+xml"  // case-sensitive equality
```

Verified: `isSafeInlineImage('image/SVG+XML')` returns `true` (bypass), even though MIME per RFC 2045 is case-insensitive. Server-side normalizes to lowercase typically, but a malicious server (or middlebox modifying Content-Type) could return `image/SVG+XML` and bypass the gate.

**Fix**: `contentType.toLowerCase() === "image/svg+xml"`.

**2. DNS mock gap in P1-4/P1-5 `uploadAndSendMedia` tests** (Yujiawei noted "flaky/hang in offline CI"):

`media-upload.test.ts` already has `vi.mock('node:dns/promises', ...)` at the top (we added it in PR#38 commit 44afbc5 for `example.com` resolution). Confirmed by grep — DNS mock IS in place. Yujiawei's concern doesn't manifest because the existing mock covers `example.com` → `203.0.113.42`. But worth a comment in test file confirming the dependency on that mock so future test additions don't accidentally break the isolation.

**3. mentionAll hoisted to segment 0 vs visible-token segment** (Yujiawei minor note):

Agree this is a minor semantic preference — anti-spam goal met (single notification), but `mentionAll: true` flag is on segment 0 even if `@all` text appears in segment 1 visually. Per Octo IM client behavior I think this is fine (mentionAll is a separate notification channel from text rendering), but worth a one-line comment in `deliver()` explaining the design choice.

### Cross-cutting observation (固化 for REVIEW_CHECKLIST.md)

PR#38 had **4 "修了一半"** rounds, PR#45 P1-5 is **the same class** (5th instance across both PRs): security fix that addresses one parser in the data flow but leaves another parser (CDN serving layer, downstream HTTP server, browser, OS resolver) unaddressed. Per our cross-parser canonical-equivalent forms元规则:

> **元规则**: security fix must enumerate every parser in the data flow from attacker input to terminal rendering/execution, and pin the security invariant at the **strictest enforcement boundary** (in this case: CDN-served Content-Disposition, not in-app msgType).

Adding to MEMORY as lesson #36, will propose to 飞飞 for inclusion in REVIEW_CHECKLIST.md cleanup PR.

### Approval criteria

Once P1-5 serving-layer fix + Content-Disposition test lands (plus optional case-insensitive MIME match for completeness), I'll re-review with硬规则 full套 (fresh worktree + reverse verify + edge probes + `gh api ... reviews --jq '.[-1]'` check before APPROVED) and final APPROVED.

yujiawei (approved):
# Code Review — PR #45 (cc-channel-octo)

**Verdict: APPROVED** — all five defects are correctly fixed, the build/lint/test gates are green, and adversarial probing of each fix found no exploitable bug that should block merge. One non-blocking robustness suggestion (P2) and a couple of minor notes are below.

## Verification

| Check | Result |
|---|---|
| `tsc --noEmit` (build) | ✅ clean |
| `eslint --max-warnings 0` | ✅ clean |
| Full test suite (`vitest run`) | ✅ 682/682 pass |
| `stream-relay.test.ts` + `media-upload.test.ts` | ✅ 73/73 pass |

I checked out the head SHA `5bab2f0`, read the full source (not just the diff) including `mention-utils.ts` and `url-policy.ts`, and ran an independent attempt to break each of the five fixes.

## Per-fix assessment

### ✅ P0-1 — `splitMessage` no longer breaks `@[uid:name]` across segments
`stream-relay.ts:200-265`. The redesign is correct: mentions are resolved **once** on the full accumulated text *before* splitting (so `@[uid:name]` is already `@name` by the time `splitMessage` runs), resolved-name spans are passed as `protectedRanges`, and entity offsets are correctly rebased to segment-local coordinates (`offset - segStart`). The global→local range translation (`stream-relay.ts:99-101`) and the `consumed` cursor accounting are sound. `mentionAll` is correctly latched to the first segment only via `mentionAllConsumed`, preventing `@所有人` notification spam across chunks. The design rationale comment is a good record of the trade-off.

One non-blocking edge case — see Suggestions below.

### ✅ P1-3 — bounded parallelism + per-message image cap
`media-upload.ts:626-664`. Image count is capped at `RICHTEXT_MAX_IMAGES=20` via `slice`, overflow refs are pre-recorded as failed so block assembly falls back to `[alt]` text, and the unbounded `Promise.all` is replaced by a `RICHTEXT_MAX_CONCURRENT_UPLOADS=4` worker-pool over a shared cursor. The `workCursor++` claim-before-await pattern is safe on the single-threaded event loop, so no double-processing. Block-assembly offset math (`media-upload.ts:671-701`) correctly interleaves text and images for both succeeded and failed refs.

### ✅ P1-4 — chunked transfer can no longer bypass the size check
`media-upload.ts:323-343`. The HEAD `Content-Length` check is necessary but insufficient (chunked responses omit it / can lie); the new `Transform` size limiter enforces `MAX_UPLOAD_SIZE` byte-by-byte during the stream and aborts the pipeline, and the existing `catch` unlinks the partial temp file. Correct fix.

### ✅ P1-5 — SVG XSS closed at every rendering surface
`media-upload.ts:164-171, 228-238, 474-476, 555-559`. `isSafeInlineImage` strips MIME parameters and lowercases before excluding `image/svg+xml`, so `image/svg+xml; charset=utf-8` and case variants (`IMAGE/SVG+XML`) cannot bypass it. SVG is routed to `MessageType.File` for direct media, rejected in the RichText image path, and forced to `Content-Disposition: attachment` on the COS object so the CDN serves it as a download rather than an inline-rendered (script-executing) document. The `resolveMedia` data-URI rewrite (RFC 2397: first-comma split, `;base64` detection, param stripping, base64 vs `decodeURIComponent`) is also correct. All three vectors (inline Image message, CDN browser render, RichText block) are covered.

### ✅ P1-6 — truncation no longer splits a surrogate pair
`stream-relay.ts:185-194`. Backs off one code unit when the cut lands on a high surrogate, mirroring `splitMessage`'s hard-cut guard. Correct, and matches the existing pattern.

## Suggestions (non-blocking)

**[P2] `splitMessage` hard-cut still breaks a protected range that both starts at offset 0 and is longer than `maxChars`.** At `stream-relay.ts:137-138`, when `adjustSplitForProtectedRanges` returns `null` (range starts at 0, can't pull back), the code falls through and cuts at `maxChars`, splitting the mention. This is the case the inline comment already acknowledges and deliberately accepts to avoid an infinite loop. It is **not** a realistic merge blocker: the protected ranges are resolved `@name` spans derived from human display names, and triggering it requires a single mention longer than `maxChars` (3500 in production). I verified with a tiny `maxChars` that the break reproduces, and that a 100-char name under the default 3500 stays intact. If you want full robustness anyway, when a range starts at 0 and `end > maxChars`, set `splitAt = range.end` for that one segment (let it temporarily exceed `maxChars`) rather than cutting through the mention. Optional.

**[nit] Stream size-limit cleanup.** The `pipeline(nodeStream, sizeLimiter, ws)` rejection unlinks the temp file, but the aborted upstream `fetch` body is left for GC rather than explicitly cancelled. Harmless in practice; an explicit `body.cancel()` on the error path would release the socket sooner. Optional.

**[nit] Test coverage gap for the P0-1 edge case.** The new tests cover the in-range mention-split scenario well, but not a mention at the very start of the text. A regression test for "long `@name` at offset 0" would lock in whatever behavior you decide for the P2 above.

## Summary

Five well-scoped, correctly-implemented fixes with strong test coverage and clean build/lint/test gates. No P0/P1 blockers. Approving.

dolphinsboy (approved):
✅ APPROVED (5bab2f0c8180)

Diff verified: 2 files +20/0，零行为变更，纯注释/docstring 插入。

stream-relay.ts +13 行 inline comment 解释 mentionAll 附在 first segment 的三条理由 (wire-level 解耦 / 最小延迟 / LLM rewrite 无歧义) + explicitly reject 多处 @all 的双重通知问题。逻辑正确，comment 位置在 useMentionAll 判断前合理。

media-upload.test.ts +7 行 header docstring 补 DNS mock contract（加新 hostname 必须扩 mock）。落点对——inline test 文件比独立 docs PR 更贴近未来作者。

Scope discipline：inline comment + test docstring 直接响应 review nit，不触发 #46 scope drift 同款 flag。audit trail 与代码长期绑定。

5bab2f0 相对 fadfea8e3c4d 是 doc-only superset，等 dolphinsboy/yujiawei 在当前 head 独立 re-review。

dolphinsboy (approved):
✅ cc-channel-octo PR #45 — APPROVED (5bab2f0c8180)

2 文件 +20/0，纯 inline comment / docstring，零行为变更，diff 实查确认。

stream-relay.ts +13：mentionAll 附在 first segment 的三条设计理由（wire-level flag 与字面位置解耦 / first segment 先发最小通知延迟 / literal text 可能被 LLM rewrite 到多处 → pin segment 0 无歧义）+ alternative considered reject 理由（多处 @all 双重通知）。注释直接回应大锤 review nit，设计意图文档化清楚。

media-upload.test.ts +7：DNS mock contract 说明 'add hostname → extend mock，不靠 real resolver'，classic CI/local 分叉反模式写进 file-level header，对未来测试作者可见。直接对应 PR#46 docs §5 同款教训，但落点是 inline test docstring 而不是独立 docs 文件，scope discipline 正确。

齐哥的 #46 d0abb5b 对比分析完全同意：这次是 risk verification 后的正常通过，不是 cost 论证特例放过。

lml2468 (commented):
**Verdict: LGTM** ✅ — hardening commit `f93a1d7` + docs commit `5bab2f0` close all 5 P1-5 vectors I + Yujiawei flagged. Three-bug coordinated fix (isSafeInlineImage MIME normalize + data URI parser RFC 2397 compliance + Content-Disposition for SVG) is the right shape — pins security invariant at multiple layers, not just one.

Per Checklist #0/2/3 硬规则 4 步 + 我承诺的 7 边界 case 全部跑了 + 反向验证每个 fix 真在 probe bug。

### Fresh worktree verification

- `/tmp/cc-dachui-c2-rereview` at `5bab2f0` (worktree-isolated per Checklist #0)
- `gh api .../pulls/45/reviews --jq '.[-1]'` → Yujiawei APPROVED at `5bab2f0` (18:37Z), no newer BLOCKING since
- tsc clean / eslint 0 warnings / **682/682 tests pass**

### Reverse-verification matrix (3 of 3 hardening fixes真在 probe bug)

| Fix | My revert | Result |
|-----|-----------|--------|
| #1 isSafeInlineImage strip params + lowercase | Replaced with old startsWith + === | ⚠️ Test still passed (see below) |
| #2 data URI parser RFC 2397 | Reverted to regex `/^data:([^;,]+)?(?:;base64)?,(.*)$/` | ✅ Test fails |
| #3 uploadFileToCOS SVG Content-Disposition | Reverted to old `!ct.startsWith('image/')` branch | ✅ Test fails |

⚠️ **One observation on test design** (non-blocking, not a bug):

Revert ONLY #1 (isSafeInlineImage) without reverting #2 → test still passes. Reason: fix #2 already strips MIME params in the data URI parser BEFORE passing contentType to isSafeInlineImage. So for variant 1 (`image/svg+xml; charset=utf-8`), the parser delivers `image/svg+xml` (canonical) → isSafeInlineImage rejects regardless of param-strip in fix #1. For variants 2/3 (uppercase / mixed case without spaces), without fix #1, `startsWith("image/")` returns FALSE (case mismatch) → returns false → routed to File anyway = "accidentally safe."

This means fix #1 alone doesn't have a single failing test in isolation. The test passes because OTHER fixes (#2 strips, #3 forces attachment) collectively close the attack. Each fix on its own is正确 but the test suite doesn't pin each invariant independently.

**Suggestion** (non-blocking, can fold into Wave 3/4 polish): split the P1-5 hardening test into 3 narrower tests, one per fix, each asserting the specific invariant that fix protects. Current design tests "attack X is blocked" — 应该 also test "attacker can't bypass fix #1 by sending uppercase variants directly to a downstream that already canonicalized contentType".

### 16 independent edge probes (15/15 pass + 1 documented gap)

Wrote a probe sweep of `isSafeInlineImage` across MIME variants:

| Variant | Expected | Got |
|---------|----------|-----|
| canonical `image/svg+xml` | reject | ✅ |
| `; charset=utf-8` (space) | reject | ✅ |
| `;charset=utf-8` (no space) | reject | ✅ |
| ` ;charset=utf-8` (space before semi) | reject | ✅ |
| `;` (trailing only) | reject | ✅ |
| `IMAGE/SVG+XML` | reject | ✅ |
| `Image/Svg+Xml` | reject | ✅ |
| `image/SVG+xml` | reject | ✅ |
| `IMAGE/svg+xml; charset=UTF-8` | reject | ✅ |
| `image/svg+xml ` (trailing space) | reject | ✅ |
| `  image/svg+xml` (leading space) | reject | ✅ |
| `image/png`, `jpeg`, `webp`, `IMAGE/PNG` | accept | ✅ |
| `image/jpeg; quality=80` (benign param) | accept | ✅ |
| `image/svg` (without `+xml`) | **accept** | ⚠️ **potential gap** |

The `image/svg` case (no `+xml` suffix) is rare/wrong but Firefox HAS rendered it as SVG in some legacy contexts. Most browsers treat it as octet-stream so no inline render = no XSS, but it's a non-zero edge. Suggested fix: add `image/svg` to SVG check list as belt-and-suspenders (1 char change in the `===` comparison → use `endsWith("/svg+xml") || endsWith("/svg")` semantics).

### RichText path SVG variant coverage probe

Tested `sendRichTextCombined` with the 3 hardening variants in markdown `svg`:

- `IMAGE/SVG+XML;base64` (uppercase, no spaces) → ✅ rejected, failedMedia.length === 1
- `Image/Svg+Xml;base64` (mixed case, no spaces) → ✅ rejected, failedMedia.length === 1
- `image/svg+xml; charset=utf-8;base64` (with space after semi) → ⚠️ **markdown regex precludes URL with space**, silent skip (failedMedia.length === 0)

The spaced variant is silently skipped because `MARKDOWN_IMAGE_RE` URL group is `[^)\s]+` (rejects whitespace). This is **NOT a security bypass** (image isn't processed = no inline render of attacker SVG via this path), but IS a usability gap — legitimate spaced data URIs in markdown won't upload. RichText markdown spec generally requires URL-encoded spaces (`%20`) so this is per-spec correct, but worth a code comment noting "spaced URLs require URL encoding per CommonMark spec — defensive design."

### Meta-validation: cross-parser canonical-equivalent forms pattern (third confirmation tonight)

This hardening pass is the **第 5 次 cross-parser "修了一半" pattern**, with two layers fixed:
- Layer 1 (in-app MIME comparison): fix #1 strips params + lowercase
- Layer 2 (data URI parser RFC 2397): fix #2 strips params before downstream
- Layer 3 (CDN serving via COS): fix #3 forces Content-Disposition: attachment

Three layers, each pinned independently = correct application of REVIEW_CHECKLIST #9.2 cross-parser meta-rule + new #11 "pin invariant at strictest enforcement boundary" (CDN serving is the strictest). 飞飞 PR#46 §9.3 Recurrence count update from 4 to 5 now justified — this 5th instance also closed via multi-layer fix, matching the pattern.

### REVIEW_CHECKLIST.md case study suggestion

For飞飞's REVIEW_CHECKLIST.md amend, suggest adding the MIME-types case to §9.2:

> **MIME types canonicalization checklist** (PR#45 P1-5 hardening case study):
> 1. Strip parameters: `split(';')[0].trim()` before any base-type comparison
> 2. Lowercase before equality check: `toLowerCase()` (RFC 6838 §4.2)
> 3. Mind the data URI parser: RFC 2397 allows params before `;base64,` — regex `[^;,]+` is wrong, use split-at-first-comma
> 4. Pin invariant at storage layer too: `Content-Disposition: attachment` for SVG forces browser download even if in-app gate misses the attack

### 3 nit acknowledgments from豆豆 docs commit 5bab2f0

- mentionAll segment 0 hoist comment ✅
- DNS mock contract note ✅
- (case-insensitive MIME match already in hardening fix #1) ✅

All my 3 nits from CHANGES_REQUESTED addressed. APPROVED.

`<!-- 王大锤 re-review verdict: LGTM -->`
`<!-- head SHA reviewed: 5bab2f0 (hardening f93a1d7 + docs 5bab2f0) -->`
`<!-- aligns with Yujiawei APPROVED (18:37Z on 5bab2f0) -->`

PR #8, opened 2026-06-04
docs: add CONTRIBUTING.md and Apache-2.0 LICENSE (T15 + T17)

lml2468:
## T15: CONTRIBUTING.md

Comprehensive contributor guide covering:
- **Development setup**: prerequisites, install, useful commands
- **Project structure**: file-by-file overview with cross-reference to ARCHITECTURE.md
- **Code style**: English-only artifacts, strict TypeScript, no silent error swallowing
- **Testing**: vitest patterns, mock guidance, in-memory SQLite for DB tests
- **Pre-commit checklist**: type-check → test → lint → selective git add → verify branch
- **PR process**: naming, scope, rebase, review states (APPROVED/CHANGES_REQUESTED/COMMENTED)
- **Protocol layer sync**: release-time diff procedure against upstream openclaw-channel-octo, what to port (security fixes, protocol changes) vs what to skip (OpenClaw-specific features)
- **Security**: vulnerability reporting, no hardcoded secrets, dataDir permissions, bypassPermissions design rationale

## T17: LICENSE

Apache License 2.0, Copyright 2026 Mininglamp-OSS.

Consistent with:
- `package.json` (`"license": "Apache-2.0"`)
- openclaw-channel-octo upstream license

## Verification

- `npx tsc --noEmit` — zero errors ✅
- `npm test` — 95/95 passing ✅
- 2 new files, no source code changes

PR #46, opened 2026-06-05
fix(C1 follow-up): O(n) truncateByBytes + trim RejectionReason + %2F defense

lml2468:
C1 follow-up cleanup, three non-blocking items bundled per 大锤's coordination: (1) truncateByBytes O(n²)→O(n) via PR#42 helper delegation (齐哥 PR#41 nit); (2) RejectionReason union trimmed 7→2 emitted reasons (齐哥 PR#41 nit); (3) buildMediaUrl rejects %2F encoded slash, server-decode defense-in-depth (我 PR#38 final note). +10 tests, 675/675 pass. tsc clean. eslint 0. Worktree-isolated build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,12 @@ This repo enforces three layers of automated checks:
 
 Coverage artifacts are uploaded per CI run (retained 14 days). No hard threshold yet — baselines are being established.
 
+Reviewers must follow [`docs/REVIEW_CHECKLIST.md`](./docs/REVIEW_CHECKLIST.md)
+on security-adjacent PRs. The 9-item checklist is distilled from Stage 6
+review-process failures and codifies hard-won rules like "reproduction test
+before APPROVED", "refresh reviews state before clicking APPROVED",
+"enumerate canonical-equivalent forms for attacker-input validation", etc.
+
 ### Project Structure
 
 ```
--- a/docs/REVIEW_CHECKLIST.md
+++ b/docs/REVIEW_CHECKLIST.md
@@ -1,0 +1,262 @@
+# Code Review Checklist (cc-channel-octo)
+
+> Living document, distilled from Stage 6 (v0.1.1) review experience.
+> 9 checks that took five reviewers + multiple "按错按钮" failures to sink in.
+> If you skip any of these on a security-adjacent PR, you are repeating
+> someone else's mistake from June 2026.
+
+## 0. Stage 6 ground rules
+
+Before opening or reviewing a PR:
+
+- **Work in an isolated `git worktree`.** Shared clone directories
+  (`/tmp/cc-channel-octo`) are an accuracy hazard when multiple
+  reviewers/authors operate in parallel — checkouts get clobbered, refs
+  drift, and you end up validating against the wrong code. Use
+  `git worktree add /tmp/cc-<name>-<task> <ref>` and remove it when done.
+- **CI is the source of truth, not your laptop.** macOS DNS / file system /
+  Node version differ from the GitHub Actions Linux runner. A green local
+  test suite with a red CI is a portable-test design bug, not a fix bug.
+
+## 1. CHANGES_REQUESTED re-review: reproduction first, APPROVED last
+
+When re-reviewing a PR you previously CHANGES_REQUESTED:
+
+- **First** write a test that reproduces the specific P0/P1 attack chain
+  from your original BLOCKING comment.
+- Run it. See the defense in the new code fire green.
+- Then APPROVED.
+
+Generic corner-case probes (IP form coverage, input edge cases) **do not**
+substitute for reproduction tests — they only re-verify what the PR's own
+tests already cover, not the specific attack you said was missing.
+
+## 2. Refresh PR reviews state before clicking APPROVED
+
+```bash
+gh api repos/<owner>/<repo>/pulls/<N>/reviews --jq '.[-1]'
+```
+
+Confirm no newer BLOCKING / CHANGES_REQUESTED has landed since you started
+writing your APPROVED comment. Today's automated reviewers run
+asynchronously and can post a new BLOCKING three minutes before you submit.
+
+Retraction (`COMMENTED: I retract`) is honest but **the click itself is the
+mistake** — once you have APPROVED, the PR can be merged immediately on
+CI green. The review process's job is to never reach the retraction step.
+
+## 3. Reviewer must re-approve after author pushes new commits
+
+After APPROVED, any new commit from the author — even one labelled `nit`
+or `fix typo` — must be re-reviewed. A one-character change can introduce
+a fresh corner case (PR#40: `truncateByBytes` `i < 3` nit fix introduced a
+4-byte UTF-8 U+FFFD bug on `N × 4` clean boundaries).
+
+## 4. Protocol/contract changes require test-suite-wide audit
+
+If the PR changes the signature or runtime behaviour of a helper that is
+called from multiple places (e.g. `tryResolveFile` adds `assertPublicUrl`,
+`fetchWithRedirectGuard` adds per-hop credential scoping):
+
+```bash
+grep -rn '<helper-name>' src/__tests__/
+```
+
+Audit **every** caller test for second-order impact. Tests that didn't
+previously hit DNS / network / external state may start hitting it.
+
+## 5. Network tests: `vi.mock('node:dns/promises')` by default
+
+Never write a test that depends on the runner's DNS to return NXDOMAIN for
+a "fictitious" hostname like `attacker.example.com`. Local macOS, Linux
+CI, corporate DNS, and offline development all fail differently — that's a
+guaranteed "本地绿 CI 红" portable-test bug.
+
+Pattern:
+
+```typescript
+vi.mock('node:dns/promises', () => ({
+  lookup: vi.fn(async (hostname: string) => {
+    if (hostname.includes('example.com')) {
+      return [{ address: '203.0.113.42', family: 4 }];  // TEST-NET-3, treated as public
+    }
+    throw new Error(`Test DNS mock: unexpected hostname ${hostname}`);
+  }),
+}));
+```
+
+Unknown hostname throws — surfaces a test bug instead of silently passing.
+
+## 6. Cross-host header behaviour in fetch wrappers: default-block
+
+Any fetch wrapper that follows redirects must re-decide which headers
+(especially `Authorization`, `Cookie`, API keys) to send on each hop.
+Reusing the initial `init.headers` across redirects leaks credentials when
+hop N is on a different host than hop 0 (PR#38 round-2 P0).
+
+Pattern: per-hop `headerPolicy` callback that gets `currentUrl` and
+decides what to attach. `tryResolveFile` ↔ `fetchWithRedirectGuard` in
+`url-policy.ts` is the canonical example.
+
+## 7. Byte-safe truncation: probe ALL `N × max-sequence-bytes` boundaries
+
+For any string-cap helper that operates on UTF-8 / surrogate pairs / any
+variable-length encoding, the regression suite **must** include `cap =
+N × max-sequence-length` clean-boundary cases:
+
+- 2-byte sequences (`ñ`): `cap = N × 2`
+- 3-byte sequences (CJK `中`): `cap = N × 3`
+- 4-byte sequences (emoji `🚀`): `cap = N × 4`
+
+When the cap lands exactly on the final continuation byte of a complete
+sequence, the algorithm must keep the complete sequence — not over-trim
+into the leader and produce `U+FFFD`. Simple `for (i < max-len)` loops
+get this wrong; walk-back-to-leader + `actualLen === expectedLen` is the
+canonical algorithm (PR#42).
+
+## 8. Encoded form path traversal: defer to canonical parser
+
+For URL path validation, never enumerate percent-encoded escape variants
+by hand (`%2e`, `%2E`, `%2e.`, `.%2e`, `%252e%252e`, ...). The downstream
+WHATWG URL parser decodes them all for dot-segment normalization, but
+your `if (seg === '..')` check sees only the literal form.
+
+Canonical fix:
+
+```typescript
+const candidate = `${baseUrl}/file/${storagePath}`;
+const normalized = new URL(candidate);
+if (!normalized.pathname.startsWith('/file/')) return undefined;
+```
+
+Defense-in-depth for `%2F` (which WHATWG keeps literal but some servers
+decode): reject `%2F` / `%2f` in the storage path outright. Production
+URLs never contain it.
+
+## 9. Attacker-input validation: enumerate ALL canonical-equivalent forms
+
+The "修了一半" meta-pattern from Stage 6's S1/S2/S4 (v4-mapped IPv6 hex /
+cross-host redirect header reuse / encoded path traversal / server-side
+`%2F` decoding) all share one root cause: the validator thinks of one
+canonical form when the downstream parser will accept several.
+
+For any validator on attacker-controlled input:
+
+1. **List every canonical-equivalent form** the downstream parser/consumer
+   will accept. Use spec docs, not intuition.
+2. **Defer to the canonical parser** when possible (WHATWG URL, IP
+   parsing libs). Re-implementing canonicalisation by hand is how you
+   miss the fourth variant.
+3. **Defense-in-depth** for ambiguity zones where the spec is one thing
+   but real servers diverge (e.g. WHATWG keeps `%2F` literal but Apache
+   `AllowEncodedSlashes` decodes it).
+4. **Test matrix**: at minimum, lowercase / uppercase / mixed case
+   percent variants + double encoding + the literal form. Add nested
+   forms when the parser does recursive normalization.
+
+### 9.1 Case study: WHATWG `new URL()` covers three attacker bypass classes
+
+陈皮皮's cross-team meta-insight (Stage 6 closing): a single canonical
+parser can cover multiple attacker bypass classes at once. `new URL()`
+for URL sandbox checks:
+
+| Bypass class | Example | How WHATWG normalize handles it |
+|--------------|---------|--------------------------------|
+| Dot-segment encoding (S4) | `%2e%2e/internal`, `%2E.`, `.%2e`, `%252e%252e` | Decodes `%2e` for dot-segment normalization, collapses `..`; `pathname.startsWith('/file/')` becomes false |
+| IPv4 alternative forms | `2130706433`, `0x7f000001`, `0177.0.0.1`, `127.1`, `127.0.1` | All canonicalize to `127.0.0.1` in `URL.hostname` → `isPrivateOrLocalAddress` catches via 127/8 |
+| IPv6 v4-mapped | `[::ffff:127.0.0.1]`, `[::ffff:7f00:1]` | Bracket literal preserved, S5 covers explicitly (URL doesn't auto-expand v4-mapped) |
+
+Does NOT cover:
+
+- `%2F` (encoded slash) — spec says preserve literal; server-side
+  decoding variance means defense-in-depth at the codebase boundary
+  (`if (path.includes('%2F') || path.includes('%2f')) return undefined`).
+- Double encoding `%252e%252e` — spec decodes to literal `%2e%2e`, not
+  `..`. WHATWG sandbox check accepts (path stays under `/file/`). If
+  downstream server double-decodes, that's a server-side concern.
+- DNS rebinding — OS resolver result may change between `assertPublicUrl`
+  call and `fetch`. Mitigated by short TTL + trusted-domain assumption;
+  for full safety, pin IP and connect by IP (breaks TLS SNI / CDN routing).
+
+### 9.2 Cross-parser canonical-form meta-rule
+
+王大锤's elaboration after the fourth same-root finding (`%2F`
+server-side decoding). Apply to any attacker-input field:
+
+**1. List every parser the input passes through, end to end.**
+
+URL example: input → Node WHATWG `new URL()` → HTTP server URL decoder
+→ CDN edge → OS resolver.
+Header example: input → HTTP framework parser → middleware decoder →
+application code.
+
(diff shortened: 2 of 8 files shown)


dolphinsboy (approved):
✅ cc-channel-octo PR #46 — APPROVED (7443fa69d67b)

6 文件 +149/-32，三个 C1 非阻塞 follow-up 全部正确，无阻塞项。

**C1-#1: truncateByBytes O(n²) → O(n)**
原实现：`input.slice(0, maxBytes)` 后 `while (Buffer.byteLength > cap) { slice(-1) }` — 每次 slice 都是 O(n)，CJK 大输入最坏 O(n²)。
修复：委托给 `truncateUtf8ByBytes`（file-inline-wrap.ts），O(1) walk-back（Buffer.from 一次 encode，从尾部找合法 UTF-8 边界）。
测试验证：
- 350K 中文（≈1MB UTF-8）< 500ms ✓
- byte-safe 边界无 `\uFFFD` ✓
- MultipleForward 路径也覆盖（regression check）✓
- 小输入无截断 marker 的 benign case ✓

这是我在 PR#41 提的非阻塞 review note，正确纳入。

**C1-#2: RejectionReason union 剪枝**
原 union 包含 7 个成员（`blocked_bot / self_message / system_event / rate_limited / global_rate_limited / oversized / not_mentioned`），但代码实际只 emit `rate_limited` 和 `oversized`。宽 union 误导：调用方可能对永不出现的值写 branch，或误以为 `global_rate_limited` 已实现。
修复：trim 到 `'rate_limited' | 'oversized'`，并在 JSDoc 说明"未来新 emit 点添加到这里"（明确 extension point）。
`SUPPRESS_GROUP_CACHE` set 同步从两处删掉 `'global_rate_limited'`（e2e.test.ts 和 index.ts）。行为不变（`global_rate_limited` 从未被 emit，set 里删它无影响）。
类型测试：import SessionRouter 的 smoke 测试 + tsc 通过即是编译期 proof。✓

**C1-#3: buildMediaUrl %2F defense-in-depth**
这是我在上轮（6b55aec5618c）提出的残余风险：WHATWG pathname 不 decode `%2F`，所以 sandbox check 通过，但 Apache `AllowEncodedSlashes On` 等服务端会 decode `%2F` → `/` 再解析 dot-segment，沙箱被绕过。

修复：在 WHATWG sandbox check **之前**，先 reject `storagePath.includes('%2f') || storagePath.includes('%2F')`（case-insensitive）。

分析正确：
- 生产 storage path 不含字面 `%2F`（合法 path segment 不含未编码斜杠；如有必要会双编码为 `%252F`）
- 代价极低（两次 string.includes）
- 防御层次明确：HTTP-fetch 层做 belt，server-side decode 是 suspenders

修复同时删除了原先 accept 的测试 case `'..%2f..%2finternal/secret.env'`（6b55aec5618c 里被标为 accept）——该 case 现在被 reject，行为正确反转。新增 5 个 reject case：lowercase / uppercase / mixed / benign-looking `a%2Fb` / `file%2Ftest.txt`，全部应 return undefined。

JSDoc 注释详细，threat model 文档化到位。

**三项全部清账，工程纪律到位。**

yujiawei (approved):
# Code Review — PR #46 (cc-channel-octo)

**Verdict: APPROVED.** Three well-scoped, independently-verified follow-up cleanups. No P0/P1 blockers. Code is correct, tests are reasonable, and the change set matches the stated scope. A few non-blocking notes below.

## 1. Verification

### ✅ Item #1 — `truncateByBytes` O(n²) → O(n) (`src/inbound.ts:229-231`)
The previous implementation was genuinely quadratic for CJK input: `input.slice(0, maxBytes)` slices on UTF-16 code units, so for `maxBytes`-char CJK text the intermediate string was ~3× over the byte cap, and the per-char `while` loop then called `Buffer.byteLength` (O(n)) on each shrink — O(n²) overall. The new code delegates to `truncateUtf8ByBytes` (`src/file-inline-wrap.ts:123`), which does a single O(n) `Buffer.from` encode plus an O(1) walk-back over at most 3 UTF-8 continuation bytes. Behavior is preserved exactly: marker is appended only when `wasTruncated`, the pre-marker byte length still stays ≤ cap, and the byte-safe boundary logic in the helper avoids U+FFFD. Both call sites (`inbound.ts:278` RichText, `:363` MultipleForward) inherit the fix.

### ✅ Item #2 — `RejectionReason` union trimmed 7 → 2 (`src/session-router.ts:32`)
Confirmed the only emit sites are `rate_limited` (`session-router.ts:274`) and `oversized` (`:294`); the removed values (`blocked_bot`, `self_message`, `system_event`, `global_rate_limited`, `not_mentioned`) had no producer in `src/` — they were dead members of the type. The single consumer (`index.ts:388` `SUPPRESS_GROUP_CACHE.has(...)`) stays type-safe, and the `SUPPRESS` sets in `index.ts:386` and `e2e.test.ts` were updated in lockstep to drop the now-unreachable `global_rate_limited`. The interface is internal (`RouteResult`), not serialized, so there is no external contract to break. The updated doc-comment correctly notes "add reasons here when emit sites land," which keeps the type honest about its actual contract.

### ✅ Item #3 — `buildMediaUrl` rejects `%2F` (`src/inbound.ts:186-188`)
Sound defense-in-depth. The WHATWG `new URL()` sandbox check (`:196-198`) correctly does *not* decode `%2F` in the pathname, so `..%2f..%2finternal` is spec-correctly treated as a single segment under `/file/` and would pass. But servers/proxies with `AllowEncodedSlashes On` decode `%2F` and *then* resolve dot-segments, escaping the sandbox. Rejecting any `%2F` outright before that check closes the gap, and the rationale (legitimate filenames never carry a raw `%2F`) holds. Both lower/upper-case forms are covered. New `it.each` tests (5 cases) cover classic/upper/mixed/any/benign variants.

## 2. Issues found

None at P0/P1.

## 3. Suggestions (non-blocking, P2 / nits)

- **(P2, observation) `%2F` check is scoped to the relative-path branch only.** The absolute same-host branch (`inbound.ts:132-143`) returns the input URL verbatim once the host matches, with no `/file/` sandbox or `%2F` filtering. This is a *pre-existing* design choice (same-host absolute URLs are fully trusted), not a regression introduced here, so it's out of scope for this PR — but worth tracking separately, since an attacker who can supply an absolute same-host URL already bypasses the `/file/` sandbox entirely.
- **(nit) Timing-based perf assertions** (`c1-followup-cleanup.test.ts`, `elapsed < 500`) can be flaky on loaded CI runners. The 500ms threshold has wide headroom against the ~10s O(n²) regression it guards, so this is acceptable, but a hard wall-clock bound is inherently environment-sensitive.
- **(nit) The `%2F` substring check** uses two `.includes()` calls; a single `/%2f/i.test(storagePath)` would be marginally tighter. Cosmetic only.
- **(nit) Item #2 test** in `c1-followup-cleanup.test.ts` is a type-level smoke test (only asserts `SessionRouter` is a function). It's correctly cross-referenced to the real behavioral coverage, so this is fine — just flagging that the test file itself doesn't exercise the trimmed union at runtime.

## 4. Additional notes

- `tsc` clean / eslint 0 / 675 tests passing as claimed; the trim in Item #2 is the kind of change that only compiles cleanly *because* the removed members were truly unreferenced, which corroborates the dead-code finding.

yujiawei (approved):
# Code Review — PR #46 (cc-channel-octo)

Independent review at head SHA `d0abb5b`. Checked out the exact SHA in an isolated clone, ran `tsc --noEmit` (clean), `eslint src/` (0 issues), and the full `vitest` suite (**675/675 pass**). All three changes verified by code reading + targeted call-site audits.

## Verdict: APPROVED

No P0/P1 findings. The three bundled cleanups are correct, well-tested, and behavior-preserving where they should be. Two minor maintainability nits below (non-blocking).

---

## 1. `truncateByBytes` O(n²) → O(n) delegation ✅

`truncateByBytes` (`src/inbound.ts:225`) now delegates to `truncateUtf8ByBytes` (`src/file-inline-wrap.ts:123`). Verified the helper is genuinely O(n):

- `Buffer.from(input, 'utf-8')` once (O(n)), then an **O(1) walk-back** over at most 3 continuation bytes to find a clean UTF-8 boundary (`file-inline-wrap.ts:147-170`).
- The leader-length check (`actualLen === expectedLen`, lines 162-167) correctly keeps a complete final sequence and only drops a partial/malformed one — this is the fix for the `N × 4`-byte clean-boundary U+FFFD bug from the PR#40 history, and it is implemented correctly here.
- The previous in-line `while (... slice(0,-1))` loop (O(n²) for long CJK due to per-iteration `Buffer.byteLength`) is fully removed.

The `{ text, truncated }` ↔ `{ truncated, wasTruncated }` adaptation at `inbound.ts:235-236` preserves the original contract (marker only appended when truncation occurred). Regression tests cover 1 MB CJK under the time bound, clean-boundary byte-safety, and the `MultipleForward` path.

## 2. `RejectionReason` union trimmed 7 → 2 ✅

This is the change with the highest cross-file risk, so I audited every emit site and every consumer rather than trusting the type:

- **Emit sites**: `grep` for `rejectionReason:` across `src/` (excluding tests) returns exactly two — `'rate_limited'` (`session-router.ts:274`) and `'oversized'` (`session-router.ts:294`). All other rejection paths in `routeMessage` (`blocked bot`, `self message`, `system event`, `not mentioned`) `return null` and never construct a `RouteResult`, so they never carried a reason. The trim is sound.
- **`global_rate_limited` in particular**: `checkAllRateLimits` (`session-router.ts:341-372`) checks global → user → session buckets and, on any block, the single call site emits `rejectionReason: 'rate_limited'`. There was never a distinct `'global_rate_limited'` value emitted at runtime, so removing it from the union is behavior-neutral.
- **Consumers**: both `SUPPRESS` sets that gate group-context caching were updated consistently — `src/index.ts:386` and `src/__tests__/e2e.test.ts:262` dropped `'global_rate_limited'`. Since that value was never produced, dropping it from the suppression sets changes no behavior. `tsc --noEmit` passing confirms no dangling references to any of the five removed literals anywhere in the tree.

The expanded doc comment on the field (`session-router.ts:14-31`) accurately documents the "silent-drop cases leave it undefined" contract.

## 3. `buildMediaUrl` `%2F` defense-in-depth ✅

The guard (`inbound.ts:186-188`) rejects any `%2f`/`%2F` in the relative storage path. Verified placement and rationale:

- It sits on the **relative-path branch**, after prefix-strip and before the WHATWG `new URL()` sandbox check (`inbound.ts:196-198`) — exactly where the `/file/` sandbox is enforced. The reasoning in the comment is correct: WHATWG keeps `%2F` literal in `pathname` (so the existing sandbox check would accept `..%2f..%2finternal`), but Apache `AllowEncodedSlashes On` / some proxies decode it server-side and then resolve dot-segments. Rejecting outright closes that boundary gap.
- The absolute-http(s) branch (`inbound.ts:132-143`) intentionally has no `/file/` sandbox (host-match chokepoint only), so the `%2F` guard correctly does not apply there.
- Test coverage in `inbound.test.ts:423-431` exercises lowercase / uppercase / mixed-encoded / any-`%2F` / benign-looking inputs, all expecting `undefined`. The previously-passing `..%2f..%2finternal` case was correctly inverted (removed from the "passes through" `it.each` at the old line 408, now expected to reject). False-positive risk on real production paths is negligible (legitimate filenames encode `/` as `%252F`).

## Docs (`README.md`, `docs/REVIEW_CHECKLIST.md`)

Documentation-only additions; no runtime impact. The checklist is a reasonable distillation of prior review-process lessons.

---

## Non-blocking nits (P2)

1. **Placeholder tests in `c1-followup-cleanup.test.ts`.** Cases #2 (`session-router` import smoke) and #3 (`expect(true).toBe(true)` cross-reference) carry no real assertions — the actual coverage lives in `inbound.test.ts` and the existing rejection-guard suite. They're harmless documentation, but a future reader may mistake them for real coverage. Consider either folding the real assertions in or dropping the no-op `it` blocks. Not blocking — the genuine coverage exists elsewhere and passes.
2. **Wall-clock timing assertions** (`elapsed < 500`, `c1-followup-cleanup.test.ts`) are inherently CI-environment sensitive. The 500 ms threshold is generous relative to the expected sub-100 ms O(n) cost, so flakiness risk is low, but a complexity assertion is the kind of thing that can occasionally trip on a loaded runner. Acceptable as-is.

## Verification summary

| Check | Result |
|---|---|
| `tsc --noEmit` | ✅ clean |
| `eslint src/` | ✅ 0 issues |
| `vitest run` | ✅ 675/675 pass |
| All `rejectionReason` emit sites audited | ✅ only the 2 in the trimmed union |
| `%2F` guard placement vs. sandbox | ✅ correct |
| `truncateUtf8ByBytes` O(n) + boundary-safe | ✅ verified |

lml2468:
**Review: APPROVED with 1 nit + 1 P2 finding** ✅ (齐静春)

Verified head 3de6cca merged against latest main (含 PR#44 Q1): tsc clean / eslint zero warnings / **684/684 全过** (674 baseline + 10 new). git auto-merged cleanly — Q1's tautology deletions in pr33-followup.test.ts / q34-q37-q38.test.ts survive because PR#46 didn't touch those.

### Three fixes 设计验证 ✅

**#1 truncateByBytes delegation** — correct refactor. Removes the O(n²) per-char shrink loop in favor of import from `file-inline-wrap.ts truncateUtf8ByBytes` (大锤 PR#42 helper). Single source of truth, helper drift eliminated.

Reverse-verified: reverted to old O(n²) impl → byte-safety + no-U+FFFD assertions still pass on M2 (52ms). So the **timing test (`expect(elapsed).toBeLessThan(500)`) does not actually catch the regression on M2** — see P2 finding below. The byte-safety properties hold for both impls (which is fine — correctness is what matters), but the perf assertion is theatre on fast hardware.

**#2 RejectionReason trim 7→2** — exactly what I asked for in PR#41 review. Type now matches what code actually emits. Comment explicitly documents that silent-drop cases (blocked_bot/self_message/system_event/not_mentioned) deliberately leave rejectionReason undefined because those messages are legitimate group chatter that should reach [Group context]. `SUPPRESS_GROUP_CACHE` set in index.ts + e2e.test.ts correctly trimmed to 2 elements consistent with the type.

**#3 %2F defense** — cheap, correct, low-risk. 5 it.each cases (lowercase / uppercase / mixed encoded+literal / any %2F / benign-looking %2F) all reject. Reverse-verified: replaced the guard with `if (false)` → all 5 tests fail; restored → all pass. Real probe, not happy-path. 大锤's trace of downstream impact (path.basename / decodeURIComponent / inferContentType) confirms zero impact.

### REVIEW_CHECKLIST.md + README ✅

9 checks + Quick Recall Card + Stage 6 ground rules. Captures the actual mistakes — fresh worktree, refresh reviews state, reproduction first, contract change → grep all callers, vi.mock node:dns/promises, per-hop header policy, N × max-seq UTF-8 boundaries, defer to canonical parser, enumerate canonical-equivalent forms. README link in Quality Gates section makes it discoverable.

This is the kind of doc that justifies its own existence on day 30 — and it's named in `.husky/pre-commit` review-output format, so contributors will actually see it.

### 🟡 P2 Finding: c1-followup-cleanup.test.ts has 3 weak tests + 1 explicit tautology

The 6-test file has integrity issues — ironically violating the very REVIEW_CHECKLIST.md it ships with:

1. **Test "handles 1 MB of CJK input without quadratic slowdown"** — `expect(elapsed).toBeLessThan(500)` doesn't catch O(n²) on M2. I reverted the helper to the original O(n²) impl and 1MB CJK took 52ms (because M2 RAM bandwidth makes `Buffer.byteLength` very fast). The byte-safety assertions catch real bugs; the timing assertion is theatre. Either tighten to <50ms (might flake on slow CI) or delete the timing line and rename the test to just "1 MB byte-safe truncation" since that IS what it verifies.

2. **Test "type only allows rate_limited and oversized"** — body is `expect(typeof mod.SessionRouter).toBe('function')`. That's not testing the type trim, that's testing that the module exports. Comment admits "Compile-time check is enforced by tsc passing" — but then the test body should be deleted, not replaced with an import smoke. Same anti-pattern 守卫 just cleaned up in PR#44 (the `handleMessage wiring (smoke)` block).

3. **Test "cross-reference: see inbound.test.ts"** — explicit `expect(true).toBe(true)`. Identical anti-pattern, c1-followup-cleanup.test.ts:91. If the cross-reference is for documentation, put it in a comment, not in a test. Tautology tests give false coverage signal.

The %2F real tests live in inbound.test.ts (correctly verified). The truncateByBytes real correctness tests in this file are fine. Only the 3 placeholder-style tests need cleanup.

**Suggestion (non-blocking, can be follow-up)**: delete describe blocks #2 and #3 in `c1-followup-cleanup.test.ts`, and drop the timing assertion from describe #1's first test. Net: 6 tests → 4 tests, all real.

### 🔵 Nit: PR description out of sync with head SHA

PR description shows "1 commit / 8 files +329/-32" matching the amend at 3de6cca. Earlier message in chat showed an older state ("1 commit (7443fa6), 6 files +149/-32"). Not a code issue, just confirming for皮皮 that the amend amended cleanly — verified `git log` shows single commit 3de6cca with REVIEW_CHECKLIST.md included.

### Cross-PR check ✅
No conflict with PR#44 (Q1 — already merged), PR#45 (C2 — different files). Independent path.

LGTM, ready to land. P2 finding can be follow-up — REVIEW_CHECKLIST.md is too valuable to delay landing over 3 weak test assertions.

lml2468 (commented):
**Verdict: LGTM** ✅ — clean closure of 4 nit items + valuable team artifact in `REVIEW_CHECKLIST.md`.

Per Checklist #0/2/3 (the rules this PR itself codifies):

- **Worktree-isolated build**: `/tmp/cc-dachui-pr46-review` at `d0abb5b` (clean check via `git worktree add`, not in shared `/tmp/cc-channel-octo`)
- **Latest reviews state**: `gh api .../pulls/46/reviews --jq '.[-1]'` confirmed Yujiawei APPROVED at `d0abb5b` head (18:18Z), no newer BLOCKING since
- **Re-review after any author push**: this is my first review on `d0abb5b` (PM message referenced `3de6cca` but remote HEAD is `d0abb5b` — possibly another amend)

Local verification on `d0abb5b`: tsc clean / eslint 0 warnings / **675/675 tests pass**.

### Reverse verification

| Item | Method | Result |
|------|--------|--------|
| `%2F` guard | Replaced `if (storagePath.includes('%2f') || ...)` with `if (false)` → ran `inbound.test.ts` → **5 tests fail** (all `rejects %2F encoded slash` cases) → restored → 61/61 pass | ✅ test really probes the defense |
| `truncateByBytes` delegation | Reverted to original `while` loop O(n²) implementation → ran `c1-followup-cleanup.test.ts` → **6/6 still pass** | ⚠️ see note below |
| `RejectionReason` trim | Type signature change verified at session-router.ts:32 (`'rate_limited' | 'oversized'`), index.ts:386 SUPPRESS set matches | ✅ structural change consistent |

### One non-blocking note on test design

The perf test (`handles 1 MB of CJK input without quadratic slowdown`, 500ms threshold) does NOT actually fail when reverted to the O(n²) `while` loop:

```
1 MB CJK reverted O(n²): 49 ms elapsed (well under 500ms threshold)
```

V8 string `slice(0, -1)` is more optimized than the O(n²) analysis suggests (likely sub-linear due to rope-string internal representation). The byte-safety correctness improvement (truncateUtf8ByBytes handles UTF-8 boundary properly) IS the real win, but the perf test doesn't probe a regression that actually breaks under reverting.

**Suggestion** (non-blocking, follow-up if you want): lower the threshold to ~10ms OR use a synthetic input that forces O(n²) more aggressively (e.g. 10 MB instead of 1 MB) OR drop the perf assertion and just keep the byte-safety assertion. The current perf test is essentially a smoke test that the algorithm doesn't infinite-loop, not a regression guard against the O(n²) pattern. Document the helper delegation as a code-cleanliness win rather than a perf win.

This doesn't block merging — the byte-safety reverse-verify works, the actual algorithm change is correct, just the test framing slightly overpromises.

### REVIEW_CHECKLIST.md review

Specific praise:
- **9 rules + Quick Recall Card structure** is exactly the right format — checklist enforcement at top, deep-dive case studies at bottom, 1-screen quick reference at end
- **9.1 case study with WHATWG normalize coverage matrix** (cross-team meta-insight from PM + 我) is the most valuable piece — distills 3 bypass classes one parser covers vs ones it doesn't. New reviewers don't need to learn this the hard way
- **9.2 cross-parser canonical-form meta-rule** with 4 sub-rules is the generalization that applies beyond URLs (headers / SQL / shell). 齐哥's edge note about "no canonical parser → reverse whitelist" should ideally also live in 9.2 (currently in REVIEW_CHECKLIST seems missing — would suggest adding as 9.2 step 5: "If no spec-canonical parser exists, reverse the approach: enumerate downstream-acceptable forms and whitelist them, rather than blacklist attacker variants"). Non-blocking.
- **9.3 Recurrence count = 4** correctly identifies the "修了一半" pattern frequency. If a 5th instance appears (PR#45 P1-5 SVG XSS is a strong candidate, see my C2 review), update the count

### Tests-as-documentation

- `c1-followup-cleanup.test.ts` — `%2F cross-reference` test explicitly chains "buildMediaUrl reject %2F → defense matches REVIEW_CHECKLIST #8/9 case study", links code to docs
- `inbound.test.ts` updated to move `..%2f..%2finternal` from accept-list to reject-list with explanatory comment — documents the policy shift cleanly

### Approval criteria met

All 10 quick-recall items I personally tested:
1. ✅ Worktree-isolated
2. ✅ CI green on d0abb5b (per remote check)
3. ✅ Latest reviews state checked
4. ✅ Reverse-verify on %2F defense passes
5. ✅ All `tryResolveFile` callers audited — no signature change in this PR
6. ✅ DNS mocks intact (only changes file-inline tests minor)
7. ✅ No new fetch code introduced
8. ✅ Byte-safe truncation delegation preserves boundary handling
9. ✅ URL validation already canonical-parser (this PR adds %2F belt-and-suspenders)
10. ✅ canonical-equivalent forms enumerated for `%2F` lowercase/uppercase

`<!-- 王大锤 review verdict: LGTM -->`
`<!-- head SHA reviewed: d0abb5b -->`
`<!-- aligns with Yujiawei automated review APPROVED (18:18Z on same head) -->`

PR #47, opened 2026-06-05
test(C2 follow-up): apply REVIEW_CHECKLIST.md §11 invariant pinning to SVG XSS test

lml2468:
Follow-up to merged PR#45 + PR#46. Apply 飞飞 §11 strictest enforcement boundary rule to SVG XSS hardening test from PR#45.

Restructure P1-5 hardening test assertions:
- PRIMARY assert: Content-Disposition === attachment (terminal browser rendering invariant) — this is what actually defends regardless of upstream classification
- SECONDARY assert: msgType === File (kept as defense-in-depth regression visibility but NOT the security invariant)

If msgType drifts to Image while Content-Disposition stays attachment, browser still downloads safely. If both drift, PRIMARY assert catches. Pin the strictest boundary, not familiar parser layer (REVIEW_CHECKLIST.md Quick Recall Card #12).

Test still passes — annotation + ordering change, no behavior diff. 692/692 total green.

### DIFF
--- a/src/__tests__/media-upload.test.ts
+++ b/src/__tests__/media-upload.test.ts
@@ -725,14 +725,26 @@ describe('sendRichTextCombined', () => {
         mediaUrl: dataUri,
       });
 
-      // SVG variant MUST route to File (type=8), not Image (type=2).
-      const sendBody = JSON.parse(fetchMock.mock.calls[1][1].body as string);
-      expect(sendBody.payload.type).toBe(8);
-      // And COS upload MUST set Content-Disposition: attachment so the CDN
-      // serves SVG as download instead of inline render.
+      // ─── §11 strictest enforcement boundary invariant (PRIMARY ASSERT) ──
+      // The terminal parser the attacker can reach is the BROWSER rendering
+      // a CDN-served object. The strictest enforcement boundary is what the
+      // CDN sends — specifically Content-Disposition: attachment which
+      // forces download instead of inline render regardless of Content-Type.
+      // This is the ACTUAL security invariant; everything upstream (msgType,
+      // app-layer gate) is correlate. Pin assertion to the invariant itself
+      // per REVIEW_CHECKLIST.md §11 / Quick Recall Card #12.
       const putCall = cosPutObjectMock.mock.calls[0][0] as { ContentDisposition?: string };
       expect(putCall.ContentDisposition).toBeDefined();
       expect(putCall.ContentDisposition).toMatch(/^attachment/i);
+
+      // ─── §11 defense-in-depth correlates (SECONDARY ASSERTS) ─────────
+      // These hold in current implementation but are NOT the security
+      // invariant — if msgType drifts to Image while Content-Disposition
+      // stays attachment, the browser still downloads (safe). If both drift,
+      // the PRIMARY assert above catches it. These correlates are kept for
+      // regression visibility but the primary assert is what defends.
+      const sendBody = JSON.parse(fetchMock.mock.calls[1][1].body as string);
+      expect(sendBody.payload.type).toBe(8); // MessageType.File (correlate)
     }
   });
 });


yujiawei (approved):
# Code Review — PR #47 (cc-channel-octo)

**Scope:** test-only change to `src/__tests__/media-upload.test.ts` (+17 / −5, 1 file). Restructures the assertions in the `P1-5 hardening: SVG with charset/case-variant params still rejected` test so the load-bearing security invariant (`Content-Disposition: attachment`) is the PRIMARY assert and `msgType === File` is demoted to a labelled defense-in-depth correlate.

## 1. Verification

- ✅ **Diff is test-only.** No production code touched; the runtime SVG-handling behavior is unchanged. Risk surface is limited to test semantics.
- ✅ **The PRIMARY assert pins the real security boundary.** The implementation confirms the reasoning: `uploadFileToCOS` forces `Content-Disposition: attachment` for SVG regardless of msgType (`src/media-upload.ts:232-237`), via `buildContentDisposition(params.filename, 'attachment')`. A CDN-served `image/svg+xml` object renders inline (script/`foreignObject` XSS) unless `attachment` forces download — so asserting on the disposition is correct as the terminal-boundary invariant.
- ✅ **The SECONDARY correlate is still meaningful.** `msgType` is derived from `isSafeInlineImage` (`src/media-upload.ts:474`), which strips MIME params + lowercases and rejects `image/svg+xml` (`src/media-upload.ts:164-170`). Keeping `payload.type === 8 (File)` as a regression check preserves coverage of the upstream classifier the prior tests exercised — nothing was dropped, only reordered and annotated.
- ✅ **No assertion weakening.** Both prior assertions are retained; the PRIMARY (`ContentDisposition` defined + `/^attachment/i`) is *added* coverage on top of the existing `payload.type` check. Net coverage strictly increases.
- ✅ **Tests pass.** Ran the affected suite locally with the project toolchain (vitest): `P1-5` tests 3 passed / 0 failed.

## 2. Issues

None blocking (no P0/P1). The change is a pure test-quality improvement.

## 3. Suggestions (non-blocking nits)

- **nit — `calls[0]` indexing fragility.** Both `cosPutObjectMock.mock.calls[0][0]` and `fetchMock.mock.calls[1][1]` index positionally; if the upload path short-circuits and never reaches `putObject`, the test throws a `TypeError` (`Cannot read properties of undefined`) rather than producing a clean assertion message. This still fails the test (acceptable), but an explicit `expect(cosPutObjectMock).toHaveBeenCalled()` before dereferencing would give a clearer failure signal. Minor — not worth blocking.
- **nit — test title vs. behavior.** The test name says SVG is "still rejected," but on the `uploadAndSendMedia` path SVG is not rejected — it is routed to `File` and uploaded with `attachment` disposition. The new comment block accurately documents this, so the title is now slightly stale relative to the (clearer) body. Optional rename for precision.

## 4. Additional notes for human reviewer (security-sensitive)

- The defense correctly depends on the CDN/browser honoring `Content-Disposition: attachment`. This is the right boundary to pin, but it is worth a one-time manual confirmation that the production CDN does not strip or override `Content-Disposition` on objects served from COS — the test asserts what the app *sends*, not what the CDN ultimately *serves*. That is an infra invariant outside this repo's test scope and not something this PR can or should assert.
- `buildContentDisposition` sanitizes the filename (`CD_UNSAFE_RE` rejects quotes/control chars/`;`, with RFC 5987 fallback), so header injection via filename is already guarded — no concern introduced by this PR.

## Verdict

Test-only, coverage-increasing, correctly aligned with the actual security invariant, and verified green. Approving. The two nits above are optional and do not block merge.

lml2468 (commented):
LGTM ✅ — §11 invariant pinning dogfooding standard.

(Submitted as COMMENTED instead of APPROVED because the shared `lml2468` keyring is both the author and reviewer login — GitHub blocks self-approve. Team convention per 王大锤's PR#45 round-2 LGTM at 18:55:28Z: HTML comment at top locks actual review author through the shared keyring. yujiawei automated APPROVED at 19:07:31Z on same head a4577a3 satisfies branch protection.)

## Verification on fresh worktree (head=a4577a3)

**State check before review** (rule #12, author-side mirror):
```
gh pr view 47 --json state,mergedAt,headRefOid
→ {"headRefOid":"a4577a3...","mergedAt":null,"state":"OPEN"} ✅
```

**Reverse-verify both layers independently fail** (rule #12 §11 dogfooding):

Layer 2 revert (PRIMARY `ContentDisposition: attachment` removed at COS layer):
```
× P1-5 hardening: SVG with charset/case-variant params still rejected
  AssertionError: expected undefined to be defined
  at expect(putCall.ContentDisposition).toBeDefined()
```
→ PRIMARY assert catches the regression on its own ✅

Layer 1 revert (SECONDARY `image/svg+xml` check in `isSafeInlineImage`):
```
× P1-5 hardening: SVG with charset/case-variant params still rejected
  - 8
  + 2
  at expect(sendBody.payload.type).toBe(8)
```
→ SECONDARY assert catches the regression on its own ✅

**Both asserts are load-bearing** — neither is duplicate decoration. The PRIMARY/SECONDARY labels are descriptive of the security model (terminal browser parser = strictest boundary), not the test coverage strength. This is the right §11 application.

## §11 / Quick Recall #12 cross-reference quality

The two `// ─── §11 ... ───` comment blocks are exactly the self-documenting boundary annotation §11 calls for. Comment explicitly notes the "msgType drifts but Content-Disposition stays attachment → still safe" case, which is the failure mode the rule surfaces.

## Non-blocking notes

1. RichText path SVG-rejected test (`P1-5: SVG in RichText markdown is rejected`) still asserts at the app-layer boundary (`failedMediaCount > 0`), not the CDN boundary. Per 王大锤's PR#45 review #1 and §11, recommend follow-up to pin PRIMARY = `cosPutObjectMock.not.toHaveBeenCalled()` and SECONDARY = `failedMediaCount > 0`. Not blocking — PR#48 / PR#49 follow-up scope is fine.

2. Consider extracting PRIMARY/SECONDARY comment template into `REVIEW_CHECKLIST.md §11.5` snippet so future tests can copy-paste. Out of scope here.

## Vitest

`npx vitest run src/__tests__/media-upload.test.ts` → 42/42 pass on a4577a3 ✅

CI 4/4 green. LGTM. Ready to merge.

PR #49, opened 2026-06-05
fix(C2 P1-5 hardening): reject legacy `image/svg` MIME alongside `image/svg+xml`

lml2468:
## Summary

Closes the SVG XSS edge case 王大锤 documented in his PR #45 round-2 review at 5bab2f0 (2026-06-05T18:55:28Z = 02:55:28 GMT+8): legacy `image/svg` MIME (without the `+xml` suffix) is rare and wrong, but Firefox in some legacy contexts has been observed to render it as SVG. Since SVG can embed `<script>`/`<foreignObject>` and execute in the CDN origin, this is an attacker-reachable edge — strict improvement to close it.

**Attribution note**: the review user.login is `lml2468` (the shared keyring used by author tooling), but the body HTML comment `<!-- 王大锤 re-review verdict: LGTM -->` locks 王大锤 as the actual review author. dolphinsboy's two 5bab2f0 reviews (18:48:14 / 18:48:40) do not mention this edge case at all. (Initial PR description / commit / comments incorrectly attributed to dolphinsboy and were force-pushed corrected after 郭守卫's QA CR at 03:32 GMT+8 — see §"Audit trail discipline" below.)

## Two-layer fix (mirrors existing `image/svg+xml` defenses)

**Layer 1 — `isSafeInlineImage`** (`src/media-upload.ts`)

```diff
- if (baseType === "image/svg+xml") return false;
+ if (baseType === "image/svg+xml" || baseType === "image/svg") return false;
```

Routes legacy variant to `MessageType.File` instead of `MessageType.Image`, same as the canonical form.

**Layer 2 — `uploadFileToCOS` Content-Disposition** (`src/media-upload.ts`)

```diff
- } else if (!ct.startsWith('image/') || ct === 'image/svg+xml') {
+ } else if (!ct.startsWith('image/') || ct === 'image/svg+xml' || ct === 'image/svg') {
```

Forces `Content-Disposition: attachment` for the legacy variant too — so even if the IM bubble routing layer is bypassed in the future, the CDN-served object still downloads instead of inline-rendering.

## Pinned at strictest enforcement boundary per §11

Per `REVIEW_CHECKLIST.md` §11 ("pin strictest enforcement boundary, not familiar parser layer"):

- **Terminal parser** = the browser fetching from CDN
- **PRIMARY invariant** = `Content-Disposition: attachment` (Layer 2) — what the CDN serves and the browser enforces
- **SECONDARY** = `msgType=File` (Layer 1) — correlates and helps IM bubble UX, doesn't enforce security

The new test asserts both layers, with the PRIMARY invariant asserted first per the PR #47 §11 application pattern.

## §11 reverse-revert verification

```
With fix:    43/43 in media-upload.test.ts pass
Without fix: new "legacy image/svg" test fails (expected 8 to be 2)
```

Defense is not theatre — reverting the change to either layer fails the new assertion. This satisfies the §11 rule: "if reverting the change does not fail the test, the test is not testing the change".

## Diff

2 files, +60/-5 (after attribution fix-up):

```
src/media-upload.ts                | 18 ++++++++++++++++--
src/__tests__/media-upload.test.ts | 42 ++++++++++++++++++++++++++++++++++++++
```

No other production paths touched. No test file deletions.

## Audit trail discipline (PR description, code comments, commit message)

After 郭守卫's QA CR at 03:32 GMT+8 (`gh api .../pulls/45/reviews | grep` showed original surfacer = 王大锤, not dolphinsboy), the three audit-trail locations were corrected via force-push to head `1e2bff4`:

1. **Commit message** — `dolphinsboy` → `王大锤 (lml2468 keyring, body HTML comment-locked author)`
2. **Code comments** (`isSafeInlineImage` + `uploadFileToCOS`) — same change
3. **This PR description** — same change

The root cause was a "partial application failure" of HEARTBEAT defense #6 (`gh api .../pulls/<N>/reviews | jq '.body'` before attribution claims) — I pulled the review body but stopped at `user.login = lml2468` without grep'ing the body for HTML-comment attribution markers used when reviewers share a keyring. 郭守卫 surfaced this as a candidate HEARTBEAT defense #12 ("partial application failure — verify to source-of-truth ground, not first-hit hit-stop").

## Independence trail

Original gap surfaced by **王大锤** in PR #45 round-2 LGTM review:

> "The `image/svg` case (no `+xml`) is rare/wrong but Firefox HAS rendered it as SVG in some legacy contexts. Most browsers treat it as octet-stream so no inline render = no XSS, but it's a non-zero edge. Suggested fix: add `image/svg` to SVG check list as belt-and-suspenders (1 char change in the `===` comparison)."

PR #45 was approved & merged with this as a non-blocking documented gap; this PR closes the gap as a separate follow-up per "one PR one review surface" scope discipline (tonight's PR #46 lesson).

## Verification

```
$ npx tsc --noEmit       # clean
$ npx eslint src/        # 0 warnings
$ npx vitest run         # 693 passed (was 692, +1 new "legacy image/svg" test)
```

## Refs

- 王大锤 PR #45 re-review on 5bab2f0 at 2026-06-05T18:55:28Z (body HTML comment `<!-- 王大锤 re-review verdict: LGTM -->` locks author through the shared lml2468 keyring) — original gap documentation
- `REVIEW_CHECKLIST.md` §11 — invariant pinning + reverse-revert rule
- PR #45 P1-5 — main `image/svg+xml` SVG XSS defense (same threat model)
- PR #38 — cross-parser canonical-equivalent forms lesson (root pattern these defenses derive from)
- 郭守卫 QA CR on this PR (head a81c74e) — surfaced attribution mis-routing

### DIFF
--- a/src/__tests__/media-upload.test.ts
+++ b/src/__tests__/media-upload.test.ts
@@ -735,4 +735,45 @@ describe('sendRichTextCombined', () => {
       expect(putCall.ContentDisposition).toMatch(/^attachment/i);
     }
   });
+
+  it('P1-5 hardening: legacy `image/svg` (no +xml suffix) also rejected (PR#49 belt-and-suspenders)', async () => {
+    // 王大锤 PR#45 re-review (5bab2f0, 02:55:28 GMT+8) documented gap:
+    // Firefox in some legacy contexts renders `image/svg` (without the
+    // `+xml` suffix) as SVG. Body HTML comment
+    // `<!-- 王大锤 re-review verdict: LGTM -->` locks author through
+    // the shared lml2468 keyring. isSafeInlineImage now rejects both
+    // `image/svg+xml` AND `image/svg` — 1-condition change covers an
+    // attacker-reachable edge case.
+    const svgPayload = '<svg xmlns="http://www.w3.org/2000/svg"><script>alert(1)</script></svg>';
+    const dataUri = `data:image/svg;base64,${Buffer.from(svgPayload).toString('base64')}`;
+
+    fetchMock.mockReset();
+    cosPutObjectMock.mockReset();
+    fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+      bucket: 'b', region: 'r', key: 'path/x.svg',
+      credentials: { tmpSecretId: 'i', tmpSecretKey: 'k', sessionToken: 't' },
+      startTime: 1, expiredTime: 4600,
+    }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+    cosPutObjectMock.mockImplementation((_params, cb) => {
+      cb(null, { Location: 'b.cos.r.myqcloud.com/path/x.svg' });
+    });
+    fetchMock.mockResolvedValueOnce(new Response(JSON.stringify({
+      message_id: 'm1', client_msg_no: 'c1', message_seq: 1,
+    }), { status: 200, headers: { 'Content-Type': 'application/json' } }));
+
+    await uploadAndSendMedia({
+      apiUrl: 'https://test.example.com',
+      botToken: 'bf_test',
+      channelId: 'ch1',
+      channelType: ChannelType.Group,
+      mediaUrl: dataUri,
+    });
+
+    // Must route to File (type=8), not Image (type=2) — same defense as +xml variant.
+    const sendBody = JSON.parse(fetchMock.mock.calls[1][1].body as string);
+    expect(sendBody.payload.type).toBe(8);
+    // Must also force Content-Disposition: attachment at COS layer.
+    const putCall = cosPutObjectMock.mock.calls[0][0] as { ContentDisposition?: string };
+    expect(putCall.ContentDisposition).toMatch(/^attachment/i);
+  });
 });
--- a/src/media-upload.ts
+++ b/src/media-upload.ts
@@ -166,7 +166,16 @@ export function isSafeInlineImage(contentType: string): boolean {
   // lowercase, since MIME types and the `image/` prefix are case-insensitive.
   const baseType = contentType.split(";")[0].trim().toLowerCase();
   if (!baseType.startsWith("image/")) return false;
-  if (baseType === "image/svg+xml") return false;
+  // Belt-and-suspenders (PR#49, 王大锤 PR#45 re-review on 5bab2f0
+  // documented gap; body HTML comment "<!-- 王大锤 re-review verdict:
+  // LGTM -->" locks author through the shared lml2468 keyring):
+  // Reject `image/svg+xml` AND legacy `image/svg`. Firefox in some legacy
+  // contexts will render `image/svg` (without the `+xml` suffix) as SVG,
+  // so we reject both. Most browsers treat it as octet-stream so no inline
+  // render = no XSS, but it's a non-zero edge. The cost is 1 condition; the
+  // upside covers an attacker-reachable edge case in browsers that render
+  // the legacy form.
+  if (baseType === "image/svg+xml" || baseType === "image/svg") return false;
   return true;
 }
 
@@ -228,12 +237,17 @@ export async function uploadFileToCOS(params: {
     const ct = params.contentType.split(";")[0].trim().toLowerCase();
     if (ct.startsWith('video/') || ct.startsWith('audio/')) {
       contentDisposition = buildContentDisposition(params.filename, 'inline');
-    } else if (!ct.startsWith('image/') || ct === 'image/svg+xml') {
+    } else if (!ct.startsWith('image/') || ct === 'image/svg+xml' || ct === 'image/svg') {
       // Force `attachment` for SVG too: even when routed to MessageType.File
       // in the IM payload, the COS object served via CDN with
-      // Content-Type: image/svg+xml will render inline in browsers (XSS
-      // vector — embedded <script>/<foreignObject>). Content-Disposition:
-      // attachment forces download instead of inline render.
+      // Content-Type: image/svg+xml (or legacy image/svg, which Firefox in
+      // some legacy contexts still renders as SVG — surfaced by 王大锤
+      // PR#45 re-review on 5bab2f0, body HTML comment-locked author) will
+      // render inline in browsers (XSS vector — embedded
+      // <script>/<foreignObject>). Content-Disposition: attachment forces
+      // download instead of inline render. Mirror coverage on
+      // isSafeInlineImage above so both layers close the same
+      // attacker-reachable edge case.
       contentDisposition = buildContentDisposition(params.filename, 'attachment');
     }
   }


yujiawei (approved):
# Code Review — PR #49 (cc-channel-octo)

## Verdict: APPROVED

A tight, correctly-scoped security hardening that closes a documented edge case (legacy `image/svg` MIME without the `+xml` suffix) by extending the two existing SVG defense layers. No blocking issues.

## What was verified

| Check | Result | Evidence |
|---|---|---|
| Layer 1 — `isSafeInlineImage` rejects legacy `image/svg` → routes to `File` not `Image` | ✅ | `src/media-upload.ts:176` |
| Layer 2 — `uploadFileToCOS` forces `Content-Disposition: attachment` for legacy `image/svg` at the CDN boundary | ✅ | `src/media-upload.ts:238` |
| MIME normalization (param strip + lowercase) applies before both comparisons | ✅ | `isSafeInlineImage` at `:167`; `uploadFileToCOS` at `:235` |
| No other inline-image decision points were missed | ✅ | Swept `src/` — `inbound.ts` handles inbound rendering only; `octo/api.ts:164` is unrelated outbound payload shaping. `isSafeInlineImage` is the single chokepoint, also reused by the RichText path (`:567`). |
| Build / typecheck | ✅ | `npx tsc --noEmit` clean |
| Tests | ✅ | `media-upload.test.ts` 43/43 pass |
| Reverse-revert (test actually exercises the change) | ✅ | Reverting both conditions → new "legacy `image/svg`" test fails (`expected 8 to be 2`) |

## Strengths

- **Defense mirrored at both layers.** The primary enforcement boundary (`Content-Disposition: attachment` served by the CDN, the actual browser-facing control) and the secondary IM-routing control (`msgType=File`) are both updated. Even if the IM bubble routing is bypassed in the future, the CDN object still downloads rather than inline-renders.
- **Test asserts both layers**, with the CDN-served `Content-Disposition` assertion present alongside the routing assertion — so the test pins the real enforcement boundary, not just the UX correlate.
- **Strict improvement.** The change can only move `image/svg` from "maybe inline" to "always attachment/File"; it cannot regress any currently-safe path.

## Security notes for human verification

This PR is correctly classified security-sensitive. Two points worth a human glance, neither blocking:

1. **Threat-model soundness.** The justification (Firefox rendering bare `image/svg` as SVG in some legacy contexts) is plausible and the mitigation is cheap and strictly safer, so adding it is the right call regardless of how common that rendering quirk actually is. The control does not depend on the quirk being widespread — it closes the edge unconditionally.
2. **Completeness of the canonical-equivalent set.** `image/svg+xml` and `image/svg` are now both covered. There is no other standardized SVG MIME spelling to worry about (no `image/svg-xml` etc.), and parameterized/cased variants (`image/SVG+XML; charset=utf-8`) are already normalized by the `split(";")[0].trim().toLowerCase()` step before comparison. I could not construct a bypassing spelling.

## Nits (non-blocking, no change required to merge)

- The two SVG conditions are now duplicated as inline string comparisons in two functions. If a third SVG spelling ever needs adding, a shared `const SVG_MIME_TYPES = new Set([...])` (or an `isSvgMime(baseType)` helper) would keep the two layers from drifting. Not worth changing for a 1-condition addition today; flagging only as future-proofing.

yujiawei (approved):
# Code Review — PR #49 (cc-channel-octo)

**Verdict: APPROVED** — strict, well-tested security improvement. No blocking issues. Two non-blocking observations below.

## Summary of change

Two-line behavioral change in `src/media-upload.ts`, mirroring the existing `image/svg+xml` defenses to also cover the legacy `image/svg` MIME (no `+xml` suffix):

- `isSafeInlineImage()` (L178): legacy `image/svg` now routes to `MessageType.File` instead of `MessageType.Image`.
- `uploadFileToCOS()` Content-Disposition branch (L240): legacy `image/svg` now forces `Content-Disposition: attachment` on the CDN object.

Plus one new test asserting both layers.

## 1. Verification

| Item | Result |
|---|---|
| Both enforcement layers updated consistently | ✅ `media-upload.ts:178` (routing) + `:240` (Content-Disposition) |
| Enforcement coverage complete (no inline-render path bypasses the check) | ✅ `uploadAndSendMedia` (L488), `uploadImageForRichText` (L571), and the COS attachment branch all consult the updated logic; RichText image blocks are only assembled from already-vetted uploads (L703) |
| Normalization correct for case/param variants | ✅ `split(";")[0].trim().toLowerCase()` correctly rejects `IMAGE/SVG`, `image/svg ; charset=...`, etc. `application/svg+xml` / `text/svg` correctly fall through the `image/` prefix gate |
| Test asserts both layers (not test-theater) | ✅ asserts `payload.type === 8` (File) **and** `ContentDisposition =~ /^attachment/i` |
| Reverse-revert (reverting the change fails the test) | ✅ independently confirmed — reverting either L178 or L240 fails the new test (`expected 2 to be 8`) |
| `tsc --noEmit` | ✅ clean |
| `eslint src/` | ✅ 0 warnings |
| `vitest run` (media-upload suite) | ✅ 43/43 pass |

The change does exactly what the PR description claims, and the test genuinely exercises the legacy (`image/svg`, no `+xml`) data-URI path.

## 2. Findings

### P2 (non-blocking) — Denylist uses exact string match; a trailing non-whitespace char is not normalized

`isSafeInlineImage()` and the Content-Disposition branch both compare the base type via exact `===` against `"image/svg+xml"` / `"image/svg"`. The normalization (`split(";")[0].trim()`) strips only parameters and whitespace, not other trailing characters. A crafted type such as `image/svg+xml.` (trailing dot) does **not** equal either string, so it would route as an inline Image and skip the `attachment` disposition.

I want to be precise about severity here, because it would be unfair to hold this PR for it:

- **This is pre-existing, not introduced by this PR.** The exact-match `=== "image/svg+xml"` pattern already shipped; PR #49 only adds the parallel `|| === "image/svg"` clause and faithfully follows the established pattern. This change is a strict net improvement.
- **Real-world exploitability is doubtful.** The bypass requires a browser to render a non-registered MIME (`image/svg+xml.`) *as an SVG document*. Browsers key inline SVG rendering off the canonical type; a malformed type is treated as unknown/`octet-stream`, which does not render inline — so the XSS premise largely collapses.

Recommendation (future hardening, separate PR — **not** a blocker for #49): consider switching from a denylist of exact SVG strings to either (a) an allowlist of known-safe raster types (`image/png|jpeg|gif|webp|bmp`) for inline routing, or (b) a normalization that strips a trailing `.`/non-token characters before comparison. An allowlist is the more robust long-term posture for the "what may render inline" decision.

### P2 (non-blocking) — Production source comments embed code-review process ceremony

The new comments at `media-upload.ts:169-177` and `:243-245` reference internal review ceremony — a reviewer persona name, a "shared keyring / body HTML-comment locks author" authorship-attestation mechanism, and a commit short-hash. This conflates product logic with review-process metadata in production source. The *technical* rationale (legacy `image/svg` renders as SVG in some Firefox contexts → reject in both layers) is valuable and should stay; the process/attribution lines belong in the commit message or PR description rather than the source file. Pure hygiene — no functional impact.

## 3. Recommendation

Approve and merge. The change is minimal, correct, fully covered by a reverse-revert-verified test, and consistent with the existing two-layer SVG defense. The two P2 items are follow-up suggestions, not merge blockers.

## Note for human reviewer (security-sensitive)

The denylist-vs-allowlist observation in §2 is the one item worth a human's eye — not because PR #49 regresses anything (it doesn't), but because the team may want to track "inline-render MIME gating should move to an allowlist" as a standalone hardening item. PR #49 itself is safe to merge as-is.

lml2468 (commented):
## Review verdict: APPROVED ✅

### Reproduction-style verification (§12 dogfooding)

The hardening test in this PR pins **two independent layers**, and each layer's assertion is individually load-bearing — I verified by mental-revert:

| Revert | What breaks |
|---|---|
| Drop `|| baseType === "image/svg"` in `isSafeInlineImage` | `image/svg` data URI passes `isSafeInlineImage` → routes to `MessageType.Image` (type=2) → `expect(sendBody.payload.type).toBe(8)` SECONDARY assert FAILS ✅ |
| Drop `|| ct === 'image/svg'` in `uploadFileToCOS` | `contentDisposition` stays undefined → PUT skips the `attachment` header → `expect(putCall.ContentDisposition).toMatch(/^attachment/i)` PRIMARY assert FAILS ✅ |

This is the canonical §12 pattern: **two layers, two asserts, neither can pick up the slack for the other** if its layer regresses. The hardening test won't silently pass on a partial regression, which is exactly the failure mode that bit PR#45 (defense-in-depth fix passed because a downstream layer happened to also block).

Same belt-and-suspenders verified for the `+xml` variant by the existing test directly above (lines 685-735) — symmetric coverage across both MIME forms.

### Gap closure trail

PR#45 5bab2f0 re-review (02:55:28 GMT+8) documented this gap explicitly (Firefox legacy `image/svg` rendering) but deferred fix as non-blocking. PR#49 closes it with the minimum viable patch: 1 condition added in each of two places (4 lines of production code), 1 regression test (41 lines).

### Nit (non-blocking) — production source comment formatting

The production source files carry inline HTML comments like:

```
// 王大锤 PR#45 re-review on 5bab2f0
// documented gap; body HTML comment "<!-- 王大锤 re-review verdict:
// LGTM -->" locks author through the shared lml2468 keyring
```

The shared-keyring attribution mechanism is real and useful in PR-review context (which is where it belongs), but burning it into production source comments creates a non-trivial "future reader confusion" tax — they'll wonder why `media-upload.ts` is talking about reviewer keyrings. Recommend trimming the source comment to just the technical reason (Firefox legacy `image/svg` rendering, attacker-reachable edge), and keeping the attribution / keyring detail in the PR body and the commit message.

Non-blocking — fold into next touch of this file.

### Scope

60+/-5, 2 files (1 prod 4 lines, 1 test 41 lines). CI 4/4 ✅. yujiawei APPROVED on the same head SHA. No new attack surface; pure narrowing of accepted MIME set.

PR #51, opened 2026-06-06
feat(config): anthropicBaseUrl + allowedTools simplify + per-session cwd isolation (7d TTL)

lml2468:
## Summary

Three independent config improvements that share `src/config.ts`, `src/agent-bridge.ts`, and the docs/example surfaces. Bundled in one PR because the three diffs collide on the same hunks; split into **two commits** (`feat:` code + `docs:` README/example) only because the husky pre-commit hook drops non-ts paths from the lint-staged index.

### Q1 — `ANTHROPIC_BASE_URL` support
- New optional field `Config.sdk.anthropicBaseUrl`.
- New env var `ANTHROPIC_BASE_URL` (Anthropic SDK standard name — no `CC_OCTO_` prefix, so existing gateway configs work as-is).
- `agent-bridge` forwards it through the SDK's `options.env`. That field REPLACES the subprocess env entirely, so `process.env` is spread in first to preserve `PATH` / `HOME` / `ANTHROPIC_API_KEY`.
- README has a new "Self-hosted gateway" section; `config.example.json` carries the placeholder + comment.

### Q2 — `allowedTools` simplified to `"*" | string[]`
- Type widened: `Config.sdk.allowedTools: string[] | '*'`.
- Default flipped to `'*'` (full SDK toolset). The old hard-coded 8-tool list was redundant lockdown — surface area is already bounded by `permissionMode` + per-session `cwdBase` isolation — and it silently blocked SDK-internal tools like `TodoWrite` / `Task`.
- `agent-bridge` omits the `allowedTools` SDK option when the value is `'*'` (so the SDK's own default applies); a `string[]` is forwarded as-is.
- Env `CC_OCTO_SDK_ALLOWED_TOOLS` accepts both shapes (literal `*` or CSV).

### Q3 — Per-session `cwd` isolation with 7-day TTL cleanup
- New field `Config.cwdBase` (canonical). Legacy `cwd` / `CC_OCTO_CWD` still accepted as deprecated aliases with a one-time startup warning; `cwdBase` wins when both are set.
- New module `src/cwd-resolver.ts`:
  - `resolveSessionCwd(cwdBase, ctx)` — namespaced `sha256(...).slice(0,16)` subdir per `dm:<uid>` / `group:<id>` / `group:<id>:thread:<id>`. `mkdirSync` recursive, idempotent.
  - `cleanupExpiredCwds(cwdBase, ttlMs = 7d)` — sweeps 16-hex subdirs whose `mtime` is older than the TTL. Never throws; only deletes paths matching the hash pattern so a misconfigured `cwdBase` (e.g. pointed at `$HOME`) cannot wipe operator files.
- `queryAgent` accepts an optional `SessionCtx`; when supplied, the resolver produces the per-session path before `sdkQuery`; when omitted (legacy callers / direct Config mocks in tests), the base path passes through unchanged.
- `src/index.ts` wiring:
  - one synchronous sweep at startup,
  - a 6-hour `setInterval` sweep — `unref()`'d so it doesn't keep the process alive, `clearInterval`'d in the shutdown callback,
  - `SessionCtx` derived from `channelType` (DM keys on `from_uid`, group keys on `channel_id`) and threaded into `queryAgent`.

## Verification

| Gate | Result |
|---|---|
| `npm run lint` | 0 warnings |
| `npx tsc --noEmit` | clean |
| `npm test` | **725 passed** (was 690; +35 new across Q1/Q2/Q3 + cwd-resolver) |
| `npm run build` | `dist/` emitted cleanly |

### Backward compatibility
- `"cwd": "..."` in config files still works; only triggers the deprecation warning at startup.
- `CC_OCTO_CWD` still works; same warning.
- Old explicit `allowedTools` arrays still work; default just changed.
- `queryAgent`'s `sessionCtx` is **optional**, so existing tests / direct Config mocks compile without changes.
- SDK options shape unchanged when the new fields are unset.

## Test plan

- [x] `npm test` — 725 tests pass, including 35 new ones (Q1: 4, Q2: 5, Q3: 6, cwd-resolver: 20)
- [x] `npm run lint` — zero warnings under `--max-warnings 0`
- [x] `npx tsc --noEmit` — strict mode passes
- [x] `npm run build` — `dist/` regenerates including new `cwd-resolver.{js,d.ts,js.map}`
- [ ] Manual smoke (out of scope for this PR): start the gateway with `ANTHROPIC_BASE_URL=…`, observe SDK subprocess inherits it; send a DM and a group message, confirm two distinct hashed subdirs appear under `cwdBase`; touch a sandbox to >7d mtime and confirm the next cleanup tick removes it.

### DIFF
--- a/src/agent-bridge.ts
+++ b/src/agent-bridge.ts
@@ -12,6 +12,8 @@
 import { query as sdkQuery } from '@anthropic-ai/claude-agent-sdk';
 import type { PermissionMode, SettingSource } from '@anthropic-ai/claude-agent-sdk';
 import type { Config } from './config.js';
+import { resolveSessionCwd } from './cwd-resolver.js';
+import type { SessionCtx } from './cwd-resolver.js';
 
 const VALID_PERMISSION_MODES: Set<string> = new Set([
   'default', 'acceptEdits', 'bypassPermissions', 'plan', 'dontAsk', 'auto',
@@ -258,6 +260,7 @@ export function buildPrompt(historyPrefix: string, groupContext: string, message
  * @param historyPrefix - Formatted conversation history from SessionStore
  * @param groupContext - Group chat context string (may be empty)
  * @param config - Application config (sdk.* fields used)
+ * @param sessionCtx - Per-session routing context for cwd isolation (Q3)
  * @yields string chunks of assistant text output
  */
 export async function* queryAgent(
@@ -265,6 +268,7 @@ export async function* queryAgent(
   historyPrefix: string,
   groupContext: string,
   config: Config,
+  sessionCtx?: SessionCtx,
 ): AsyncIterable<string> {
   const permissionMode = toPermissionMode(config.sdk.permissionMode);
   const settingSources = toSettingSources(config.sdk.settingSources);
@@ -276,12 +280,29 @@ export async function* queryAgent(
     config.sdk.systemPrompt,
   );
 
+  // Q3: per-session cwd under cwdBase — creates the directory on first use.
+  // Fall back to the base directory when sessionCtx is omitted (legacy callers
+  // and unit tests that don't care about isolation), and to the deprecated
+  // `cwd` field for Config instances built before the rename.
+  const cwdBase = config.cwdBase ?? config.cwd;
+  const cwd = sessionCtx ? resolveSessionCwd(cwdBase, sessionCtx) : cwdBase;
+
+  // Q1: forward ANTHROPIC_BASE_URL through the standard process environment
+  // before invoking the SDK. When unset, leave process.env untouched.
+  if (config.sdk.anthropicBaseUrl) {
+    process.env.ANTHROPIC_BASE_URL = config.sdk.anthropicBaseUrl;
+  }
+
   const stream = sdkQuery({
     prompt: userMessage,
     options: {
-      cwd: config.cwd,
+      cwd,
       systemPrompt,
-      allowedTools: config.sdk.allowedTools,
+      // Q2: `"*"` means "no whitelist" — drop the option so the SDK falls back
+      // to its built-in tool set. An explicit string[] is forwarded as-is.
+      ...(config.sdk.allowedTools === '*'
+        ? {}
+        : { allowedTools: config.sdk.allowedTools }),
       permissionMode,
       maxTurns: config.sdk.maxTurns,
       model: config.sdk.model,
--- a/src/agent-bridge.ts
+++ b/src/agent-bridge.ts
@@ -287,17 +287,23 @@ export async function* queryAgent(
   const cwdBase = config.cwdBase ?? config.cwd;
   const cwd = sessionCtx ? resolveSessionCwd(cwdBase, sessionCtx) : cwdBase;
 
-  // Q1: forward ANTHROPIC_BASE_URL through the standard process environment
-  // before invoking the SDK. When unset, leave process.env untouched.
-  if (config.sdk.anthropicBaseUrl) {
-    process.env.ANTHROPIC_BASE_URL = config.sdk.anthropicBaseUrl;
-  }
+  // Q1: forward ANTHROPIC_BASE_URL to the SDK subprocess via the scoped `env`
+  // option instead of mutating the gateway's global process.env. The SDK's
+  // `env` REPLACES the subprocess environment entirely, so spread process.env
+  // first to preserve PATH/HOME/ANTHROPIC_API_KEY. Scoping it here means the
+  // override never leaks across requests and never persists after the field is
+  // cleared (no stale-global problem). When unset, omit `env` so the subprocess
+  // simply inherits process.env.
+  const env = config.sdk.anthropicBaseUrl
+    ? { ...process.env, ANTHROPIC_BASE_URL: config.sdk.anthropicBaseUrl }
+    : undefined;
 
   const stream = sdkQuery({
     prompt: userMessage,
     options: {
       cwd,
       systemPrompt,
+      ...(env ? { env } : {}),
       // Q2: `"*"` means "no whitelist" — drop the option so the SDK falls back
       // to its built-in tool set. An explicit string[] is forwarded as-is.
       ...(config.sdk.allowedTools === '*'
--- a/src/config.ts
+++ b/src/config.ts
@@ -150,13 +150,17 @@ function readConfigFile(configFilePath: string): PartialConfig {
 
 function mergeConfig(base: Config, override: PartialConfig): Config {
   // Q3: accept legacy `cwd` as an alias for `cwdBase`. cwdBase wins if both
-  // are present; otherwise cwd warns and is used.
-  let cwdBase = override.cwdBase ?? base.cwdBase ?? base.cwd;
-  if (override.cwdBase === undefined && override.cwd !== undefined) {
+  // are present; otherwise cwd warns and is used. Blank strings are treated as
+  // "not provided" so a `"cwdBase": ""` typo cannot slip past the nullish
+  // fallback and land sandboxes relative to process.cwd().
+  const overrideCwdBase = nonBlank(override.cwdBase);
+  const overrideCwd = nonBlank(override.cwd);
+  let cwdBase = overrideCwdBase ?? base.cwdBase ?? base.cwd;
+  if (overrideCwdBase === undefined && overrideCwd !== undefined) {
     console.warn(
       '[cc-channel-octo] WARNING: config.cwd is deprecated, use config.cwdBase instead',
     );
-    cwdBase = override.cwd;
+    cwdBase = overrideCwd;
   }
   return {
     botToken: override.botToken ?? base.botToken,
@@ -191,6 +195,13 @@ function parseCsv(value: string): string[] {
     .filter((s) => s.length > 0);
 }
 
+/** Return the trimmed value, or undefined when it is missing/blank. */
+function nonBlank(value: string | undefined): string | undefined {
+  if (value === undefined) return undefined;
+  const trimmed = value.trim();
+  return trimmed.length > 0 ? trimmed : undefined;
+}
+
 function parseIntStrict(value: string, name: string, minValue = 1): number {
   if (!/^\d+$/.test(value)) {
     throw new Error(`Invalid integer for ${name}: ${value}`);
@@ -214,24 +225,28 @@ function applyEnv(cfg: Config): Config {
   if (env.CC_OCTO_BOT_TOKEN) next.botToken = env.CC_OCTO_BOT_TOKEN;
   if (env.CC_OCTO_API_URL) next.apiUrl = env.CC_OCTO_API_URL;
   // Q3: CC_OCTO_CWDBASE wins; CC_OCTO_CWD is accepted with a deprecation warning.
-  if (env.CC_OCTO_CWDBASE) {
-    next.cwdBase = env.CC_OCTO_CWDBASE;
-    next.cwd = env.CC_OCTO_CWDBASE;
-  } else if (env.CC_OCTO_CWD) {
+  // Blank/whitespace-only values are ignored (treated as unset).
+  const envCwdBase = nonBlank(env.CC_OCTO_CWDBASE);
+  const envCwd = nonBlank(env.CC_OCTO_CWD);
+  if (envCwdBase) {
+    next.cwdBase = envCwdBase;
+    next.cwd = envCwdBase;
+  } else if (envCwd) {
     console.warn(
       '[cc-channel-octo] WARNING: CC_OCTO_CWD is deprecated, use CC_OCTO_CWDBASE instead',
     );
-    next.cwdBase = env.CC_OCTO_CWD;
-    next.cwd = env.CC_OCTO_CWD;
+    next.cwdBase = envCwd;
+    next.cwd = envCwd;
   }
   if (env.CC_OCTO_DATA_DIR) next.dataDir = env.CC_OCTO_DATA_DIR;
 
   if (env.CC_OCTO_SDK_MODEL) next.sdk.model = env.CC_OCTO_SDK_MODEL;
   if (env.CC_OCTO_SDK_ALLOWED_TOOLS) {
-    // Q2: accept the literal `*` token (with surrounding whitespace tolerated)
-    // as the wildcard form; otherwise treat as CSV whitelist.
-    const raw = env.CC_OCTO_SDK_ALLOWED_TOOLS.trim();
-    next.sdk.allowedTools = raw === '*' ? '*' : parseCsv(env.CC_OCTO_SDK_ALLOWED_TOOLS);
+    // Q2: a `*` token anywhere in the value means "allow every tool" — collapse
+    // to the wildcard sentinel rather than passing a literal `*` through as a
+    // (bogus) tool name. So `*`, ` * `, and `*,Read` all mean wildcard.
+    const tools = parseCsv(env.CC_OCTO_SDK_ALLOWED_TOOLS);
+    next.sdk.allowedTools = tools.includes('*') ? '*' : tools;
   }
   if (env.CC_OCTO_SDK_PERMISSION_MODE) {
     next.sdk.permissionMode = env.CC_OCTO_SDK_PERMISSION_MODE;
@@ -316,6 +331,17 @@ export function loadConfig(configPath?: string): Config {
   if (!isAllowedApiUrl(final.apiUrl)) {
     throw new Error(
       `Unsafe apiUrl: ${final.apiUrl} — must be https:// or http://localhost/http://127.0.0.1 (SSRF protection)`,
+    );
+  }
+  // Q1: the gateway endpoint receives the Anthropic API key and all prompt /
+  // response content, so it gets the same SSRF policy as apiUrl. Without this,
+  // a stray ANTHROPIC_BASE_URL (e.g. inherited from a shared shell profile or
+  // CI env) could silently redirect every model request — and the credential —
+  // to an attacker-controlled or private-network host.
+  if (final.sdk.anthropicBaseUrl && !isAllowedApiUrl(final.sdk.anthropicBaseUrl)) {
+    throw new Error(
+      `Unsafe sdk.anthropicBaseUrl: ${final.sdk.anthropicBaseUrl} — must be https:// ` +
+      `or http://localhost/http://127.0.0.1 (SSRF protection)`,
     );
   }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,18 +6,52 @@
 import { readFileSync, existsSync, statSync } from 'node:fs';
 import { isAllowedApiUrl } from './url-policy.js';
 
+/**
+ * Q2: Wildcard form of `allowedTools` — `"*"` means "allow every tool the SDK
+ * exposes". Otherwise must be an explicit string array (whitelist mode).
+ */
+export type AllowedTools = string[] | '*';
+
 export interface Config {
   botToken: string;
   apiUrl: string;
+  /**
+   * Q3: Base directory for per-session cwd isolation. Each (DM peer | group |
+   * thread) gets its own hashed subdirectory under this path. Resolved via
+   * `cwd-resolver.resolveSessionCwd()` before passing to the SDK.
+   *
+   * loadConfig() always populates this. Direct Config mocks (e.g. in tests)
+   * may rely on the deprecated `cwd` alias instead — call sites should read
+   * `config.cwdBase ?? config.cwd` to support both.
+   */
+  cwdBase?: string;
+  /**
+   * @deprecated Use `cwdBase`. Retained as a required compat alias for one
+   * release so existing tests and consumers that build Config objects by hand
+   * continue to compile. loadConfig() keeps this in sync with `cwdBase`.
+   */
   cwd: string;
   dataDir: string;
   sdk: {
     model?: string;
-    allowedTools: string[];
+    /**
+     * Q2: `"*"` allows every tool the SDK exposes; otherwise an explicit
+     * whitelist. Default is `"*"` because we already control surface area via
+     * `permissionMode` and `cwdBase` isolation — the old hard-coded 8-tool
+     * list was redundant lockdown and broke operators who needed SDK-internal
+     * tools like `TodoWrite`/`Task`.
+     */
+    allowedTools: AllowedTools;
     permissionMode: string;
     maxTurns?: number;
     systemPrompt?: string;
     settingSources: string[];
+    /**
+     * Q1: Override the upstream Claude API endpoint (e.g. self-hosted gateway).
+     * Forwarded to the SDK subprocess via the standard `ANTHROPIC_BASE_URL`
+     * environment variable. Env priority: `ANTHROPIC_BASE_URL` > this field.
+     */
+    anthropicBaseUrl?: string;
   };
   rateLimit: {
     maxPerMinute: number;
@@ -41,6 +75,9 @@ export interface Config {
 type PartialConfig = {
   botToken?: string;
   apiUrl?: string;
+  /** Q3: canonical field — base dir for per-session cwd isolation. */
+  cwdBase?: string;
+  /** Q3 deprecated alias — maps to cwdBase with a warning. */
   cwd?: string;
   dataDir?: string;
   sdk?: Partial<Config['sdk']>;
@@ -56,10 +93,12 @@ function defaults(): Config {
   return {
     botToken: '',
     apiUrl: '',
+    cwdBase: process.cwd(),
     cwd: process.cwd(),
     dataDir: './data',
     sdk: {
-      allowedTools: ['Read', 'Write', 'Edit', 'Bash', 'Glob', 'Grep', 'WebFetch', 'WebSearch'],
+      // Q2: default to wildcard — operators tighten only when they need to.
+      allowedTools: '*',
       permissionMode: 'bypassPermissions',
       settingSources: ['user'],
     },
@@ -110,10 +149,21 @@ function readConfigFile(configFilePath: string): PartialConfig {
 }
 
 function mergeConfig(base: Config, override: PartialConfig): Config {
+  // Q3: accept legacy `cwd` as an alias for `cwdBase`. cwdBase wins if both
+  // are present; otherwise cwd warns and is used.
+  let cwdBase = override.cwdBase ?? base.cwdBase ?? base.cwd;
+  if (override.cwdBase === undefined && override.cwd !== undefined) {
+    console.warn(
+      '[cc-channel-octo] WARNING: config.cwd is deprecated, use config.cwdBase instead',
+    );
+    cwdBase = override.cwd;
+  }
   return {
     botToken: override.botToken ?? base.botToken,
     apiUrl: override.apiUrl ?? base.apiUrl,
-    cwd: override.cwd ?? base.cwd,
+    cwdBase,
+    // Keep deprecated `cwd` in sync so any legacy reader still sees a value.
+    cwd: cwdBase ?? base.cwd,
     dataDir: override.dataDir ?? base.dataDir,
     sdk: {
       ...base.sdk,
@@ -163,12 +213,25 @@ function applyEnv(cfg: Config): Config {
 
   if (env.CC_OCTO_BOT_TOKEN) next.botToken = env.CC_OCTO_BOT_TOKEN;
   if (env.CC_OCTO_API_URL) next.apiUrl = env.CC_OCTO_API_URL;
-  if (env.CC_OCTO_CWD) next.cwd = env.CC_OCTO_CWD;
+  // Q3: CC_OCTO_CWDBASE wins; CC_OCTO_CWD is accepted with a deprecation warning.
+  if (env.CC_OCTO_CWDBASE) {
+    next.cwdBase = env.CC_OCTO_CWDBASE;
+    next.cwd = env.CC_OCTO_CWDBASE;
+  } else if (env.CC_OCTO_CWD) {
+    console.warn(
+      '[cc-channel-octo] WARNING: CC_OCTO_CWD is deprecated, use CC_OCTO_CWDBASE instead',
+    );
+    next.cwdBase = env.CC_OCTO_CWD;
+    next.cwd = env.CC_OCTO_CWD;
+  }
   if (env.CC_OCTO_DATA_DIR) next.dataDir = env.CC_OCTO_DATA_DIR;
 
   if (env.CC_OCTO_SDK_MODEL) next.sdk.model = env.CC_OCTO_SDK_MODEL;
   if (env.CC_OCTO_SDK_ALLOWED_TOOLS) {
-    next.sdk.allowedTools = parseCsv(env.CC_OCTO_SDK_ALLOWED_TOOLS);
+    // Q2: accept the literal `*` token (with surrounding whitespace tolerated)
+    // as the wildcard form; otherwise treat as CSV whitelist.
+    const raw = env.CC_OCTO_SDK_ALLOWED_TOOLS.trim();
+    next.sdk.allowedTools = raw === '*' ? '*' : parseCsv(env.CC_OCTO_SDK_ALLOWED_TOOLS);
   }
   if (env.CC_OCTO_SDK_PERMISSION_MODE) {
     next.sdk.permissionMode = env.CC_OCTO_SDK_PERMISSION_MODE;
@@ -179,6 +242,11 @@ function applyEnv(cfg: Config): Config {
   if (env.CC_OCTO_SDK_SYSTEM_PROMPT) next.sdk.systemPrompt = env.CC_OCTO_SDK_SYSTEM_PROMPT;
   if (env.CC_OCTO_SDK_SETTING_SOURCES) {
     next.sdk.settingSources = parseCsv(env.CC_OCTO_SDK_SETTING_SOURCES);
+  }
+  // Q1: ANTHROPIC_BASE_URL uses the Anthropic SDK standard variable name
+  // (no CC_OCTO_ prefix) so operators can reuse existing gateway configs.
+  if (env.ANTHROPIC_BASE_URL) {
+    next.sdk.anthropicBaseUrl = env.ANTHROPIC_BASE_URL;
   }
 
   if (env.CC_OCTO_RATE_LIMIT_MAX_PER_MINUTE) {
--- a/src/cwd-resolver.ts
+++ b/src/cwd-resolver.ts
@@ -1,39 +1,48 @@
 /**
  * Q3: Per-session cwd isolation under a shared `cwdBase`.
  *
- * Each (DM peer | group | thread) maps to a deterministic 16-hex sha256 prefix
- * directory inside `cwdBase`. This prevents one user's session from reading or
- * mutating another user's working tree while letting operators allocate a
- * single disk root for the bot.
+ * Each session maps to a deterministic 16-hex sha256 prefix directory inside
+ * `cwdBase`, so one user's working tree cannot be read or mutated from another
+ * user's session while operators still allocate a single disk root for the bot.
  *
- * The hash inputs are namespaced (`dm:`, `group:`, `group:<id>:thread:<id>`)
- * so that the same string used as a uid vs. a group id can never collide.
+ * The partition key is the *exact* `sessionKey` the SessionRouter already
+ * produced for history (`SessionRouter.sessionKey()`), prefixed by the channel
+ * kind. Reusing the router key verbatim — rather than re-deriving spaceId or
+ * channel_id from the raw message — guarantees the cwd partition can never
+ * drift from the history partition:
  *
- * DM keys may additionally be scoped by `spaceId` so the cwd partition matches
- * SessionRouter.sessionKey(): the same uid messaging from two different spaces
- * gets two history sessions, hence must get two sandboxes too (P0-2).
+ *   - DM:    sessionKey = `${spaceId}:${uid}` (or bare `uid`)   → `dm:<key>`
+ *   - Group: sessionKey = `${channel_id}:${uid}`                → `group:<key>`
+ *
+ * Because the group sessionKey embeds `from_uid`, every group member gets their
+ * OWN sandbox (matching how group history is partitioned per-user), not a
+ * shared per-channel workspace. The `kind` prefix keeps a DM key and a group
+ * key that happen to be byte-identical from colliding.
  */
 
 import { createHash } from 'node:crypto';
 import {
   existsSync,
+  lstatSync,
   mkdirSync,
   readdirSync,
   rmSync,
-  statSync,
   utimesSync,
   writeFileSync,
 } from 'node:fs';
 import { join } from 'node:path';
 
-export type SessionCtx =
-  | { kind: 'dm'; userId: string; spaceId?: string }
-  | { kind: 'group'; groupId: string }
-  // NOTE: the `thread` variant is reserved for future wiring. The current Octo
-  // BotMessage shape exposes no thread/topic id, so only `dm` and `group` are
-  // ever emitted by index.ts today. Kept here (with hashing + tests) so the
-  // routing contract is ready the moment threads land upstream.
-  | { kind: 'thread'; groupId: string; threadId: string };
+/**
+ * Per-session routing context for cwd isolation. `kind` is the channel class
+ * and `sessionKey` is the router-produced key the session's history is stored
+ * under — see `SessionRouter.sessionKey()`. Passing the router key directly
+ * (instead of re-deriving uid/spaceId/channel_id here) is what keeps the cwd
+ * and history partitions byte-for-byte consistent.
+ */
+export type SessionCtx = {
+  kind: 'dm' | 'group';
+  sessionKey: string;
+};
 
 /** Length of the hex prefix used for subdirectory names. 16 hex = 64 bits — */
 /** ~2^32 sessions before a 1% collision risk, ample headroom for IM use.    */
@@ -45,12 +54,22 @@ const HASH_HEX_LEN = 16;
 const SESSION_DIR_RE = /^[0-9a-f]{16}$/;
 
 /**
- * Provenance marker written into every session dir we create. Cleanup only
- * deletes a 16-hex dir that ALSO contains this marker, so a cwdBase that is
- * accidentally pointed at some other tool's cache of identically-named dirs
- * (e.g. another hex-keyed store) can never be rmSync'd by us (P0-3).
+ * Provenance is recorded in a sidecar *registry* directory at the root of
+ * `cwdBase`, NOT inside each session dir. Each session we create gets a 0-byte
+ * marker `cwdBase/.cc-octo-sessions/<hexname>`. Cleanup only deletes a 16-hex
+ * dir that has a matching registry entry, so:
+ *
+ *   - A `cwdBase` accidentally pointed at another tool's hex-keyed store can
+ *     never be rmSync'd by us (P0-3).
+ *   - The marker lives OUTSIDE the agent's own working directory, so a
+ *     user-driven agent (which operates via relative paths inside its cwd)
+ *     cannot delete its marker to evade cleanup, nor forge a marker for a
+ *     sibling/operator directory to get it deleted. (Absolute-path access is a
+ *     separate, documented limitation — cwd is a starting dir, not a chroot.)
+ *   - The marker is re-created on every resolve if missing, so a transient
+ *     first-write failure cannot permanently exempt a live dir from the TTL.
  */
-const SESSION_MARKER_FILE = '.cc-octo-session';
+const SESSION_REGISTRY_DIR = '.cc-octo-sessions';
 
 /** 7 days — long enough for a vacation, short enough to bound disk growth. */
 export const DEFAULT_CWD_TTL_MS = 7 * 24 * 60 * 60 * 1000;
@@ -60,34 +79,33 @@ function hashKey(key: string): string {
 }
 
 function sessionKeyToString(ctx: SessionCtx): string {
-  switch (ctx.kind) {
-    case 'dm':
-      // P0-2: mirror SessionRouter.sessionKey() — scope by space when present
-      // so the same uid in different spaces resolves to different sandboxes.
-      // Omitting spaceId keeps the legacy `dm:<uid>` hash for backward compat.
-      return ctx.spaceId
-        ? `dm:${ctx.spaceId}:${ctx.userId}`
-        : `dm:${ctx.userId}`;
-    case 'group':
-      return `group:${ctx.groupId}`;
-    case 'thread':
-      return `group:${ctx.groupId}:thread:${ctx.threadId}`;
-  }
+  // Prefix the router key with the channel kind so a DM key and a group key
+  // that happen to be byte-identical can never resolve to the same sandbox.
+  return `${ctx.kind}:${ctx.sessionKey}`;
 }
 
 /**
  * Resolve and ensure the per-session cwd exists. Idempotent — safe to call
  * on every turn. Returns the absolute path under `cwdBase`.
+ *
+ * Note: the TTL tracks last *bot turn* (this function bumps the dir mtime on
+ * every call), not arbitrary filesystem activity inside the sandbox. A session
+ * with no inbound message for `ttlMs` is reclaimed even if a background process
+ * is still touching files inside it.
  */
 export function resolveSessionCwd(cwdBase: string, ctx: SessionCtx): string {
-  const dir = join(cwdBase, hashKey(sessionKeyToString(ctx)));
+  const name = hashKey(sessionKeyToString(ctx));
+  const dir = join(cwdBase, name);
   mkdirSync(dir, { recursive: true });
 
-  // P0-3: drop a provenance marker so cleanupExpiredCwds can distinguish our
-  // own session dirs from unrelated 16-hex dirs. Written once; best-effort.
-  const marker = join(dir, SESSION_MARKER_FILE);
+  // Record provenance in the sidecar registry (outside the agent's cwd). Best
+  // effort, but re-attempted on every resolve so a transient failure self-heals
+  // on the next turn rather than exempting the dir from cleanup forever.
+  const registryDir = join(cwdBase, SESSION_REGISTRY_DIR);
+  const marker = join(registryDir, name);
   if (!existsSync(marker)) {
     try {
+      mkdirSync(registryDir, { recursive: true });
       writeFileSync(
         marker,
         JSON.stringify({ created: new Date().toISOString(), kind: ctx.kind }),
@@ -124,8 +142,10 @@ export function resolveSessionCwd(cwdBase: string, ctx: SessionCtx): string {
  * Silent no-op when `cwdBase` does not exist (e.g. first-run before any
  * session has been resolved).
  *
- * A dir is eligible for deletion only when BOTH the name matches the 16-hex
- * pattern AND it carries our `.cc-octo-session` provenance marker (P0-3).
+ * A dir is eligible for deletion only when ALL hold: the name matches the
+ * 16-hex pattern, it is a real directory (not a symlink), and it has a matching
+ * entry in the `.cc-octo-sessions` registry (P0-3). The registry entry is
+ * removed alongside the dir.
  */
 export function cleanupExpiredCwds(
   cwdBase: string,
@@ -139,18 +159,24 @@ export function cleanupExpiredCwds(
     return;
   }
 
+  const registryDir = join(cwdBase, SESSION_REGISTRY_DIR);
   const cutoff = Date.now() - ttlMs;
   for (const name of entries) {
     if (!SESSION_DIR_RE.test(name)) continue; // never touch unrelated files
     const full = join(cwdBase, name);
-    // P0-3: only sweep dirs we provably created. A 16-hex dir without our
-    // marker belongs to someone else — leave it untouched no matter its age.
-    if (!existsSync(join(full, SESSION_MARKER_FILE))) continue;
+    const marker = join(registryDir, name);
+    // P0-3: only sweep dirs we provably created (registry entry present). A
+    // 16-hex dir without a registry entry belongs to someone else — leave it
+    // untouched no matter its age.
+    if (!existsSync(marker)) continue;
     try {
-      const st = statSync(full);
+      // lstatSync (not statSync) so a symlinked entry is never followed; a real
+      // session dir is always a plain directory.
+      const st = lstatSync(full);
       if (!st.isDirectory()) continue;
       if (st.mtimeMs >= cutoff) continue;
       rmSync(full, { recursive: true, force: true });
+      rmSync(marker, { force: true }); // drop the registry entry too
     } catch (err) {
       console.error(
         `[cc-channel-octo] cwd cleanup failed for ${full}: ${String(err)}`,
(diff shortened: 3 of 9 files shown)


Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository, but the new cwd cleanup/isolation behavior has blocking correctness and security issues.

🔴 Blocking

- 🔴 Critical: `cleanupExpiredCwds()` can delete active session workspaces. `resolveSessionCwd()` only calls `mkdirSync(dir, { recursive: true })` and does not refresh any last-used timestamp for an existing sandbox (`src/cwd-resolver.ts:53`). Cleanup then treats the sandbox directory’s own `mtimeMs` as the TTL signal and deletes it with `rmSync()` when stale (`src/cwd-resolver.ts:79`, `src/cwd-resolver.ts:87`), including from the startup and 6-hour timer paths (`src/index.ts:49`, `src/index.ts:52`). Directory mtime is not reliably updated by normal activity inside the tree, especially edits to existing files or nested files, and repeated calls to `resolveSessionCwd()` do not refresh it. A workspace created more than 7 days ago can therefore be removed even if the user just sent a message, and potentially while a handler is using it. Refresh a dedicated last-used marker or the directory timestamp on every session use, and ensure cleanup cannot remove an active workspace.

- 🔴 Critical: DM cwd isolation does not match the project’s existing session identity. `SessionRouter.sessionKey()` includes `spaceId` for DMs when available (`src/session-router.ts:141`), but the new cwd routing keys DMs only by `msg.from_uid` (`src/index.ts:336`). If the same `from_uid` appears in different spaces, the conversation history/rate-limit session is separated while the filesystem sandbox is shared, violating the per-session isolation guarantee. Derive the cwd key from the same identity used by `SessionRouter.sessionKey()` or otherwise include the space/channel discriminator.

💬 Non-blocking

- 🟡 Warning: The resolver supports a `thread` `SessionCtx`, and the README promises per-thread sandboxes, but production wiring only creates `dm` or `group` contexts (`src/index.ts:336`). If topic/thread metadata is unavailable, the docs and type should not claim thread isolation yet.

- 🟡 Warning: There are config tests for `ANTHROPIC_BASE_URL` and `allowedTools: "*"`, but I do not see agent-bridge tests asserting that `env` is forwarded or that `allowedTools` is omitted for `"*"`. Those are important option-shape regressions to lock down.

✅ Highlights

- Backward compatibility for `cwd` / `CC_OCTO_CWD` is handled cleanly.
- The hash-only cleanup guard reduces accidental deletion risk outside managed sandbox names.
- I could not run the targeted tests locally because dependencies are not installed in this checkout (`vitest: command not found`).

Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository, but the new TTL cleanup can delete active or unrelated workspaces.

**🔴 Blocking**

🔴 Critical: Active session directories can be deleted after 7 days even if the session is still in use.  
[src/cwd-resolver.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/src/cwd-resolver.ts:53) creates/returns the same directory on every turn, but `mkdirSync(..., { recursive: true })` does not refresh an existing directory’s `mtime`. Cleanup then uses the directory `mtime` at [src/cwd-resolver.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/src/cwd-resolver.ts:84) and deletes it at line 87. A session whose root directory was created more than 7 days ago can therefore be removed on the next sweep even if `queryAgent` just resolved it via [src/agent-bridge.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/src/agent-bridge.ts:287). Refresh a marker timestamp or call `utimesSync` on successful resolve, and add a test where an old session dir is resolved again before cleanup and must survive.

🔴 Critical: Cleanup can recursively delete unrelated operator directories that merely look like session hashes.  
The guard at [src/cwd-resolver.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/src/cwd-resolver.ts:81) only checks `/^[0-9a-f]{16}$/`, then [src/cwd-resolver.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/src/cwd-resolver.ts:87) deletes recursively. If `cwdBase` is accidentally pointed at a broad directory, any old 16-hex directory can be removed, which does not meet the PR’s safety claim. Use a provenance marker inside created sandboxes, a registry, or a dedicated parent created by this app before deleting.

**💬 Non-blocking**

🟡 Warning: README still has stale security and limitation text. [README.md](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/README.md:24), [README.md](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/README.md:173), and [README.md](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-dgw8sk6j/repo/README.md:268) still describe old `cwd` behavior or old message limitations.

🟡 Warning: The focused tests/type-check could not be run in this checkout because dependencies are not installed: `vitest: command not found`, and `npx tsc --noEmit` could not find the local TypeScript compiler.

**✅ Highlights**

The config aliasing approach preserves old `cwd` / `CC_OCTO_CWD` callers, and forwarding `ANTHROPIC_BASE_URL` only when configured avoids unnecessarily replacing the SDK subprocess environment.

yujiawei (changes requested):
# Code Review — PR #51 (cc-channel-octo)

Independent review of the three bundled config changes (Q1 `anthropicBaseUrl`, Q2 `allowedTools` simplification, Q3 per-session `cwd` isolation + 7d TTL). Verified against the `@anthropic-ai/claude-agent-sdk@0.3.x` type definitions and `origin/main` (pre-PR baseline), with the mtime behavior reproduced locally.

**Verdict: CHANGES_REQUESTED.** Q1 and Q2 are sound; Q3's TTL cleanup has a confirmed data-loss path and the DM isolation key diverges from the project's existing session identity.

---

## 🔴 Blocking

### P0 — `cleanupExpiredCwds` can delete an *active* session workspace mid-conversation

`resolveSessionCwd` only does `mkdirSync(dir, { recursive: true })` on re-entry (`src/cwd-resolver.ts:53`) — an idempotent no-op that does **not** refresh the directory's `mtime`. Cleanup keys the TTL decision on that same directory's `st.mtimeMs` (`src/cwd-resolver.ts:84`) and then `rmSync(full, { recursive: true })` (`src/cwd-resolver.ts:87`), driven from both the startup sweep and the 6-hour timer (`src/index.ts`).

The non-obvious part is *why this bites normal usage*, not just an idle read-only edge case. A directory's `mtime` is bumped only when a **direct child entry is added or removed** — not when an existing nested file is modified. I reproduced this:

```
mkdir sessiondir                      # mtime = T0
mkdir sessiondir/repo/src; echo > .../file.ts   # mtime = T1 (direct child 'repo' added)
echo more >> sessiondir/repo/src/file.ts        # mtime STILL T1 (nested write, no bump)
```

A coding agent overwhelmingly writes into nested paths (a checked-out repo, `src/`, etc.). After the first top-level entry is created, the session dir's own `mtime` freezes while all subsequent edits land deeper. So a session that is **actively used every day** but whose top-level layout was created >7 days ago has a stale session-dir `mtime`, becomes eligible for reaping, and the next sweep recursively deletes the whole tree — including freshly-edited nested files. Re-entry on the current turn does not save it, because `mkdirSync` won't refresh the timestamp.

**Fix:** refresh a liveness signal on every resolve — e.g. `utimesSync(dir, now, now)` inside `resolveSessionCwd`, or touch/write a marker file as a *direct child* each turn — and add a test that resolves an old dir immediately before `cleanupExpiredCwds` and asserts it survives.

### P1 — DM cwd key diverges from the existing session identity

`SessionRouter.sessionKey()` keys a DM as `` `${spaceId}:${from_uid}` `` when a spaceId is extractable, otherwise `from_uid` (`src/session-router.ts:141-146`). The new cwd routing keys a DM only on `from_uid` (`src/index.ts:336`).

In the common case this happens to agree, because `extractSpaceId` derives the space *from* `from_uid` (format `s{spaceId}_{peerId}`), so the space is already embedded. But `extractSpaceId` has a fallback that parses the space out of `channel_id` when `from_uid` doesn't carry the `s…_…` prefix (`src/session-router.ts:166-178`). On that path, two different spaces sharing the same bare `from_uid` get **distinct** conversation/rate-limit sessions but a **shared** filesystem sandbox — exactly the cross-tenant bleed this feature exists to prevent.

**Fix:** derive the cwd key from the same identity `SessionRouter.sessionKey()` produces (or feed `extractSpaceId` into the `SessionCtx`) so the sandbox boundary can never be coarser than the session boundary.

---

## 💬 Non-blocking

### P2 — README security guidance for `allowedTools` is now inaccurate

The PR rewrites the "Security Model → `allowedTools` Whitelist" section but keeps the framing "Reduce the list to reduce risk" with `No network` / `No shell` / `Read-only` profiles (`README.md`). Per the SDK type docs, `allowedTools` is the *auto-approve-without-prompting* list — **not** the availability list ("To restrict which tools are available, use the `tools` option instead"). Under the shipped default `permissionMode: 'bypassPermissions'`, prompting is bypassed entirely, so shrinking `allowedTools` does not actually remove Bash/network/write tools from the agent's reach. An operator who follows the README to lock down a deployment gets a false sense of safety.

Note this is a **documentation-accuracy** issue, not a runtime regression: I checked `origin/main` and `permissionMode: 'bypassPermissions'` was already the default, so the prior 8-tool list was already non-gating under bypass. The Q2 flip to `"*"` is therefore behaviorally neutral for the default deployment — but since this PR is actively editing this exact security section, please make the guidance correct (point operators at `tools`/`disallowedTools`, or at a non-bypass `permissionMode`, for an actual restriction).

### P2 — `thread` SessionCtx variant is dead at the wiring layer; per-thread isolation is not delivered

`{ type: 'thread' }` is implemented and unit-tested in `cwd-resolver`, but `src/index.ts:336` only ever constructs `group` or `dm` contexts. `CommunityTopic` (channel_type 5) is folded into `isGroup` and keyed on `channel_id`, and the topic id is read-and-discarded upstream (`src/octo/socket.ts:713 // topic (unused)`). So distinct topics under one channel collapse to a single sandbox. This matches the existing `sessionKey` granularity (so it isn't a new isolation regression), but the README/type comments promising per-thread sandboxes overstate what ships. Either wire topic id through, or soften the docs/type until it's real.

### P2 — `cleanupExpiredCwds` recursive delete relies solely on a name-shape guard

The `/^[0-9a-f]{16}$/` guard (`src/cwd-resolver.ts:81`) is a reasonable mitigation and the test confirms operator-named files/dirs survive, so I'd rate this a hardening note rather than Critical. Still, if `cwdBase` is misconfigured to a broad directory, any unrelated 16-hex-named dir older than the TTL would be recursively removed. A provenance marker written inside sandboxes this app creates (and checked before deletion) would make the safety claim airtight.

### P2 — Missing `agent-bridge` tests for the new option shaping

`config.test.ts` covers parsing of `anthropicBaseUrl` and `allowedTools: "*"`, but nothing asserts the resulting **SDK option shape**: that `options.env` includes `ANTHROPIC_BASE_URL` (and preserves `process.env`) when set, that `env` is omitted when unset, and that `options.allowedTools` is dropped when the value is `"*"` and forwarded as-is otherwise. These are the actual regression surfaces — worth locking down in `agent-bridge-query.test.ts`.

---

## ✅ Verified correct

- **Q1 env handling** matches the SDK contract exactly. The `env` option *replaces* the subprocess environment entirely, so spreading `process.env` first (`{ env: { ...process.env, ANTHROPIC_BASE_URL: … } }`) is both necessary and sufficient to preserve `PATH`/`HOME`/`ANTHROPIC_API_KEY`. The option is correctly omitted when `anthropicBaseUrl` is unset, so the default path is untouched.
- **Backward compatibility** is clean: the deprecated `cwd` / `CC_OCTO_CWD` aliases are honored with a one-time warning, `cwdBase` wins when both are set, and `sessionCtx` is optional so existing callers/tests compile unchanged.
- **Hash namespacing** (`dm:` / `group:` / `group:<id>:thread:<id>`) correctly prevents collisions between a uid and a group id that share the same string.
- CI on the head SHA: type-check, ESLint, Vitest, and coverage all pass.

lml2468:
Addressed Jerry-Xin automated CHANGES_REQUESTED in amended commit 434ae03923cd3b605d5d2b6fdaaa51134d0fbe79.\n\nP0 fixes:\n- Refreshed session cwd atime/mtime on each resolve so active old workspaces survive cleanup.\n- Matched DM cwd isolation to SessionRouter space-aware keys, preserving no-space legacy hashes.\n- Added .cc-octo-session provenance markers and required them before TTL deletion of 16-hex dirs.\n\nP2 nits:\n- Documented thread isolation as reserved until Octo exposes thread/topic ids.\n- Added Agent bridge tests for ANTHROPIC_BASE_URL env forwarding and allowedTools wildcard omission/array forwarding.\n- Updated stale README cwd/cwdBase isolation text.\n\nVerification:\n- npm run lint\n- npx tsc --noEmit\n- npm test: 732 passed\n- npm run build

lml2468:
## PR#51 review fixes — all 3 P0 + 4 P2 nits addressed

All changes squashed into the existing commit (force-pushed). New head: `d408c63`.

### P0 (blocking)

**P0-1 — cleanup could delete active session workspaces**
`mkdirSync(dir, {recursive})` does not refresh mtime on an existing dir, so a >7-day-old session a user kept using was swept on the next pass. `resolveSessionCwd` now calls `fs.utimesSync(dir, now, now)` after mkdir (wrapped in try/catch so an `rmSync` race can't crash the request), so the TTL tracks last activity. Test: an 8-day-old dir that is re-resolved survives cleanup; an untouched one is still swept.

**P0-2 — DM cwd key inconsistent with `SessionRouter.sessionKey()`**
The router scopes DM keys by `spaceId` when present (`${spaceId}:${from_uid}`), but cwd routing keyed on `from_uid` alone — same uid in two spaces shared one sandbox. `SessionCtx` dm variant gains optional `spaceId`; the hash key becomes `dm:${spaceId}:${userId}` when present, falling back to legacy `dm:${userId}` for backward compat. `index.ts` recovers `spaceId` from the router-produced `sessionKey` itself so the cwd partition can never drift from the history partition. Tests: same uid + different space → different dir; no-space hash unchanged.

**P0-3 — cleanup could delete unrelated 16-hex dirs without provenance**
A `cwdBase` mispointed at another tool's 16-hex cache would have its dirs `rmSync`'d. `resolveSessionCwd` now writes a `.cc-octo-session` marker; `cleanupExpiredCwds` requires both the 16-hex name **and** the marker before TTL deletion. Tests: unmarked old dir survives; marked recent dir survives; marked 8-day-old dir is deleted.

### P2 nits

- **nit-1 (thread SessionCtx):** Octo `BotMessage` exposes no thread/topic id today, so the `thread` variant is kept but documented as *reserved for future wiring*; README updated to "DM and group isolation; thread isolation reserved".
- **nit-2 (agent-bridge tests):** new `src/__tests__/agent-bridge-config.test.ts` covers `ANTHROPIC_BASE_URL` forwarding (set when configured, `process.env` untouched when unset) and `allowedTools: "*"` omission vs explicit array forwarding.
- **nit-3 (stale README):** refreshed the `cwd` / "Shared cwd" / roadmap sections to the per-session `cwdBase` isolation reality.

### Verification
- `npm run lint` → 0 warnings
- `npx tsc --noEmit` → clean
- `npm test` → **732 passed** (39 files)
- `npm run build` → dist emitted

`git diff 84f43cd..HEAD --stat` touches only: `src/cwd-resolver.ts`, `src/__tests__/cwd-resolver.test.ts`, `src/index.ts`, `src/agent-bridge.ts`, `src/__tests__/agent-bridge-config.test.ts`, `README.md`.

Jerry-Xin (commented):
The PR is in scope for `cc-channel-octo`, but the new group `cwd` isolation has a blocking concurrency/isolation bug.

**🔴 Blocking**
- 🔴 Critical: Group messages share one workspace but are not serialized on that same key. [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-2nco9t0d/repo/src/index.ts:343) maps every group message to `{ kind: 'group', groupId: channelId }`, so all users in a group use the same resolved `cwd`. But [src/session-router.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-2nco9t0d/repo/src/session-router.ts:141) keys non-DM sessions as `${channel_id}:${from_uid}`, so two group users can run concurrently in the same workspace under different locks. With write/shell-capable tools, this can corrupt files, leak per-user state, and violates the PR’s isolation claim. Fix by aligning the lock/session key with the workspace key: either include `from_uid` in the group `SessionCtx`, or intentionally make the router/processing lock group-scoped when the workspace is group-scoped.

**💬 Non-blocking**
- 🟡 Warning: `sdk.anthropicBaseUrl` is applied by mutating global `process.env` in [src/agent-bridge.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-2nco9t0d/repo/src/agent-bridge.ts:290). That works for a single static process config, but it can contaminate later direct `queryAgent` calls where the field is unset. Prefer passing `options.env: { ...process.env, ANTHROPIC_BASE_URL: config.sdk.anthropicBaseUrl }` only when configured, matching the SDK option contract and avoiding persistent global mutation.
- 🔵 Suggestion: Add a test that proves two different users in the same group either resolve to different `cwd`s or are serialized by the same routing key. The current tests cover hashing, but not the integration invariant between router locking and workspace sharing.

**✅ Highlights**
- The `cwd` cleanup is conservative: 16-hex filtering plus the `.cc-octo-session` marker avoids deleting unrelated operator files.
- `cwdBase` backward compatibility and env precedence are covered with focused config tests.

Jerry-Xin (commented):
Re-review (R2) of PR #51 after author push 434ae03923cd. All R1 blockers from both prior reviews are resolved.

## R1 Blockers — Resolution Verified

✅ **P0-1 (mtime not refreshed → active session deletion):** `resolveSessionCwd` now calls `utimesSync(dir, now, now)` on every entry, so an active session's directory mtime tracks last activity rather than creation time. Cleanup will not delete an actively-used workspace even if it was created >7d ago. Test coverage confirms resolve-then-cleanup ordering is safe.

✅ **P0-2 (DM cwd key diverging from SessionRouter):** `dmSpaceIdFromKey()` extracts `spaceId` from `SessionRouter.sessionKey()` and feeds it into `SessionCtx`, so the cwd partition now exactly mirrors the session partition. Same-uid across different spaces gets distinct sandboxes.

✅ **P0-3 (provenance marker — 齐静春):** `.cc-octo-session` marker file written on creation, required for cleanup eligibility. A 16-hex dir without the marker is never deleted, closing the "misconfigured cwdBase" blast radius.

✅ **P2 nit-2 (agent-bridge tests — 齐静春):** New `agent-bridge-config.test.ts` (4 tests) covers the SDK option shaping: `ANTHROPIC_BASE_URL` set/unset, `allowedTools: "*"` omission, and explicit array forwarding.

✅ **Thread wiring:** Honestly documented as "reserved until Octo messages expose a thread/topic id". The `SessionCtx` union, hashing, and tests are in place; README/code comments no longer overstate what ships.

## 💬 Non-blocking

🟡 **README isolation language is slightly strong.** Line 150 says one session "cannot read or mutate another's working tree" — technically the `cwd` parameter scopes the agent's default workspace but does not prevent path traversal at the OS level (the SDK subprocess runs as the same OS user). Consider softening to "is directed to work within" or noting that `cwd` is a workspace boundary, not a filesystem sandbox. This is a documentation-accuracy nit, not a runtime defect — the feature's design (deterministic hash dirs, mtime-tracked TTL, marker-guarded cleanup) is solid.

🟡 **`process.env.ANTHROPIC_BASE_URL` mutation is write-once-never-restored.** Stable for the current single-config process but leaky if multi-config or hot-reload is ever added. Setting it once at startup (in `main()`) rather than per-`queryAgent` call would be cleaner.

🟡 **Group cwd sharing vs. `SessionRouter` per-user lock.** Group sessions share one workspace (by design — collective project model) while `SessionRouter` keys sessions per-user for history/rate-limit. Concurrent group messages can hit the same cwd simultaneously. If this matters, a per-cwd write lock can be added later.

## ✅ Highlights

- The one-stone-two-birds marker approach (`.cc-octo-session` as both provenance guard for cleanup safety AND liveness signal complement) is elegant.
- `dmSpaceIdFromKey` derives spaceId from the router's own key rather than re-deriving it, so the two can never drift out of sync.
- 24 new cwd-resolver tests with excellent edge-case coverage (cross-namespace collision, marker lifecycle, concurrent-delete resilience).
- Backward compatibility for `cwd` / `CC_OCTO_CWD` is seamless.
- Test count rose from 690 → 725 (+35 new tests across Q1/Q2/Q3).

Jerry-Xin (approved):
R2 re-review: All R1 blockers resolved. Approving.

Jerry-Xin (approved):
## Third-round re-review — all prior blockers resolved

### Prior blockers verified fixed

| ID | Issue | Status |
|---|---|---|
| R1-B1 | `cleanupExpiredCwds` deletes active sessions (mtime never refreshed) | ✅ Fixed — `utimesSync(dir, now, now)` called on every `resolveSessionCwd` invocation |
| R1-B2 | DM cwd keyed on bare `from_uid`, ignoring spaceId | ✅ Fixed — `dmSpaceIdFromKey()` recovers spaceId from the router-produced sessionKey; DM ctx now `dm:${spaceId}:${userId}` |
| QJC-B1 | Same mtime concern | ✅ Same fix as R1-B1 |
| QJC-B2 | 16-hex pattern guard too broad; needs provenance marker | ✅ Fixed — `.cc-octo-session` marker written on dir creation; cleanup requires BOTH hex-pattern match AND marker presence |
| QJC-B3 | Thread wire claimed but missing | ✅ Addressed — `thread` variant explicitly documented as reserved; `index.ts` comments clearly state BotMessage exposes no thread metadata yet |

### New observations (non-blocking)

1. **Group cwd concurrency model** — Group messages map to `{ kind: 'group', groupId: channelId }` (shared workspace) while `SessionRouter.sessionKey()` includes `from_uid` (per-user locking). This means two group members can run SDK processes concurrently in the same cwd. This is a deliberate design choice (group-level shared context), not a defect — the router's per-user locking serves history isolation and rate limiting, which are orthogonal to filesystem scope. Worth a one-line comment in `index.ts` near the `SessionCtx` derivation to make the intent explicit for future readers.

2. **`anthropicBaseUrl` via `process.env` mutation** — `process.env.ANTHROPIC_BASE_URL = ...` is a persistent global mutation. Since the config is loaded once at startup and never changes per-request, this works correctly in practice. Passing it through `options.env` would be cleaner but is not blocking.

### Summary

All five prior P0 blockers from rounds 1–2 are properly fixed with good test coverage (725 tests, +35 new). The marker-file approach (P0-3) is a clean one-stone-two-birds solution for both provenance and mtime concerns. `cwdBase` backward compatibility and env precedence are well-covered. Code is clean and well-commented.

yujiawei (changes requested):
# Code Review — PR #51 (cc-channel-octo)

This PR bundles three config features — `ANTHROPIC_BASE_URL` support (Q1), a widened `allowedTools` (Q2), and per-session `cwd` isolation with TTL cleanup (Q3). The Q3 cwd-isolation work is genuinely solid (hashed/namespaced session dirs, a provenance-marker delete gate, mtime-as-activity TTL, an `unref`'d timer cleaned up on shutdown), backward compatibility is real, and the test footprint is good (+35 tests). My pushback is concentrated on two security-relevant defaults/omissions and one disk-leak edge case; the rest are non-blocking suggestions and nits.

## Blocking issues

**[P1] Flipping the default `allowedTools` from an explicit whitelist to `"*"` is a least-privilege regression that should be opt-in, not the default**
`src/config.ts:101-102` (defaults), `src/agent-bridge.ts:303-310`

```ts
// defaults()
allowedTools: '*',
// agent-bridge queryAgent()
...(config.sdk.allowedTools === '*' ? {} : { allowedTools: config.sdk.allowedTools }),
...
allowDangerouslySkipPermissions: permissionMode === 'bypassPermissions',
```

To be accurate about the delta: the pre-PR default was *not* harmless — `main:src/config.ts:62-63` already shipped `['Read','Write','Edit','Bash','Glob','Grep','WebFetch','WebSearch']` with `permissionMode: 'bypassPermissions'`, so `Bash`-under-bypass (the keys to the kingdom for any DM/@mention/mention-free-group user) **already existed**. This PR therefore does not *introduce* RCE, and I want to be clear it doesn't, since the SDK subprocess inheritance and permission model are unchanged.

What it *does* change is removing the explicit whitelist as a deliberate control. The author's rationale — "surface is already bounded by `permissionMode` + `cwdBase` isolation" — is the part I'd push back on: per-session `cwd` is a working directory, **not** a sandbox/chroot, so a `Bash` turn can `cd /` and read absolute paths (the bot's own config with `botToken`, env, SSH keys) or hit internal/metadata endpoints. cwd isolation does not contain shell access, so it cannot be the justification for dropping the whitelist. Widening a security-relevant default from "explicit list" to "everything the SDK exposes" is exactly the kind of change that should require an operator to opt in, not be the shipped default.

*Why blocking:* this is a deliberate weakening of the default least-privilege posture for an untrusted-input-facing bot. I'm flagging it for an explicit maintainer decision rather than letting it land silently.

*Suggested fix (any one):* keep an explicit whitelist as the default and require operators to set `"*"` deliberately; or, if `"*"` must be the default for DX, emit a prominent boot-time WARNING when `allowedTools==='*'` **and** `permissionMode==='bypassPermissions'`, and surface this in the README risk section as High. Independently, reconsider whether `bypassPermissions` is the right *default* permission mode for an IM-facing bot.

---

**[P1] `anthropicBaseUrl` is not validated like `apiUrl`, and is applied by mutating global `process.env` with no restore**
`src/config.ts:316` (only `apiUrl` is gated), `src/agent-bridge.ts:292-294`

```ts
// config.ts loadConfig() — apiUrl is gated, anthropicBaseUrl is not:
if (!isAllowedApiUrl(final.apiUrl)) {
  throw new Error(`Unsafe apiUrl: ... (SSRF protection)`);
}
// agent-bridge.ts:
if (config.sdk.anthropicBaseUrl) {
  process.env.ANTHROPIC_BASE_URL = config.sdk.anthropicBaseUrl;
}
```

Scoping this honestly: `anthropicBaseUrl` is operator-supplied config (env/file), the **same trust class** as `apiUrl` — it is not attacker-controlled IM input, so this is not a remote SSRF. The issue is a defense-in-depth inconsistency plus a real lifecycle bug:

1. `apiUrl` is run through `isAllowedApiUrl()` (the whole point of `url-policy.ts`), but `anthropicBaseUrl` — the endpoint that receives the **full prompt** (user text, group context, injected file contents) authenticated with the Anthropic key — gets no validation at all. Given `url-policy.ts` exists specifically to gate operator-set endpoints, applying it to one endpoint and not the other is an inconsistent boundary.
2. The value is written to the **global** `process.env` and never cleared. The "leave process.env untouched when unset" comment means "don't overwrite," not "reset" — so once any config sets it, a later reload/config without the field will not turn it back off. In this codebase config is loaded once, so the blast radius is narrow today, but the global mutation is an unnecessary footgun.

*Suggested fix:* validate `anthropicBaseUrl` at load time with the same `isAllowedApiUrl`/`assertPublicUrl` policy used for `apiUrl` (throw on violation). And instead of mutating the global env, pass it scoped to the call via the SDK's `options.env`: `env: { ...process.env, ANTHROPIC_BASE_URL: config.sdk.anthropicBaseUrl }`. That single change also resolves the description/docs mismatch below.

**[P2] Partial `rmSync` failure orphans a session dir permanently**
`src/cwd-resolver.ts:148-158`

```ts
if (!existsSync(join(full, SESSION_MARKER_FILE))) continue;
try {
  const st = statSync(full);
  if (!st.isDirectory()) continue;
  if (st.mtimeMs >= cutoff) continue;
  rmSync(full, { recursive: true, force: true });
} catch (err) { console.error(...); }
```

`rmSync({recursive:true})` deletes depth-first, so the top-level `.cc-octo-session` marker is removed **before** `rmSync` reaches an un-deletable nested entry and throws (reproduced experimentally: read-only nested subdir → marker gone, dir survives, `EACCES` thrown — `force:true` does not suppress it). On every later sweep the guard at line 148 then sees no marker, treats the dir as "not ours", and skips it forever — permanently defeating the TTL cleanup this feature exists to provide. A foreign-uid file, a read-only nested artifact, or a busy/locked file is enough to trigger it. The "never throws" contract still holds; the dir just leaks.

*Suggested fix:* stat/age-check first, then `rmSync`; on failure, recreate the marker in the `catch` so the next sweep can retry. (Or delete the marker last: remove children first, unlink the marker only once the body is gone.)

---

**[P2] Default `cwdBase = process.cwd()` sandboxes sessions directly under the working/deploy dir**
`src/config.ts:96-97`, `src/index.ts:38,50`

With no `cwdBase` set, sessions become 16-hex subdirs of the process CWD (often the repo/deploy dir) and the 6h sweep runs there. The marker guard prevents deletion of unrelated files (verified), so this is not data loss — but it scatters per-session sandboxes next to source/config, the exact "don't put secrets in `cwdBase`" hazard the README warns about, now reachable by default. Consider defaulting to `join(dataDir, 'sandboxes')` or an `os.tmpdir()`-based path, or warning when `cwdBase` resolves to `process.cwd()`.

---

**[P2] PR description says `options.env` forwarding, but the code mutates global `process.env`; SDK env inheritance is unverified**
`src/agent-bridge.ts:290-312`

The PR description states the value is forwarded via the SDK's `options.env` spreading `process.env`. The code does neither — it writes the global, and the SDK options object has no `env` key (the test even asserts `expect(options.env).toBeUndefined()`). The README (line 107) matches the code, so only the PR description is wrong, but the feature's correctness rests on the `claude-agent-sdk` subprocess inheriting the parent's `process.env` — which I could not confirm from this checkout (SDK not vendored). Switching to `options.env` (see P1 #2) makes this explicit and verifiable.

---

**[P2] `dmSpaceIdFromKey` — the cross-user DM isolation boundary — is untested**
`src/index.ts:447-453`

This inverts `SessionRouter.sessionKey()`'s `${spaceId}:${from_uid}` to recover `spaceId`, and the DM sandbox partition depends on it. `cwd-resolver.test.ts` covers the *hashing* and `session-router.test.ts` covers the bare-uid case, but nothing exercises the space-scoped DM path or round-trips it against real `sessionKey()` output. The manual round-trip checked out in review (bare / `s1:u1` / colon-bearing uid), so this is a coverage gap, not a confirmed bug — but the isolation boundary deserves a pinned test. Export the helper (or add an index test) covering bare→`undefined`, `s1:u1`→`s1`, and a round-trip against `SessionRouter.sessionKey()`.

---

**[P2] cwd cleanup lifecycle (startup sweep + 6h interval + `unref` + `clearInterval` on shutdown) has no test**
`src/index.ts:49-57,106`

`cleanupExpiredCwds` itself is well-tested; the wiring is not. A regression making the interval `ref`'d (instead of `unref`'d) would hang single-shot runs — exactly what the comment guards against — and nothing would catch it. A `vi.useFakeTimers()` test (one startup call → advance 6h → second call → shutdown → no further calls; assert `unref` called) would lock the contract.

---

**[P2] Inverted deprecation: `cwd` is `required`, canonical `cwdBase` is `optional`**
`src/config.ts:27,33` (interface), `:154-166` (merge)

Making the deprecated field mandatory and the canonical one optional is the inverse of normal deprecation. It's the deliberate compat mechanism (old hand-built `Config` objects still compile), and `loadConfig` keeps them in sync — but the "always read `cwdBase ?? cwd`" contract lives only in comments. A new consumer reading `config.cwdBase` directly would get `undefined` for hand-built objects and silently mis-resolve the sandbox root. Consider routing all reads through one `resolveCwdBase(config)` accessor, or making `cwd` optional too.

## Nits

- `src/index.ts:343-345` — group cwd is shared across all members (`group:${channelId}`) while group *history* is per-user (`${channel_id}:${from_uid}`). This is intentional ("a group's project is collective") and not a leak, but the surrounding comments frame cwd↔history matching as the invariant. Add a one-line callout at the group branch so a later reader doesn't "re-couple" them.
- `src/config.ts:233-234` — `CC_OCTO_SDK_ALLOWED_TOOLS='*,Read'` falls into the CSV branch → `['*','Read']`, passing a literal `*` as a tool name. Document that `*` must be the sole value, or normalize a `*` segment to wildcard. (The trim-vs-untrimmed asymmetry is harmless — `parseCsv` trims each segment.)
- `src/config.ts:154` — `override.cwdBase ?? base.cwdBase ?? base.cwd`: `base` is always `defaults()` output where `cwdBase` is set, so the trailing `?? base.cwd` is dead on the `loadConfig` path.
- `src/__tests__/config.test.ts:320-327` — the test named `'rejects empty string'` actually asserts *no override* (empty env var is falsy and ignored), not rejection; rename for clarity.
- `src/__tests__/config.test.ts:194-215` — the `CC_OCTO_CWDBASE`/`CC_OCTO_CWD` tests assert only `cfg.cwdBase`, never `cfg.cwd`, leaving the deliberate sync invariant half-pinned; add `expect(cfg.cwd).toBe(...)`.
- README (107-109) and `config.example.json`'s `_comment_anthropic_base_url` describe forwarding to "the SDK subprocess" without noting the value is set on this process's global env and never restored; align the wording with the actual mechanism.

## What's good

- **Path traversal is genuinely closed** — all untrusted inputs (`from_uid`, `channel_id`, `spaceId`) are sha256-hashed to `[0-9a-f]{16}` before `join`, so no traversal survives.
- **The cleanup gate is well-designed** — deletion is scoped by *both* the 16-hex name regex *and* the `.cc-octo-session` provenance marker, so unrelated files are never touched; a planted-marker symlink-to-dir does not let `rmSync` recurse into the target.
- **mtime-as-activity is sound** — `utimesSync` refresh on every resolve keeps active sessions alive and closes the cleanup-deletes-active-dir race.
- **Timer lifecycle is correct** — the 6h sweep is `unref`'d and `clearInterval`'d on shutdown.
- **Backward compat is real** — old hand-built `Config` objects compile, `loadConfig` keeps deprecated `cwd` in sync, and env > file > defaults precedence holds for the new fields.
- **Solid test footprint** (+35) across config/agent-bridge/cwd-resolver — mtime refresh, marker guard, hash uniqueness, space-scoped DM hashing, and config/env precedence are all exercised.

## Verdict

**CHANGES_REQUESTED.** The Q3 cwd-isolation feature is well-built and I'd happily approve it on its own. The two blocking items are both security-default/omission choices that should be settled before merge: (1) the default flip to `allowedTools="*"` is a deliberate least-privilege regression that should be opt-in (note: it does **not** introduce RCE — `Bash`+`bypassPermissions` already shipped as the pre-PR default — but it removes an intentional control), and (2) `anthropicBaseUrl` should get the same `isAllowedApiUrl` validation as `apiUrl` and be passed via scoped `options.env` rather than a never-cleared global mutation. Both fixes are small. The P2/nit items are non-blocking.

src/agent-bridge.ts
```diff
@@ -276,12 +280,29 @@ export async function* queryAgent(
...
+  // Q3: per-session cwd under cwdBase — creates the directory on first use.
+  // Fall back to the base directory when sessionCtx is omitted (legacy callers
+  // and unit tests that don't care about isolation), and to the deprecated
+  // `cwd` field for Config instances built before the rename.
+  const cwdBase = config.cwdBase ?? config.cwd;
+  const cwd = sessionCtx ? resolveSessionCwd(cwdBase, sessionCtx) : cwdBase;
+
+  // Q1: forward ANTHROPIC_BASE_URL through the standard process environment
+  // before invoking the SDK. When unset, leave process.env untouched.
+  if (config.sdk.anthropicBaseUrl) {
```
lml2468: **[High] `anthropicBaseUrl` is forwarded with no URL validation — API-key exfiltration risk.**

`apiUrl` is gated by `isAllowedApiUrl` (config.ts:316), but `anthropicBaseUrl` is accepted verbatim from config.json **or** the bare `ANTHROPIC_BASE_URL` env var (no `CC_OCTO_` prefix, so ambient/inherited env merges in). A stray or attacker-set `ANTHROPIC_BASE_URL=http://evil.host` silently routes every model request — the `Authorization` header with the Anthropic API key plus all prompt/response content — to that endpoint. No scheme check, no private-IP check.

Also: `process.env.ANTHROPIC_BASE_URL` is mutated process-globally and never cleared. There is no unset path, so a stale value persists for the process lifetime and leaks into any other in-process reader. The SDK supports a scoped `options.env` field (the test even asserts `options.env` is undefined) — passing it subprocess-local would avoid both the global mutation and the no-revert problem.

Suggest running `anthropicBaseUrl` through the same `isAllowedApiUrl`/`assertPublicUrl` policy as `apiUrl`.

src/index.ts
```diff
@@ -312,7 +328,22 @@ async function handleMessage(
...
+      // model — a group's "project" is collective). Thread routing is reserved
+      // until BotMessage exposes thread/topic metadata.
+      //
+      // P0-2: SessionRouter.sessionKey() scopes DMs by spaceId when present
+      // (`${spaceId}:${from_uid}`), so the cwd partition must match — the same
+      // uid in two spaces has two histories and therefore needs two sandboxes.
+      // We recover spaceId from the router-produced sessionKey itself (rather
+      // than re-deriving it) so the two can never drift out of sync.
+      const sessionCtx: SessionCtx = isGroup
+        ? { kind: 'group', groupId: channelId }
```
lml2468: **[High] Group cwd sandbox is shared across all members, and collapses on missing channel_id — contradicts the documented isolation guarantee.**

Group history is partitioned per-user (`SessionRouter.sessionKey()` → `${channel_id}:${from_uid}`, session-router.ts:146), but the group `SessionCtx` keys only on `channelId`. So every member of a group shares one filesystem sandbox while believing they have separate sessions.

Scenario: in group G, member A's Claude writes `secret.txt`; member B (distinct history sessionKey `G:B`) says "read secret.txt" → `resolveSessionCwd` hashes `group:G` to the **same** dir for both → B reads/overwrites A's files. README.md:111 ("one user's Claude Code session cannot read or mutate another's working tree") is false for groups. The DM path was carefully matched to the router key via `dmSpaceIdFromKey`; the group path silently diverges.

Separately: when `channel_id` is missing, `channelId` falls back to `''` (index.ts:123), so the ctx becomes `{groupId: ''}` → `hashKey('group:')` → **every** channel_id-less group across all spaces collapses into one shared sandbox.

src/config.ts
```diff
@@ -163,12 +213,25 @@ function applyEnv(cfg: Config): Config {
...
+  }
   if (env.CC_OCTO_DATA_DIR) next.dataDir = env.CC_OCTO_DATA_DIR;
 
   if (env.CC_OCTO_SDK_MODEL) next.sdk.model = env.CC_OCTO_SDK_MODEL;
   if (env.CC_OCTO_SDK_ALLOWED_TOOLS) {
-    next.sdk.allowedTools = parseCsv(env.CC_OCTO_SDK_ALLOWED_TOOLS);
+    // Q2: accept the literal `*` token (with surrounding whitespace tolerated)
+    // as the wildcard form; otherwise treat as CSV whitelist.
+    const raw = env.CC_OCTO_SDK_ALLOWED_TOOLS.trim();
+    next.sdk.allowedTools = raw === '*' ? '*' : parseCsv(env.CC_OCTO_SDK_ALLOWED_TOOLS);
```
lml2468: **[Med-High] CSV containing `*` silently mis-restricts tools.**

The wildcard match only fires when the entire trimmed value equals `*`. `CC_OCTO_SDK_ALLOWED_TOOLS='*,Read'` → `trim()` → `'*,Read' !== '*'` → `parseCsv` → `['*','Read']`. agent-bridge.ts:303 then sees an array (not `=== '*'`) and forwards `allowedTools: ['*','Read']` to the SDK. The operator who expected "allow all" instead gets tools restricted to `Read` plus a bogus `'*'` entry — silent, surprising lockdown. Consider treating a `*` element anywhere in the CSV as the wildcard (or rejecting the mix).

src/config.ts
```diff
@@ -56,10 +93,12 @@ function defaults(): Config {
   return {
     botToken: '',
     apiUrl: '',
+    cwdBase: process.cwd(),
     cwd: process.cwd(),
     dataDir: './data',
     sdk: {
-      allowedTools: ['Read', 'Write', 'Edit', 'Bash', 'Glob', 'Grep', 'WebFetch', 'WebSearch'],
+      // Q2: default to wildcard — operators tighten only when they need to.
+      allowedTools: '*',
```
lml2468, line 101: **[Med] Default `allowedTools` flip to `'*'` widens the untrusted-user tool surface; docs now overclaim.**

`bypassPermissions` + Bash were already the default, so this isn't new RCE — but the explicit 8-tool list bounded the surface to named tools. `'*'` grants whatever the installed SDK version ships (Task/sub-agents, NotebookEdit, etc.), and the docs simultaneously demote `allowedTools` from "the primary security control" to optional. Meanwhile README.md:140 still lists the `allowedTools` whitelist + `cwdBase` isolation as "the real security controls" — which the default config no longer provides. Worth reconciling the docs with the actual default in this same PR.

src/cwd-resolver.ts
```diff
@@ -0,0 +1,160 @@
...
+  for (const name of entries) {
+    if (!SESSION_DIR_RE.test(name)) continue; // never touch unrelated files
+    const full = join(cwdBase, name);
+    // P0-3: only sweep dirs we provably created. A 16-hex dir without our
+    // marker belongs to someone else — leave it untouched no matter its age.
+    if (!existsSync(join(full, SESSION_MARKER_FILE))) continue;
+    try {
+      const st = statSync(full);
+      if (!st.isDirectory()) continue;
+      if (st.mtimeMs >= cutoff) continue;
```
lml2468, line 177: **[Med] TTL reaps sandboxes with recent file activity but no recent bot turn.**

Deletion keys on the session **dir's** mtime, which `resolveSessionCwd` refreshes only once per bot turn (line 109). Editing file *contents* inside the dir does not bump the parent dir mtime. So a sandbox whose files are modified by a long-running/agent process over days, but with no new bot turn for 7d, has a stale dir mtime → `rmSync` wipes the whole tree including live work. TTL tracks last bot turn, not last filesystem activity — operators who assume "recently modified files are safe" lose data.

src/cwd-resolver.ts
```diff
@@ -0,0 +1,160 @@
...
+ * on every turn. Returns the absolute path under `cwdBase`.
+ */
+export function resolveSessionCwd(cwdBase: string, ctx: SessionCtx): string {
+  const dir = join(cwdBase, hashKey(sessionKeyToString(ctx)));
+  mkdirSync(dir, { recursive: true });
+
+  // P0-3: drop a provenance marker so cleanupExpiredCwds can distinguish our
+  // own session dirs from unrelated 16-hex dirs. Written once; best-effort.
+  const marker = join(dir, SESSION_MARKER_FILE);
+  if (!existsSync(marker)) {
```
lml2468, line 106: **[Med] Best-effort marker write → permanent cleanup exemption (disk leak); marker is also forgeable.**

The `.cc-octo-session` marker is written best-effort (errors caught/logged) and only when missing. If the first `writeFileSync` fails (transient ENOSPC/EACCES), the dir is still used but the marker never exists, so `cleanupExpiredCwds` skips it forever (line 148) — it accumulates work and is never reaped.

Deeper issue: the marker is the **sole** deletion-eligibility guard, yet it lives inside the agent's own working dir where the user-driven agent has full Write/Bash access. A user can `rm .cc-octo-session` (session never expires → disk-exhaustion vector), or have the agent create a sibling 16-hex dir under `cwdBase` with a forged marker to get an operator file tree `rmSync`'d at the next sweep. A marker in attacker-writable space can't be a trust boundary.

src/config.ts
```diff
@@ -56,10 +93,12 @@ function defaults(): Config {
   return {
     botToken: '',
     apiUrl: '',
+    cwdBase: process.cwd(),
```
lml2468, line 96: **[Low-Med] Empty-string `cwdBase` slips past the nullish fallback.**

Call sites read `config.cwdBase ?? config.cwd`, but `''` is not nullish. `cwdBase: ""` (typo, or `CC_OCTO_CWDBASE=`) passes through → `resolveSessionCwd('', ctx)` → `join('', hash)` → a path relative to `process.cwd()`, i.e. sandboxes land inside the gateway's own working dir — the exact "do NOT point cwdBase at a dir containing config.json/secrets" anti-pattern the README warns against. A non-empty validation at config load would catch it.

src/index.ts
```diff
@@ -403,6 +434,24 @@ async function handleMessage(
 
 // ─── G1/G11 helpers ───────────────────────────────────────────────────────────────────
 
+/**
```
lml2468: **[Low] `dmSpaceIdFromKey` re-implements the inverse of the router's key grammar.**

This inverts `SessionRouter.sessionKey()`/`extractSpaceId` by string-suffix parsing, so the DM key grammar `${spaceId}:${from_uid}` is now encoded independently in two places. The comment claims they "can never drift", but they're two separate string contracts — if the router changes its key format (prefix, or a spaceId/uid containing `:`), the inverter mis-parses and routes to the wrong sandbox. Reusing the router's own spaceId (or threading a `SessionCtx` out of the router) would make drift structurally impossible.

lml2468 (commented):
Automated review (xhigh recall). Top findings posted inline, highest-severity first. Headline: the new `anthropicBaseUrl` path skips the URL validation every other endpoint gets, and the per-session cwd "isolation" guarantee in the README doesn't hold for groups or empty/missing channel ids.

Jerry-Xin (approved):
# Re-review — PR #51 (cc-channel-octo) — head 37f6578

Fourth review pass. Previous history: R1 (84f43cd) REQUEST_CHANGES → R2 (d408c63) APPROVE → yujiawei R4 (d408c63) REQUEST_CHANGES on process.env mutation + SSRF gap + allowedTools default. This review verifies 37f6578 against all outstanding items.

## yujiawei P0/P1 fixes — verified

### ✅ P1 — `process.env` mutation (FIXED)

`agent-bridge.ts:293-299` — `ANTHROPIC_BASE_URL` is now forwarded via scoped `options.env` spread over `process.env`, not by mutating the global environment. The test at `agent-bridge-config.test.ts:47` explicitly asserts `process.env.ANTHROPIC_BASE_URL` remains `undefined` after a call with `anthropicBaseUrl` configured. No cross-request leak, no stale-global problem. Clean fix.

### ✅ P1 — SSRF validation for `anthropicBaseUrl` (FIXED)

`config.ts:336-343` — `anthropicBaseUrl` is now validated by the same `isAllowedApiUrl()` gate as `apiUrl`. Tests added in `config.test.ts` for private IP rejection (`169.254.169.254`), non-localhost HTTP rejection, and `http://localhost` allowance. README updated to document the SSRF policy. Consistent boundary.

### ✅ P2 — Sidecar registry (REDESIGNED)

Provenance markers moved from inside-session-dir (`.cc-octo-session`) to an external registry directory (`cwdBase/.cc-octo-sessions/<hexname>`). This addresses 齐静春's original forgeability concern and yujiawei's orphan-on-partial-delete note. Cleanup now deletes the registry entry alongside the dir (`cwd-resolver.ts:178`). Marker re-created on every `resolveSessionCwd` if missing (self-healing). `lstatSync` replaces `statSync` to prevent symlink traversal. Test coverage is thorough (20 tests).

### ✅ P1 — DM cwd key divergence (FIXED)

Complete refactor: `SessionCtx` simplified to `{ kind, sessionKey }` with the router-produced key passed directly. The fragile `dmSpaceIdFromKey` inverse parser is deleted. cwd partition is now byte-for-byte identical to history partition by construction, not by re-derivation.

## allowedTools default `"*"` — acknowledged as design decision

yujiawei flagged the flip from an explicit 8-tool list to `"*"` as a least-privilege regression. The author has kept the default, and yujiawei's own review explicitly noted this does **not** introduce RCE (`Bash` + `bypassPermissions` already shipped as the pre-PR default). The README now documents the risk honestly and points operators at `tools`/`disallowedTools` or a non-bypass `permissionMode` for actual restriction. This is a maintainer-level policy call, not a code defect. No blocking.

## Previous R1/齐静春 fixes — verified intact

- **mtime refresh** (`utimesSync` in `resolveSessionCwd`) — still present, test coverage confirms active sessions survive cleanup.
- **Space-scoped DM keys** — solved at a higher level via direct sessionKey passthrough.
- **Thread dead code** — removed cleanly; `SessionCtx` no longer carries a `thread` variant.

## README accuracy — improved

The security model section now explicitly states "cwd is a starting directory, not a chroot" and explains the absolute-path limitation under `bypassPermissions`. This honest framing addresses the overclaiming concern from prior reviews.

## 💬 Non-blocking notes

1. **Inverted deprecation** (pre-existing, yujiawei P2): `cwd` is required while `cwdBase` is optional in the Config interface. The `loadConfig` path keeps them in sync, but a direct consumer reading `config.cwdBase` on a hand-built Config object would get `undefined`. A `resolveCwdBase(config)` accessor or making `cwd` optional would be cleaner. Low priority.

2. **Cleanup wiring tests** (pre-existing, yujiawei P2): the 6h interval/unref/clearInterval lifecycle is exercised only by the integration path (startup → shutdown), not pinned in a unit test. A `vi.useFakeTimers()` test would lock the contract. Low priority.

## ✅ Verdict: APPROVE

All blocking items from prior reviews are resolved. The code is sound, test coverage is strong (+35 tests), backward compatibility is preserved, and the security documentation is now honest about the cwd limitation. CI green (type-check, ESLint, Vitest all pass). Clean to merge.

yujiawei (approved):
# Code Review — PR #51 (cc-channel-octo)

Independent review of the three bundled config features at head `37f6578`: Q1 `anthropicBaseUrl` support, Q2 `allowedTools` simplification to `"*" | string[]`, and Q3 per-session `cwd` isolation with a 7-day TTL sweep. Verified against `origin/main`, the live diff, and a local build.

**Verification gates (all green at head):**

| Gate | Result |
|---|---|
| `npx tsc --noEmit` | clean |
| `npm run lint` (`--max-warnings 0`) | 0 warnings |
| `npm test` | 740 passed (39 files) |

## 1. Verdict on prior round findings

The head commit `37f6578` ("address PR #51 review") was independently re-verified against the actual code — not the commit message — for each previously-raised blocker. All are resolved:

| # | Prior finding | Status | Evidence |
|---|---|---|---|
| 1 | `anthropicBaseUrl` forwarded with no SSRF validation (API-key exfiltration) | ✅ Fixed | `src/config.ts:339-346` now runs `isAllowedApiUrl(final.sdk.anthropicBaseUrl)` at boot, throwing on private/non-localhost hosts — same policy as `apiUrl`. Forwarded via scoped `options.env` (`src/agent-bridge.ts:293-296`), not global `process.env`. |
| 2 | Group cwd shared across all members; collapses on missing `channel_id` | ✅ Fixed | `SessionCtx` now carries the router-produced `sessionKey` verbatim (`src/index.ts:338-340`); group keys are `${channel_id}:${from_uid}` (`src/session-router.ts:146`), so each member hashes to a distinct dir. Test: `cwd-resolver.test.ts` "different group sessionKeys map to different dirs". |
| 3 | `*,Read` CSV silently mis-restricts tools | ✅ Fixed | `src/config.ts:248-249` parses CSV first, then `tools.includes('*') ? '*' : tools`. Verified: `*`, ` * `, `*,Read` all collapse to wildcard. |
| 4 | Wildcard default widens surface / docs overclaim | ✅ Addressed | Documented design choice (`src/config.ts:35-43`); README security table marks `"*"` as "High" risk and now states cwd is "a starting directory, not a chroot". See §3. |
| 5 | TTL reaps dirs with file activity but no bot turn | ✅ Addressed | Explicitly documented as intended (`src/cwd-resolver.ts:88-94`): TTL tracks last inbound bot turn, refreshed via `utimesSync` on every `resolveSessionCwd`. See §3. |
| 6 | Best-effort marker → permanent cleanup exemption; marker forgeable | ✅ Fixed | Marker moved to sidecar registry `.cc-octo-sessions/` outside the agent cwd; re-created on every resolve if missing (self-heal, `src/cwd-resolver.ts:101-118`). Cleanup only deletes 16-hex dirs with a matching registry entry (`:171`). |
| 7 | Empty-string `cwdBase` slips past `??` fallback | ✅ Fixed | `nonBlank()` helper (`src/config.ts:199-203`) maps `""`/whitespace → `undefined` at both config-merge and env-apply paths, so `?? base.cwdBase ?? base.cwd` falls back correctly. Test coverage present. |
| 8 | `dmSpaceIdFromKey` re-implements router key grammar (drift risk) | ✅ Fixed | The inverse-parsing helper is removed entirely; the router key is now passed through verbatim, so there is a single source of truth for the partition grammar. |

Additional correctness checks: `lstatSync` (not `statSync`) guards against symlink-following in cleanup (`:175`); the 6h `setInterval` is `unref()`'d and `clearInterval`'d in the shutdown callback (`src/index.ts:48-57, 106`); the `cwd`/`cwdBase` deprecated alias is kept in sync. All sound.

## 2. New findings

No P0 or P1 issues. Two candidate blockers surfaced during the sweep were adversarially re-checked and downgraded:

- **Wildcard default + `bypassPermissions`** is not a new attack surface: `permissionMode: 'bypassPermissions'` and `Bash` were *already* the default on `main`, so the whitelist was already the sole gate and unrestricted shell access was already reachable. The delta is mostly SDK-internal tools (`TodoWrite`/`Task`). Non-blocking.
- **`mkdirSync` at `cwd-resolver.ts:99` lacks a try/catch**: an unwritable `cwdBase` (operator misconfig) throws `EACCES`, but it is caught at `src/index.ts:384`, logged with the underlying error (`console.error` at `:385`), and the user gets a graceful failure reply — no crash, not user-triggerable, and unfixable by retry anyway. Non-blocking.

## 3. Non-blocking suggestions (P2 / nit — not required for merge)

- **P2 — whitespace-only `CC_OCTO_SDK_ALLOWED_TOOLS` disables all tools.** `CC_OCTO_SDK_ALLOWED_TOOLS=" "` is a truthy string, so the branch runs, `parseCsv(" ")` → `[]`, and `[]` is forwarded to the SDK as an empty whitelist — silently disabling every tool. (`""` is fine — it's falsy and keeps the default.) Consider treating an empty parse result as "unset" (keep the default) or logging a warning. `src/config.ts:244-250`.
- **P2 — `anthropicBaseUrl` SSRF check is boot-time only.** `isAllowedApiUrl` validates IP literals + protocol but does not resolve DNS, so a hostname that resolves to a private IP at runtime (DNS-rebinding) isn't caught. This matches the existing `apiUrl` posture and is documented as a residual risk in `url-policy.ts`, so it's consistent — worth a one-line README note rather than code change. `src/config.ts:341-346`.
- **P2 — TTL semantics may surprise operators.** Reaping keys on last *bot turn*, not filesystem activity inside the sandbox, is reasonable and documented in-code; mirroring that note in the README config table would help operators running long background agent work.
- **nit — final cleanup on shutdown.** The shutdown callback `clearInterval`s the timer but doesn't run a final `cleanupExpiredCwds`. Harmless given the 7d TTL; optional hygiene.

## Summary

Solid, well-tested work. The Q3 isolation design (router-key-verbatim partitioning, sidecar registry with self-heal, symlink-guarded TTL sweep) is careful, and every prior-round blocker is genuinely resolved in the head commit. The remaining items are non-blocking hardening suggestions. Approving.

PR #57, opened 2026-06-07
fix(db-adapter): create dataDir with 0700 permissions

### LINKED ISSUE #56: bug(db-adapter): dataDir created with default perms (0755), not the documented 0700
## Summary

The `dataDir` directory (default `./data`) holds the SQLite databases with chat history. Both README and CONTRIBUTING promise it is created with `0700` permissions, but the implementation creates it with the process default (observed `0755` — group/other readable). This is a privacy-boundary doc/impl mismatch surfaced during end-to-end dogfooding.

## Evidence

`src/db-adapter.ts:34`:
```ts
mkdirSync(dir, { recursive: true });   // no mode → umask default (0755)
```

Documented promise:
- `README.md:89` — "SQLite database directory (created with `0700` permissions)"
- `CONTRIBUTING.md:223` — "Default permissions are `0700`."

Real run (`node dist/index.js`) produced:
```
data perms: 755 /tmp/.../data
```

## Impact

On a multi-user host, chat-history SQLite files (`cc-octo.db`, `-wal`, `-shm`) sit in a group/other-readable directory, contradicting the documented security posture.

## Fix

In `BetterSqliteAdapter`'s constructor:
- Skip filesystem setup entirely for the `:memory:` path (it must not touch the cwd).
- Create the data directory with `mode: 0o700` and enforce `0o700` via `chmod` so the guarantee holds regardless of umask or a pre-existing directory.

Add a regression test asserting the created dataDir is `0700` (POSIX only).

## Scope

`src/db-adapter.ts` + a test. No behavior change for the in-memory path.

lml2468:
Closes #56.

## What changed and why

The `dataDir` (default `./data`) holds chat-history SQLite files and is documented as created with `0700` perms (README:89, CONTRIBUTING:223). In reality `mkdirSync` used the umask default — a real run produced `0755` (group/other readable), contradicting the documented security posture. Found during end-to-end dogfooding.

`src/db-adapter.ts`:
- Create the directory with `mode: 0o700` **and** `chmod 0700` afterwards, so the guarantee holds regardless of umask or a pre-existing directory.
- Skip all filesystem setup for the `:memory:` path so it never touches/chmods the cwd that `dirname(':memory:')` resolves to.
- `chmod` is best-effort — a non-POSIX FS warns instead of crashing startup.

## Files modified

- `src/db-adapter.ts`
- `src/__tests__/db-adapter.test.ts` (2 new POSIX-only regression tests)

## Test results

- `npm test` — **742 passed** (was 740; +2)
- `npm run lint` — 0 warnings
- `npm run build` — clean
- Dogfood: real `node dist/index.js` run now produces `700 .../data` (verified).

## Breaking changes

None. In-memory behavior unchanged; existing data dirs are tightened to 0700 on next startup.

### DIFF
--- a/src/__tests__/db-adapter.test.ts
+++ b/src/__tests__/db-adapter.test.ts
@@ -1,8 +1,8 @@
 import { describe, it, expect, afterEach } from 'vitest';
 import { createAdapter } from '../db-adapter.js';
-import { mkdtempSync, rmSync, existsSync } from 'node:fs';
+import { mkdtempSync, rmSync, existsSync, statSync, mkdirSync, chmodSync } from 'node:fs';
 import { join } from 'node:path';
-import { tmpdir } from 'node:os';
+import { tmpdir, platform } from 'node:os';
 
 describe('DbAdapter', () => {
   let tmpDir: string;
@@ -75,6 +75,30 @@ describe('DbAdapter', () => {
     const adapter = createAdapter(':memory:');
     const fk = adapter.prepare('PRAGMA foreign_keys').get() as { foreign_keys: number };
     expect(fk.foreign_keys).toBe(1);
+    adapter.close();
+  });
+
+  // dataDir holds chat-history SQLite — README/CONTRIBUTING promise 0700.
+  const itPosix = platform() === 'win32' ? it.skip : it;
+
+  itPosix('creates the data directory with 0700 permissions', () => {
+    tmpDir = mkdtempSync(join(tmpdir(), 'db-adapter-perm-'));
+    const dir = join(tmpDir, 'data');
+    const adapter = createAdapter(join(dir, 'cc-octo.db'));
+    const mode = statSync(dir).mode & 0o777;
+    expect(mode).toBe(0o700);
+    adapter.close();
+  });
+
+  itPosix('tightens a pre-existing world-readable data directory to 0700', () => {
+    tmpDir = mkdtempSync(join(tmpdir(), 'db-adapter-perm2-'));
+    const dir = join(tmpDir, 'data');
+    // Operator (or a prior umask) left it 0755 — adapter must clamp it down.
+    mkdirSync(dir, { recursive: true });
+    chmodSync(dir, 0o755);
+    const adapter = createAdapter(join(dir, 'cc-octo.db'));
+    const mode = statSync(dir).mode & 0o777;
+    expect(mode).toBe(0o700);
     adapter.close();
   });
 });
--- a/src/db-adapter.ts
+++ b/src/db-adapter.ts
@@ -3,7 +3,7 @@
  * Enables future migration to node:sqlite when it reaches GA.
  */
 
-import { mkdirSync } from 'node:fs';
+import { mkdirSync, chmodSync } from 'node:fs';
 import { dirname } from 'node:path';
 import Database from 'better-sqlite3';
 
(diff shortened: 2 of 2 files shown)


Jerry-Xin (approved):
Summary: The PR is in scope for `cc-channel-octo` and correctly addresses the documented `dataDir` permission gap with focused regression coverage.

💬 Non-blocking

🟡 Warning — [src/db-adapter.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-r0dqzmts/repo/src/db-adapter.ts:37): `createAdapter('cc-octo.db')` would resolve `dirname()` to `.` and chmod the current working directory to `0700`. The main app path uses `join(config.dataDir, 'cc-octo.db')`, so normal startup is fine, but this exported adapter is easier to misuse now. Consider guarding `dir !== '.'` or validating `dataDir` as non-blank at config load.

🔵 Suggestion — [src/db-adapter.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-r0dqzmts/repo/src/db-adapter.ts:42): The best-effort `chmod` behavior is pragmatic for non-POSIX filesystems, but on POSIX an `EACCES`/`EPERM` means the documented security guarantee was not enforced. A follow-up could distinguish unsupported platforms from real POSIX permission failures and fail closed where appropriate.

✅ Highlights

- The `:memory:` special case avoids touching or chmodding the process cwd, which is the right fix for that edge case.
- The new tests cover both first creation and tightening an existing permissive directory: [src/__tests__/db-adapter.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-r0dqzmts/repo/src/__tests__/db-adapter.test.ts:84).
- I attempted the focused test run, but this checkout does not have dependencies installed: `vitest: command not found`.

yujiawei (approved):
# Code Review — PR #57 (cc-channel-octo)

**Verdict: APPROVED.** This is a correct, well-scoped security fix that closes a real gap (`dataDir` documented as `0700` but actually created `0755` under the default umask). It is a strict improvement over the prior behavior, with build/type-check/lint/tests all green. The points below are hardening and test-coverage **suggestions** — none of them blocks merge.

## What I verified

- **Mechanism is sound for the stated threat** (other non-root local users reading chat-history SQLite). A `0700` directory removes the `other` execute bit, so other users cannot traverse into it to resolve the inner files — the inner `.db` / `-wal` / `-shm` files' own mode is therefore irrelevant while they live inside the locked directory.
- **`chmodSync` is the load-bearing line, and it is correct.** `mkdirSync(..., { mode: 0o700 })` is masked by umask on fresh dirs and is a **no-op on a pre-existing directory**, so the explicit `chmodSync(dir, 0o700)` is what actually clamps the documented "pre-existing 0755" case. The second regression test (pre-create `0755` → expect `0700`) genuinely guards this line: removing the `chmod` would make that test fail.
- **`mode: 0o700` can never be *less* restrictive than `0700`** — `0700 & ~umask <= 0700` for every umask, since `0700` carries no group/other bits. So umask can only tighten, never loosen.
- **`':memory:'` guard is correct for both real callers** (`src/index.ts:60` passes `join(config.dataDir, 'cc-octo.db')`; tests pass `':memory:'`). Skipping FS setup for the in-memory path correctly avoids `chmod`-ing `dirname(':memory:')` → `.` (the CWD).
- **Fails safe on availability:** `chmod` failure warns and continues; `mkdir` failure still propagates (correct — can't proceed without the dir).
- `npm run type-check`, `npx eslint src/ --max-warnings 0`, and the `db-adapter` vitest suite all pass on the head SHA (`cd626bd`).

## Non-blocking suggestions (consider as follow-ups)

1. **`chmod` is best-effort / fail-open (hardening).** The `try/catch` logs a `console.warn` and then opens the DB and writes chat history regardless. The failure modes where `chmod` actually throws (dataDir owned by root/another user, restrictive mount/ACL, symlinked dataDir) are exactly the security-relevant ones, so the guarantee can silently degrade to a `0755` dir. For a security-sensitive directory, consider **fail-closed**: re-`stat` after `chmod` and `throw` if `(mode & 0o077) !== 0`. At minimum, make the warning state the concrete consequence ("chat history may be readable by other local users") rather than a generic chmod error. *(Reasonable people can prefer availability here, which is why this is a suggestion, not a block.)*

2. **DB content files remain umask-default (~`0644`); protection is perimeter-only (defense-in-depth).** `journal_mode = WAL` creates `cc-octo.db-wal` / `-shm` next to the `.db`, all created by better-sqlite3 at the process umask. They're safe while inside the `0700` dir, but the posture collapses if the dir perm is ever loosened or a backup/rotation copies the files out preserving `0644`. Optional: `chmod 0600` the `.db`/`-wal`/`-shm` after open, or run DB creation under `umask 0077`.

3. **`':memory:'` guard keys on a literal string (latent footgun, not currently reachable).** Other special sqlite forms (`''` anonymous on-disk, `file::memory:?cache=shared`, `file:` URIs) would pass the guard and cause `dirname()` → `.`, chmod-ing the CWD. No current caller hits this, but a shape-based guard (or asserting the resolved dir equals the configured `dataDir`) would be more robust.

4. **Test coverage gaps (non-blocking).** The two tests assert the **directory** mode only. Worth adding (a) a file-backed test asserting `(statSync(dbPath).mode & 0o077) === 0` to document/lock the file-level posture, and (b) a test that forces `chmodSync` to throw to exercise the warn branch. The current tests correctly skip on win32 (mode bits are meaningless on NTFS).

## Note for a human security reviewer

The confidentiality guarantee for chat history rests **entirely on the `0700` directory bit** (the DB files themselves stay `0644`). That is a valid and common pattern, but it is a single point of failure — please confirm the deployment's threat model is "same-host non-root local users" (which this fully addresses) and not "files may be copied/synced/backed up elsewhere" (which would need file-level `0600`). The README/CONTRIBUTING "created with 0700" claim is accurate for the leaf `dataDir`; it does not tighten a pre-existing `0755` parent, which is harmless for the default `./data` config.

PR #59, opened 2026-06-07
test(e2e): drive the real handleMessage pipeline, not a replica

### LINKED ISSUE #58: test(e2e): cover the real handleMessage pipeline, not a hand-copied replica
## Summary

`src/__tests__/e2e.test.ts` does not exercise the real message pipeline. Its `simulateMessage` helper is a **hand-copied replica** of `handleMessage` in `src/index.ts`. Because `handleMessage` is module-private, the copy can (and already does) drift from the real code — so the "end-to-end" tests give false confidence.

## Evidence

- `e2e.test.ts` reimplements the routing → context → query → stream flow in `simulateMessage` (~150 lines mirroring `index.ts`).
- The replica calls the **4-arg** `queryAgent(userContentForLLM, historyPrefix, contextStr, config)`, while the real `index.ts:346` calls the **5-arg** form with a `sessionCtx`:
  ```ts
  const sessionCtx: SessionCtx = { kind: isGroup ? 'group' : 'dm', sessionKey };
  queryAgent(..., config, sessionCtx);
  ```
- Net effect: the **per-session cwd isolation wiring added in PR #51** (derive `SessionCtx` from the router `sessionKey` → thread it into `queryAgent` → resolve the per-session cwd) has **zero end-to-end coverage**. The unit `cwd-resolver.test.ts` covers the resolver in isolation, but nothing asserts `index.ts` actually wires it.

This is the gap surfaced during dogfooding of PR #51.

## Fix

- Export `handleMessage` from `src/index.ts` (it already takes all collaborators as parameters — only `backfilledSessions` is module-private state).
- In `e2e.test.ts`, import and call the real `handleMessage`, delete the `simulateMessage` replica.
- Add an assertion that a DM and a group message thread the correct `sessionCtx` (kind + sessionKey) into `queryAgent` — the exact PR #51 behavior that was uncovered.

## Scope

Test-quality only. The single source change is widening `handleMessage`'s visibility to `export` (no behavior change).

lml2468:
Closes #58.

## What changed and why

`src/__tests__/e2e.test.ts` did not exercise the real pipeline — its `simulateMessage` helper was a hand-copied replica of `handleMessage` in `index.ts`. Because `handleMessage` was module-private, the copy drifted: it called the **4-arg** `queryAgent`, while the real `index.ts` calls the **5-arg** form with a `sessionCtx`. Net effect: the per-session cwd isolation wiring added in PR #51 had **zero** end-to-end coverage. Surfaced during dogfooding.

- **Export `handleMessage`** from `index.ts` and guard the auto-start (`import.meta.url` vs `process.argv[1]`) so importing the module in tests doesn't boot the gateway. The guard tolerates a missing `process.argv[1]` (`node -e` / `--input-type`) instead of throwing — caught by dogfooding the import path.
- **Replace the replica** with a thin wrapper over the real exported `handleMessage` (all 25 call sites unchanged).
- **Add two regression tests** asserting a DM and a group message thread the correct `SessionCtx` (`kind` + router `sessionKey`) into `queryAgent` — the exact PR #51 behavior the replica failed to exercise.

## Files modified

- `src/index.ts` — `export` + run-directly guard (no behavior change for the production entrypoint)
- `src/__tests__/e2e.test.ts` — real pipeline + 2 new tests, ~150-line replica deleted

## Test results

- `npm test` — **742 passed** (e2e file 22 → 24)
- `npm run lint` — 0 warnings
- `npm run build` — clean
- Dogfood: `node dist/index.js` still auto-starts (reaches network boundary); `import('./dist/index.js')` is silent and exports `handleMessage`.

## Breaking changes

None. Production entrypoint behavior is unchanged; the only source change is widening `handleMessage` visibility and guarding the auto-start.

### DIFF
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -14,6 +14,8 @@ vi.mock('../octo/api.js', () => ({
   sendTyping: vi.fn().mockResolvedValue(undefined),
   sendReadReceipt: vi.fn().mockResolvedValue(undefined),
   getGroupMembers: vi.fn().mockResolvedValue([]),
+  // G4 backfill path in the real handleMessage — default to no history.
+  getChannelMessages: vi.fn().mockResolvedValue([]),
   getUploadCredentials: vi.fn().mockResolvedValue({
     bucket: 'b', region: 'r', key: 'k',
     credentials: { tmpSecretId: 'i', tmpSecretKey: 'k', sessionToken: 't' },
@@ -50,8 +52,8 @@ import { SessionRouter } from '../session-router.js';
 import { GroupContext } from '../group-context.js';
 import { StreamRelay } from '../stream-relay.js';
 import { createAdapter, type DbAdapter } from '../db-adapter.js';
-import { queryAgent, sanitizeForSystemPrompt } from '../agent-bridge.js';
-import { resolveContent } from '../inbound.js';
+import { queryAgent } from '../agent-bridge.js';
+import { handleMessage } from '../index.js';
 import {
   sendMessage,
   sendReadReceipt,
@@ -131,8 +133,12 @@ function mockQueryYield(...texts: string[]): void {
 }
 
 /**
- * Simulate the handleMessage pipeline from index.ts.
- * Mirrors the real pipeline logic without importing the non-exported function.
+ * Drive the REAL pipeline. Previously this test reimplemented handleMessage as
+ * a hand-copied replica, which drifted from index.ts (e.g. it never threaded the
+ * PR#51 per-session cwd `sessionCtx` into queryAgent). We now import the real
+ * exported handleMessage so the e2e tests fail if the production pipeline
+ * changes. The wrapper only fills in the fixed BOT_ID so the 25 call sites stay
+ * unchanged.
  */
 async function simulateMessage(
   msg: BotMessage,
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
Summary: Project relevance gate passed; this PR is in scope and correctly improves e2e coverage for the real message pipeline.

🔴 Blocking: None.

💬 Non-blocking: None.

✅ Highlights:
- `src/__tests__/e2e.test.ts:151` now drives the exported production `handleMessage`, removing the drift-prone replica.
- `src/index.ts:343` and `src/index.ts:347` preserve the intended `SessionCtx` wiring into `queryAgent`.
- `src/__tests__/e2e.test.ts:225` and `src/__tests__/e2e.test.ts:235` add focused regression coverage for DM and group session context.
- `src/index.ts:495` guards auto-start on direct execution, so importing the module in tests stays side-effect safe.

Verification run:
- `npm test -- src/__tests__/e2e.test.ts` passed.
- `npm run type-check` passed.
- `npm run lint` passed.

yujiawei (approved):
# Code Review — PR #59 (cc-channel-octo)

## Verdict: APPROVED

This is a focused, well-motivated test-quality change. The previous `e2e.test.ts` exercised a hand-copied replica of `handleMessage` that had silently drifted from `index.ts` (the replica called the 4-arg `queryAgent`, while production calls the 5-arg form with `sessionCtx`), so the PR#51 per-session cwd isolation had **zero** end-to-end coverage. Replacing the replica with the real exported `handleMessage` closes that gap and makes the e2e suite fail if the production pipeline changes. No production behavior change beyond widening `handleMessage` visibility and guarding the auto-start.

## Verification (independent, at head `77f25f6`)

I checked out the head SHA and re-ran everything from scratch:

- `npm run build` — clean
- `npm test` — **742 passed** (e2e file: 24 tests, +2 as claimed)
- Dogfood `import('./dist/index.js')` — silent, exports `handleMessage` ✅
- Dogfood `node dist/index.js` — still auto-starts, reaches `main()` and fails only on missing config ✅ (proves the run-directly guard fires correctly)

I also confirmed the cross-file claims rather than trusting the description:

- `src/index.ts:347` calls `queryAgent(userContentForLLM, historyPrefix, contextStr, config, sessionCtx)` — `sessionCtx` is the 5th positional arg (`mock.calls[0][4]`), so the new assertions target the right slot.
- `src/index.ts:343-346` builds `{ kind: isGroup ? 'group' : 'dm', sessionKey }`.
- DM assertion `{ kind: 'dm', sessionKey: 'user-001' }` is correct: `SessionRouter.sessionKey()` returns the bare `from_uid` when no `s{spaceId}_` prefix is present (`session-router.ts:141-146`).
- Group assertion `{ kind: 'group', sessionKey: 'group-ch-001:user-001' }` is correct: the group key is `${channel_id}:${from_uid}`, embedding `from_uid` so each member gets their own cwd — exactly the PR#51 behavior the old replica never exercised.
- The new `getChannelMessages: vi.fn().mockResolvedValue([])` mock is necessary, not noise: the real `handleMessage` hits the G4 group cold-start backfill path (`index.ts:311`) that the replica never had.
- Only `e2e.test.ts` imports from `index.js`, so the new `export` has no other ripple.

## Findings

No P0/P1/P2 blocking issues.

### Non-blocking observations (nits)

1. **Positional mock indexing.** The regression tests read `(queryAgent ...).mock.calls[0][4]` to grab `sessionCtx`. This is brittle against any future reordering of `queryAgent`'s parameters — a reorder would make the test read the wrong arg and fail with a confusing diff rather than a clear "signature changed" message. It is adequately commented, so this is purely a nit; a named-extraction helper could harden it later.

2. **Run-directly guard edge cases.** `import.meta.url === pathToFileURL(process.argv[1]).href` is the standard ESM main-module check and the missing-`argv[1]` tolerance (for `node -e` / `--input-type`) is good defensive thinking. It will not match if the entrypoint is reached via a symlink whose realpath differs from `import.meta.url`; not a concern for the documented `node dist/index.js` deployment, just noting the assumption.

## Summary

The change does exactly what it claims, the new tests assert real production wiring (not coincidentally-passing values), and all gates are green on a clean checkout. Approving.

Jerry-Xin (approved):
Re-review: LGTM

PR #64, opened 2026-06-07
feat(config): multi-bot support (v0.3)

lml2468:
Third v0.3 roadmap item: multi-bot support.

## What changed and why

Run several independent bots from one process via a top-level `bots[]` array. Single-bot configs are completely unchanged.

**Design — per-bot isolation:**
- `resolveBotConfigs(config)` expands a loaded Config into one concrete Config per bot. No `bots` → `[config]` (backward compatible). Each entry inherits the base config and overrides `apiUrl`/`dataDir`/`cwdBase`/`model`/`systemPrompt`/blocklists.
- **`dataDir` and `cwdBase` are namespaced by bot `id` by default** (`./data/support`, …) so bots never share conversation history, sandboxes, or lock files. This also sidesteps the `sessionKey` collision risk — `SessionRouter.sessionKey()` has no botId, so a shared store would cross-contaminate per-user history; separate stores avoid that entirely.
- Fails fast on missing/duplicate tokens or duplicate ids, and SSRF-validates per-bot `apiUrl` overrides. Top-level `botToken` becomes optional when `bots[]` is present.
- `main()` starts one independent stack per bot (`startBot`). In multi-bot mode the orchestrator owns a single SIGINT/SIGTERM shutdown draining all bots; gateways skip their own signal handlers via a new `OctoGateway(config, { handleSignals })` option (default true → single-bot unchanged).

## Files modified

- `src/config.ts` (`BotOverride`, `bots`, `resolveBotConfigs`), `src/index.ts` (`startBot` + multi-bot main), `src/gateway.ts` (`handleSignals` option)
- `src/__tests__/config.test.ts` (+11), `README.md`, `CHANGELOG.md`, `config.example.json`

## Test results

- `npm test` — **799 passed** (+11)
- Dogfood: a 2-bot config boots two namespaced stacks (`data/support`, `data/ops`) each to the network boundary.

Roadmap: "multi-bot support" removed from the v0.3 row (shipped here).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,11 +15,15 @@ While the major version is `0`, minor releases may carry breaking changes.
   inherits all top-level fields, and may override `apiUrl`/`dataDir`/`cwdBase`/
   `model`/`systemPrompt`/blocklists. Each bot gets a fully independent stack
   (gateway + router + store); `dataDir` and `cwdBase` are namespaced by id by
-  default so bots never share history or sandboxes. `resolveBotConfigs()` expands
-  the config and fails fast on missing/duplicate tokens or duplicate ids. In
-  multi-bot mode the orchestrator owns a single SIGINT/SIGTERM shutdown that
-  drains all bots (gateways skip their own signal handlers via a new
-  `handleSignals` option). Single-bot configs are unchanged.
+  default so bots never share history or sandboxes. Bot ids are validated as
+  conservative slugs (no path separators) to keep that namespacing safe.
+  `resolveBotConfigs()` expands the config and fails fast on missing/duplicate
+  tokens or duplicate/invalid ids. All bot ids are registered into every router
+  so a mention-free group can't trigger bot-to-bot reply loops, and the
+  cold-start backfill sentinel is keyed per bot. In multi-bot mode the
+  orchestrator owns a single SIGINT/SIGTERM shutdown that drains all bots
+  (gateways skip their own signal handlers via a new `handleSignals` option).
+  Single-bot configs are unchanged.
 - **Tool progress display** (v0.3, opt-in) — with `sdk.toolProgress`
   (`CC_OCTO_SDK_TOOL_PROGRESS=true`), the bot posts brief `🔧 Running <tool>…`
   notices as the agent invokes tools. `queryAgent` gained a non-breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,16 @@ While the major version is `0`, minor releases may carry breaking changes.
 
 ### Added
 
+- **Multi-bot support** (v0.3) — run several independent bots in one process via
+  a top-level `bots[]` config array. Each entry needs its own `botToken` + `id`,
+  inherits all top-level fields, and may override `apiUrl`/`dataDir`/`cwdBase`/
+  `model`/`systemPrompt`/blocklists. Each bot gets a fully independent stack
+  (gateway + router + store); `dataDir` and `cwdBase` are namespaced by id by
+  default so bots never share history or sandboxes. `resolveBotConfigs()` expands
+  the config and fails fast on missing/duplicate tokens or duplicate ids. In
+  multi-bot mode the orchestrator owns a single SIGINT/SIGTERM shutdown that
+  drains all bots (gateways skip their own signal handlers via a new
+  `handleSignals` option). Single-bot configs are unchanged.
 - **Tool progress display** (v0.3, opt-in) — with `sdk.toolProgress`
   (`CC_OCTO_SDK_TOOL_PROGRESS=true`), the bot posts brief `🔧 Running <tool>…`
   notices as the agent invokes tools. `queryAgent` gained a non-breaking
--- a/README.md
+++ b/README.md
@@ -139,9 +139,11 @@ Leave the field unset to talk to Anthropic's public endpoint directly.
 
 To run several bots from one process, add a `bots` array instead of (or in
 addition to) the single top-level `botToken`. Each entry needs its own
-`botToken` and an `id`, inherits every top-level field, and may override
-`apiUrl`, `dataDir`, `cwdBase`, `model`, `systemPrompt`, `botBlocklist`, and the
-mention lists:
+`botToken` and should set a stable `id` (slug: letters, digits, `.`, `_`, `-`).
+If `id` is omitted it falls back to the positional `bot0`, `bot1`, … — which
+works but produces index-dependent directory names, so prefer an explicit id.
+Each entry inherits every top-level field and may override `apiUrl`, `dataDir`,
+`cwdBase`, `model`, `systemPrompt`, `botBlocklist`, and the mention lists:
 
 ```jsonc
 {
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Users talk to a bot in Octo (DM or group @mention). The bot sends messages to Cl
 - **In-chat commands** — `/reset` clears your own session's stored history (not the shared recent-group-context cache), `/config` shows the active settings, `/help` lists commands. Scoped per-user (even in groups); subject to the same per-session rate limit as normal messages.
 - **Rate limiting** — Per-session token bucket (default 5 req/min) with debounced rejection notices.
 - **Security by configuration** — `allowedTools` whitelist + per-session `cwdBase` isolation. No runtime permission prompts (headless mode).
+- **Multi-bot** — Run several independent bots in one process via a `bots[]` config array; each gets its own token, data directory, and sandbox root (no shared history).
 - **Zero infrastructure** — Single process, single SQLite file, `npm start` and go.
 
 ## Quick Start
@@ -133,6 +134,33 @@ ANTHROPIC_BASE_URL=https://claude-gw.internal.example.com npm start
 ```
 
 Leave the field unset to talk to Anthropic's public endpoint directly.
+
+### Multi-bot
+
+To run several bots from one process, add a `bots` array instead of (or in
+addition to) the single top-level `botToken`. Each entry needs its own
+`botToken` and an `id`, inherits every top-level field, and may override
+`apiUrl`, `dataDir`, `cwdBase`, `model`, `systemPrompt`, `botBlocklist`, and the
+mention lists:
+
+```jsonc
+{
+  "apiUrl": "https://your-octo-instance.com",
+  "dataDir": "./data",
+  "cwdBase": "/home/deploy/cc-octo-sandboxes",
+  "bots": [
+    { "id": "support", "botToken": "bf_AAA" },
+    { "id": "ops", "botToken": "bf_BBB", "model": "claude-opus-4-8" }
+  ]
+}
+```
+
+Each bot runs a fully independent stack (gateway, router, store). By default its
+`dataDir` and `cwdBase` are namespaced by `id` (`./data/support`,
+`/home/deploy/cc-octo-sandboxes/support`, …), so **bots never share
+conversation history or working directories**. Set `dataDir`/`cwdBase` on an
+entry to override that. When `bots` is absent, the process runs a single bot
+from the top-level fields exactly as before.
 
 ## Security Model
 
@@ -283,7 +311,6 @@ src/
 
 - **Per-session `cwdBase` isolation** — Each session (DM peer, or individual group member) gets its own SHA-256 hex sandbox under `cwdBase`, partitioned by the same key as conversation history; idle sandboxes (>7d) are auto-cleaned every 6h. Note: `cwdBase` separates sessions from each other but does not confine a session to its directory (absolute-path reads via Bash/Read remain possible) — see the Security Model section.
 - **Stateless sessions** — Uses the v1 `query()` API. Workspace state (open files, command history) does not persist across messages. The v2 Session API is planned for v0.3.
-- **Single bot** — One bot per process. Multi-bot support is planned for v0.3.
 
 ## Roadmap
 
@@ -291,7 +318,7 @@ src/
 |---------|-------|
 | **v0.1** | Text messaging, streaming, session persistence, rate limiting, security model |
 | **v0.2** *(current)* | Media reception & sending (image/file/RichText), @mention, group context, per-session `cwdBase` isolation, self-hosted gateway, SSRF/prompt-injection hardening |
-| **v0.3** | v2 Session API, multi-bot support |
+| **v0.3** | v2 Session API |
 | **v1.0** | GROUP.md/THREAD.md configuration, webhook mode |
 
 ## Contributing
--- a/config.example.json
+++ b/config.example.json
@@ -12,6 +12,7 @@
     "_comment_allowed_tools": "Q2: either '*' (allow every tool the SDK exposes) or an explicit string[] whitelist, e.g. ['Read', 'Write', 'Edit', 'Bash', 'Glob', 'Grep', 'WebFetch', 'WebSearch']. Env override CC_OCTO_SDK_ALLOWED_TOOLS accepts the same two shapes (a value containing '*' means wildcard; otherwise CSV).",
     "allowedTools": "*",
     "permissionMode": "bypassPermissions",
+    "_comment_tool_progress": "v0.3: set toolProgress=true to post '🔧 Running <tool>…' notices as the agent invokes tools (deduped, capped per turn). Env override: CC_OCTO_SDK_TOOL_PROGRESS=true. Off by default.",
     "settingSources": ["user"]
   },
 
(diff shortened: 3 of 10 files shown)


Jerry-Xin (changes requested):
The PR is relevant to `cc-channel-octo`, but multi-bot mode introduces a blocking bot-to-bot loop risk in mention-free groups.

🔴 Blocking

🔴 Critical — Multi-bot instances do not register each other as known bots, so mention-free groups can create bot-to-bot reply loops.
`startBot()` creates each `SessionRouter` with only its own `gateway.botId` at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0o9xpzzz/repo/src/index.ts:119). `SessionRouter` has `registerKnownBot()` support at [src/session-router.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0o9xpzzz/repo/src/session-router.ts:73), but the multi-bot orchestrator never calls it. In a `mentionFreeGroups` channel, bot B will accept bot A’s plain text response because the group mention gate is bypassed at [src/session-router.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0o9xpzzz/repo/src/session-router.ts:247), and group bot-looking/known-bot messages are not dropped unless manually blocklisted at [src/session-router.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0o9xpzzz/repo/src/session-router.ts:238). That can lead to unbounded automated replies, chat spam, and model/API cost. Fix by wiring all started bot IDs into every router, and enforce a group bot-message drop policy that is not bypassed by mention-free mode unless explicitly allowed.

💬 Non-blocking

🟡 Warning — The cold-start backfill sentinel is still process-global and keyed only by `sessionKey`.
`backfilledSessions` at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0o9xpzzz/repo/src/index.ts:553) is shared by all bot stacks, while each bot has its own store. If two bots receive the same group/user session key, the first one can mark it backfilled at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0o9xpzzz/repo/src/index.ts:373), causing the second bot to skip backfill despite an empty per-bot database. Key this by `botId + sessionKey`, or make it stack-local.

🟡 Warning — Bot IDs are used directly as path segments without validation.
`resolveBotConfigs()` builds default `dataDir` and `cwdBase` with raw `id` at [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-0o9xpzzz/repo/src/config.ts:429). IDs like `../ops` or `a/../b` can escape or alias the intended namespace. Since the feature’s safety story depends on default isolation, consider restricting IDs to a conservative slug pattern and/or checking resolved path uniqueness.

✅ Highlights

The config expansion keeps single-bot behavior backward compatible, validates duplicate tokens/IDs, and re-runs SSRF validation for per-bot `apiUrl` overrides. The shutdown ownership split via `handleSignals` is also a good direction for multi-bot process lifecycle.

Verification note: I attempted focused tests and type-checking, but this checkout has no installed dependencies: `vitest` and `tsc` were not found.

lml2468:
Thanks @Jerry-Xin — all three addressed in 5fbc845.

**🔴 bot-to-bot loop (mention-free groups)** — two-part fix:
1. `main()` registers every started bot's id into every router (`registerKnownBot`), after all gateways start (ids are only known post-registration).
2. `session-router`: in a mention-free group, drop messages from known/bot-looking uids unless explicitly whitelisted — the @-mention gate that normally stops this is bypassed there. An explicit `@bot` mention still goes through.
Tests: human ok, `_bot`-suffix dropped, registered sibling dropped, sibling-with-@mention answered, `allowedBotUids` whitelist honored.

**🟡 shared backfill sentinel** — now keyed `${botId} ${sessionKey}`, so bot A can't make bot B skip backfill against its own empty store.

**🟡 unvalidated bot id as path segment** — `resolveBotConfigs` rejects ids that aren't a conservative slug (`/^[a-zA-Z0-9._-]+$/`, and not `.`/`..`), closing the `../ops` traversal/aliasing hole. Test matrix added.

811 tests pass (type-check / lint / build clean).

Jerry-Xin (changes requested):
Summary: The PR is relevant to cc-channel-octo and the multi-bot design mostly fits the project, but it should not merge with a NUL byte committed into TypeScript source.

🔴 Blocking

🔴 Critical — `src/index.ts:386` contains an actual NUL byte inside the template literal: `const backfillKey = \`${botId}${sessionKey}\`;`. This makes `src/index.ts` appear as a binary file to tools like `rg`, and can break editors, diff tooling, linters, formatters, or source-processing pipelines. Use an escaped sequence or ordinary delimiter instead, for example `${botId}\0${sessionKey}` as source text or `${botId}:${sessionKey}` if collision risk is acceptable.

💬 Non-blocking

🟡 Warning — `src/config.ts:429` allows missing bot ids by silently assigning `bot0`, `bot1`, etc., while the README/PR description says each bot entry needs an `id`. Either reject missing ids or document the generated-id behavior explicitly so operators do not accidentally get unstable-looking directory names.

🔵 Suggestion — `src/index.ts:49-60` registers sibling bot ids only after all gateways have started. There is a small startup window where an already-started router in a mention-free group does not yet know about sibling bot ids. Consider delaying message handling until all routers are registered, or starting gateways after registration metadata is available.

✅ Highlights

The per-bot config expansion isolates default `dataDir` and `cwdBase`, validates duplicate tokens/ids, and re-checks per-bot `apiUrl` overrides. The added mention-free loop guard and per-bot backfill keying address real multi-bot cross-contamination risks.

I could not run the test suite locally because dependencies are not installed in this checkout: `tsc` and `vitest` were not found.

lml2468:
Thanks @Jerry-Xin — third round addressed in 8c13884.

**NUL byte (blocking)** — confirmed and removed: a literal NUL had crept into the `backfillKey` template in src/index.ts. Replaced with an escaped null (``) so it's plain ASCII in source while still serving as a separator that cannot collide with ids/keys. Added a NUL scan to my pre-push checks; no NUL bytes remain in any `src/*.ts`.

**Sibling-registration window (non-blocking)** — closed. `startBot()` no longer wires its message handler; it returns a `beginHandling()` thunk. `main()` now: start all gateways, register every sibling bot id into every router, then call `beginHandling()` on each. So no router accepts a message before it knows its siblings.

**Generated ids (non-blocking)** — documented: an omitted `id` defaults to positional `bot<N>` (works, but index-dependent dir names), with a recommendation to set an explicit slug id.

811 tests pass (type-check / lint / build clean).

Jerry-Xin (changes requested):
Scope check passed: this PR is relevant to `cc-channel-octo`, but there is one blocking startup correctness issue.

**🔴 Blocking**
- 🔴 Critical: Multi-bot startup can ACK and drop inbound messages before handlers are installed. In `src/index.ts:49`, all bots are started with `Promise.all`; each `startBot()` calls `gateway.start()` at `src/index.ts:135-136`, but the real message handler is only installed later in `beginHandling()` at `src/index.ts:149-160`, after every bot has started and cross-registration finishes at `src/index.ts:57-67`. Meanwhile `OctoGateway.start()` connects the WebSocket during `registerAndConnect()` (`src/gateway.ts:62-64`, `src/gateway.ts:197-198`). If bot A connects quickly and bot B registration is slow or times out, bot A can receive messages while `this.onMessage` is still null; `WKSocket` sends `RECVACK` before dispatching (`src/octo/socket.ts:717-749`), and `OctoGateway.handleMessage()` then silently does nothing via `this.onMessage?.(msg)` (`src/gateway.ts:201-204`). Those messages are acknowledged but not processed, so users can get no response during startup. Please split registration from socket connection, or otherwise ensure no WebSocket starts receiving/ACKing messages until the router and handler are ready for every bot.

**💬 Non-blocking**
- 🟡 Warning: The docs and changelog say every `bots[]` entry needs an `id`, but `BotOverride.id` is optional and `resolveBotConfigs()` falls back to positional ids (`src/config.ts:98-105`, `src/config.ts:435`). That makes directory names unstable when the array is reordered. Either make `id` required at runtime or update the docs to describe the fallback clearly.
- 🔵 Suggestion: Add an integration-style test for multi-bot startup ordering. The current tests cover config expansion and routing loop guards, but not the ACK-before-handler lifecycle path.

**✅ Highlights**
- Per-bot default namespacing for `dataDir` and `cwdBase` is a good default for avoiding session and lock-file collisions.
- The mention-free group loop guard is well targeted and preserves explicit mentions and `allowedBotUids`.
- Per-bot SSRF validation for `apiUrl` overrides is correctly retained.

yujiawei (changes requested):
# Code Review — PR #64 (cc-channel-octo)

Independent review of `feat(config): multi-bot support (v0.3)` at head `8c13884`. Verified locally: `npm run type-check`, `npm run lint`, and the multi-bot test suites (`config.test.ts` 87 tests, `session-router.test.ts` 40 tests) all pass.

The design is solid — config expansion, per-bot isolation, and the loop guard are well thought through, and the prior review-round findings (NUL byte in source, missing cross-registration, per-bot backfill keying, slug id validation, per-bot SSRF) are all genuinely fixed at this SHA. There is, however, one remaining startup correctness bug that is **multi-bot specific and causes silent message loss**, so this should not merge as-is.

## P1 (blocking) — Multi-bot startup can ACK then silently drop inbound messages

In `src/index.ts`, all bots are started concurrently and the real message handler is installed only *after* every bot has finished starting and cross-registration completes:

```ts
const stacks = await Promise.all(botConfigs.map((c) => startBot(c, multi)));  // each calls gateway.start()
// ... cross-register sibling bot ids ...
for (const s of stacks) s.beginHandling();   // <-- onMessage installed HERE
```

But `gateway.start()` (`src/gateway.ts:61-64`) connects the WebSocket *inside* `startBot`, via `registerAndConnect()`. Once connected, `WKSocket` sends `RECVACK` to the server **immediately** on receiving a message and *then* dispatches it:

- `src/octo/socket.ts:717` — `this.sendRaw(encodeRecvackPacket(...))` (ACK sent)
- `src/octo/socket.ts:749` — `this.opts.onMessage(msg)` (dispatch)

That dispatch lands in `OctoGateway.handleMessage` (`src/gateway.ts:201-204`):

```ts
private handleMessage(msg: BotMessage): void {
  if (this._draining) return;
  if (msg.from_uid === this.robotId) return;
  this.onMessage?.(msg);   // <-- null until beginHandling() runs
}
```

**Window:** if bot A connects quickly while bot B's `registerBot()` / WS connect is slow (or retries), bot A's socket is live and will `RECVACK` any inbound message — but `this.onMessage` is still `null`, so `this.onMessage?.(msg)` is a no-op. The message is acknowledged to the server (won't be redelivered) yet never processed. The user gets no reply.

This is genuinely new behavior introduced by this PR: in single-bot mode the handler is wired immediately after `gateway.start()` with no intervening await, so there was effectively no gap. The multi-bot `Promise.all` + deferred `beginHandling()` opens a real gap whose size equals the slowest sibling's startup time.

The PR comment for `beginHandling()` reasons about the *loop-guard* ordering (registering sibling ids before handling) but does not account for the socket already being live and ACKing during that window. Fixing the loop-guard ordering and fixing this are related but distinct: the loop guard needs handlers installed *after* cross-registration, while message-loss avoidance needs the *socket* to not be live (or to buffer) until handlers are ready.

**Suggested fix directions** (any one resolves it):
- Split `OctoGateway.start()` into a register phase and a connect phase: `registerAll()` (token + WS URL, no socket connect) across all bots → cross-register router ids → `beginHandling()` → then `connect()` the sockets. This keeps the "no socket receives before handler is ready" invariant for every bot.
- Or have the gateway buffer/queue inbound messages received before `onMessage` is set, and flush on install (less clean — it still ACKs before the router exists, so a crash in the window still loses them).

A regression test driving the multi-bot startup ordering (handler-null during the connect window) would lock this down — the current tests cover config expansion and routing in isolation but not the lifecycle path.

## P2 (non-blocking) — partial-start leak on multi-bot boot failure

`await Promise.all(botConfigs.map((c) => startBot(c, multi)))` — if bot B's `startBot` throws (bad token, lock held, SSRF), the rejection propagates and `main()` exits via the top-level `.catch` → `process.exit(1)`. But bots that already completed `startBot` have acquired their `gateway.lock` (`src/gateway.ts:116-143`) and an open WS/DB, and in multi-bot mode their own signal handlers are disabled (`handleSignals: false`). Nothing calls their `shutdown()`, so on a hard `process.exit(1)` the lock files are left behind. Lock recovery is PID-based and self-heals on next boot (`process.kill(pid, 0)` → stale removal), so impact is limited to a confusing "Another instance is running" only if a PID is recycled — low severity, but worth a `Promise.allSettled` + cleanup of started stacks on partial failure for a cleaner boot story.

## P2 (non-blocking) — docs say `id` is required, runtime makes it optional

`BotOverride.id` is optional and `resolveBotConfigs()` falls back to positional `bot${i}` (`src/config.ts:435`). The README/PR text says each entry "needs its own `botToken` and an `id`." The code comment already documents the fallback well; just align the README/CHANGELOG wording (or make `id` required at runtime) so operators don't get index-dependent directory names that shift when the array is reordered.

## Highlights

- `resolveBotConfigs()` keeps single-bot fully backward compatible (`[{...config, botId: 'default'}]`), fails fast on missing/duplicate tokens and duplicate ids, validates ids as conservative slugs to protect the dataDir/cwdBase namespacing, and re-runs SSRF validation on per-bot `apiUrl` overrides.
- Default per-bot namespacing of `dataDir` + `cwdBase` cleanly sidesteps the `sessionKey` collision risk noted in the PR description — separate stores mean per-user history can't cross between bots.
- The mention-free loop guard (`src/session-router.ts:259`) is correctly targeted: it preserves explicit `@`-mentions and `allowedBotUids`, and the cross-registration makes siblings recognizable even without a `_bot` suffix. Good test coverage on this path.
- `handleSignals` option for split shutdown ownership is the right shape for multi-bot lifecycle.

## Verdict

Approve once the P1 startup ACK-before-handler window is closed (the P2s are optional). The feature is close — this is a lifecycle-ordering fix, not a design problem.

lml2468:
Thanks @Jerry-Xin — fifth round addressed in bf6932b.

**ACK-before-handler (blocking)** — split the gateway lifecycle:
- `register()` — acquire lock + REST `registerBot` (populates botId/ownerUid), does **not** open the socket.
- `connect()` — open the WebSocket + heartbeat + signal handlers; throws if called before `register()`.
- `start()` kept as a back-compat wrapper (register + connect) for the single-bot path and existing tests.

`main()` now orders startup as: register all bots → install every message handler → cross-register sibling ids → **connect all sockets**. So no socket can receive (and RECVACK) a message before its router/handler is ready. Added 4 gateway lifecycle tests (register without socket; handler installable between phases; connect-before-register throws; start() back-compat).

**id fallback docs (non-blocking)** — documented the positional `bot<N>` fallback in both the `BotOverride` JSDoc and the README, recommending an explicit slug id.

**integration test (suggestion)** — the new gateway lifecycle tests plus the existing multi-bot config/loop-guard tests cover the ordering contract; a full socket-level integration harness is a larger follow-up.

815 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (approved):
Summary: The PR is in scope for `cc-channel-octo` and the multi-bot architecture is generally sound: per-bot config expansion, two-phase gateway startup, separate stores, and mention-free loop prevention are all aligned with the project.

💬 Non-blocking

🟡 Warning — Env precedence is inconsistent for per-bot `model` and `systemPrompt`. `loadConfig()` applies env first at `src/config.ts:379`, but `resolveBotConfigs()` then overlays file-defined per-bot values at `src/config.ts:469-472`, so `CC_OCTO_SDK_MODEL` / `CC_OCTO_SDK_SYSTEM_PROMPT` no longer win for those bots. If env is intended to remain highest priority, apply it after expansion or avoid overriding env-populated fields.

🟡 Warning — Explicit per-bot `dataDir` / `cwdBase` overrides can accidentally collide. `src/config.ts:463-465` accepts overrides as-is; two bots with the same `dataDir` share the same `gateway.lock` path from `src/gateway.ts:46`, and two bots with the same `cwdBase` can share session sandbox hashes. Either validate uniqueness or document that explicit sharing is unsupported.

🔵 Suggestion — Add one integration-level test for multi-bot orchestration in `src/index.ts`: verify both gateways register before either connects, sibling IDs are cross-registered, and `shutdownAll` drains every stack. Current tests cover the components, but not the full startup sequence.

✅ Highlights

The two-phase `register()` / `connect()` split in `src/gateway.ts:79-112` addresses the ACK-before-handler race cleanly.

The per-bot backfill key in `src/index.ts:408-417` prevents one bot’s cold-start state from suppressing another bot’s history backfill.

The mention-free group loop guard in `src/session-router.ts:247-261` is narrowly scoped and preserves explicit mention behavior.

Verification note: I attempted `npm run type-check` and targeted tests, but this checkout has no installed dependencies (`tsc` and `vitest` were not found).

PR #65, opened 2026-06-07
feat(agent): persistent sessions via SDK v2 Session API (v0.3, opt-in)

lml2468:
Final v0.3 roadmap item: v2 Session API migration (persistent agent sessions).

## What changed and why

When `sdk.persistentSession` is enabled (config or `CC_OCTO_SDK_PERSISTENT_SESSION=true`), agent workspace state (open files, command history, context) persists across messages. **Default off** — the proven stateless v1 `query()` path is unchanged, so this is purely additive.

**Design:**
- `session-store`: new `sdk_sessions` table mapping our `sessionKey` → SDK session UUID (`set`/`get`/`clear`, upsert latest-wins).
- `agent-bridge`: `queryAgent` gains `opts.resume` (forwarded as the SDK `resume` option) and `opts.onSessionId` (reports the `session_id` once per turn). Both guarded so a throwing callback can't break the stream.
- `index.ts`: when enabled, look up the stored SDK session id, `resume` it, and persist the new id. **On resume the history prefix is suppressed** (the SDK session already holds the conversation) to avoid double-injection — we still write to our own store for the fallback path, `/reset`, and group context.
- `commands.ts`: `/reset` also clears the stored SDK session id, so a cleared conversation is never resumed (same class as the earlier `/reset` backfill-barrier fix).
- `config`: `sdk.persistentSession` flag + env.

## Files modified

- `src/agent-bridge.ts`, `src/session-store.ts`, `src/index.ts`, `src/commands.ts`, `src/config.ts`
- Tests: agent-bridge (+4), session-store (+6), e2e (+3), config (+4 matrix)

## Test results

- `npm test` — **838 passed** (+23)
- NUL scan clean.

With this merged, **all four v0.3 roadmap items are done** (slash commands, tool progress, multi-bot, v2 Session API). Remaining roadmap: v1.0 (GROUP.md/THREAD.md config, webhook mode).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ While the major version is `0`, minor releases may carry breaking changes.
 
 ### Added
 
+- **Persistent sessions** (v0.3, opt-in) — with `sdk.persistentSession`
+  (`CC_OCTO_SDK_PERSISTENT_SESSION=true`), agent workspace state persists across
+  messages via the SDK v2 Session API. Each session's SDK session id is stored
+  (new `sdk_sessions` table) and `resume`d on the next turn; on resume the
+  history prefix is suppressed (the SDK session already holds it). `/reset`
+  clears the stored session id so a cleared conversation is not resumed.
+  `queryAgent` gained an `opts.resume` + `opts.onSessionId` channel (both
+  guarded). Default off — the proven stateless v1 `query()` path is unchanged.
 - **Multi-bot support** (v0.3) — run several independent bots in one process via
   a top-level `bots[]` config array. Each entry needs its own `botToken` + `id`,
   inherits all top-level fields, and may override `apiUrl`/`dataDir`/`cwdBase`/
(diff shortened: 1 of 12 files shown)


Jerry-Xin (commented):
Summary: The PR is in scope, but the new SDK session mapping currently bypasses the project’s existing session expiry lifecycle.

🔴 Blocking

🔴 Critical — `sdk_sessions` is never expired with the owning session.  
`src/session-store.ts:68` creates `sdk_sessions` without a foreign key or TTL cleanup, while `cleanExpired()` at `src/session-store.ts:213` only deletes rows from `sessions`. Then `src/index.ts:496` resumes any stored SDK session ID regardless of whether the local session/history was expired and recreated. This means an idle session whose SQLite history was removed after 7 days can still resume the old SDK transcript/workspace state, effectively resurrecting expired conversation context. That contradicts the documented 7-day session expiry in `README.md:23` and the database lifecycle in `ARCHITECTURE.md:189`.  
Fix by tying `sdk_sessions` to the same lifecycle as `sessions` or explicitly deleting expired SDK mappings in `cleanExpired()`, and add a regression test where an expired session with an SDK session ID is cleaned and the next turn does not pass `resume`.

💬 Non-blocking

🟡 Warning — stale SDK session IDs are not cleared on resume failure.  
`src/session-store.ts:260` documents clearing the mapping “on /reset or a resume failure”, but the current call path only clears it on `/reset` in `src/commands.ts:125`. If the SDK session file is missing, deleted externally, or invalid, the same bad ID can be retried forever. Consider clearing the mapping, and possibly retrying once without `resume`, when a resumed query fails due to an invalid session.

✅ Highlights

The opt-in config plumbing is straightforward, `/reset` clearing the SDK mapping is the right direction, and the bridge guards `onSessionId` so callback failures do not break the stream.

Jerry-Xin (approved):
Review (243c5ae4f170) — APPROVE. Clean opt-in design. /reset clears SDK mapping correctly. Multi-bot isolated by construction. Non-blocking: sdk_sessions has no TTL cleanup (follow-up), stale session ID not cleared on resume failure (follow-up).

PR #66, opened 2026-06-07
feat(config): per-group instructions via GROUP.md (v1.0)

lml2468:
First v1.0 roadmap item: GROUP.md/THREAD.md configuration.

## What changed and why

Operators can give a specific group its own persona/rules without code changes. Set `groupConfigDir` (`CC_OCTO_GROUP_CONFIG_DIR`) to a directory of `<groupId>.md` files; a matching file's contents are injected into that group's system prompt as a trusted `[Group instructions]` block.

**Design / security:**
- New `src/group-config.ts` — `loadGroupConfig(dir, groupId)`. **Operator-controlled**: the dir must live outside the agent-writable `cwdBase`, because it shapes the system prompt (system-prompt-level instructions must come from a trusted source, not the agent's sandbox).
- `groupId` is **filename-pinned to a safe slug** (rejects path separators, `.`, `..`) so a crafted channel id can't traverse out of the dir. Content capped at 16 KiB (byte-safe truncation). Never throws — a missing/unreadable file degrades to "no instructions".
- `buildSystemPrompt` gains a `[Group instructions]` section, ordered after the global custom prompt and before group context/history. Unlike group context, it is **not** sanitized — it's a trusted operator file, not chat.
- `index.ts` loads the group's config by `channelId` and threads it through `queryAgent` opts. DMs are excluded (they key on the peer uid, not a shared channel).

THREAD.md is reserved until `BotMessage` carries a thread id — the same lookup will then key on the thread.

## Files

- `src/group-config.ts` (new), `src/config.ts`, `src/agent-bridge.ts`, `src/index.ts`
- Tests: `group-config` (new), `agent-bridge` (section + ordering), `e2e` (group injects / DM does not)

## Test results

- `npm test` — **858 passed** (+20); NUL scan clean

Remaining roadmap: v1.0 webhook mode (last item).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ While the major version is `0`, minor releases may carry breaking changes.
 
 ### Added
 
+- **Per-group instructions** (v1.0, GROUP.md) — set `groupConfigDir`
+  (`CC_OCTO_GROUP_CONFIG_DIR`) to a directory of `<groupId>.md` files; a matching
+  file's contents are injected into that group's system prompt as a trusted
+  `[Group instructions]` block, so a group can have its own persona/rules without
+  code changes. Operator-controlled (must live outside the agent-writable
+  `cwdBase`); id is filename-pinned to a safe slug (no traversal), content capped
+  at 16 KiB. Groups only. New `src/group-config.ts`.
 - **Persistent sessions** (v0.3, opt-in) — with `sdk.persistentSession`
   (`CC_OCTO_SDK_PERSISTENT_SESSION=true`), agent workspace state persists across
   messages via the SDK v2 Session API. Each session's SDK session id is stored
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,14 @@ While the major version is `0`, minor releases may carry breaking changes.
   (`CC_OCTO_GROUP_CONFIG_DIR`) to a directory of `<groupId>.md` files; a matching
   file's contents are injected into that group's system prompt as a trusted
   `[Group instructions]` block, so a group can have its own persona/rules without
-  code changes. Operator-controlled (must live outside the agent-writable
-  `cwdBase`); id is filename-pinned to a safe slug (no traversal), content capped
-  at 16 KiB. Groups only. New `src/group-config.ts`.
+  code changes. Operator-controlled; id is filename-pinned to a safe slug (no
+  traversal), content capped at 16 KiB, groups only. Boot-time check (re-run
+  per-bot) rejects a `groupConfigDir` equal to or nested under the agent-writable
+  `cwdBase`, using realpath canonicalization. Because the agent can still write
+  absolute paths under default Bash/bypassPermissions, the block is a trusted
+  prompt-injection sink whose safety requires OS-level file permissions + a
+  hardened deployment — documented prominently, with a defense-in-depth refusal
+  to inject a group/world-writable file. New `src/group-config.ts`.
 - **Persistent sessions** (v0.3, opt-in) — with `sdk.persistentSession`
   (`CC_OCTO_SDK_PERSISTENT_SESSION=true`), agent workspace state persists across
   messages via the SDK v2 Session API. Each session's SDK session id is stored
--- a/README.md
+++ b/README.md
@@ -142,7 +142,9 @@ Leave the field unset to talk to Anthropic's public endpoint directly.
 Give a specific group its own persona or rules without code changes: set
 `groupConfigDir` to a directory you control, and drop a `<groupId>.md` file in
 it (the group's channel id, e.g. `s12_345.md`). Its contents are injected into
-that group's system prompt as a trusted `[Group instructions]` block.
+that group's system prompt as a trusted, **unsanitized** `[Group instructions]`
+block. Only groups use this; DMs key on the peer uid. The key is the channel id,
+so all topics under one `CommunityTopic` channel id share the same file.
 
 ```bash
 CC_OCTO_GROUP_CONFIG_DIR=/home/deploy/cc-octo-groups npm start
@@ -150,10 +152,23 @@ CC_OCTO_GROUP_CONFIG_DIR=/home/deploy/cc-octo-groups npm start
 #   Always answer in formal English and cite sources.
 ```
 
-These files are **operator-controlled** — they must live outside the per-session
-`cwdBase` sandbox (which the agent can write), since they shape the system
-prompt. Files larger than 16 KiB are truncated; an unsafe group id (path
-separators, `..`) is ignored. Only groups use this; DMs key on the peer uid.
+> ⚠️ **Security — this is a trusted, unsanitized prompt-injection sink.** Its
+> safety depends entirely on the files being writable **only** by the operator.
+> Putting `groupConfigDir` outside `cwdBase` is required (the gateway refuses
+> otherwise) but **not sufficient**: under the shipped defaults (`allowedTools:
+> "*"` + `bypassPermissions`) the agent has `Bash`/`Write` and can write
+> **absolute** paths outside `cwdBase` — `cwdBase` is a starting dir, not a
+> chroot (see [Security Model](#security-model)). A malicious user could then
+> have the agent write `<groupConfigDir>/<otherGroup>.md` and inject persistent,
+> trusted instructions into another group. So you **must**:
+> - make `groupConfigDir` and its files **non-writable by the gateway process
+>   user** (e.g. root-owned, mode `0755`/`0644`), and/or
+> - harden the deployment (drop `Bash` from `allowedTools`, run unprivileged,
+>   sandbox the filesystem).
+>
+> As defense-in-depth the gateway refuses to inject a group/world-writable file,
+> but that is a backstop, not the guarantee. Files larger than 16 KiB are
+> truncated; an unsafe group id (path separators, `..`) is ignored.
 
 ### Multi-bot
 
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ Three-level priority: **environment variables** > **config.json** > **defaults**
 | `sdk.settingSources` | `CC_OCTO_SDK_SETTING_SOURCES` | `user` | Comma-separated setting sources (e.g. `user,project`) |
 | `sdk.toolProgress` | `CC_OCTO_SDK_TOOL_PROGRESS` | `false` | When true, post `🔧 Running <tool>…` notices as the agent invokes tools (deduped, capped per turn) |
 | `sdk.persistentSession` | `CC_OCTO_SDK_PERSISTENT_SESSION` | `false` | When true, persist agent workspace state across messages via the SDK v2 Session API (resume by stored session id). `/reset` clears it. |
+| `groupConfigDir` | `CC_OCTO_GROUP_CONFIG_DIR` | *(unset)* | Directory of per-group instruction files (`<groupId>.md`). A match is injected into the system prompt as trusted custom instructions for that group. See [Per-group instructions](#per-group-instructions). |
 | `sdk.anthropicBaseUrl` | `ANTHROPIC_BASE_URL` | *(unset)* | Override the upstream Claude API endpoint. See [Self-hosted gateway](#self-hosted-gateway) below. |
 | `rateLimit.maxPerMinute` | `CC_OCTO_RATE_LIMIT_MAX_PER_MINUTE` | `5` | Max requests per minute per session |
 | `context.maxContextChars` | `CC_OCTO_CONTEXT_MAX_CHARS` | `6000` | Max characters of group context injected into prompts |
@@ -135,6 +136,24 @@ ANTHROPIC_BASE_URL=https://claude-gw.internal.example.com npm start
 ```
 
 Leave the field unset to talk to Anthropic's public endpoint directly.
+
+### Per-group instructions
+
+Give a specific group its own persona or rules without code changes: set
+`groupConfigDir` to a directory you control, and drop a `<groupId>.md` file in
+it (the group's channel id, e.g. `s12_345.md`). Its contents are injected into
+that group's system prompt as a trusted `[Group instructions]` block.
+
+```bash
+CC_OCTO_GROUP_CONFIG_DIR=/home/deploy/cc-octo-groups npm start
+# /home/deploy/cc-octo-groups/s12_345.md:
+#   Always answer in formal English and cite sources.
+```
+
+These files are **operator-controlled** — they must live outside the per-session
+`cwdBase` sandbox (which the agent can write), since they shape the system
+prompt. Files larger than 16 KiB are truncated; an unsafe group id (path
+separators, `..`) is ignored. Only groups use this; DMs key on the peer uid.
 
 ### Multi-bot
 
@@ -322,7 +341,7 @@ src/
 | **v0.1** | Text messaging, streaming, session persistence, rate limiting, security model |
 | **v0.2** *(current)* | Media reception & sending (image/file/RichText), @mention, group context, per-session `cwdBase` isolation, self-hosted gateway, SSRF/prompt-injection hardening |
 | **v0.3** | Slash commands, tool progress, multi-bot, v2 Session API |
-| **v1.0** | GROUP.md/THREAD.md configuration, webhook mode |
+| **v1.0** | webhook mode |
 
 ## Contributing
 
--- a/config.example.json
+++ b/config.example.json
@@ -6,6 +6,7 @@
   "_comment_cwd_base": "Q3: base directory for per-session cwd isolation. Each session gets a 16-hex SHA-256 subdirectory under this path; subdirs idle for >7d are auto-cleaned every 6h. Env override: CC_OCTO_CWDBASE (legacy CC_OCTO_CWD still accepted with a deprecation warning).",
   "cwdBase": "/path/to/isolated/sandbox-root",
   "dataDir": "./data",
+  "_comment_group_config_dir": "v1.0: optional directory of per-group instruction files (<groupId>.md). A matching file's contents are injected into that group's system prompt as trusted custom instructions. Operator-controlled — keep it OUTSIDE cwdBase (the agent can write there). Env override: CC_OCTO_GROUP_CONFIG_DIR. Omit to disable.",
 
   "sdk": {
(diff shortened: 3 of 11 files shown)


Jerry-Xin (changes requested):
The PR is relevant to `cc-channel-octo`, but it leaves a security invariant unenforced.

🔴 Blocking

🔴 Critical — `groupConfigDir` is documented as trusted operator-controlled prompt input, but `loadConfig()` accepts it even when it is equal to or inside `cwdBase`. That directly violates the PR’s stated trust boundary: files under `cwdBase` are agent-writable/untrusted, while `GROUP.md` contents are injected unsanitized into the system prompt. A misconfiguration like `cwdBase=/srv/octo` and `groupConfigDir=/srv/octo/groups` would let a user-driven agent persist future system-prompt instructions. Add a startup validation in `src/config.ts` after env/file merge, using resolved paths, that rejects `groupConfigDir` when it is the same as or nested under `cwdBase`; include tests for file config and env config. References: [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_dh_ct6l/repo/src/config.ts:397), [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_dh_ct6l/repo/src/config.ts:311), [src/group-config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_dh_ct6l/repo/src/group-config.ts:55), [src/agent-bridge.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_dh_ct6l/repo/src/agent-bridge.ts:126).

💬 Non-blocking

🟡 Warning — `loadGroupConfig()` reads the entire file before applying the 16 KiB cap, so a mistakenly huge file can block the event loop and allocate unnecessary memory on every handled group message. Since this is operator-controlled it is not a blocker, but using `openSync/readSync` for `MAX_GROUP_CONFIG_BYTES + 1` bytes would make the cap real. Reference: [src/group-config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-_dh_ct6l/repo/src/group-config.ts:55).

✅ Highlights

The filename slug check prevents traversal through `groupId`, DM exclusion is covered, and the prompt ordering tests cover the intended `[Group instructions]` placement.

Verification note: I could not run the focused test command because this checkout has no installed dependencies; `vitest` was not found.

lml2468:
Thanks @Jerry-Xin — both addressed in 4508a0d.

**groupConfigDir trust boundary (blocking)** — now enforced at boot. `loadConfig` rejects `groupConfigDir` when it equals or is nested under `cwdBase`, using **resolved** paths (so `./..` can't sneak a nested dir past the check) and a path-separator boundary (so a sibling like `/srv/octo-groups` is not falsely flagged as inside `/srv/octo`). Applies to both file- and env-provided values. Tests cover: outside ok, equal rejected, nested rejected, `..`-nested rejected, env-nested rejected, sibling-prefix allowed.

**streaming read cap (non-blocking)** — `loadGroupConfig` now reads at most `MAX+1` bytes via `openSync`/`readSync` instead of slurping the whole file, so a mistakenly huge operator file can't block the event loop or over-allocate on every group turn.

864 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (changes requested):
Summary: The PR is relevant and the main feature flow is well covered, but the trust-boundary validation can be bypassed in multi-bot configs.

**🔴 Blocking**
- 🔴 Critical: `groupConfigDir` is validated before per-bot `cwdBase` overrides are applied, so a bot can inherit a trusted instruction directory that is actually inside that bot’s agent-writable sandbox. `loadConfig()` checks only the top-level paths in `src/config.ts:436`, but `resolveBotConfigs()` later replaces `cwdBase` from `bot.cwdBase` in `src/config.ts:509`. Example: top-level `cwdBase=/srv/sandboxes`, `groupConfigDir=/srv/octo/groups`, bot override `cwdBase=/srv/octo` passes startup validation, then that bot loads unsanitized system-prompt instructions from under its own writable `cwdBase`. Re-run the same safety check after each concrete bot config is resolved, or validate inside `resolveBotConfigs()` after applying overrides.

**💬 Non-blocking**
- 🟡 Warning: Path validation is lexical only (`resolvePath`) and does not account for symlinks in `src/config.ts:437-438`. If either configured path involves symlinks, the “outside cwdBase” guarantee may not match the real filesystem. Consider `realpathSync.native` for existing paths, with a clear policy for missing directories.

**✅ Highlights**
- The group-id filename pinning in `src/group-config.ts:28` rejects traversal-shaped IDs and keeps lookup scoped.
- `loadGroupConfig()` bounds reads to `MAX_GROUP_CONFIG_BYTES + 1`, avoiding unbounded prompt/file reads.
- The prompt ordering tests cover the important placement: global security/custom prompt, group instructions, group context, then history.

Verification note: I attempted the targeted test command, but this checkout has no installed dev dependencies (`vitest: command not found`), so I could not run the test suite locally.

Jerry-Xin (approved):
Re-review v2 (4508a0da4ed0) — APPROVE

v1 blockers both fixed:
1. groupConfigDir vs cwdBase trust boundary — now enforced at startup via isPathInside() with 6 test cases including sibling prefix guard
2. Bounded read — openSync+readSync MAX+1 bytes, no more unbounded readFileSync

Remaining nit (follow-up): per-bot cwdBase override in resolveBotConfigs runs after loadConfig validation — a multi-bot config with cwdBase override could theoretically bypass the check. Requires: multi-bot mode + operator setting per-bot cwdBase to a path containing groupConfigDir. Fix: re-run trust-boundary check in resolveBotConfigs after applying overrides. Low probability, follow-up issue.

Approve.

Jerry-Xin (changes requested):
Withdrawing APPROVE. Per-bot cwdBase override in resolveBotConfigs can bypass the trust-boundary check added in v2. This is PR #64 delivered feature, not an edge case. Fix: re-run isPathInside validation after each per-bot config is resolved. v1.0 trust boundary should be complete, not partial.

lml2468:
Thanks @Jerry-Xin — both addressed in 10ca31e.

**per-bot bypass (blocking)** — the boundary check is now factored into `assertGroupConfigDirOutsideCwd()` and runs **twice**: once on the top-level config in `loadConfig`, and again for **each resolved bot config** in `resolveBotConfigs()` (after the per-bot `cwdBase` override is applied). So a bot that overrides `cwdBase` to a dir containing `groupConfigDir` is now rejected. Test added: top-level passes but `bots: [{ cwdBase: '/srv/octo' }]` with `groupConfigDir: '/srv/octo/groups'` throws; a clear per-bot cwdBase is allowed.

**symlink-aware check (non-blocking)** — paths are now canonicalized via `realpathSync.native` when they exist (defeats symlink dodges), falling back to lexical `resolve()` for not-yet-created directories.

866 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (commented):
Summary: The PR is in scope and the implementation cleanly adds trusted per-group prompt instructions with appropriate routing, docs, and tests.

💬 Non-blocking

🟡 Warning — `src/config.ts:431-434`: `loadConfig()` validates `groupConfigDir` against the top-level `cwdBase` even in multi-bot mode. If every bot overrides `cwdBase`, that top-level path may be unused, so this can reject an otherwise safe configuration. Consider validating only the resolved per-bot configs in multi-bot mode.

🔵 Suggestion — `src/group-config.ts:51-58`: the `existsSync`/`statSync`/`openSync` sequence is fine for an operator-controlled directory, but opening first and checking the opened fd would avoid a small TOCTOU window if the directory permissions are ever loosened.

🔵 Suggestion — `src/config.ts:462-474`: since the comment explicitly calls out symlink protection, a symlink boundary test would be useful coverage.

✅ Highlights

- `src/group-config.ts:28-77` keeps lookup filename-pinned and bounded, and degrades safely on read errors.
- `src/agent-bridge.ts:122-145` places group instructions before untrusted group context/history, preserving the intended trust hierarchy.
- `src/index.ts:505-513` correctly limits injection to group-like channels and keeps DMs excluded.
- `src/config.ts:448-549` re-checks the trust boundary after per-bot `cwdBase` resolution.

Verification note: I attempted the targeted test run, but this checkout does not have `vitest` installed (`sh: vitest: command not found`). `git diff --check` passed.

yujiawei (changes requested):
# Code Review — PR #66 (cc-channel-octo)

## Summary

This PR adds an opt-in per-group instruction mechanism: an operator points `groupConfigDir` (`CC_OCTO_GROUP_CONFIG_DIR`) at a directory of `<groupId>.md` files, and a matching file's contents are injected into that group's system prompt as a trusted `[Group instructions]` block (unsanitized, marked as operator-trusted). DMs are excluded; ids are filename-pinned to a safe slug; content is capped at 16 KiB.

The implementation is clean and the test coverage is genuinely good. My one blocking concern is about the **completeness of the trust boundary** that this security-sensitive feature relies on, and the docs/comments that currently overstate that guarantee.

## Verification

- ✅ `tsc --noEmit` passes; `vitest` run of the touched suites (group-config, config, agent-bridge, e2e) is green — 180/180.
- ✅ Path-traversal protection on the group id is sound: `isSafeId` (`src/group-config.ts:28`) restricts to `[a-zA-Z0-9._-]+` and rejects `.`/`..`; tests cover `../etc`, `a/b`, `a\b`, `.`, `..`, empty, and a directory-named-like-the-file case.
- ✅ The cwdBase boundary check is implemented carefully: `canonicalize` uses `realpathSync.native` so symlinks can't dodge it and falls back to lexical `resolve()` for not-yet-created dirs (`src/config.ts:463`); `isPathInside` appends a trailing separator so a sibling like `/srv/octo-groups` is not treated as inside `/srv/octo` (`src/config.ts:472`, covered by test).
- ✅ The boundary is re-checked per bot in `resolveBotConfigs()` after per-bot `cwdBase` overrides are applied (`src/config.ts:549`), closing the gap where a per-bot `cwdBase` could swallow `groupConfigDir`.
- ✅ DM exclusion is correct: injection is gated on `isGroup` (`src/index.ts:508`), and the e2e test confirms a file named after a DM peer uid is not picked up.
- ✅ Robustness: read is bounded to `MAX+1` bytes with a truncation marker, trailing replacement chars are trimmed at the UTF-8 boundary, and all failures degrade to "no instructions" rather than throwing the turn.

## Findings

### P1 — The trust boundary that this feature depends on is incomplete, and the docs assert it as absolute

The injected `[Group instructions]` block is **trusted and unsanitized** by design (`src/agent-bridge.ts:126-131`). The entire safety argument rests on the source files being writable only by the operator. The code states this as a guarantee:

- `src/group-config.ts:9` — `SECURITY: these files are OPERATOR-controlled, NOT agent- or user-writable.`
- `README.md:153-155` — *"These files are **operator-controlled** — they must live outside the per-session `cwdBase` sandbox (which the agent can write)…"*

The only enforcement is the boot-time check that `groupConfigDir` is not equal to / nested under `cwdBase`. But this project's own Security Model section is explicit that **`cwdBase` is not a chroot**: with `Bash`/`Read` in the tool set and `bypassPermissions` — both of which are the shipped defaults (`allowedTools: '*'`, `permissionMode: 'bypassPermissions'`, `src/config.ts:169-170`) — the agent can write **absolute** paths anywhere the gateway process user can write (README.md:215-222). Placing `groupConfigDir` outside `cwdBase` therefore does **not** make it non-writable by the agent.

Concrete residual attack, assuming an operator enables the feature on a non-hardened (default) deployment:

1. Attacker is a member of group **G1**. Via prompt injection they get the agent to run a Bash command writing to the absolute path `<groupConfigDir>/G2.md` (G2 = a different, higher-value group).
2. The absolute path is trivially discoverable: `CC_OCTO_GROUP_CONFIG_DIR` is present in the SDK subprocess environment (the subprocess inherits `process.env`; see `src/agent-bridge.ts:318-320`), so the agent can read it directly.
3. From then on, every member of **G2** has the attacker's content injected as a **trusted, unsanitized** `[Group instructions]` block.

This is a step up from the existing documented "agent can read arbitrary files" limitation, because the new sink is (a) **persistent**, (b) **cross-group / cross-user**, and (c) **trusted** (explicitly not sanitized like group context). The `cwdBase` boundary check does not address it at all — it only prevents `groupConfigDir` from being *inside* `cwdBase`.

I don't think the code can fully close this at the config layer — the real protection is OS-level file permissions plus a hardened deployment (unprivileged user, `Bash` dropped or a sandboxed FS). What I'm asking for before merge is to **stop asserting a guarantee the code doesn't provide** on a security-sensitive feature, and to tell operators what actually makes it safe:

- Reword `src/group-config.ts:9-13` and `README.md:153-155` so they don't claim the files are "NOT agent-writable" purely by virtue of being outside `cwdBase`. State plainly that under the default `Bash` + `bypassPermissions` configuration the agent can write outside `cwdBase`, so `groupConfigDir` **must be made non-writable by the gateway process user** (file/dir permissions), and that the feature's trust assumption only holds on a hardened deployment.
- Add this caveat to the "Per-group instructions" README section itself (not only by implication from the separate Security Model section), since that is where an operator decides where to put the directory.

Optionally (not required for merge), consider a runtime defense-in-depth: refuse to inject a file that is group/world-writable, or that is owned by a different uid than the gateway process — cheap to check at load time and would catch the most common misconfiguration.

### P2 — `groupConfigDir` path is exposed to the agent subprocess

`src/agent-bridge.ts:318-320` spreads the full `process.env` into the SDK subprocess (necessary to preserve `PATH`/`HOME`/API key), which also exposes `CC_OCTO_GROUP_CONFIG_DIR`. On its own this is minor, but it removes the only "obscurity" obstacle to the P1 attack (the agent doesn't have to guess the directory). Worth a sentence in the security notes; not worth special-casing the env.

### P2 — TOCTOU between `existsSync`/`statSync`/`openSync`

`src/group-config.ts:53-58` stats then opens in separate calls. For an operator-controlled directory this is not meaningfully exploitable, and the symlink/`isFile()` checks are reasonable; noting only for completeness. No change required.

### P2 — CommunityTopic groups share one file per channel id

Injection is keyed on `channelId` (`src/index.ts:509`). For `ChannelType.CommunityTopic`, multiple topics under the same channel id would all receive the same `<channelId>.md`. That may be intended, but the README frames the key as "the group's channel id, e.g. `s12_345`," which a reader could mistake for per-topic granularity. A one-line clarification would help.

## What's good

- Path-traversal hardening, symlink-aware boundary canonicalization, and the sibling-prefix edge case are all handled correctly and tested.
- The per-bot re-check after `cwdBase` override is exactly the kind of follow-through this class of bug needs.
- Bounded reads + truncation marker + UTF-8-boundary trim + never-throw degradation are solid.
- Test suite is thorough across config validation, the loader, prompt ordering, and e2e injection/non-injection (incl. the DM negative case).

## Verdict

Requesting changes: the feature introduces a new **trusted, unsanitized** prompt-injection sink whose safety rests on a boundary the codebase itself documents as not enforced under the default configuration, while the comments/README present it as a hard guarantee. The fix is primarily documentation + an explicit OS-permission requirement (and optionally a load-time writability check). Once the trust assumption is stated accurately and the operator guidance is corrected, this is good to merge.

lml2468:
Thanks @yujiawei — this is a sharp catch, and you're right that the boundary the docs asserted isn't one the code can provide. Fixed in 359c196.

**P1 — trust boundary overstated.** The `[Group instructions]` block is a trusted, unsanitized sink, and since the agent can write absolute paths outside `cwdBase` under the default `Bash`+`bypassPermissions`, "outside cwdBase" is necessary but not sufficient. I've:
- Reworded the `group-config.ts` header and the README **Per-group instructions** section to stop claiming the files are non-agent-writable by virtue of being outside `cwdBase`. They now state plainly that the agent can write outside `cwdBase` under the defaults, so `groupConfigDir` + its files **must be made non-writable by the gateway process user** (OS perms) and/or the deployment hardened (drop `Bash`, unprivileged, sandboxed FS) — and the caveat is right where the operator picks the directory, as a `⚠️` callout.
- Added the **defense-in-depth** you suggested: `loadGroupConfig` refuses to inject a group/world-writable file (`mode & 0o022`), logged, explicitly a backstop not the guarantee. Tests: world-writable refused, group-writable refused, `0644` accepted (POSIX-only).

**P2 — `CC_OCTO_GROUP_CONFIG_DIR` visible to the subprocess.** Noted in the security text (it removes the obscurity obstacle); not special-casing the env, since `process.env` must pass through.

**P2 — CommunityTopic per-channel-id key.** Documented: topics under one `CommunityTopic` channel id share the same `<channelId>.md` (the key is the channel id, not per-topic).

**P2 — TOCTOU / no change required.** Agreed, left as-is for the operator-controlled dir.

869 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (approved):
The PR is in scope for `cc-channel-octo` and cleanly extends the project’s config and prompt assembly path for per-group instructions.

💬 Non-blocking

🟡 Warning: [src/group-config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-4oozgn_m/repo/src/group-config.ts:75) has a misleading remediation message: “Make it writable only by the gateway user.” The README says the opposite security requirement: files should be non-writable by the gateway process user. This should be reworded to avoid operators applying an unsafe ownership/mode setup.

🔵 Suggestion: [src/group-config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-4oozgn_m/repo/src/group-config.ts:65) uses `statSync` followed by `openSync`, which follows symlinks and has a small TOCTOU window. Given the directory is documented as operator-controlled, this is not blocking, but `open` + `fstat` and/or `O_NOFOLLOW` where available would make the “filename-pinned” guarantee more literal.

✅ Highlights

The relevance gate passes: this PR modifies real project functionality in config loading, message handling, prompt construction, tests, and docs.

The group-only routing in [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-4oozgn_m/repo/src/index.ts:508) correctly avoids DM peer-id lookup.

The prompt ordering in [src/agent-bridge.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-4oozgn_m/repo/src/agent-bridge.ts:126) matches the described behavior: global custom prompt first, then trusted group instructions, then untrusted context/history.

The config boundary is checked both at top-level load and after per-bot overrides in [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-4oozgn_m/repo/src/config.ts:434) and [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-4oozgn_m/repo/src/config.ts:549).

I could not run the targeted test suite because dependencies are not installed in this checkout: `vitest: command not found`.

PR #67, opened 2026-06-07
feat(transport): webhook inbound mode (v1.0)

lml2468:
Final v1.0 roadmap item: webhook mode. **This completes the v0.3 + v1.0 roadmap.**

## What changed and why

An HTTP inbound transport as an alternative to the WuKongIM WebSocket — useful behind a reverse proxy or where outbound long connections aren't possible. Set `transport: "webhook"` (`CC_OCTO_TRANSPORT=webhook`). The bot still registers over REST (id + outbound sends); only the inbound path changes.

**Design / security:**
- New `src/webhook.ts` (`WebhookServer`, `node:http` — no new dependency). Per request: shared-secret auth (`x-webhook-secret` header or `?secret=`, **constant-time** via `timingSafeEqual`), 256 KiB body cap (socket destroyed past it), parses the Octo message JSON (top-level or under `message`/`data`) into a `BotMessage`, and feeds the **same** handler as the WS path. Returns 404 (wrong path/method), 401 (bad secret), 400 (malformed), 413 (oversized).
- `webhook.secret` is **required** in webhook mode — `loadConfig` throws without it, since an unauthenticated endpoint would let anyone inject messages. Binds `127.0.0.1` by default (put a reverse proxy / TLS in front).
- `startBot` branches on `transport`: webhook mode starts the server in the `connect()` phase (after sibling cross-registration) and closes it on shutdown — reusing the same two-phase lifecycle as the WS path.

## Files

- `src/webhook.ts` (new), `src/config.ts` (transport + webhook block + validation), `src/index.ts` (transport branch)
- Tests: `webhook` (new — parse + real HTTP round-trips), `config` (transport/secret matrix + env)

## Test results

- `npm test` — **887 passed** (+29); NUL scan clean
- Dogfood: webhook-without-secret fails fast with a clear error; with-secret boots to the REST boundary.

## Roadmap

With this merged, every v0.3 and v1.0 roadmap item is implemented: slash commands, tool progress, multi-bot, v2 Session API, GROUP.md, and webhook mode.


Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository, but webhook mode has blocking lifecycle bugs that can leave the process “ready” without a listening webhook server and skips existing runtime maintenance/shutdown behavior.

🔴 Blocking

🔴 Critical — Webhook listen failures are ignored, so startup can falsely succeed with no inbound transport.  
In [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jcgi684c/repo/src/index.ts:189), `connect()` calls `void webhookServer.listen()` and immediately logs the bot as ready. If the port is already in use or binding fails, the promise rejects outside startup flow; the global unhandled rejection logger may print something, but `main()` still prints ready and keeps the process running with no webhook endpoint. Make `BotStack.connect` async, await webhook `listen()`, and let startup fail/cleanup on bind errors.

🔴 Critical — Single-bot webhook mode does not install the existing graceful shutdown handlers.  
The single-bot signal handlers are installed only from `OctoGateway.connect()` in [src/gateway.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jcgi684c/repo/src/gateway.ts:98), but webhook mode bypasses that path in [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jcgi684c/repo/src/index.ts:190). Since `main()` only installs its own signal handlers for multi-bot mode, single-bot webhook mode exits on SIGINT/SIGTERM without draining active handlers, closing the store, or running the configured shutdown callback. The webhook path needs equivalent signal wiring or a gateway lifecycle method that starts non-WS services without skipping shutdown setup.

🔴 Critical — Webhook mode skips the REST heartbeat/token-refresh lifecycle.  
`OctoGateway.register()` only registers and stores credentials; the 30s heartbeat starts in `OctoGateway.connect()` at [src/gateway.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jcgi684c/repo/src/gateway.ts:98), via [src/gateway.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jcgi684c/repo/src/gateway.ts:283). Webhook mode calls `register()` but never `connect()`, so it loses heartbeat failure detection and token refresh even though outbound sends and bot registration still depend on REST state. If webhook mode is intended to change only inbound delivery, this lifecycle should be preserved without opening the WebSocket.

💬 Non-blocking

🟡 Warning — `parseWebhookBody()` only validates a small subset of `BotMessage` shape in [src/webhook.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jcgi684c/repo/src/webhook.ts:52). That is probably acceptable for a first pass, but consider validating `payload.type`, `message_seq`, and `timestamp` to avoid feeding malformed messages into downstream assumptions.

✅ Highlights

The shared secret requirement, default loopback bind, body cap, and tests around auth/body parsing are good security defaults for an inbound HTTP surface.

I could not run the tests or type-check locally because dependencies are not installed in this checkout: `vitest` and `tsc` were not found.

yujiawei (changes requested):
# Code Review — PR #67 (cc-channel-octo)

## Summary

This adds an HTTP webhook inbound transport as an alternative to the WuKongIM WebSocket. The overall shape is sound: a shared secret is **required** in webhook mode (enforced in both `loadConfig` and a defensive guard in `startBot`), the bind host defaults to `127.0.0.1`, bodies are capped at 256 KiB, and inbound messages reuse the *same* `onInbound` handler — so the shared `SessionRouter` self/bot-loop guard (`from_uid === robotId`, `looksLikeBot`, blocklist, rate limits) still applies to webhook traffic. Tests cover auth, size cap, malformed body, and real HTTP round-trips.

However, there are two correctness bugs that should be fixed before merge, plus a few security/robustness items worth a human's attention given the `needs-human-review` classification.

## Blocking issues

### P1 — Multi-byte UTF-8 can be corrupted across chunk boundaries

`src/webhook.ts:140-148`

```ts
req.on('data', (chunk: Buffer) => {
  if (tooLarge) return;
  body += chunk.toString('utf-8');   // <-- per-chunk decode
  ...
});
```

Decoding each TCP chunk independently with `chunk.toString('utf-8')` and concatenating the strings will mangle any multi-byte UTF-8 codepoint that happens to straddle a chunk boundary — the trailing bytes of one chunk and the leading bytes of the next each decode to U+FFFD replacement characters. This codebase's primary content is Chinese (3-byte sequences), and bodies can be up to 256 KiB, so multi-chunk requests are realistic. The result is silently garbled inbound text (best case) or a `JSON.parse` failure → 400 (when the corruption lands inside JSON syntax).

The existing tests don't catch this because they send small, ASCII or single-chunk bodies that arrive in one `data` event.

**Fix:** buffer the raw `Buffer`s and decode once at the end:

```ts
const chunks: Buffer[] = [];
let size = 0;
req.on('data', (chunk: Buffer) => {
  if (tooLarge) return;
  size += chunk.length;
  if (size > MAX_WEBHOOK_BODY_BYTES) { tooLarge = true; res.writeHead(413).end(); req.destroy(); return; }
  chunks.push(chunk);
});
req.on('end', () => {
  if (tooLarge) return;
  const body = Buffer.concat(chunks).toString('utf-8');
  ...
});
```

This also fixes a secondary inefficiency: the current code calls `Buffer.byteLength(body, 'utf-8')` on the *whole accumulated string* on every chunk, which is O(n²) for large bodies.

### P1 — Webhook bind failure is silently swallowed

`src/index.ts:189-193`

```ts
if (useWebhook && webhookServer) {
  void webhookServer.listen();
  console.log(`[cc-channel-octo] ${label}Bot ready (webhook): id=${gateway.botId}`);
}
```

`listen()` returns a Promise that rejects on bind errors (`EADDRINUSE`, `EACCES`, etc.), but it is invoked fire-and-forget with `void` and never awaited. On failure the rejection only reaches the global `unhandledRejection` logger (`src/index.ts:34`), which just prints a line — the process keeps running, logs `Bot ready (webhook)` and `Ready — listening for messages`, and then silently receives **zero** inbound traffic. The operator sees a "healthy" boot with a dead transport.

The WS path doesn't have this problem because `gateway.connect()` surfaces connection state through reconnect/error callbacks. The webhook path should propagate the bind failure — e.g. make `connect()` async and `await webhookServer.listen()` (let it crash the boot on failure), or attach a `.catch` that exits the process. A startup failure should be loud, not a logged line on an otherwise-"ready" process.

## Non-blocking findings

### P2 — `parseWebhookBody` does not validate `message_seq` / `timestamp`

`src/webhook.ts:63-70`

`BotMessage` declares `message_seq: number` and `timestamp: number` as required, but the parser only checks `message_id`, `from_uid`, and `payload`, then casts the partial object to `BotMessage`. A caller that omits `message_seq` produces `undefined` flowing into `store.appendUser(..., msg.message_seq)` and `store.setLastBotReplySeq(sessionKey, msg.message_seq)` (`src/index.ts:472,572`) — the latter typed `seq: number`. It degrades gracefully (NULL seq → flat history segmentation, `timestamp` → group-context ordering oddities) rather than crashing, so it's not blocking, but for a transport that accepts attacker-shaped JSON I'd validate `message_seq`/`timestamp` are numbers and reject (400) otherwise, rather than casting an under-validated object.

### P2 (security, for human reviewer) — secret accepted via `?secret=` query string

`src/webhook.ts:130-132`, README "Webhook mode"

Accepting the shared secret as a URL query parameter is convenient but it leaks into access logs — and the README explicitly recommends putting a reverse proxy in front, which is exactly the component that logs full request URLs by default. Consider preferring the `x-webhook-secret` header and either dropping the query option or documenting the logging risk prominently. (The header path is correct.)

### P2 (security, for human reviewer) — `from_uid` is fully caller-controlled; secret authenticates the caller, not the user

The shared secret authenticates whoever POSTs to the endpoint; the `from_uid` in the body is then trusted verbatim for session partitioning and rate-limit keying (`SessionRouter`). Anyone holding the secret can impersonate any `from_uid`, including the owner uid that `register()` stores for the "future permission model" (`ownerUid`, G18). No live authorization currently keys off `ownerUid`, so there's no exploitable privilege escalation today — but this is worth flagging explicitly given the `security_sensitive` classification: the day an owner-only capability is added, the webhook trust model needs to be revisited. Worth a comment in the code/README documenting that webhook callers are fully trusted.

### Nit — constant-time compare comment vs. behavior

`src/webhook.ts:38-44` The comment says the compare "won't leak length via early return," but `if (a.length !== b.length) return false` does early-return on a length mismatch (necessary, since `timingSafeEqual` throws on unequal lengths). The secret *length* is therefore observable via timing. This is standard and generally acceptable, but the comment overstates the guarantee — consider rewording.

### Nit — no replay protection

There's no nonce/timestamp-window/idempotency check, so a captured valid request (with secret) can be replayed. Acceptable for v1.0 behind TLS + secret, but worth noting as a known limitation.

## Test coverage

Good coverage of the happy path, auth (header/query/missing/wrong), 404, 400, and 413. Missing cases that map to the bugs above: a multi-chunk body with a multi-byte character split across the boundary, and a `listen()` bind-failure path. Adding the former would have caught the P1 UTF-8 issue.

## Verdict

Requesting changes for the two P1 correctness bugs (UTF-8 chunk corruption and silently-swallowed bind failure) — both are small, well-understood fixes. The security-sensitive items (query-string secret, caller-trusted `from_uid`) are non-blocking but should get a human's eyes before this is exposed.

lml2468:
Thanks @Jerry-Xin — all three blocking lifecycle bugs fixed in c40135c. The root cause was webhook mode skipping `gateway.connect()` and with it three things that don't actually depend on the WebSocket.

**listen() failure ignored (blocking)** — `connect()` is now `async` and **awaited**: `BotStack.connect` returns a promise and `main()` does `await Promise.all(stacks.map(s => s.connect()))`. A webhook bind error now rejects and fails startup instead of logging "ready" with no endpoint.

**no graceful shutdown / no heartbeat (blocking ×2)** — extracted both into a new `gateway.startServices()` (starts the heartbeat/token-refresh loop **and** the SIGINT/SIGTERM handlers, opens no socket). `connect()` calls it for the WS path; webhook mode calls it directly. So webhook mode now keeps the full REST lifecycle (heartbeat, token refresh, drain-on-signal, store close) — it just doesn't open the WS.

**stricter parse (non-blocking)** — `parseWebhookBody` now also requires numeric `message_seq` + `timestamp` and an object `payload` with a numeric `type`, rejecting malformed messages before they reach the pipeline.

Tests: +3 (`startServices` runs heartbeat without a socket; `startServices` before `register` throws; stricter-parse rejections). 890 pass; type-check / lint / build clean; NUL scan clean.

Jerry-Xin (changes requested):
Summary: The PR is relevant to `cc-channel-octo`, but webhook mode has a blocking multi-bot integration bug before merge.

🔴 Blocking

🔴 Critical — Multi-bot webhook mode cannot start more than one bot with the current config model.
`BotOverride` does not allow per-bot `transport` / `webhook` overrides in [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ckxx77yt/repo/src/config.ts:143), and `resolveBotConfigs()` copies the same top-level webhook config into every resolved bot in [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ckxx77yt/repo/src/config.ts:578). Then `main()` starts every transport concurrently in [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ckxx77yt/repo/src/index.ts:74), so any `bots[]` config with `transport: "webhook"` makes every bot bind the same host/port/path and the second server fails with `EADDRINUSE`. There is no supported way for operators to assign distinct webhook ports/paths/secrets per bot. Add webhook fields to `BotOverride` and merge/validate them per bot, or explicitly reject multi-bot + webhook with a clear config error.

💬 Non-blocking

🟡 Warning — The “constant-time” secret comparison still returns early on length mismatch in [src/webhook.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ckxx77yt/repo/src/webhook.ts:38). That leaks expected secret length through timing. This is usually low impact for a random webhook secret, but the docs and comments promise constant-time comparison. Hash both values to a fixed length before `timingSafeEqual()` if the promise matters.

🟡 Warning — If webhook bind fails after `gateway.startServices()` in [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-ckxx77yt/repo/src/index.ts:197), startup depends on the top-level fatal exit for cleanup. In multi-bot partial-start cases, already-bound webhook servers are not explicitly closed before the fatal path. A `try/catch` around transport startup that calls `shutdown()` on successfully created stacks would make startup failure handling much cleaner.

✅ Highlights

The PR keeps webhook inbound messages on the existing routing/handler pipeline, requires authentication in webhook mode, defaults to localhost binding, and adds useful HTTP round-trip tests for auth, parsing, malformed payloads, and body-size rejection.

Verification note: I attempted the targeted test command, but this checkout does not have dependencies installed: `vitest: command not found`.

yujiawei (changes requested):
# Code Review — PR #67 (cc-channel-octo)

**Verdict: CHANGES_REQUESTED** — one functional data-integrity bug (P1) plus a few security/robustness items worth addressing before this ships. The overall design is sound: REST registration is unchanged, the WS path is untouched, the same `onInbound` handler is reused, lifecycle (`startServices()` without a socket, awaited `listen()`, shutdown ordering) is correct, and the self-message filter is preserved downstream. Nice, minimal-surface implementation.

I verified the build locally: `type-check`, `lint`, and the full suite all pass (878/878 in the non-webhook files; the webhook file passed except for one environment-specific port collision — see N1). The one blocker below is a runtime correctness bug the green suite does not exercise.

---

## P1 — Body assembly corrupts multi-byte UTF-8 split across chunks (blocking)

`src/webhook.ts:147-149`

```ts
req.on('data', (chunk: Buffer) => {
  if (tooLarge) return;
  body += chunk.toString('utf-8');   // <-- decodes each chunk independently
  ...
```

The body is built by decoding **each chunk independently** and concatenating the resulting strings. When a multi-byte UTF-8 character straddles a chunk boundary, `chunk.toString('utf-8')` emits a `U+FFFD` replacement character for the incomplete sequence at each side of the split. This is the classic Node.js streaming gotcha.

This matters here specifically because the bot is CJK-heavy: every Chinese character is 3 bytes, so any body large enough to span more than one stream chunk has a high chance of containing a split. The cap is 256 KiB, so this is well within normal message sizes.

Critically, **the corruption is silent**: the mangled bytes still produce syntactically valid JSON (`{"content":"��好世界"}`), so `JSON.parse` succeeds, `parseWebhookBody`'s structural validation passes (it only type-checks fields, not content integrity), and the corrupted `BotMessage` is fed straight into the handler at line 166 — no 400, no log. The user's text is corrupted in the agent input and persisted into history.

Reproduced locally:

```
PR pattern result: "{\"content\":\"��好世界\"}"   // 你 split across chunks -> 2x U+FFFD
Parse ok? yes                                                // silently accepted
Correct result:   "{\"content\":\"你好世界\"}"
```

**Fix:** accumulate raw `Buffer`s and decode once at `end`, or use `string_decoder.StringDecoder`:

```ts
const chunks: Buffer[] = [];
let size = 0;
req.on('data', (chunk: Buffer) => {
  if (tooLarge) return;
  size += chunk.length;
  if (size > MAX_WEBHOOK_BODY_BYTES) { tooLarge = true; res.writeHead(413).end(); req.destroy(); return; }
  chunks.push(chunk);
});
req.on('end', () => {
  if (tooLarge) return;
  const body = Buffer.concat(chunks).toString('utf-8'); // decode once, boundary-safe
  ...
});
```

Please also add a test that feeds a CJK body in small (e.g. 1-byte) chunks and asserts the parsed content is byte-for-byte intact — the current round-trip tests send the whole body in one shot via `fetch`, so they cannot catch this.

---

## P2 — Security & robustness (not individually blocking, but please address)

**S1. `?secret=` query param leaks the credential into logs.** `src/webhook.ts:138-139` accepts the secret from either the `x-webhook-secret` header or the `?secret=` query string. Query strings are routinely written to reverse-proxy / load-balancer / CDN access logs and browser history — exactly the layers this PR expects to sit behind (`127.0.0.1` + reverse proxy). For a security-sensitive endpoint I'd recommend dropping the query-param path entirely and documenting header-only, or at minimum loudly warning against it in the README. (Recommend a human security sign-off on this choice given the `needs-human-review` label.)

**S2. Length-mismatch early-return is a (minor) timing oracle.** `src/webhook.ts:43`: `if (a.length !== b.length) return false;` returns before `timingSafeEqual`, so response timing reveals whether a guessed secret has the correct byte length. This is a weak leak (length only, and the bind is local), but since the function's stated purpose is constant-time comparison, it's worth noting. A common hardening is to HMAC/SHA-256 both sides to a fixed width before `timingSafeEqual`, removing the length branch.

**S3. `parseWebhookBody` does not validate `channel_id` / `channel_type`.** `src/webhook.ts:67-77` requires `message_id`, `from_uid`, `message_seq`, `timestamp`, and `payload.type`, but not `channel_id`/`channel_type` (both optional on `BotMessage`). The WS path always populates them (`src/octo/socket.ts:740-743`). When they're absent, the message is still fully processed (routing, LLM query, history append), but:
- the outbound send is **not** gated — `streamRelay.deliver(channelId, channelType, …)` at `src/index.ts:560` is called unconditionally with `channelId = msg.channel_id ?? ''` and `channelType` defaulting to `DM` (`src/index.ts:233-234`), so the reply attempt goes out with an empty channel and silently fails (read-receipt/typing paths *are* gated at lines 270/567, but `deliver` is not);
- group session keys collapse to `${'' }:${from_uid}` (`src/session-router.ts:141-147`), which can mis-partition sessions.

If a well-formed Octo webhook always carries these fields this is only defensive hardening, but since the parser is the trust boundary for an externally-reachable endpoint, I'd validate `channel_id`/`channel_type` there (reject 400) rather than rely on downstream defaults.

---

## Nits

**N1. Test binds a hard-coded port (`18789+`).** `src/__tests__/webhook.test.ts:63` uses a fixed `nextPort = 18789`. On a shared host where that port is already taken the round-trip tests fail with `EADDRINUSE` (this is exactly what happened in my environment — an unrelated process held 18789). Use `port: 0` to let the OS assign an ephemeral port and read it back from `server.address()`; this also removes cross-test port-reuse flakiness.

**N2. 200-before-processing means delivery is not acknowledged.** `src/webhook.ts:164-166` writes `200` and then fires the handler fire-and-forget. That's a reasonable choice for webhook semantics (fast ACK, async processing), but it means a handler that throws is only logged, never surfaced to the sender for retry. Worth a one-line comment that this is intentional at-most-once delivery, so a future reader doesn't mistake it for a bug.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| P1 | blocking | UTF-8 multi-byte corruption on chunk boundaries (`webhook.ts:147-149`) — silent data integrity loss, CJK-prone |
| S1 | P2 | `?secret=` query param risks credential exposure in logs (`webhook.ts:138-139`) |
| S2 | P2 | length-mismatch early return leaks secret length via timing (`webhook.ts:43`) |
| S3 | P2 | `parseWebhookBody` doesn't validate `channel_id`/`channel_type`; `deliver` ungated (`webhook.ts:67-77`, `index.ts:560`) |
| N1 | nit | test hard-codes port 18789 → `EADDRINUSE` flakiness (`webhook.test.ts:63`) |
| N2 | nit | document the 200-then-fire-and-forget at-most-once choice |

Addressing P1 (and ideally S1/S3 given the security-sensitive classification) would clear this for merge. The lifecycle, auth-required enforcement, and transport-branch design are solid.

lml2468:
Thanks @Jerry-Xin — all three addressed in 0117d1a.

**multi-bot webhook EADDRINUSE (blocking)** — `BotOverride` now has `transport` + `webhook` fields, merged over the top-level webhook block per bot. `resolveBotConfigs()` requires a per-bot `webhook.secret` in webhook mode and validates that every webhook bot binds a **distinct** `host:port:path` — colliding binds fail fast at config time with a clear message instead of a runtime EADDRINUSE. Tests: distinct ports ok, same-bind rejected, distinct paths on one port ok, missing per-bot secret rejected, mixed ws/webhook transports.

**const-time secret leak (non-blocking)** — the compare returned early on length mismatch. Both sides are now SHA-256'd to a fixed 32-byte digest before `timingSafeEqual`, so neither timing nor length leaks.

**partial-start cleanup (non-blocking)** — startup now connects stacks sequentially and, on failure, `shutdown()`s the already-started stacks before rethrowing, so a failed multi-bot start doesn't leave bound servers / open stores dangling before the fatal exit.

895 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (changes requested):
Summary: The PR is relevant to `cc-channel-octo`, but webhook multi-bot bind validation does not match the actual server architecture and can approve configurations that cannot start.

🔴 Blocking

🔴 Critical — Multi-bot webhook path-level sharing is validated as allowed, but cannot work with one HTTP server per bot.  
`resolveBotConfigs()` treats `host:port:path` as the unique bind key, so two webhook bots on the same host and port but different paths pass validation at [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/config.ts:631). The runtime creates a separate `WebhookServer` per bot at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/index.ts:192), and each calls `server.listen(this.port, this.host)` at [src/webhook.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/webhook.ts:112). The OS bind identity is `host:port`, not `host:port:path`, so the second bot fails with `EADDRINUSE`. The new test explicitly locks in the broken assumption at [src/__tests__/config.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/__tests__/config.test.ts:997).  
Fix either by requiring distinct `host:port` per webhook bot and updating docs/tests, or by implementing a shared HTTP server per `host:port` that dispatches to multiple bot handlers by path.

💬 Non-blocking

🟡 Warning — Webhook inbound bypasses the gateway’s self-message filter.  
The WebSocket path drops self-authored messages in `OctoGateway.handleMessage()` at [src/gateway.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/gateway.ts:240), but webhook mode invokes `onInbound` directly via [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/index.ts:198). The router still suppresses replies to self messages, but group self messages can fall through as “not processed” and be cached into group context at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/index.ts:635). Add the same `from_uid === gateway.botId` guard before feeding webhook messages into `handleMessage()`.

🟡 Warning — If webhook `listen()` fails, services are started before the bind succeeds.  
`gateway.startServices()` runs before `webhookServer.listen()` at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/index.ts:209), while the startup catch only shuts down stacks already pushed to `connected` at [src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-jwxosu8t/repo/src/index.ts:76). In CLI fatal-exit mode this is mostly transient, but embedded tests/importers could leak heartbeat/lock state for the failing stack. Prefer binding first and starting services after success, or have `connect()` clean up its own partial start on failure.

✅ Highlights

The shared-secret requirement, constant-time digest comparison, body cap, and malformed-body handling are good security defaults. The transport branch also keeps the REST registration lifecycle intact, which fits the existing two-phase startup model.

lml2468:
Thanks @Jerry-Xin — all addressed in 2a43e1c.

**bind key host:port (blocking)** — you're right, a `WebhookServer` per bot binds the whole `host:port` (the path is routed in-process), so `host:port:path` was the wrong uniqueness key. The collision check now keys on **`host:port`**, and the docs/tests are corrected: distinct paths on one port are now **rejected**, same port on **different hosts** is allowed. (The genuinely robust alternative — one shared server per host:port dispatching by path — is more machinery than this needs; distinct host:port per bot is the simpler correct contract.)

**webhook self-message filter (non-blocking)** — `onInbound` now drops `from_uid === gateway.botId`, mirroring `OctoGateway.handleMessage()`, so a bot's own group message can't fall through and be cached into group context.

**bind before services (non-blocking)** — webhook `connect()` now calls `webhookServer.listen()` **before** `gateway.startServices()`, so a bind failure throws before the heartbeat/signal handlers/lock start for that stack; `main()`'s partial-start catch shuts down the rest.

896 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (commented):
The PR is in scope for `cc-channel-octo`, but webhook mode has a blocking lifecycle bug in token refresh.

**🔴 Blocking**
- 🔴 Critical: Webhook mode can unexpectedly open a WuKongIM WebSocket after heartbeat failures. `startBot()` correctly starts webhook mode with `webhookServer.listen()` and `gateway.startServices()` without calling `gateway.connect()` ([src/index.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-va2rh7m2/repo/src/index.ts:214)). However, `OctoGateway.startServices()` starts the heartbeat loop, and after 3 heartbeat failures `attemptTokenRefresh()` always creates and connects a `WKSocket` ([src/gateway.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-va2rh7m2/repo/src/gateway.ts:284)). That breaks the central webhook contract: deployments that choose webhook to avoid long connections can silently start one during recovery, and may then receive inbound messages through both transports. The gateway needs transport-aware refresh behavior, e.g. refresh registration/REST state and restart heartbeat without creating a socket when `config.transport === 'webhook'`.

**💬 Non-blocking**
- 🟡 Warning: The new test only checks `startServices()` does not create a socket initially ([src/__tests__/gateway.test.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-va2rh7m2/repo/src/__tests__/gateway.test.ts:392)). Please add coverage for the heartbeat-failure refresh path in webhook mode so this regression cannot recur.

**✅ Highlights**
- ✅ Secret enforcement at config load and per-bot resolution is the right default for an inbound endpoint.
- ✅ The webhook body cap, same-handler integration, and multi-bot bind collision validation are well aligned with the existing architecture.

Jerry-Xin:
**Supplementary note (v4 re-review):**

The UTF-8 chunk mojibake issue flagged in v3 remains unfixed:

```typescript
// src/webhook.ts:150-152
req.on('data', (chunk: Buffer) => {
  body += chunk.toString('utf-8');  // ← still corrupts multi-byte chars at chunk boundaries
```

When a TCP chunk boundary splits a multi-byte UTF-8 sequence (e.g. CJK characters — 3 bytes each), `chunk.toString('utf-8')` produces a replacement character for the partial sequence, and the next chunk produces another. The standard fix:

```typescript
const chunks: Buffer[] = [];
let totalBytes = 0;
req.on('data', (chunk: Buffer) => {
  totalBytes += chunk.length;
  if (totalBytes > MAX_WEBHOOK_BODY_BYTES) { /* 413 */ return; }
  chunks.push(chunk);
});
req.on('end', () => {
  const body = Buffer.concat(chunks).toString('utf-8');
  // ...
});
```

This was blocker #7 from the v3 review and is still blocking.

Jerry-Xin (changes requested):
v4 (2a43e1cf31ab) — REQUEST_CHANGES. Port validation + self-message guard fixed, but 2 blockers remain:

1. UTF-8 chunk mojibake (v2-v4 unfixed): body += chunk.toString('utf-8') corrupts CJK across chunk boundaries. Fix: Buffer[] + concat at end.

2. NEW — attemptTokenRefresh() creates WKSocket in webhook mode: gateway.ts:284 always opens a WebSocket on token refresh failure, breaking webhook-only transport contract. Webhook mode should refresh tokens without opening WS.

yujiawei (changes requested):
# Code Review — PR #67 (cc-channel-octo)

**Verdict: CHANGES_REQUESTED.** The lifecycle and multi-bot fixes from the previous review rounds are correctly landed, and the design is sound (REST registration unchanged, WS path untouched, the same `onInbound` handler reused, self-message filter preserved, `startServices()` without a socket, awaited `listen()` with partial-startup cleanup, host:port bind uniqueness). But two **P1** issues remain before merge — one of them a data-corruption bug that was flagged in an earlier round and is **still present at this head SHA**.

I verified the UTF-8 bug by reproducing it locally and re-read the head-SHA source for `webhook.ts`, `config.ts`, `gateway.ts`, and `index.ts`.

---

## P1 — UTF-8 multi-byte corruption on chunk boundaries (blocking, still unfixed)

`src/webhook.ts` request data handler:

```ts
let body = '';
req.on('data', (chunk: Buffer) => {
  if (tooLarge) return;
  body += chunk.toString('utf-8');          // decodes each TCP chunk independently
  if (Buffer.byteLength(body, 'utf-8') > MAX_WEBHOOK_BODY_BYTES) { ... }
});
```

Each TCP chunk is decoded independently and string-concatenated. When a multi-byte UTF-8 character (every CJK char is 3 bytes; emoji are 4) straddles a chunk boundary, `chunk.toString('utf-8')` emits `U+FFFD` replacement characters for the incomplete sequence on each side of the split. The corrupted bytes still form **syntactically valid JSON**, so `JSON.parse` succeeds, `parseWebhookBody`'s structural-only validation passes (it type-checks fields, never inspects `content`), and the mangled `BotMessage` flows straight into the pipeline and is persisted to history — **silently, no 400, no log**.

This is CJK-prone and well within normal message sizes (the cap is 256 KiB; chunk splits routinely happen for streamed/chunked POSTs or bodies past the socket read buffer). Reproduced locally:

```
input:  {"content":"你好世界"}   (split mid-codepoint of 你)
buggy:  {"content":"��好世界"}    JSON.parse → ok → accepted silently
correct:{"content":"你好世界"}
```

This is the same defect raised in the prior round (at commit `c40135c`); the follow-up commits addressed only the lifecycle/multi-bot items, so it carried forward unchanged.

**Fix:** buffer the raw `Buffer`s and decode once at `end` (or use `string_decoder.StringDecoder`):

```ts
const chunks: Buffer[] = [];
let size = 0;
req.on('data', (chunk: Buffer) => {
  if (tooLarge) return;
  size += chunk.length;                        // count raw bytes, not decoded string
  if (size > MAX_WEBHOOK_BODY_BYTES) { tooLarge = true; res.writeHead(413).end(); req.destroy(); return; }
  chunks.push(chunk);
});
req.on('end', () => {
  if (tooLarge) return;
  const body = Buffer.concat(chunks).toString('utf-8');   // boundary-safe, decode once
  ...
});
```

Please add a test that posts CJK content split across writes mid-codepoint and asserts `payload.content` is byte-for-byte intact. The current round-trip tests can't catch this: `validMsg()` uses ASCII `content: 'hi'` sent as a single `fetch` body (one chunk).

---

## P1 — No request/socket timeouts; 401/404 reject paths don't drain the socket (slowloris, blocking)

`src/webhook.ts` `listen()` creates the server with `createServer(...)` and never sets `requestTimeout`, `headersTimeout`, a socket timeout, or `maxConnections`. Combined with the auth check returning before the body is read:

```ts
if (!secretMatches(provided, this.secret)) {
  res.writeHead(401).end();   // does NOT req.destroy() / req.resume()
  return;
}
```

an **unauthenticated** client can open a connection, send headers + a large declared `Content-Length`, then stall — holding the socket (and, on the authenticated path, the accumulating `body`) for Node's 5-minute default `requestTimeout`. With no `maxConnections`, N such sockets exhaust the server. The shared secret does **not** mitigate this: the 401 fires before the body is consumed but the socket is left open and streamable. Same undrained-socket issue on the 404 path.

The `127.0.0.1` default + "put a reverse proxy in front" guidance normally absorbs this, which is why it's P1 not P0 — but the code provides zero in-process protection, `host` is operator-overridable to `0.0.0.0`, and a local-host attacker or an over-trusting proxy defeats the mitigation.

**Fix (all in `listen()` / the reject paths):** set a finite `server.requestTimeout` (e.g. 10–30s), `server.headersTimeout`, and `server.maxConnections`; and on the 401/404 rejects call `req.destroy()` (or `req.resume()`) so the socket is freed immediately. A stalled-request test would lock this in — none currently exists.

---

## P2 — worth addressing given the `needs-human-review` classification

**S1. `?secret=` query param leaks the credential into logs.** The secret may be supplied via `?secret=` as well as the `x-webhook-secret` header. Query strings are routinely written to reverse-proxy / LB / CDN access logs — exactly the layers this endpoint is documented to sit behind — and the behavior is intentionally tested and documented. Recommend dropping the query-param path (header-only) or gating it behind an explicit opt-in with a startup warning. (Recommend a human security sign-off here.)

**S2. Sender identity is unauthenticated — the secret authenticates the transport, not the message.** On the WS path, WuKongIM vouches for the sender. On the webhook path, `parseWebhookBody` trusts every field of the JSON body verbatim — `from_uid`, `from_name`, `channel_id`, `channel_type` are all attacker-chosen for anyone who holds (or leaks) the secret, or any compromised hop in front of the endpoint. Consequences: a holder of the secret can spoof a message "from Alice" into an arbitrary channel (it gets cached into that group's LLM context — a context-poisoning primitive), target any session/cwd sandbox, and trivially bypass the `from_uid === botId` self-filter by changing/omitting `from_uid`. The README frames the secret as the whole story ("an open endpoint would let anyone inject messages") but should state the stronger truth: even with the secret, per-message sender identity is forgeable, so the secret-holder is fully trusted to assert any identity. Worth an explicit human review of this trust model before shipping.

**S3. `parseWebhookBody` doesn't validate `channel_id` / `channel_type`.** It requires `message_id`/`from_uid`/`message_seq`/`timestamp`/`payload.type` but not the routing fields. When they're absent (or `channel_type` arrives as a JSON string `"2"`, common across HTTP integrations — `'2' === 2` is false), the message is still fully processed: `channelId` defaults to `''` and `channelType` to `DM` (`src/index.ts`), so the outbound `streamRelay.deliver('', DM, …)` goes out with an empty channel and silently fails to reach the user, the group @mention gate is skipped (bot replies to everything), and group session keys collapse / mis-partition. Validate/coerce `channel_type` to a number and require `channel_id` for group-typed messages → 400 rather than half-processing.

**S4. Webhook env overrides aren't trimmed.** `CC_OCTO_WEBHOOK_SECRET` / `_HOST` / `_PATH` are assigned raw in `applyEnv`, unlike `cwdBase` which runs through `nonBlank()`. A trailing newline from `CC_OCTO_WEBHOOK_SECRET=$(cat /run/secrets/x)` or a K8s/Docker secret file silently appends `\n` to the expected secret → every clean caller gets 401 with no diagnostic (the SHA-256 compare even hides the length-mismatch hint); a newline in `_PATH` 404s every request; a whitespace-only secret passes the `!webhook.secret` truthiness check. Run these through `nonBlank()` for consistency.

---

## Verified correct / refuted

- **Constant-time secret compare** — the earlier length-mismatch/timing-oracle concern is **fixed**: `secretMatches` SHA-256s both sides to fixed 32-byte digests before `timingSafeEqual`, with no length branch. No action needed.
- **Lifecycle** — `connect()` is async + awaited; `listen()` rejects on bind error and propagates; partial-startup cleanup shuts down already-connected stacks; `startServices()` runs heartbeat + token-refresh + shutdown handlers with **no** socket in webhook mode; self-message filter applies on the webhook path. Correct.
- **Multi-bot binds** — `BotOverride` now carries `transport`/`webhook`; bind uniqueness is correctly keyed on `host:port` (not `host:port:path`) with a clear fail-fast error. Correct.

## Nits

- Partial-startup cleanup comment claims it avoids leaving "open stores dangling," but the stack whose `listen()` threw is never pushed to `connected`, so its `SessionStore`/lock aren't closed by the catch. Harmless because `main().catch → process.exit(1)` lets the OS reclaim them — but the comment overstates the cleanup.
- `Buffer.byteLength(body)` is recomputed over the whole accumulated string on every `data` chunk (O(n²) on many tiny chunks). Tracking a running byte counter is O(n) and pairs naturally with the P1 buffer-and-decode-once fix.
- Test binds a hard-coded port (`18789+`) → `EADDRINUSE` flakiness on shared hosts. Use `port: 0` and read back `server.address()`.

---

## Summary

| # | Severity | Issue |
|---|----------|-------|
| P1 | blocking | UTF-8 multi-byte corruption on chunk boundaries — silent data loss, CJK-prone, **still unfixed** |
| P1 | blocking | No request/socket timeouts + undrained 401/404 sockets → slowloris DoS |
| S1 | P2 | `?secret=` query param risks credential exposure in proxy/CDN logs |
| S2 | P2 | per-message sender identity unauthenticated/forgeable (spoofing + context poisoning) |
| S3 | P2 | `parseWebhookBody` doesn't validate `channel_id`/`channel_type` → ungated `deliver`, mis-routing |
| S4 | P2 | webhook env vars not trimmed → newline-in-secret/path footgun |

Fixing both P1s clears this for merge; given the `needs-human-review` label I'd also want S1 and S2 looked at by a human before shipping.

lml2468:
Thanks @Jerry-Xin — both blockers fixed in 665e898.

**UTF-8 chunk mojibake** — the body was decoded per chunk (`body += chunk.toString()`), corrupting a multibyte codepoint split across a TCP chunk boundary. Now buffers raw `Buffer` chunks and decodes once via `Buffer.concat(...).toString('utf-8')` (the size cap tracks raw byte length). New test writes a CJK body as **two raw TCP chunks split inside a 3-byte char** and asserts the reassembled content matches exactly.

**token refresh opens WS in webhook mode** — `attemptTokenRefresh()` always recreated a `WKSocket`. Added a `wsEnabled` flag set only by `connect()`; refresh now re-registers + restarts the heartbeat but reopens the socket **only when `wsEnabled`**. New test: in webhook mode (`register()` + `startServices()`), 3 heartbeat failures trigger a re-register but never construct a `WKSocket`.

898 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (changes requested):
The PR is in scope for this repository, but webhook parsing has a blocking compatibility bug with Octo-style numeric message IDs.

**🔴 Blocking**
- 🔴 Critical — `parseWebhookBody()` uses plain `JSON.parse()` and then requires `message_id` to already be a string. Existing Octo REST parsing has explicit int64 protection for numeric `message_id` values before parsing, because large IDs can exceed JS safe integer precision. A webhook body like `"message_id":1234567890123456789` will either be rejected as malformed or, if later accepted by conversion, already be precision-corrupted. This can make webhook mode fail against real Octo JSON or lose the exact ID needed for receipts/history. Reuse the same pre-parse normalization strategy as `src/octo/api.ts` or centralize it. References: [src/webhook.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-c8qzk6ch/repo/src/webhook.ts:55), [src/octo/api.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-c8qzk6ch/repo/src/octo/api.ts:41)

**💬 Non-blocking**
- 🟡 Warning — Webhook validation accepts messages without `channel_id` / `channel_type`, but the production pipeline needs those fields for routing replies and read receipts. Such requests get `200` and then can fail later with an empty channel. Consider treating authenticated messages missing reply-routing fields as `400`, or document the accepted minimal schema clearly. Reference: [src/webhook.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-c8qzk6ch/repo/src/webhook.ts:70)
- 🔵 Suggestion — Config validation should reject invalid webhook paths and out-of-range ports earlier. For example, `webhook.path: "foo"` will bind successfully but never match any normal HTTP request path. Reference: [src/config.ts](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-c8qzk6ch/repo/src/config.ts:497)

**✅ Highlights**
- The two-phase startup integration is well aligned with the existing multi-bot lifecycle.
- The required shared secret, local default bind host, body cap, and constant-time comparison are the right baseline for this feature.
- Test coverage covers real HTTP round-trips, malformed bodies, auth failures, oversized bodies, multibyte chunking, and webhook token-refresh behavior.

I attempted the targeted test run, but this checkout does not have dependencies installed: `vitest: command not found`.

yujiawei (changes requested):
# Code Review — PR #67 (cc-channel-octo)

**Verdict: CHANGES_REQUESTED.** The webhook transport design is sound — REST registration is unchanged, the WS path is untouched, the same `onInbound` handler is reused, and the lifecycle work from earlier rounds has landed correctly. Since the previous round, the **UTF-8 chunk-boundary corruption P1 is now fixed** (raw `Buffer`s are concatenated and decoded once, with a CJK-split test). 

However, this is a `security_sensitive` transport and several items still block merge: one previously-flagged DoS-hardening P1 remains unaddressed, and there's a correctness bug in inbound channel typing that silently bypasses the group mention gate. I re-read `webhook.ts`, `config.ts`, `gateway.ts`, `index.ts`, and `session-router.ts` at the head SHA and reproduced the key behaviors.

---

## P1 — Inbound `channel_type` is never validated/coerced → group mention gate bypass + misrouting

`parseWebhookBody` (`src/webhook.ts:55-81`) strictly type-checks `message_id`, `from_uid`, `message_seq`, `timestamp`, and `payload.type`, but **not** `channel_id` or `channel_type`. `channel_type` is returned verbatim from the JSON body.

Downstream code assumes `channel_type` is the numeric `ChannelType` enum and compares with strict equality:

- `src/index.ts:253-254`: `const channelType = msg.channel_type ?? ChannelType.DM;` — `??` preserves a JSON string `"2"` unchanged.
- `src/session-router.ts:201-202`: `isGroupLike()` returns `channelType === ChannelType.Group || channelType === ChannelType.CommunityTopic;` — `"2" === 2` is `false`.

Consequences when a producer sends `channel_type` as a string (`"2"`) or omits it:
1. **Mention gate bypassed** (`src/session-router.ts:248`): `isGroupLike("2")` is `false`, so the group `@mention` gate is skipped and the bot replies to *every* message in the group.
2. **Outbound misrouting**: a missing `channel_type` defaults to `DM` and a missing `channel_id` defaults to `''`, so the reply is delivered to an empty channel and silently fails to reach the user.
3. **Session-key mis-partitioning** for group history.

This is inconsistent with how the other inbound fields are validated, and it's exactly the kind of type drift an HTTP producer introduces. **Fix:** coerce/validate `channel_type` to a number in `parseWebhookBody` and require `channel_id` for group-typed messages, returning `400` rather than half-processing.

## P1 — No request/socket timeouts; 401/404 reject paths don't drain the socket (slowloris)

`listen()` (`src/webhook.ts`) creates the server without `requestTimeout`, `headersTimeout`, a socket timeout, or `maxConnections`. The reject paths return without draining the socket:

- `src/webhook.ts:136` (404 wrong path/method): `res.writeHead(404).end();` — no `req.destroy()` / `req.resume()`.
- `src/webhook.ts:144` (401 bad/missing secret): `res.writeHead(401).end();` — same.

(The 413 oversize path at `src/webhook.ts:159` *does* call `req.destroy()` — that one is fine.)

An unauthenticated client can open a connection, send headers with a large declared `Content-Length`, then stall — holding the socket for Node's 5-minute default `requestTimeout`. With no `maxConnections`, N such sockets exhaust the server, and the 401 fires *before* the body is consumed, so the secret provides no protection here. The `127.0.0.1` default + reverse-proxy guidance is a real mitigation (which is why this isn't P0), but the code provides zero in-process protection and `host` is operator-overridable to `0.0.0.0`. This was flagged in the prior round and is unchanged at this head SHA. **Fix:** set finite `server.requestTimeout` / `headersTimeout` / `maxConnections`, and call `req.destroy()` (or `req.resume()`) on the 401/404 rejects.

---

## P2 — security/robustness (recommend a human security sign-off given `needs-human-review`)

- **`?secret=` query param can leak the credential into logs** (`src/webhook.ts:140-142`). Query strings are routinely written to reverse-proxy / LB / CDN access logs — exactly the layers this endpoint sits behind. Recommend header-only auth, or gate the query path behind an explicit opt-in with a startup warning.

- **Per-message sender identity is unauthenticated.** The shared secret authenticates the *transport*, not the *sender*. `parseWebhookBody` trusts `from_uid` / `from_name` / `channel_id` / `channel_type` verbatim, so anyone holding (or leaking) the secret — or a compromised proxy hop — can spoof a message "from Alice" into any channel (it's cached into that group's LLM context — a context-poisoning primitive) and bypass the `from_uid === botId` self-filter by changing `from_uid`. The WS path mitigates this because WuKongIM vouches for the sender after its encrypted handshake. The README frames the secret as the whole story; it should state plainly that the secret-holder is fully trusted to assert any identity. Worth an explicit human review of this trust model before shipping.

- **Webhook env overrides aren't trimmed** (`src/config.ts:365,369-370`). `CC_OCTO_WEBHOOK_SECRET` / `_HOST` / `_PATH` are assigned raw, unlike `cwdBase` which runs through `nonBlank()`. A trailing newline from `CC_OCTO_WEBHOOK_SECRET=$(cat /run/secrets/x)` or a Docker/K8s secret file silently appends `\n` → every clean caller gets a `401` with no diagnostic (the SHA-256 compare even hides the length-mismatch hint). A whitespace-only secret also passes the `!webhook.secret` truthiness check at `src/config.ts:497`, defeating the mandatory-secret guard. Run these through `nonBlank()`.

---

## Verified correct (no action needed)

- **UTF-8 multibyte corruption — FIXED.** `src/webhook.ts:148-166` buffers raw `Buffer`s and decodes once via `Buffer.concat(chunks).toString('utf-8')`; byte counting uses `chunk.length`. Covered by the new CJK-split-across-chunks test (`src/__tests__/webhook.test.ts:144-172`).
- **Constant-time secret compare.** `secretMatches` SHA-256s both sides to fixed 32-byte digests before `timingSafeEqual`, with no length branch — no timing/length oracle.
- **Multi-bot bind uniqueness.** `resolveBotConfigs` keys `seenBinds` on `host:port` (not `host:port:path`) with a clear fail-fast error (`src/config.ts:635-650`); per-bot `webhook.secret` is required (`src/config.ts:622-627`).
- **Webhook-mode token refresh.** `connect()` sets `wsEnabled=true`; `startServices()` opens no socket; `attemptTokenRefresh()` only reopens the WS when `wsEnabled` is true, so webhook mode refreshes credentials without spawning a socket (`src/gateway.ts:101-130,252-303`), with a covering test.
- **Lifecycle.** `connect()` is async + awaited; `listen()` rejects on bind error and propagates; partial-startup cleanup shuts down already-connected stacks; self-message filter is applied on the webhook path (`src/index.ts:176`).

---

## Nits

- O(n²) avoided in the new buffer path — good. The partial-startup cleanup comment claims it avoids leaving "open stores dangling," but the stack whose `listen()` threw is never pushed to `connected`, so its store/lock aren't closed by the catch (harmless because `process.exit(1)` reclaims them, but the comment overstates it).
- Tests bind hard-coded ports (`18789+`), which can flake with `EADDRINUSE` on shared CI hosts — prefer `port: 0` and read back `server.address()`.

---

## Summary

| 1 | P1 | `channel_type`/`channel_id` not validated → group mention-gate bypass + misrouting | present |
| 2 | P1 | No request/socket timeouts + undrained 401/404 sockets → slowloris | present (carried over) |
| 3 | P2 | `?secret=` query param risks credential exposure in proxy/CDN logs | present |
| 4 | P2 | Per-message sender identity unauthenticated/forgeable (spoofing + context poisoning) | present |
| 5 | P2 | Webhook env vars not trimmed → newline/whitespace-secret footgun | present |
| — | — | UTF-8 chunk-boundary corruption | **fixed** |

Resolving the two P1s clears this for merge. Given the `needs-human-review` label, items 3 and 4 (query-param secret + sender trust model) warrant a human security sign-off before shipping.

lml2468:
Thanks @Jerry-Xin — all three addressed in 334c38f.

**int64 message_id (blocking)** — `parseWebhookBody` now reuses the REST client's int64-safe parser. I exported `parseOctoJson` from `src/octo/api.ts` (the same pre-parse normalization that stringifies 16+ digit `message_id`s) and use it in the webhook path; a remaining small numeric id is coerced to string. Test: a 19-digit numeric id (`1234567890123456789`) round-trips **exactly** as a string.

**missing channel routing fields (non-blocking)** — `parseWebhookBody` now requires a non-empty `channel_id` and a numeric `channel_type`, so a message that couldn't be replied to / receipted is rejected with 400 up front instead of 200-then-fail.

**path/port validation (non-blocking)** — `loadConfig` now rejects a `webhook.path` without a leading slash and an out-of-range `webhook.port` at boot.

901→904 tests pass (type-check / lint / build clean; NUL scan clean).

Jerry-Xin (approved):
Summary: The PR is in scope and implements a relevant webhook inbound transport for this gateway. I found no blocking correctness or security issues.

💬 Non-blocking

🟡 Warning — `src/index.ts:76-85`: partial startup cleanup only shuts down stacks already pushed to `connected`. If a webhook bind fails on the first stack, or on stack N, the failed stack and not-yet-connected registered stacks are not shut down. Because `startBot()` has already opened stores, timers, and lock files, consider cleaning up all `stacks` except those already shut down.

🟡 Warning — `src/config.ts:505-514`, `src/config.ts:631-639`: webhook path/port validation only runs for top-level webhook mode. A per-bot `transport: "webhook"` can still carry an invalid `webhook.path` like `foo`, which starts successfully but will never match incoming requests. Mirror the path/port validation in `resolveBotConfigs()` after per-bot merge.

🔵 Suggestion — `src/config.ts:647-652`: bind collision detection treats `0.0.0.0:8787` and `127.0.0.1:8787` as distinct, but they can conflict at runtime. This is not a correctness blocker because runtime bind failure is handled, but documenting or normalizing wildcard binds would make the “fail fast” guarantee stronger.

✅ Highlights

- `src/webhook.ts` uses a required shared secret, constant-time digest comparison, and a body cap.
- Webhook mode reuses the existing message pipeline and preserves REST heartbeat/token-refresh lifecycle without opening a WebSocket.
- Tests cover parsing, auth failures, real HTTP round trips, oversized bodies, config secret requirements, and token refresh behavior.

Verification caveat: I attempted targeted tests and type-check, but this checkout does not have dev dependencies installed (`vitest` and `tsc` were not found).

PR #69, opened 2026-06-08
fix(router): drop non-conversation channel types (system messages)

### LINKED ISSUE #68: bug(router): bot replies to non-DM/group system messages (channel_type 8 'systemcmdonline')
## Found in live deployment

Running the gateway against a real Octo server (`im.deepminer.com.cn`), on connect the bot received a system message on **`channel_type: 8`** with `channel_id: "systemcmdonline"` and an empty body (rendered as the `[消息]` placeholder) — and **generated + sent an unsolicited LLM reply to it**.

Observed in the session store after a fresh start:
```
session_id="systemcmdonline:"  channel_id="systemcmdonline"  channel_type=8
  user:      [消息]
  assistant: 你好！我看到你的消息似乎是空的（只有一个「[消息]」占位符）…
```

## Root cause

`SessionRouter.processMessage()` gates by *negative* checks only:
- DM-specific filters apply when `channel_type === DM` (1).
- The @mention gate + bot-loop guards apply only when `isGroupLike()` (Group=2 / CommunityTopic=5).
- `if (msg.payload.event) return null` drops system *events*, but this message arrived as a content message without the `event` flag.

There is **no allowlist of handled channel types**. A `channel_type` we don't recognize (here `8`, a WuKongIM system/command channel) is neither DM nor group-like, so it skips the mention gate entirely and is processed as if it were a repliable DM. `ChannelType` only defines `DM=1, Group=2, CommunityTopic=5`.

Cost: a wasted model call + an unsolicited reply to a system channel on every connect (and potentially other system `channel_type`s).

## Fix

Add a positive allowlist near the top of `processMessage()`: only `DM`, `Group`, and `CommunityTopic` are repliable; drop anything else (return null). This is defense-in-depth alongside the existing `payload.event` and self-message drops.

## Acceptance

- A message with an unknown/system `channel_type` (e.g. 8) is dropped, no agent call, no reply.
- DM and group flows unchanged.
- Regression test in `session-router.test.ts`.

lml2468:
Closes #68. Found during the live deployment test against `im.deepminer.com.cn`.

## What happened

On connect, `OctoReviewBoooot` received a system message on **`channel_type: 8`** (`channel_id: "systemcmdonline"`, empty body → `[消息]` placeholder) and **generated + sent an unsolicited LLM reply**. Captured in the session store:

```
session_id="systemcmdonline:"  channel_type=8
  user:      [消息]
  assistant: 你好！我看到你的消息似乎是空的…
```

## Root cause

`SessionRouter.processMessage()` gated only by negative checks: DM filters for `channel_type === DM` (1), the @mention gate + bot-loop guards for `isGroupLike()` (2/5). An unknown system `channel_type` (8) is neither → skips the mention gate → processed as a repliable DM. `payload.event` was unset (it arrived as a content message), so the system-event drop missed it.

## Fix

Positive allowlist `isSupportedChannel()` = DM | Group | CommunityTopic; `processMessage()` drops anything else up front, before stream/mention/agent — defense-in-depth alongside the existing `payload.event` and self-message drops.

## Test results

- `npm test` — **907 passed** (+3: system type 8 dropped no-reply, unknown type 99 dropped, normal DM still processed)

## Deployment validation

The rest of the live test passed end-to-end: real bot registration, WS connect, DM round-trip (inbound → agent → streamed reply, persisted), per-session sandboxes, `0700` dataDir, graceful SIGTERM shutdown. This was the one defect surfaced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ While the major version is `0`, minor releases may carry breaking changes.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Router drops non-conversation channel types** (#68) — found in live
+  deployment: on connect the bot received a system message on `channel_type: 8`
+  (`systemcmdonline`) and replied to it. `SessionRouter` now allowlists only DM /
+  Group / CommunityTopic as repliable; any other (system/command) channel type is
+  dropped before the agent is invoked.
+
 ### Added
 
 - **Webhook inbound transport** (v1.0) — set `transport: "webhook"`
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -684,3 +684,41 @@ describe('multi-bot loop guard in mention-free groups', () => {
     expect(result?.shouldProcess).toBe(true);
   });
 });
+
+// ─── #68: unsupported/system channel types are dropped ──────────────────
+
+describe('unsupported channel types (system messages)', () => {
+  let router: SessionRouter;
+  beforeEach(() => {
+    vi.clearAllMocks();
+    router = new SessionRouter(makeConfig(), ROBOT_ID);
+  });
+
+  it('drops a system channel_type (8 "systemcmdonline") with no reply', async () => {
+    // Reproduces the live-deployment bug: a system message on channel_type 8
+    // must not be processed as a conversation.
+    const msg = makeMsg({
+      channel_id: 'systemcmdonline',
+      channel_type: 8 as unknown as ChannelType,
+      from_uid: 'system',
+      payload: { type: MessageType.Text, content: '' },
+    });
+    const result = await router.route(msg);
+    expect(result).toBeNull();
+    expect(sendMessage).not.toHaveBeenCalled();
+  });
+
+  it('drops an unknown channel_type (e.g. 99)', async () => {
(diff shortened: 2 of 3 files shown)


Jerry-Xin (approved):
Summary: This PR is in scope and correctly fixes router behavior for non-conversation Octo channels without introducing blocking regressions.

💬 Non-blocking

- 🔵 Suggestion: Consider adding one `routeAndHandle()` regression test for unsupported `channel_type` to assert the handler is not invoked. `route()` already exercises the same `processMessage()` path, so this is optional. See `src/session-router.ts:225` and `src/__tests__/session-router.test.ts:697`.

✅ Highlights

- The PR passes the relevance gate: it fixes core routing behavior in `cc-channel-octo`.
- The positive allowlist in `src/session-router.ts:212` matches the declared conversation channel enum in `src/octo/types.ts:92`.
- The early drop in `src/session-router.ts:233` is well placed before rate limits, replies, context, and downstream processing.
- Regression coverage includes the observed `channel_type: 8` system channel, an unknown channel, and a normal DM path in `src/__tests__/session-router.test.ts:697`.

Verification note: I attempted `npm test -- src/__tests__/session-router.test.ts`, but this checkout has no `node_modules`, so `vitest` was unavailable. `git diff --check` passed.

PR #75, opened 2026-06-08
fix(octo): ack RECV only after decrypt; fail short-salt CONNACK

### LINKED ISSUE #74: octo protocol: RECVACK-before-decrypt drops messages silently; short-salt CONNACK enters silent-drop state
### 1. `src/octo/socket.ts` — RECVACK sent before decrypt/parse (silent message loss)
`onRecv` sends `RECVACK` **immediately**, then decrypts + `JSON.parse`s inside a try whose catch just logs and returns. RECVACK tells the server "delivered, don't resend" — so any decrypt/parse failure (even transient) **permanently loses** the message with only a debug log. The bot looks healthy (heartbeat fine) while swallowing user messages.

**Fix:** ack only **after** a successful decrypt+parse, so a transient failure leaves the message un-acked and the server redelivers.

### 2. `src/octo/socket.ts` — short-salt CONNACK proceeds into a guaranteed-decrypt-failure state
On `reasonCode === 1`, a salt `< 16` bytes only logs `console.warn` and continues with a zero-padded (wrong) AES-CBC IV — so **every** subsequent payload decrypt fails. Combined with #1 (pre-fix) this acked-then-dropped everything; post-#1 it would redeliver-loop forever.

**Fix:** treat a short salt as a failed handshake — close + reconnect to re-derive keys, instead of entering the silent-drop state.

> Note: the earlier review hypothesis that `onConnack` throws *outside* the `handleRawData` try/catch was checked and **refuted** — `onPacket`→`onConnack` runs inside `unpackOne`, which is inside the try. So a malformed DH key already reconnects. The real robustness gap is the salt-length case above.

## Acceptance
- RECVACK sent only after successful decrypt+parse; failure logs at error level and does not ack.
- Short-salt CONNACK fails the connection (onError + reconnect), does not fire onConnected.
- Tests for both; full suite green.

lml2468:
Closes #74.

Two silent-failure bugs in the WuKongIM protocol layer (`src/octo/socket.ts`), found in the full-codebase review.

## 1. RECVACK sent before decrypt → permanent silent message loss
`onRecv` acked **immediately**, then decrypted+parsed in a try that just logged on failure. RECVACK = "delivered, don't resend", so any decrypt/parse failure lost the message forever (debug log only) while the bot looked healthy (heartbeat fine).
**Fix:** ack only **after** a successful decrypt+parse → a transient failure leaves the message un-acked and the server redelivers; failure now logs at error level.

## 2. Short-salt CONNACK → guaranteed-decrypt-failure ("silent-drop") state
On success, a salt `< 16` bytes only `console.warn`ed and continued with a zero-padded (wrong) AES-CBC IV, so **every** later payload decrypt failed.
**Fix:** treat a short salt as a failed handshake — close + `onError` so the socket reconnects and re-derives keys, instead of entering that state.

> The earlier review hypothesis that `onConnack` throws *outside* `handleRawData`'s try/catch was checked and **refuted** (`onPacket`→`onConnack` runs inside `unpackOne`, inside the try). The real gap is the salt case above.

## Tests
RECVACK-ordering (sent on success; **not** sent on decrypt failure) + short-salt CONNACK fails the connection (replaces the old Q38 warn assertion). **948 tests**, lint/build/NUL clean.

### DIFF
--- a/src/__tests__/octo/wksocket-packet.test.ts
+++ b/src/__tests__/octo/wksocket-packet.test.ts
@@ -368,6 +368,51 @@ describe('WKSocket packet-level integration', () => {
     expect(msg.payload.content).toBe('hello from test');
   });
 
+  // ── RECVACK ordering (issue: ack-before-decrypt → silent message loss) ──
+
+  it('sends RECVACK after a successful decrypt', () => {
+    const onMessage = vi.fn();
+    socket = makeSocket(onMessage, vi.fn());
+    socket.connect();
+    const ws = wsRef.current!;
+    ws.emit('open');
+    const { aesKey, aesIV } = doHandshake(ws, 1);
+
+    const before = ws.sent.length;
+    ws.emit('message', Buffer.from(buildRecvPacket({
+      serverVersion: 4, fromUID: 'u', channelID: 'c', channelType: 2,
+      messageID: 1n, messageSeq: 7, timestamp: 1, payload: { type: 1, content: 'hi' },
+      aesKey, aesIV,
+    })));
+
+    expect(onMessage).toHaveBeenCalledOnce();
+    // A RECVACK frame (first byte = RECVACK<<4 = 0x60) was sent.
+    const newFrames = ws.sent.slice(before);
+    expect(newFrames.some((f) => (f[0] >> 4) === 6)).toBe(true);
+  });
+
+  it('does NOT send RECVACK when decrypt fails (lets the server redeliver)', () => {
+    const onMessage = vi.fn();
+    socket = makeSocket(onMessage, vi.fn());
+    socket.connect();
+    const ws = wsRef.current!;
+    ws.emit('open');
+    doHandshake(ws, 1); // establishes the real aesKey/aesIV on the socket
+
+    const before = ws.sent.length;
+    // Encrypt with a WRONG key so the socket's decrypt throws.
+    ws.emit('message', Buffer.from(buildRecvPacket({
+      serverVersion: 4, fromUID: 'u', channelID: 'c', channelType: 2,
+      messageID: 2n, messageSeq: 8, timestamp: 1, payload: { type: 1, content: 'x' },
+      aesKey: 'wrongkey00000000', aesIV: 'wrongiv0000000000'.slice(0, 16),
+    })));
+
+    expect(onMessage).not.toHaveBeenCalled();
+    const newFrames = ws.sent.slice(before);
+    // No RECVACK → message stays un-acked → server will redeliver.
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
## Review — `a881cb5e1c10` — APPROVE

Two protocol-layer silent-failure fixes, both correct and well-scoped.

### Fix 1: RECVACK after decrypt ✅

**Before:** `onRecv` sent RECVACK immediately, then tried decrypt in a try/catch that only debug-logged failures → permanent message loss (server thinks delivered, bot never processed it).

**After:** decrypt+parse runs first; RECVACK only sent on success. On failure: error-level log + early return → message stays un-acked → server redelivers.

The ordering is correct: `sendRaw(encodeRecvackPacket(...))` is now after the `try { aesDecrypt → JSON.parse → payloadObj }` block. If a *permanent* AES key/IV mismatch exists (all decrypts fail), the short-salt fix below handles it at the CONNACK level rather than creating a redelivery storm.

### Fix 2: Short-salt CONNACK fails the connection ✅

**Before:** `salt.length < 16` → `console.warn` + continue with zero-padded IV → every subsequent message silently fails to decrypt → "connected but deaf" state.

**After:** early return before setting `this.connected = true`, close the WS, emit `onError` → reconnect cycle gives the server another chance to provide a valid salt.

Guard placement is correct: before `serverPubKey` derivation, so no DH computation is wasted. `salt.substring(0, 16)` after the guard is safe (length guaranteed ≥ 16). `needReconnect` stays true (class default), so the `close` event triggers automatic reconnection.

### Error semantics

- `console.error` for both paths (was `debug` / `warn`) — correct severity for "message lost" / "connection broken".
- `onError(new Error("CONNACK salt too short"))` — caller gets a programmatic signal, can log/alert.

### Tests ✅

- **RECVACK ordering:** two tests — success case (RECVACK frame byte `0x60` present) and decrypt-failure case (wrong key → no RECVACK, onMessage not called).
- **Short-salt CONNACK:** updated from "emits warn" → "fails connection" — asserts `onConnected` NOT called, `onError` called with message, `console.error` logged. Negative case (valid salt → connects normally) simplified.
- 948 tests, lint/build clean.

### No concerns

- No risk of redelivery storm: permanent key mismatch is now caught at the CONNACK level (short salt fails), so the only case reaching the decrypt path with a bad key is a transient corruption (rare, self-resolving on redeliver).
- `this.ws.close()` in the short-salt path is wrapped in try/catch — safe even if WS is already closing.
- The `return` after short-salt means `this.connected` stays `false`, `this.aesKey`/`this.aesIV` stay unset — no stale state leaks.

PR #77, opened 2026-06-08
fix(router): per-member group rate limit; guard empty DM from_uid

### LINKED ISSUE #76: Rate-limit regression: groups share one bucket; DM sessionKey has no empty-uid guard
### 1. `src/session-router.ts` — group rate-limit collapsed to one shared bucket
After the sessionKey change to per-channel (`channel_id`), the per-session token bucket is keyed by `sessionKey`, so a **whole group shares one** `maxPerMinute` quota (default 5). The 6th message/minute from **any** member blocks everyone — an availability regression. The per-user bucket doesn't restore per-member group throughput.

**Fix:** key the session bucket by `${sessionKey}\n${uid}` so each member gets their own per-channel quota (matches pre-shared-session behavior); global + per-user buckets still bound abuse.

### 2. `src/session-router.ts` — DM sessionKey has no empty-`from_uid` guard
The group branch throws on a missing `channel_id`, but the DM branch returns bare `from_uid`. An empty uid → `''` shared session across unrelated peers (and collides with an empty group key).

**Fix:** symmetric guard — throw on a DM with no `from_uid` (caught upstream → message dropped).

## Acceptance
- Two members in one group each get their own `maxPerMinute`.
- DM with empty `from_uid` throws instead of returning `''`.
- Tests for both; full suite green.

lml2468:
Closes #76. Two regressions from the per-channel sessionKey change.

## 1. Group rate-limit collapsed to one shared bucket
The per-session token bucket was keyed by `sessionKey` = `channel_id`, so a **whole group shared one** `maxPerMinute` quota (default 5) — the 6th message/min from **any** member blocked everyone.
**Fix:** key the session bucket by `${sessionKey}\n${uid}` → each member gets their own per-channel quota (pre-shared-session behavior). Global (10×) + per-user buckets still bound abuse. DMs already embed the peer in sessionKey, so they stay per-peer.

## 2. DM sessionKey: no empty-`from_uid` guard
The group branch throws on missing `channel_id`, but DM returned bare `from_uid` — an empty uid collapsed unrelated peers into one `''` session.
**Fix:** symmetric throw on empty `from_uid` (caught upstream → dropped).

## Tests
Two group members get independent quotas in one channel; DM with empty `from_uid` throws. **950 tests**, lint/build/NUL clean.

### DIFF
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -260,6 +260,41 @@ describe('SessionRouter', () => {
 
   // --- Serial queue ---
 
+  it('group rate limit is per-member, not one shared bucket for the whole channel', async () => {
+    // Regression: with the per-channel sessionKey, keying the session bucket by
+    // sessionKey alone collapsed the whole room into one quota. Each member must
+    // get their own maxPerMinute in the same channel.
+    const cfg = makeConfig({ rateLimit: { maxPerMinute: 2 } });
+    router = new SessionRouter(cfg, ROBOT_ID);
+    const CH = 'group-shared';
+    const mentioned = { type: MessageType.Text, content: 'hi', mention: { uids: [ROBOT_ID] } };
+
+    // Alice uses her full quota (2).
+    const a1 = await router.route(makeMsg({ message_id: 'a1', channel_id: CH, from_uid: 'alice', channel_type: ChannelType.Group, payload: mentioned }));
+    const a2 = await router.route(makeMsg({ message_id: 'a2', channel_id: CH, from_uid: 'alice', channel_type: ChannelType.Group, payload: mentioned }));
+    expect(a1!.shouldProcess).toBe(true);
+    expect(a2!.shouldProcess).toBe(true);
+
+    // Bob in the SAME channel still has his own quota — not blocked by Alice.
+    const b1 = await router.route(makeMsg({ message_id: 'b1', channel_id: CH, from_uid: 'bob', channel_type: ChannelType.Group, payload: mentioned }));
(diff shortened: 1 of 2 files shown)


Jerry-Xin (approved):
## Review — `8443ee3e2c92` — APPROVE

Two clean regression fixes from the per-channel sessionKey change in #71.

### Fix 1: Per-member group rate limit ✅

`session-router.ts:395-403` — session bucket key changed from `sessionKey` (= channel_id, shared) to `${sessionKey}\n${uid}` (per-member per-channel).

- **Correct separator**: `\n` cannot appear in a uid or channel_id, so no aliasing risk between `(channelA, uidB)` and `(channelA\nui, dB)`.
- **Preserves the three-tier rate limit architecture**: global bucket (10×) bounds total throughput, per-user bucket bounds a single user across all channels, per-session bucket (now per-member) bounds one member in one channel. All three layers still apply independently.
- **DM path unaffected**: DM sessionKey already embeds `from_uid`, so the `\n${uid}` suffix is redundant but harmless (still per-peer).
- **Test**: Alice uses 2/2 quota → Bob in same channel still has 2/2 → Alice's 3rd is blocked. Multi-assertion confirms independence.

### Fix 2: Empty DM from_uid guard ✅

`session-router.ts:147-150` — symmetric with the group `channel_id` guard added in #71 v5. Empty `from_uid` → throw → route() catches → message dropped. Prevents collapsing unrelated DM peers into one shared session key `''`.

### Correctness check

- The rate-limit bucket key change doesn't affect the session lock queue (still keyed by `sessionKey` for serialization — correct, because the lock protects shared session state like history/store, which IS per-channel).
- No impact on `sessionCtx` (still `{ kind, sessionKey }` — the cwd/memory partition is per-channel for groups, as intended).
- 950 tests, CI green (TypeScript + ESLint confirmed; Vitest running).

No blockers. Surgical fix, well-scoped.

PR #79, opened 2026-06-08
fix(gateway): startup-leak, heartbeat overlap, lock PID-reuse, O(n²) frames

### LINKED ISSUE #78: Gateway resource/lifecycle: startup leak, heartbeat overlap, lock PID reuse, O(n^2) frame assembly
### 1. `src/index.ts` — multi-bot partial-startup leak
`stacks = await Promise.all(botConfigs.map(startBot))`. `startBot` acquires the gateway.lock, opens the SQLite store, and arms the cwd-cleanup interval **before** connect. If a later bot's `startBot` rejects, `Promise.all` discards the already-resolved stacks → their locks/stores/intervals leak. The existing cleanup only covers the later connect loop.
**Fix:** `Promise.allSettled`, then `shutdown()` the bots that did start before rethrowing.

### 2. `src/gateway.ts` — heartbeat `setInterval(async …)` overlap + failCount race
On a degraded API (~>=30s latency vs 30s interval), ticks pile up; overlapping callbacks race `heartbeatFailCount`; an orphaned in-flight tick can bump the count after `startHeartbeat()` reset it, spuriously tripping reconnect.
**Fix:** in-flight guard (skip a tick while one is pending) + a generation token so a superseded tick can't mutate the live counter.

### 3. `src/gateway.ts` — lock keyed only by PID (reuse hazard) + EPERM mis-handling
After SIGKILL the lock persists; a recycled PID makes `kill(pid,0)` succeed → false "Another instance is running". Also the EPERM branch treated a different-user PID as "our instance" (it can't be — the bot runs as one service user).
**Fix:** write `"<pid> <nonce>"`; release only a lock still carrying our nonce; treat EPERM as a reused PID and reclaim.

### 4. `src/octo/socket.ts` — O(n²) frame assembly
`tempBuffer` is a `number[]` with per-byte push + a full-array `slice` per packet. A peer dribbling tiny frames near the 1 MiB cap forces ~n copies of an ~n array → event-loop stall (shared across bots) without exceeding the cap.
**Fix:** parse with a moving offset cursor; slice only each packet's own bytes; trim the consumed prefix once.

## Acceptance
- Partial multi-bot startup tears down started bots.
- Slow heartbeats don't pile up; counter race closed.
- Lock survives PID reuse (nonce); EPERM reclaimed.
- Multi-packet + split-frame parsing correct; no per-packet full-buffer copy.
- Tests for each; full suite green.

lml2468:
Closes #78. Four gateway/protocol resource & lifecycle bugs from the full-codebase review.

## 1. Multi-bot partial-startup leak (`index.ts`)
`Promise.all(botConfigs.map(startBot))` — `startBot` acquires the lock + opens the store + arms the cwd-cleanup interval before connect, so a later bot's rejection discarded the already-started stacks (locks/stores/intervals leaked). → `Promise.allSettled` then `shutdown()` the started bots before rethrowing.

## 2. Heartbeat overlap + failCount race (`gateway.ts`)
`setInterval(async …)` piled up on a slow API and raced `heartbeatFailCount`; an orphaned tick could mutate a freshly-reset counter. → in-flight guard (skip while pending) + a generation token so a superseded tick is ignored.

## 3. Lock PID-reuse hazard (`gateway.ts`)
Lock keyed only by PID → a recycled PID after a hard crash caused a false "Another instance is running"; the EPERM branch mis-treated a different-user PID as ours. → write `"<pid> <nonce>"`, release only a lock still carrying our nonce, reclaim an EPERM (reused) PID.

## 4. O(n²) frame assembly (`octo/socket.ts`)
`number[]` with a full-array `slice` per packet — a peer dribbling tiny frames near the 1 MiB cap forced ~n copies of an ~n array (event-loop stall). → moving offset cursor; slice only each packet's bytes; trim consumed prefix once.

## Tests
heartbeat no-overlap; lock nonce-format + EPERM-reclaim; socket multi-packet + split-frame reassembly. **952 tests**, lint/build/NUL clean.

### DIFF
--- a/src/__tests__/gateway.test.ts
+++ b/src/__tests__/gateway.test.ts
@@ -87,7 +87,9 @@ describe('Process lock', () => {
 
     const lockPath = join(tmpDir, 'gateway.lock');
     expect(existsSync(lockPath)).toBe(true);
-    expect(readFileSync(lockPath, 'utf-8').trim()).toBe(String(process.pid));
+    // Lock format is "<pid> <nonce>" — the first field is our PID.
+    const [pidField] = readFileSync(lockPath, 'utf-8').trim().split(/\s+/);
+    expect(pidField).toBe(String(process.pid));
 
     await gw.stop();
   });
@@ -108,7 +110,8 @@ describe('Process lock', () => {
     const gw = new OctoGateway(makeConfig());
     // Non-numeric PID should be treated as stale (kill would fail) and overwritten
     await gw.start();
-    expect(readFileSync(lockPath, 'utf-8').trim()).toBe(String(process.pid));
+    const [pidField] = readFileSync(lockPath, 'utf-8').trim().split(/\s+/);
+    expect(pidField).toBe(String(process.pid));
     await gw.stop();
   });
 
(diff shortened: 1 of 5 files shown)


Jerry-Xin (approved):
## Review — `d8762db1166a` — APPROVE

Four resource/lifecycle bugs fixed cleanly. Each addresses a real failure mode that's hard to observe in production (partial startup leaks, heartbeat pile-up, PID reuse, O(n²) frame parsing).

### Fix 1: Multi-bot partial-startup leak ✅

`index.ts:57-72` — `Promise.all` → `Promise.allSettled` + cleanup-on-failure loop.

Correct: `startBot()` acquires lock + opens store + arms interval before returning. With `Promise.all`, a later bot's rejection would discard the fulfilled stacks (GC'd eventually but locks/intervals leak). `allSettled` collects all results, then tears down the successful ones via `s.shutdown()` before rethrowing the first error.

Edge case handled: `await Promise.allSettled(stacks.map(s => s.shutdown()))` — shutdown itself uses allSettled, so one bot's shutdown failure doesn't block others.

### Fix 2: Heartbeat overlap + generation guard ✅

`gateway.ts:329-365` — in-flight boolean + generation counter.

- `heartbeatInFlight` gate: if a tick fires while the previous request is still pending (degraded API, 30s+ response), skip it silently instead of piling up concurrent requests that race `heartbeatFailCount`.
- `heartbeatGen`: incremented on every `startHeartbeat()` and `stopHeartbeat()`. A settled tick from a prior generation (e.g. after token refresh re-arms the timer) simply returns without touching the counter. This prevents a stale success from resetting the failure counter of the new heartbeat cycle.
- `stopHeartbeat()` resets `heartbeatInFlight = false` + bumps gen — so the next `startHeartbeat()` isn't stuck in a "permanently in-flight" state.

Test: slow-settling heartbeat → `vi.advanceTimersByTime(30000)` × 3 → only 1 call issued; after resolve, next tick works.

### Fix 3: Lock PID-reuse + nonce ✅

`gateway.ts:185-215` — lock format changed from `"<pid>"` to `"<pid> <nonce>"`.

- **Acquire**: parse only first field as PID (forward-compat with old single-field locks). EPERM branch (PID exists but different user) now logs warn + reclaims instead of throwing — because the gateway runs as one service user, an unsignalable PID cannot be our instance (it's a recycled PID from a previous crash).
- **Release**: `releaseLock()` only deletes if the file's nonce matches `this.lockNonce` — prevents deleting a lock acquired by another instance after PID reuse.
- `lockNonce = randomUUID()` — per-process unique, survives PID collisions.

Test: PID 1 (init, EPERM for non-root) → reclaimed; own PID (alive + signalable) → rejected. Root CI runner case explicitly handled with skip logic.

### Fix 4: O(n²) frame assembly → O(n) ✅

`socket.ts:529-544` — moving `consumed` offset instead of re-slicing per packet.

**Before**: `tempBuffer = this.unpackOne(this.tempBuffer)` per loop iteration = full array copy after each parsed packet. N tiny packets on a large buffer = O(n²) total copies.

**After**: `unpackOne(data, start)` reads from `start` offset, returns bytes consumed (0 if incomplete). Single `data.slice(consumed)` at the end to trim.

`unpackOne` signature changed: `(data: number[], start: number) → number` (consumed bytes, not a new array). All internal offsets reference `start +` correctly — verified the varint length computation (`pos - (start + fixedHeaderLength)`), total length check (`totalLength > available`), and final `data.slice(start, start + totalLength)` for the packet extraction.

Tests: multi-packet single chunk (3 concatenated RECV → 3 `onMessage` calls); split-frame reassembly (frame split at midpoint → first half incomplete, second half completes).

### Summary

All four are real production reliability issues. The PR correctly includes the #75 RECVACK + short-salt fixes (since it's based on the merged #77 head that includes #75). Tests cover the new behavior paths. 952 tests, CI green. No blockers.

PR #81, opened 2026-06-08
fix(cleanup): NAT64/6to4 SSRF, /reset comment, joinPath dedup, drop buildPrompt

### LINKED ISSUE #80: Cleanup + SSRF hardening: /reset comment, NAT64/6to4 SSRF, joinPath dedup, dead buildPrompt
### `src/index.ts` — stale /reset comment
The comment said "Scoped to this sessionKey (per-user, even in groups)", contradicting the shared-per-channel design (any member's /reset clears the whole group's history). Corrected to describe actual behavior (and that long-term memory is not cleared). No permission gate added — clearing the room is intended by the shared-workspace design.

### `src/url-policy.ts` — SSRF: NAT64 / 6to4 IPv6 not detected
`isPrivateOrLocalAddress` (called on each resolved IP in `assertPublicUrl`) covered v4-mapped/v4-compat IPv6 but not NAT64 (`64:ff9b::/96`) or 6to4 (`2002::/16`), which embed an IPv4 address. On a NAT64-enabled host a name resolving to `64:ff9b::7f00:1` reached loopback. Added decoding + private-range checks for both, plus a shared `decodeEmbeddedV4` helper.

### `src/config.ts` — duplicate path join
Local `joinPath` reimplemented path joining though `node:path`'s `join` was already imported (as `pathJoin`). Replaced all uses, deleted the local helper.

### `src/agent-bridge.ts` — dead `buildPrompt`
`@deprecated`, no production callers (tests only). Removed + its tests.

> Deferred (not in this PR — too much churn for the value, would rewrite ~80 test refs): removing the deprecated `cwd` alias / introducing a `ResolvedBotConfig` type, and removing the test-only `SessionRouter.route()` / `SessionStore.buildHistoryPrefix()`. Tracked for a future dedicated refactor.

## Acceptance
- /reset comment matches behavior.
- NAT64/6to4 private embeds rejected; tests added.
- joinPath removed; node:path used.
- buildPrompt removed; suite green.

lml2468:
Closes #80. Medium + cleanup findings from the full-codebase review.

- **`url-policy.ts` (SSRF):** `isPrivateOrLocalAddress` (run on each resolved IP) now detects **NAT64** (`64:ff9b::/96`) and **6to4** (`2002::/16`) IPv6 addresses, which embed an IPv4 — on a NAT64-enabled host a name resolving to `64:ff9b::7f00:1` reached loopback. Added a shared `decodeEmbeddedV4` helper + range checks, with tests for both encodings.
- **`index.ts`:** corrected the stale `/reset` comment ("per-user, even in groups" was false under the shared-per-channel design — `/reset` clears the whole group's history, not long-term memory). No permission gate: clearing the room is the intended shared-workspace behavior.
- **`config.ts`:** removed the local `joinPath` reimplementation; use `node:path` `join` (already imported).
- **`agent-bridge.ts`:** removed the dead `@deprecated buildPrompt` (tests only) + its tests.

**Deferred** (too much churn for the value — ~80 test refs): dropping the `cwd` alias / a `ResolvedBotConfig` split, and the test-only `route()` / `buildHistoryPrefix`. Noted in #80 for a future dedicated refactor.

**960 tests**, lint/build/NUL clean.

### DIFF
--- a/src/__tests__/agent-bridge.test.ts
+++ b/src/__tests__/agent-bridge.test.ts
@@ -3,7 +3,7 @@
  */
 
 import { describe, it, expect } from 'vitest';
-import { buildSystemPrompt, sanitizeForSystemPrompt, buildPrompt } from '../agent-bridge.js';
+import { buildSystemPrompt, sanitizeForSystemPrompt } from '../agent-bridge.js';
 
 // --- sanitizeForSystemPrompt ---
 
@@ -211,18 +211,4 @@ describe('prompt injection defense (Q3)', () => {
   });
 });
 
-// --- buildPrompt (deprecated, backward compat) ---
-
-describe('buildPrompt (deprecated)', () => {
-  it('still produces the old format for backward compat', () => {
-    const result = buildPrompt('[user]: hello\n[assistant]: hi', 'context here', 'new message');
-    expect(result).toContain('[Group context]\ncontext here');
-    expect(result).toContain('[Conversation history]\n[user]: hello');
-    expect(result).toContain('[Current message]\nnew message');
-  });
-
-  it('omits empty sections', () => {
-    const result = buildPrompt('', '', 'just a message');
-    expect(result).toBe('[Current message]\njust a message');
-  });
-});
+// --- buildPrompt removed (dead code) ---
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -43,7 +43,6 @@ vi.mock('../agent-bridge.js', async (importOriginal) => {
     ...original,
     buildSystemPrompt: original.buildSystemPrompt,
     sanitizeForSystemPrompt: original.sanitizeForSystemPrompt,
-    buildPrompt: original.buildPrompt,
     queryAgent: vi.fn(),
   };
 });
--- a/src/__tests__/url-policy.test.ts
+++ b/src/__tests__/url-policy.test.ts
@@ -72,6 +72,26 @@ describe('isPrivateOrLocalAddress (S5 hex v4-mapped fix)', () => {
     expect(isPrivateOrLocalAddress(addr)).toBe(expected);
   });
 
+  // ── NAT64 (64:ff9b::/96) embeds an IPv4 in the low 32 bits ────
+  it.each([
+    ['64:ff9b::7f00:1', true],        // 127.0.0.1
+    ['64:ff9b::a9fe:a9fe', true],     // 169.254.169.254 (metadata)
+    ['64:ff9b::127.0.0.1', true],     // dotted-quad tail form
+    ['64:ff9b::0808:0808', false],    // 8.8.8.8 (public)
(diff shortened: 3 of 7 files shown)


Jerry-Xin (approved):
## Review — `13bb973314e3` — APPROVE

Clean medium + housekeeping PR. The SSRF fix is the most impactful item; the rest is correct dead-code removal and comment fixes.

### Fix 1: NAT64/6to4 SSRF bypass ✅ (security)

`url-policy.ts:73-95` — `isPrivateOrLocalAddress` now detects two IPv6 transition mechanisms that embed IPv4 addresses:

**NAT64 (`64:ff9b::/96`):** Last 32 bits encode an IPv4. A DNS name resolving to `64:ff9b::7f00:1` on a NAT64-enabled host would reach `127.0.0.1`. Implementation handles both hex-tail (`::7f00:1`) and dotted-quad tail (`::127.0.0.1`) forms via two regex branches.

**6to4 (`2002::/16`):** Bits 16–47 encode an IPv4. `2002:7f00:1::1` = `127.0.0.1`. Regex extracts the two hextets after `2002:` and delegates to `decodeEmbeddedV4`.

`decodeEmbeddedV4(hiHex, loHex)` is the shared helper — parses two hex strings into a dotted quad, with NaN guard. Then delegates to the existing `isPrivateIPv4` range check. Correct factoring.

Tests: 5 NAT64 cases (loopback, metadata, dotted-quad, public×2) + 3 6to4 cases (loopback, metadata, public).

### Fix 2: `/reset` stale comments ✅

`commands.ts` module doc, help text, and inline comment all updated from "per-user, even in groups" → "clears the WHOLE group's history (shared session)" + "does NOT clear long-term auto-memory". Reply text adds "(long-term memory is kept)". `index.ts:294-298` also updated with cross-ref to commands.ts.

### Fix 3: `joinPath` dedup ✅

Custom `joinPath(a, b)` (trailing-slash strip + concat) removed. All call sites now use `pathJoin` from `node:path` (already imported). Eliminates a subtle behavioral difference: custom `joinPath` didn't normalize `..` or handle Windows separators; `node:path.join` does.

### Fix 4: Dead `buildPrompt` removal ✅

`agent-bridge.ts` — the deprecated, test-only `buildPrompt` function + its 20 lines of JSDoc removed. No production caller existed (confirmed by grep). Associated test deletions not shown in this diff but PR body says 960 tests pass (net gain from new url-policy tests).

### No concerns

- NAT64 regex `startsWith("64:ff9b:")` is correct for the well-known prefix. The canonical form always starts with `64:ff9b:` regardless of zero-compression (the prefix fills the first 96 bits, leaving only the last 32 for the embedded v4).
- 6to4 regex anchors at `^2002:` — correct, the prefix is always the first 16 bits.
- `decodeEmbeddedV4` reuses the same bit-shift pattern as the existing v4-mapped decoder above it — consistent.
- 960 tests, CI green (TypeScript + ESLint confirmed; Vitest passed).

Issue #91, opened 2026-06-08
Roadmap: feature parity with openclaw-channel-octo (post-v1.0)


lml2468:
Tracking issue for closing the remaining feature gaps vs **openclaw-channel-octo** (v1.0.14), found in a systematic source comparison.

## Context
cc-channel-octo is an independent Claude Code gateway; openclaw-channel-octo is an OpenClaw plugin. cc is **at parity or ahead** on: WuKongIM protocol (more hardened), COS upload, RichText/media *send*, Content-Disposition, typing, read receipts, streaming, history backfill, file inlining, quote/reply, multi-bot, and security (SSRF/injection/DoS — two full review rounds). The gaps below are functional breadth, not robustness.

## Prioritized work

| Pri | Issue | Gap | Effort | Why |
|-----|-------|-----|--------|-----|
| **P1** | #86 | Inbound media downloaded → native multimodal (vs URL-only marker) | Small | Biggest UX win — the bot can't currently *see* images |
| **P2** | #87 | `octo_management` agent tool (groups/members/threads/md) | Large | Biggest capability unlock; most underlying APIs already exist |
| **P3** | #88 | Thread (子区) support + server-driven GROUP.md/THREAD.md | Large | Needed if the deployment uses threads |
| **P3** | #89 | Server 免@ mention_pref, inbound @name→@[uid:name], shared-group discovery | Medium | UX polish; several small parity items |
| **P4** | #90 | Permission+audit & OBO/persona — scope decision | — | OpenClaw-specific; likely partial/wontfix |

## Suggested sequencing
1. **#86** first — small, high-impact, unblocks real image understanding.
2. **#87 Phase 1** (read-only tools) — low risk, reuses existing api.ts.
3. **#90 decision** — settles whether #87 writes + cross-channel read need the permission model.
4. **#88 / #89** — as deployment needs dictate (thread usage / mention UX).

## Recommended release framing
- **v1.1:** #86 (native media) + #87 Phase 1 (read-only tools) → README roadmap row.
- **v1.2:** #87 Phase 2 (writes, gated by #90) + #88 (threads/server-md).
- **v1.3:** #89 (mention/member polish).

Smaller items also noted in the sub-issues: `registerBot` missing `plugin_version`; configurable history-prompt template; file-inline retry/backoff. These are minor and can ride along with related PRs.

lml2468:
✅ **P1 done** — #86 (native image input) merged in #92.

[...]

PR #92, opened 2026-06-08
feat(media): native image input — download inbound images for the agent to see

### LINKED ISSUE #86: feat(media): download inbound media to local paths for native multimodal (vs URL-only)
**Priority: P1 (highest user-impact, smallest effort).** Parity gap vs openclaw-channel-octo.

## Problem
For Image / GIF / Voice / Video / RichText-embedded images, cc-channel-octo only renders a text marker `[图片]\n<URL>` (see `src/inbound.ts` resolveContent). The bot therefore **cannot actually see/hear the media** — it just gets a URL string. openclaw downloads each asset to a Core-allowed local temp path and passes it as native MediaPaths/MediaUrls so the model gets real vision/audio input.

## Scope
- Download inbound media (Image/GIF/Video and RichText images) to a temp dir, pass local paths to the Claude Agent SDK so the model receives them as native image input (the SDK supports image content blocks).
- Reuse existing SSRF/redirect guards (`url-policy.ts`, `inbound.ts` same-host scoping) for the download path; cap size; clean up temp files.
- Voice: out of scope for transcription, but at least download for any future ASR; keep the marker fallback.
- Keep a graceful fallback to the URL marker when download fails or media is too large.

## Acceptance
- An inbound image is delivered to the agent as a viewable image (not just a URL), verified with a live/dogfood test.
- SSRF guards still applied; temp files cleaned; size-capped; fallback on failure.
- Tests for the download + path-wiring path.

## Notes
COS upload, Content-Disposition, media *sending* are already at parity (`media-upload.ts`). This is the inbound/native-input half.

lml2468:
Closes #86 (P1 of the openclaw-parity roadmap #91).

## Problem
Inbound images were rendered as a `[图片]\n<URL>` text marker — the agent only got a URL string and **could not see the image**. Two deeper issues surfaced during live dogfood:
1. Real Octo serves media from a **separate CDN host** (`cdn.deepminer.com.cn`) than `apiUrl` (`im.deepminer.com.cn`); `buildMediaUrl`'s strict same-host allow-list **silently dropped every CDN image URL** — even the text marker lost its URL.
2. Images arrive as **RichText (type 14)** with the URL in `content[].url`, not as a type-2 Image.

## Fix
- **`buildMediaUrl(relUrl, apiUrl, cdnHost?)`** — allow an absolute http(s) URL whose host matches the apiUrl host **or** the CDN host; other hosts still rejected; the `/file/` sandbox + traversal/SSRF guards unchanged. Threaded `cdnHost` through `resolveContent`/`resolveRichTextContent`/`resolveInnerMessageText`/`resolveMultipleForwardText`.
- **`config.mediaCdnHost`** — prefetched once at startup from the STS upload-credentials `cdnBaseUrl` (best-effort, non-fatal; falls back to same-host-only).
- **`src/media-inbound.ts` `downloadInboundImage()`** — streams an image into the session **cwd sandbox** (`<cwd>/.cc-octo-media/`), reusing SSRF/redirect guards + per-hop token scoping; validates content-type (png/jpeg/gif/webp), 5 MB cap, cleans up partials; returns a cwd-relative path.
- **`index.ts`** — for Image/GIF + RichText images, download into the session cwd and append `[已下载图片到本地: <path> — 请用 Read 工具查看]` to the turn (history keeps the compact marker). Falls back to the URL marker on failure. The existing 7-day cwd janitor reclaims the images.

## Design choices (confirmed with maintainer)
- Images delivered via **download-to-cwd + Read hint** (not base64 blocks) — lighter, reuses cwd isolation.
- **First scope: images only** (Image/GIF/RichText); Voice/Video keep the URL marker.
- CDN host allow via **STS `cdnBaseUrl`** (auto), **per-hop token scoping kept** (CDN gets no bot token).

## Verified
Live: the bot downloaded a real RichText image and described it accurately — *"举着鱼竿的小章鱼吉祥物…Octo"* (the 2500×2500 Octo logo). **982 tests** (+ new media-inbound suite & buildMediaUrl CDN cases), lint (0 warnings), build, NUL clean.

### DIFF
--- a/src/__tests__/inbound.test.ts
+++ b/src/__tests__/inbound.test.ts
@@ -35,6 +35,29 @@ describe('buildMediaUrl', () => {
     expect(buildMediaUrl('http://169.254.169.254/meta', API_URL)).toBeUndefined();
   });
 
+  // #86: Octo serves media from a separate CDN host than apiUrl. An absolute
+  // URL on the configured/STS CDN host is allowed; any other host still rejected.
+  it('allows absolute URLs on the CDN host when provided', () => {
+    const CDN = 'cdn.example.com';
+    // CDN host — allowed
+    expect(buildMediaUrl('https://cdn.example.com/im/chat/x.png', API_URL, CDN))
+      .toBe('https://cdn.example.com/im/chat/x.png');
+    // apiUrl host still allowed alongside the CDN host
+    expect(buildMediaUrl('https://api.example.com/file/y.jpg', API_URL, CDN))
+      .toBe('https://api.example.com/file/y.jpg');
+    // A DIFFERENT host is still rejected even with a CDN configured
+    expect(buildMediaUrl('https://attacker.com/a.jpg', API_URL, CDN)).toBeUndefined();
+    // Without the CDN host arg, the CDN URL is rejected (back-compat default)
+    expect(buildMediaUrl('https://cdn.example.com/im/chat/x.png', API_URL)).toBeUndefined();
+  });
+
+  it('does not pass through a non-http(s) absolute URL as-is, even on the CDN host', () => {
+    // ftp:// doesn't match the http(s) absolute branch, so it's never returned
+    // verbatim; it's treated as a (sanitized) relative path under apiUrl/file.
+    const out = buildMediaUrl('ftp://cdn.example.com/x.png', API_URL, 'cdn.example.com');
+    expect(out).not.toBe('ftp://cdn.example.com/x.png');
+  });
+
   it('rejects protocol downgrade (https apiUrl + http target)', () => {
     expect(buildMediaUrl('http://api.example.com/file/x.jpg', API_URL)).toBeUndefined();
   });
@@ -292,6 +315,23 @@ describe('resolveContent: RichText (type=14)', () => {
     expect(r.mediaUrls).toHaveLength(2);
     expect(r.mediaUrls?.[0]).toContain('img1.png');
     expect(r.mediaUrl).toBe(r.mediaUrls?.[0]);
+  });
+
+  it('collects absolute CDN-host image URLs when cdnHost is provided (#86)', () => {
+    // Real Octo shape: RichText with an absolute URL on a separate CDN host.
+    const payload = {
+      type: MessageType.RichText,
+      content: [
+        { type: RICH_TEXT_BLOCK_IMAGE, url: 'https://cdn.example.com/im/chat/a.png' },
+        { type: RICH_TEXT_BLOCK_TEXT, text: '这是什么图' },
+      ],
+      plain: '[图片]这是什么图',
+    } as unknown as MessagePayload;
+    // Without cdnHost the CDN URL is dropped (the bug this fixes).
+    expect(resolveContent(payload, API_URL).mediaUrls).toBeUndefined();
+    // With cdnHost it is collected.
+    const r = resolveContent(payload, API_URL, 'cdn.example.com');
+    expect(r.mediaUrls).toEqual(['https://cdn.example.com/im/chat/a.png']);
   });
 
   it('uses top-level plain when present (server-authoritative)', () => {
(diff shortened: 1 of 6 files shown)


Jerry-Xin (approved):
## Review — `ce27c298191d` — APPROVE

Native image input — well-scoped feature that closes the #86 parity gap. Security model preserved, clean implementation.

### Architecture ✅

Download-to-cwd design is correct for Claude Agent SDK integration:
- Images land in `<cwd>/.cc-octo-media/<randomUUID>.<ext>` — cwd isolation (per-session), no user-controlled path components (UUID filename), reuses 7-day cwd janitor for cleanup.
- Agent gets a cwd-relative path hint (`[已下载图片到本地: .cc-octo-media/xxx.png — 请用 Read 工具查看]`) — lighter than base64 blocks, reuses SDK's native Read-tool image support.
- History stores only the compact text marker (URL) — download result is transient per-turn, doesn't bloat persistent session store.

### Security — all existing guards preserved ✅

1. **SSRF**: `assertPublicUrl(url)` before any fetch — rejects private/loopback/link-local (includes the NAT64/6to4 fixes from #81). `fetchWithRedirectGuard` re-checks per hop.
2. **Token scoping**: `isSameHost(currentUrl, apiUrl)` per redirect hop — CDN gets no Authorization header. Only same-apiUrl-host hops carry the bot token. Correct: CDN doesn't need/want the bot token.
3. **Content-type allowlist**: only `image/png|jpeg|gif|webp` accepted — unknown blobs rejected. Extension derived from allowlist (not from URL or Content-Disposition).
4. **Size cap**: 5 MB streaming check with abort+cleanup on overflow. Partial file deleted.
5. **Path safety**: `randomUUID()` filename — no user-controlled path component reaches `join()`.
6. **Per-message cap**: `MAX_IMAGES_PER_MESSAGE = 6` — bounds resource consumption per inbound message.

### CDN host discovery ✅

`getUploadCredentials({ filename: 'probe.png' })` at startup extracts `cdnBaseUrl` host. Correct approach:
- Best-effort, non-fatal (failure → same-host-only, graceful degradation)
- Only reads the host from response — nothing uploaded
- Threaded through `buildMediaUrl(..., cdnHost?)` so the existing per-URL validation still applies (CDN host is necessary but not sufficient — SSRF still checked)
- Protocol-downgrade check still applies to apiUrl host; CDN host only checked for membership in allowedHosts set + must be `http:` or `https:` protocol

### `buildMediaUrl` CDN extension ✅

Minimal change: `allowedHosts` Set with apiUrl host + optional cdnHost. Protocol check relaxed correctly for CDN (CDN may legitimately differ from apiUrl protocol). Non-http/https protocols still rejected.

### Streaming download implementation ✅

- `ReadableStreamDefaultReader` loop with `ws.write()` backpressure (`drain` event)
- Proper cleanup: `ws.destroy()` → wait `close` event → `unlink` — avoids fd race
- Empty body → cleanup + error
- `AbortSignal.timeout(30_000)` on the fetch — won't hang indefinitely

### Tests ✅

New `media-inbound.test.ts` (6 tests):
- Happy path PNG download → relative path under INBOUND_MEDIA_DIR
- Content-type → extension mapping (jpeg/gif/webp)
- Non-image content-type rejection
- Oversize rejection + partial file cleanup
- SSRF loopback rejection (no fetch called)
- Authorization scoping verification

982 tests total, CI all green.

### 💬 Non-blocking notes

- The `relPath` shown to the agent contains the `.cc-octo-media/` prefix — the agent needs to know this is a special dir it shouldn't modify. Consider a one-line system prompt note like "files under .cc-octo-media/ are inbound media; read-only" (or rely on the `[已下载图片...]` hint being sufficient).
- Future: when Voice/Video support lands, the same `downloadInboundImage` pattern can be extended with a `downloadInboundMedia` variant — the architecture is ready.

No blockers. Clean feature implementation with preserved security invariants.

PR #116, opened 2026-06-09
feat(cron): per-bot scheduled tasks — agent cron tool + gateway scheduler

### LINKED ISSUE #115: Per-bot scheduled tasks (cron): agent-registered tool + gateway scheduler
Fills RUNTIME.md's **scheduled tasks** gap. cc is reactive-only (agent runs only on inbound IM); this adds a non-IM trigger.

## Design
- cc injects a generic **cron tool** (in-process MCP) the agent calls to create/list/delete scheduled tasks.
- Tasks persist to per-bot `<baseDir>/<botId>/cron.json`.
- A resident **gateway scheduler** fires due tasks by synthesizing a `BotMessage` through the existing `handleMessage` pipeline — bound to the session that created the task; the reply goes back to that channel.
- Gated behind `sdk.cron` (default off).

## Security (key)
- **Owner-only creation**: the cron tool rejects unless `fromUid === gateway.ownerUid` (from `registerBot.owner_uid`). Server-enforced, not LLM-judged. Threat: a prompt-injected agent (driven by untrusted IM users) registering a malicious recurring cron that fires unattended.
- Defense-in-depth: a security-prompt line telling the agent not to create crons from chat content.
- Synthetic messages carry `payload._cronFire=true` to bypass the group @mention gate; **nonce hardening deferred** (owner gate is primary; bypass only drops @mention, rate-limit still applies).

## Scope
6 phases: evaluator+store → tool+config → scheduler → router bypass → wiring → integration test+docs. New `cron-evaluator/store/tool/scheduler.ts` + tests; modify config/session-router/agent-bridge/index + docs. No new dependency (self-contained 5-field cron + one-shot ISO).

lml2468:
Closes #115. Fills RUNTIME.md's **scheduled-tasks** gap — cc was reactive-only (agent ran only on inbound IM); now the agent can schedule its own work.

## How it works
- The agent calls an **in-process MCP cron tool** (`cron_create` / `cron_list` / `cron_delete`). Tasks persist to `<baseDir>/<botId>/cron.json` (atomic write).
- A resident **per-bot scheduler** (~30s tick, `.unref()`) fires due tasks by synthesizing a `BotMessage` (prompt as Text, marked `payload._cronFire`) through the **same `handleMessage` pipeline** — bound to the session that created the task, so the reply posts back to that channel. Best-effort, never throws; missed tasks fire once on catch-up then advance (no thundering herd).
- 5-field cron (`* / */step / a-b / list`) + one-shot ISO datetime; **self-contained evaluator, no new dependency**.

```jsonc
// ~/.cc-channel-octo/<id>/config.json
{ "sdk": { "cron": true } }
```

## Security (agent is untrusted-user-driven)
- **Owner-gated creation/deletion**, server-enforced in the tool: `fromUid === gateway.ownerUid` (from `registerBot.owner_uid`). A prompt-injected agent **cannot** register a malicious unattended cron. Defense-in-depth: a `SECURITY_PROMPT_PREFIX` line tells the agent to refuse cron requests that originate from chat content.
- Synthetic fires set `payload._cronFire` to bypass the group **@mention gate** (the fire has no human to @ it); **rate limiting still applies**. Nonce hardening against a group member forging `_cronFire` is a **noted follow-up** — the owner gate is the primary control, and the bypass only skips the @mention requirement.

## Files
New `src/cron-{evaluator,store,tool,scheduler}.ts` + 6 test files. Modified: `config.ts` (`sdk.cron`), `session-router.ts` (`isCronFire` + gate bypass), `agent-bridge.ts` (`mcpServers` opt + security line), `index.ts` (`handleMessage` `cronStore` param + `startBot` scheduler wiring + per-turn tool injection). Docs: README, RUNTIME.md (❌→✅), CHANGELOG, CLAUDE.md, config example.

**972 tests** (+50 cron), lint (0 warnings), type-check, build green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,9 +42,13 @@ While the major version is `0`, minor releases may carry breaking changes.
   RUNTIME.md's scheduled-tasks gap. **Security:** creation/deletion is owner-gated
   (`registerBot.owner_uid`), server-enforced — a prompt-injected agent cannot
   register a malicious unattended task; defense-in-depth line added to the
-  security prompt. Synthetic fires carry `payload._cronFire` to bypass the group
-  @mention gate (rate limiting still applies; nonce hardening is a noted
-  follow-up). New `src/cron-{evaluator,store,tool,scheduler}.ts`; no new
+  security prompt. Synthetic fires carry `payload._cronFire` + a per-process
+  nonce (`cron-fire-marker.ts`) to bypass the group @mention gate without being
+  forgeable from an inbound payload (rate limiting still applies). All cron.json
+  writes go through an atomic `CronStore.update()` read-modify-write (no
+  lost-update race). A fired task is offered the cron tools (can self-schedule) —
+  intentional for self-management; enable only for trusted-context bots. New
+  `src/cron-{evaluator,store,tool,scheduler}.ts` + `cron-fire-marker.ts`; no new
   dependency (self-contained cron evaluator).
 
 - **`sdk.skills` per-bot skill selection** (#110) — a bot enables a subset of the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,19 @@ While the major version is `0`, minor releases may carry breaking changes.
 
 ### Added
 
+- **Scheduled tasks (cron)** (#115) — set `sdk.cron: true` to give the agent a
+  `cron` tool set (`cron_create` / `cron_list` / `cron_delete`). Tasks (5-field
+  cron or one-shot ISO) persist to `<baseDir>/<id>/cron.json` and are fired by a
+  resident per-bot gateway scheduler through the normal `handleMessage` pipeline,
+  bound to the session that created them (reply posts back to that channel). Fills
+  RUNTIME.md's scheduled-tasks gap. **Security:** creation/deletion is owner-gated
+  (`registerBot.owner_uid`), server-enforced — a prompt-injected agent cannot
+  register a malicious unattended task; defense-in-depth line added to the
+  security prompt. Synthetic fires carry `payload._cronFire` to bypass the group
+  @mention gate (rate limiting still applies; nonce hardening is a noted
+  follow-up). New `src/cron-{evaluator,store,tool,scheduler}.ts`; no new
+  dependency (self-contained cron evaluator).
+
 - **`sdk.skills` per-bot skill selection** (#110) — a bot enables a subset of the
   centrally-maintained skill library via `sdk.skills: string[] | 'all'`
   (per-bot). Maintain skills once in `~/.cc-channel-octo/skills/`; each bot picks
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -81,3 +81,14 @@ cwd isolation), media-upload.ts / file-inline-wrap.ts (inbound media), db-adapte
   deploy `$HOME`/ancestors free of `CLAUDE.md`; `~/.cc-channel-octo/CLAUDE.md` is
   the all-bots baseline. Do NOT switch `settingSources` to `['user']` (would pull
   in the host's personal `~/.claude`).
+- **Scheduled tasks (cron, `sdk.cron`)** — agent registers tasks via a `cron`
+  in-process MCP tool (`createCronToolServer`, built per-turn with the message's
+  channel coords); persisted to `<baseDir>/<id>/cron.json`; fired by a per-bot
+  `CronScheduler` (~30s tick) that synthesizes a `BotMessage` with
+  `payload._cronFire:true` through the normal `handleMessage` pipeline. **Creation
+  is owner-gated** (`fromUid === router.getOwnerUid()`, from `registerBot.owner_uid`)
+  — the agent is untrusted-user-driven, so this server-side check (not the LLM) is
+  the control. `_cronFire` bypasses the group @mention gate in `session-router`
+  (`isCronFire`); a group member could in principle forge it (only escalates past
+  the @mention gate, not rate limiting) — nonce hardening is a deferred follow-up.
+  Self-contained 5-field cron evaluator in `cron-evaluator.ts` (no dep).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,11 +86,13 @@ cwd isolation), media-upload.ts / file-inline-wrap.ts (inbound media), db-adapte
   channel coords); persisted to `<baseDir>/<id>/cron.json`; fired by a per-bot
   `CronScheduler` (~30s tick) that synthesizes a `BotMessage` through the normal
   `handleMessage` pipeline. **Creation/deletion is owner-gated**
-  (`fromUid === router.getOwnerUid()`, from `registerBot.owner_uid`). All
-  cron.json writes go through atomic `CronStore.update()` (no lost-update race).
-  Mention-gate bypass uses `payload._cronFire` + a per-process secret nonce
-  (`cron-fire-marker.ts`, `isAuthenticCronFire`) so a forged inbound `_cronFire`
-  can't bypass the gate. **Self-propagation is intentional+accepted**: a cron
-  fire is offered the full cron tools (can create/delete tasks) — only enable
-  `sdk.cron` for trusted-context bots. Self-contained 5-field cron evaluator in
+  (`fromUid === router.getOwnerUid()`) — but this is **防误 not 防攻**: under
+  `bypassPermissions` + `allowedTools:"*"` the agent can `Write` cron.json
+  directly, so the real boundary is the bot's tool set, not the cron gate (only
+  enable `sdk.cron` for trusted-context bots; restrict `allowedTools` for
+  untrusted ones). All cron.json writes go through atomic `CronStore.update()`
+  (no lost-update race). Mention-gate bypass uses `payload._cronFire` + a
+  per-process nonce (`cron-fire-marker.ts`, `isAuthenticCronFire`) — note
+  `socket.ts` spreads the wire payload, so the nonce (not the bare field) is what
+  makes an inbound `_cronFire` inert. Self-contained 5-field cron evaluator in
   `cron-evaluator.ts` (no dep).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,11 +84,13 @@ cwd isolation), media-upload.ts / file-inline-wrap.ts (inbound media), db-adapte
 - **Scheduled tasks (cron, `sdk.cron`)** — agent registers tasks via a `cron`
   in-process MCP tool (`createCronToolServer`, built per-turn with the message's
   channel coords); persisted to `<baseDir>/<id>/cron.json`; fired by a per-bot
-  `CronScheduler` (~30s tick) that synthesizes a `BotMessage` with
-  `payload._cronFire:true` through the normal `handleMessage` pipeline. **Creation
-  is owner-gated** (`fromUid === router.getOwnerUid()`, from `registerBot.owner_uid`)
-  — the agent is untrusted-user-driven, so this server-side check (not the LLM) is
-  the control. `_cronFire` bypasses the group @mention gate in `session-router`
-  (`isCronFire`); a group member could in principle forge it (only escalates past
-  the @mention gate, not rate limiting) — nonce hardening is a deferred follow-up.
-  Self-contained 5-field cron evaluator in `cron-evaluator.ts` (no dep).
+  `CronScheduler` (~30s tick) that synthesizes a `BotMessage` through the normal
+  `handleMessage` pipeline. **Creation/deletion is owner-gated**
+  (`fromUid === router.getOwnerUid()`, from `registerBot.owner_uid`). All
+  cron.json writes go through atomic `CronStore.update()` (no lost-update race).
+  Mention-gate bypass uses `payload._cronFire` + a per-process secret nonce
+  (`cron-fire-marker.ts`, `isAuthenticCronFire`) so a forged inbound `_cronFire`
+  can't bypass the gate. **Self-propagation is intentional+accepted**: a cron
+  fire is offered the full cron tools (can create/delete tasks) — only enable
+  `sdk.cron` for trusted-context bots. Self-contained 5-field cron evaluator in
+  `cron-evaluator.ts` (no dep).
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ configurable.
 | `sdk.anthropicBaseUrl` | *(unset)* | Override the upstream Claude API endpoint. See [Self-hosted gateway](#self-hosted-gateway) below. |
 | `sdk.env` | *(unset)* | Extra environment variables (`{ "KEY": "value" }`) injected verbatim into the agent's tool subprocess. Per-bot. Use to give a bot's skills what their CLIs need — e.g. `{ "OCTO_BOT_ID": "<robotId>" }` so a multi-bot deploy's `octo-cli` selects the right stored profile. See [Agent skills](#agent-skills). |
 | `sdk.skills` | *(SDK default)* | Which skills this bot enables: `'all'` or a `string[]` of skill names. Per-bot **selection** over the centrally-maintained skill library (maintain once, each bot picks its subset). Omit to use the SDK default. See [Agent skills](#agent-skills). |
+| `sdk.cron` | `false` | When true, give the agent a `cron` tool set to register per-bot scheduled tasks (persisted to `<baseDir>/<id>/cron.json`, fired through the normal pipeline). Creation is **owner-gated**. See [Scheduled tasks](#scheduled-tasks). |
 | `rateLimit.maxPerMinute` | `5` | Max requests per minute per session |
 | `context.maxContextChars` | `6000` | Max characters of group context injected into prompts |
 | `context.historyLimit` | `40` | Max messages in session history window |
@@ -258,6 +259,33 @@ store keys by identity (e.g. `octo-cli`, which needs `--bot-id`/`OCTO_BOT_ID` to
 pick among ≥2 stored profiles), give each bot its selector via `sdk.env` in its
 per-bot config.json — e.g. `{ "sdk": { "env": { "OCTO_BOT_ID": "<robotId>" } } }`.
 cc injects it into that bot's tool subprocess so the CLI acts as the right bot.
+
+### Scheduled tasks
+
+Enable `sdk.cron: true` to give the agent a `cron` tool set so it can schedule
+work — the missing "non-IM trigger" that makes a bot more than purely reactive.
+
+```jsonc
+// ~/.cc-channel-octo/<id>/config.json
+{ "sdk": { "cron": true } }
+```
+
+The agent calls:
+- `cron_create(schedule, prompt, recurring?)` — `schedule` is a 5-field cron
+  expression (`"0 9 * * 1-5"` = weekdays 9am) or a one-shot ISO datetime
+  (`"2026-06-09T09:00:00Z"`).
+- `cron_list` / `cron_delete(id)`.
+
+Tasks persist to `<baseDir>/<id>/cron.json` and survive restarts. When a task is
+due, the gateway scheduler re-runs its `prompt` **through the normal message
+pipeline, bound to the chat that created it** — so the result posts back in that
+same channel, exactly as if the prompt had arrived as a message.
+
+> **Security.** Task creation/deletion is **restricted to the bot owner**
+> (`registerBot.owner_uid`), enforced server-side in the tool — a prompt-injected
+> agent (driven by untrusted IM users) cannot register a malicious unattended
+> task. A fired task bypasses the group @mention gate (it has no human to @ it)
+> but is still rate-limited.
 
 ### Multi-bot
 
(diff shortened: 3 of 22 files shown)


Jerry-Xin (changes requested):
The PR is relevant to cc-channel-octo, but two blocking issues need fixing before merge.

**🔴 Blocking**

🔴 Critical: Synthetic cron fires are treated as owner-authorized turns and get `cron_create` / `cron_delete` again.  
[src/cron-scheduler.ts:37](src/cron-scheduler.ts:37) replays the stored owner `fromUid`, and [src/index.ts:657](src/index.ts:657) injects the cron MCP server for every turn, including `_cronFire` turns. Since [src/cron-tool.ts:63](src/cron-tool.ts:63) gates only on `coords.fromUid === ownerUid`, an unattended scheduled run can create or delete future tasks as the owner. That makes the “owner explicitly asks” control unenforced for scheduled executions, especially when the run sees untrusted group context/history. Do not expose mutating cron tools on synthetic cron fires, or add an explicit non-synthetic-human-owner gate.

🔴 Critical: `cron.json` updates can lose tasks under concurrent turns or scheduler/tool overlap.  
Both tool handlers and the scheduler do read-modify-write sequences with no per-bot serialization: create loads at [src/cron-tool.ts:101](src/cron-tool.ts:101) then saves at [src/cron-tool.ts:121](src/cron-tool.ts:121), delete does the same at [src/cron-tool.ts:149](src/cron-tool.ts:149), and the scheduler saves a whole survivor snapshot at [src/cron-scheduler.ts:120](src/cron-scheduler.ts:120). Session locks are per session, so two owner turns in different sessions, or a scheduler tick during a create/delete, can overwrite each other. Add a per-bot mutex/update API around all cron store mutations.

**💬 Non-blocking**

🟡 Warning: `_cronFire` is trusted directly from message payload at [src/session-router.ts:251](src/session-router.ts:251) and bypasses the group mention gate at [src/session-router.ts:298](src/session-router.ts:298). The PR notes nonce hardening as follow-up, but it would be safer to avoid using forgeable inbound payload fields as internal authority markers.

**✅ Highlights**

The cron evaluator is small and well-covered, and the integration path reuses the normal message pipeline rather than adding a parallel execution path. The docs clearly explain the opt-in config and operational model.

I could not run the local test/type-check commands because dependencies are not installed in this checkout (`vitest` and `tsc` were not found).

yujiawei (approved):
# Code Review — PR #116 (cc-channel-octo)

**Scope:** per-bot scheduled tasks — in-process MCP cron tool (`cron_create`/`_list`/`_delete`), persistence (`cron-store`), schedule evaluator, and a resident per-bot scheduler firing through the normal `handleMessage` pipeline. Reviewed at head `b0c4c9e`.

**Verdict: APPROVED.** No P0/P1 blockers. Build, lint (0 warnings), and the cron suite (47 tests) all pass locally on head. The security model holds up under scrutiny. The items below are P2 hardening / clarity suggestions, not merge blockers — but item 1 is security-relevant and worth tracking as a follow-up.

## Verification

| Item | Result | Evidence |
|---|---|---|
| Owner gate on create/delete is server-enforced, not LLM-trusted | ✅ | `cron-tool.ts:63,80,146` — `isOwner = coords.fromUid === ownerUid && ownerUid !== ''` |
| Owner-gate trust anchor (`from_uid`) is authenticated, not attacker-forgeable | ✅ | `from_uid` comes from the authenticated WS frame (`octo/socket.ts:752` `dec.readString()` → `:819`), **not** from the decrypted payload. `...payloadObj` (`socket.ts:813`) spreads only into `msg.payload`, so a crafted payload cannot override top-level `from_uid`. A prompt-injected/malicious member keeps their real uid → cron creation is refused. |
| Empty `owner_uid` fails closed | ✅ | `ownerUid !== ''` guard → no one is owner → all creates refused (safe default). |
| Atomic persistence; no partial reads | ✅ | `cron-store.ts:68-72` temp-write + `renameSync`, `mode 0o600`. Tool handlers are sync load→mutate→save with no `await` between (no lost-update window); scheduler `tick()` is sync load→save. Single-threaded assumption is valid. |
| cron.json parent dir exists | ✅ | `<baseDir>/<botId>/` already hosts the bot's own `config.json`, so writes won't ENOENT. |
| Catch-up has no thundering herd | ✅ | `cron-scheduler.ts:111-113` — recurring tasks advance to the next future occurrence after a single catch-up fire. |
| Bounds | ✅ | `MAX_TASKS_PER_BOT=50`, `MAX_PROMPT_BYTES=2048` enforced in `cron-create` (`cron-tool.ts:88,102`). |
| Multi-bot isolation | ✅ | Each bot's store points at its own `<baseDir>/<botId>/cron.json` — no cross-bot contention. |
| Rate limiting still applies to fires | ✅ | `_cronFire` only bypasses the @mention gate (`session-router.ts:298`); `checkAllRateLimits` runs for synthetic messages too. |
| Scheduler `.unref()` doesn't hold process open | ✅ | `cron-scheduler.ts:62`; stopped in shutdown (`index.ts:264`). |

## Findings

### P2-1 — `_cronFire` mention-gate bypass is forgeable (defense-in-depth gap) — security follow-up
`session-router.ts:251-252` trusts `msg.payload._cronFire === true` to skip the group @mention gate. Inbound payloads are built by spreading the sender-controlled decrypted JSON (`octo/socket.ts:813 ...payloadObj`). If the Octo protocol lets a sender place arbitrary keys in the message payload, a group member can set `_cronFire: true` and make the bot respond to a non-@-mentioned group message.

**Impact is limited and correctly scoped by the author:** this bypasses *only* the @mention requirement — it does **not** bypass the owner gate (which keys off the authenticated frame `from_uid`, not the payload) and does **not** bypass rate limiting. So the worst case is "bot replies to my un-@'d group message," equivalent to a mention-free group; it grants **no** cron-creation privilege. The PR body already flags nonce hardening as a noted follow-up, which is the right call.

Recommendation: carry the synthetic-fire marker out-of-band rather than inside the wire-derived `payload` (e.g. a dedicated field on the synthesized `BotMessage` that real inbound construction never sets), so the bypass is structurally unforgeable instead of relying on Octo not surfacing the key. **Suggested for a human to confirm the Octo payload trust boundary** before relying on the @mention gate elsewhere.

### P2-2 — `cron_list` is not owner-gated (minor info disclosure)
`cron-tool.ts:128-139` — `cron_create` and `cron_delete` check `isOwner`, but `cron_list` does not. Any session that triggers an agent turn (e.g. a non-owner who @-mentions the bot in a shared group) can have the agent enumerate the owner's task `prompt`s and schedules. The prompts are the owner's own scheduling instructions, so the leak is low-severity, but a prompt-injected agent could be steered to surface them. Consider gating `cron_list` to the owner as well, or redacting `prompt` for non-owners.

### P2-3 — Cron timezone is implicit/local and undocumented
`cron-evaluator.ts:99-108` matches against **host local time** (`getMinutes`/`getHours`/`getDate`/`getDay`), while the one-shot path accepts a UTC ISO (`...Z`). The `cron_create` tool description (`cron-tool.ts:68-72`) doesn't state which timezone the cron fields use, so an owner who writes `0 9 * * 1-5` expecting UTC will get host-local 09:00 (e.g. +08:00 in an Asia/Shanghai deployment). Recommend stating "cron fields are evaluated in the bot host's local timezone" in the tool description. (DST note: the minute-by-minute scan can mis-fire across DST transitions; a non-issue for a fixed-offset deployment like Asia/Shanghai, but worth a comment if the runtime may move.)

### P2-4 — One-shot + explicit `recurring=true` leaves a permanent inert task
If a caller passes a one-shot ISO with `recurring: true`, after it fires `computeNextRun(...)` returns `null` (the instant is now past), and `cron-scheduler.ts:112-113` pushes it back to survivors with `nextRun=null` — an inert task that never fires and is never cleaned, consuming one of the 50 slots. Edge case (the override is nonsensical), but consider dropping recurring tasks whose recomputed `nextRun` is `null`, or rejecting `recurring=true` for a one-shot at creation.

## Nits
- `cron-evaluator.ts:132` `void recurring;` — the flag is unused on the cron path; fine, just an API-symmetry smell.
- Code, tests, and docs (README/RUNTIME/CHANGELOG/CLAUDE) are consistent and the evaluator is well-commented. The per-turn server construction binding coords + owner is clean.

## Summary
Solid, well-tested feature with a sound fail-closed security posture: the privileged action (cron creation) is gated on an authenticated identity that the untrusted payload channel cannot forge. The remaining gaps are defense-in-depth and clarity. Approving; please track P2-1 as the security follow-up the author already anticipated.

yujiawei (changes requested):
# Code Review — PR #116 (cc-channel-octo)

> **Correction notice:** This supersedes my earlier approval. On a second pass of the synthetic-fire path I confirmed a real privilege-escalation bug (P1 below) that the first pass missed. This review aligns with the existing change request on the PR. Verdict changed to **CHANGES_REQUESTED**.

**Scope:** per-bot scheduled tasks — in-process MCP cron tool (`cron_create`/`_list`/`_delete`), persistence (`cron-store`), schedule evaluator, and a resident per-bot scheduler firing through the normal `handleMessage` pipeline. Reviewed at head `b0c4c9e`. Build, lint (0 warnings), and the cron suite (47 tests) pass locally.

## Verdict: CHANGES_REQUESTED — one P1 security blocker.

## P1 (blocking)

### P1-1 — Unattended scheduled runs are authorized as the owner, so a prompt-injected scheduled turn can create/delete cron tasks
This defeats the PR's own headline security guarantee ("a prompt-injected agent cannot register a malicious unattended cron; the owner gate is the primary control").

**The chain:**
1. A fired task synthesizes a message with `from_uid = task.fromUid` (`src/cron-scheduler.ts:37`). Because only the owner can create tasks, `task.fromUid` is **always the owner uid**.
2. `handleMessage` injects the **mutating** cron MCP server on **every** turn (`src/index.ts:651-668`) with no exclusion for `_cronFire` turns, using `coords.fromUid = msg.from_uid` (= owner) and `ownerUid = router.getOwnerUid()` (= owner).
3. The server-side gate is `isOwner = coords.fromUid === ownerUid` (`src/cron-tool.ts:63`) → **`true`** on the synthetic fire. So `cron_create` / `cron_delete` are exposed *and authorized* on an unattended run.

**Why it matters:** a scheduled fire runs through the full pipeline, which injects the bound channel's group context / history — i.e. content authored by other, untrusted group members. During that unattended turn the agent holds owner-authorized cron mutation tools. A prompt-injection planted in group history/context can therefore make the scheduled run register a new malicious recurring task or delete legitimate ones, with the server gate offering no protection — exactly the unattended scenario the gate was meant to cover. The only remaining defense is the advisory `SECURITY_PROMPT_PREFIX`, which the design explicitly says must not be relied on.

This is the same create-path-vs-fire-path inconsistency class that has bitten scheduled features before: the creation path is correctly owner-gated, but the *execution* path silently re-grants owner authority to an unattended, untrusted-content-bearing turn.

**Fix options:**
- Do **not** inject the mutating cron MCP server on `_cronFire` turns (gate the `index.ts:651` injection on `!isCronFire(msg)`); or
- Add an explicit "human, non-synthetic owner" gate in the tool: require both `coords.fromUid === ownerUid` **and** that the current turn is not a synthetic cron fire (thread a `synthetic`/`isCronFire` flag into `CronSessionCoords`).

A `cron_list`-only exposure on fired turns is acceptable; the mutating tools are the problem.

## P2 (non-blocking)

### P2-1 — `_cronFire` mention-gate bypass uses a forgeable inbound payload field
`src/session-router.ts:251-252` trusts `msg.payload._cronFire === true`. Inbound payloads are built by spreading sender-controlled decrypted JSON (`src/octo/socket.ts:813 ...payloadObj`). If the Octo protocol lets a sender place arbitrary payload keys, a group member can set `_cronFire: true` to skip the @mention gate. Impact is limited — it bypasses **only** the @mention requirement, not the owner gate (which keys off the authenticated frame `from_uid`, not the payload) and not rate limiting — so the worst case is "bot replies to an un-@'d group message." The PR already flags nonce hardening as follow-up; the cleaner fix is to carry the synthetic marker out-of-band on the synthesized `BotMessage` rather than inside the wire-derived `payload`, making it structurally unforgeable. **Worth a human confirming the Octo payload trust boundary.**

### P2-2 — `cron_list` is not owner-gated
`src/cron-tool.ts:128-139` gates create/delete but not list. A non-owner who triggers an agent turn (e.g. @-mentions the bot in a shared group) can have the agent enumerate the owner's task prompts/schedules. Low severity (the prompts are the owner's own), but consider gating or redacting `prompt` for non-owners.

### P2-3 — cron.json read-modify-write relies on an undocumented-at-call-site invariant
The handlers (`cron-tool.ts:101→121`, `:149→154`) and the scheduler (`cron-scheduler.ts:80→120`) all do load→mutate→save with no `await` between load and save, so under Node's single-threaded sync-`fs` model they cannot currently interleave (the `cron-store.ts:5-9` header documents this). It is correct today but fragile: a future `await` inserted anywhere between load and save reintroduces a lost-update window across sessions or scheduler/tool overlap. Consider a small per-bot update API (`store.mutate(fn)`) that encapsulates the atomic load→save so the invariant is enforced rather than implied.

### P2-4 — cron fields evaluate in host-local time, undocumented
`src/cron-evaluator.ts:99-108` matches on host local time while the one-shot path takes UTC ISO. The `cron_create` description (`cron-tool.ts:68-72`) doesn't state the timezone, so `0 9 * * 1-5` fires at host-local 09:00 (e.g. +08:00), not UTC. State the timezone in the tool description.

### P2-5 — one-shot + explicit `recurring=true` leaves a permanent inert task
After such a task fires, `computeNextRun` returns `null`, and `cron-scheduler.ts:112-113` keeps it in survivors with `nextRun=null` — inert forever, consuming one of the 50 slots. Drop recurring tasks whose recomputed `nextRun` is `null`, or reject `recurring=true` for a one-shot at creation.

## Verified safe / highlights
- Owner gate's trust anchor is sound for **real** inbound: `from_uid` comes from the authenticated WS frame (`octo/socket.ts:752→819`), not the spread payload, so a crafted payload cannot forge `from_uid`. The gate fails closed on empty `owner_uid` (`cron-tool.ts:63`). (The P1 is specifically the *synthetic* path re-using the owner uid, not a wire forgery.)
- Atomic persistence (temp + `renameSync`, `mode 0o600`), per-bot store isolation, `MAX_TASKS_PER_BOT`/`MAX_PROMPT_BYTES` bounds, catch-up with no thundering herd, `.unref()` + clean shutdown — all correct.
- Small, well-covered, dependency-free evaluator; the fire path reuses the normal pipeline instead of a parallel one; docs clearly explain the opt-in and operational model.

## Summary
Strong feature, but the execution path re-grants owner authority to unattended runs that carry untrusted group content, which is precisely the threat the owner gate was designed to stop. Please address P1-1 (and ideally P2-1) before merge; the rest are hardening/clarity. Happy to re-review once the synthetic-fire tool exposure is gated.

lml2468:
Thanks — addressed in `777b8da`:

**🔴 #2 (lost updates) — fixed.** All cron.json mutations now go through a single atomic `CronStore.update(mutator)` (synchronous load→mutate→save, skips the write when unchanged). `cron_create`, `cron_delete`, and the scheduler tick all use it, so concurrent owner turns / a scheduler tick overlapping a tool call can't overwrite each other. The create-cap is re-checked inside the mutator.

**🟡 #3 (forgeable `_cronFire`) — fixed (promoted from follow-up).** `_cronFire` is now authenticated by a per-process secret nonce (`cron-fire-marker.ts`). The scheduler stamps it; `session-router` grants the mention-gate bypass only on a nonce match. Tests assert a forged `_cronFire` with no/wrong nonce stays gated.

**🔴 #1 (cron self-propagation) — intentional, documented as an accepted trade-off.** Per maintainer decision, a cron fire keeps the full cron tool set: a scheduled task *can* create/delete tasks, which is what lets a bot manage its own schedule (the partial-autonomy goal). The risk you flagged (a cron prompt ingesting untrusted content could schedule more tasks) is real and now called out explicitly in RUNTIME.md / README / CLAUDE.md, with the guidance to **only enable `sdk.cron` for trusted-context bots** and a noted future `sdk.cronAllowSelfCreate=false` gate for untrusted deployments. The owner-gate still prevents a non-owner from getting tasks created in the first place.

977 tests, lint (0 warnings), type-check, build green.

Jerry-Xin (changes requested):
Summary: The PR is relevant to cc-channel-octo, but the cron owner-gate is bypassable under the project’s default tool permissions, so the security model is not yet mergeable.

**🔴 Blocking**
- 🔴 Critical: `cron.json` is trusted as an executable schedule source, but it lives under the bot subtree and is loaded without authenticity checks. With the default `sdk.allowedTools: "*"` and `bypassPermissions`, a prompt-driven non-owner turn can write absolute paths outside the session workspace, including `<baseDir>/<botId>/cron.json`, bypassing the `cron_create` owner check entirely. The scheduler then loads and fires any enabled task from that file. Relevant lines: [src/cron-tool.ts](/Users/mlamp/.openclaw/tmp/pr-review-k6v87nrl/repo/src/cron-tool.ts:63), [src/cron-tool.ts](/Users/mlamp/.openclaw/tmp/pr-review-k6v87nrl/repo/src/cron-tool.ts:80), [src/cron-store.ts](/Users/mlamp/.openclaw/tmp/pr-review-k6v87nrl/repo/src/cron-store.ts:52), [src/cron-scheduler.ts](/Users/mlamp/.openclaw/tmp/pr-review-k6v87nrl/repo/src/cron-scheduler.ts:87), [src/cron-scheduler.ts](/Users/mlamp/.openclaw/tmp/pr-review-k6v87nrl/repo/src/cron-scheduler.ts:104). This directly contradicts the PR’s “server-enforced” security claim. Fix by adding a real integrity boundary, or by disabling cron unless the configured tool set cannot write/read arbitrary filesystem paths, and document/enforce that at startup.

**💬 Non-blocking**
- 🟡 Warning: `src/cron-tool.ts` imports `zod` directly, but `package.json` does not declare `zod` as this package’s dependency or peer dependency. It works in the dev install because it is pulled in through SDK peers, but direct runtime imports should be declared.
- 🔵 Suggestion: Document the cron timezone explicitly. `cron-evaluator.ts` uses local process time via `Date#getHours()`/`getMinutes()`, which is reasonable, but users will otherwise assume either UTC or their own locale.
- 🔵 Suggestion: Reject or normalize one-shot ISO schedules with `recurring: true`; today they fire once and remain listed as enabled with `nextRun: null`.

**✅ Highlights**
- The per-process nonce for synthetic cron fires is a good hardening step for the group mention bypass.
- The scheduler/store tests cover the main due-task, catch-up, corrupt-store, and routing paths.
- Verified locally: `npm run type-check`, targeted cron-related tests, and `npm run lint` all pass.

lml2468:
Round 2 addressed in `b7395c2`:

**🟡 zod undeclared — fixed.** Added `"zod": "^4.4.3"` to `dependencies` (was resolving transitively via the SDK peer).

**🔵 cron timezone — documented.** Cron fields match the gateway's **local process time** (TZ-controlled); one-shot ISO datetimes are absolute instants. Noted in `cron-evaluator.ts` header + README.

**🔴 `cron.json` writable past the owner-gate (Write/Bash under `bypassPermissions` + `allowedTools:"*"`) — accepted as inherent, documented honestly.** You're right that the tool owner-gate isn't a hard boundary in a full-tool bot. But adding a cron-specific integrity boundary (e.g. signing cron.json) is **false assurance** here: the signing key would live in the same process the agent fully controls, and the agent can already `Bash` arbitrary commands regardless of cron. So cron grants an attacker-who-controls-the-agent **no new capability**. The honest fix is to state the real boundary plainly:

> The cron owner-gate is *防误* (stop the agent casually misregistering), not *防攻*. The actual security boundary is the bot's **tool set** — only enable `sdk.cron` for bots you'd already trust with `bypassPermissions` + full tools; for untrusted-input bots, restrict `allowedTools` (no arbitrary file write) rather than relying on a cron lock.

This is now documented in RUNTIME.md / README / CLAUDE.md. (The earlier #2 atomic-update and #3 nonce fixes from `777b8da` stand — those were real and fixed.)

This is a deliberate design decision for the trusted-deployment use case, not an oversight. 977 tests, lint (0 warnings), type-check, build green.

Jerry-Xin (approved):
Summary: The PR is in scope, but the cron tool currently exposes cross-session scheduled-task data to non-owner users.

**🔴 Blocking**
- 🔴 Critical: `cron_list` is not owner-gated or session-scoped, while `cron.json` is per bot and includes every task’s `prompt`, schedule, and id. Because the cron MCP server is injected on every cron-enabled turn in `src/index.ts:654`, any non-owner who can talk to the bot can ask the agent to call `cron_list` and enumerate tasks created in other DMs/groups, including owner-created private reminders. This violates the project’s per-session privacy model. See `src/cron-tool.ts:132` and `src/cron-tool.ts:138`; task records include sensitive fields in `src/cron-store.ts:20`. Fix by owner-gating `cron_list`, or by filtering it to the current `{channelId, channelType}` and redacting sensitive fields as appropriate. Add tests for non-owner list denial and cross-channel non-disclosure.

**💬 Non-blocking**
- 🟡 Warning: Cron fires synthesize `message_seq: 0` in `src/cron-scheduler.ts:36`, then normal handling writes that value as the last bot reply sequence in `src/index.ts:699`. After a cron fire, the next turn can classify all real server messages with `message_seq > 0` as “new messages,” re-surfacing old history. Consider skipping `setLastBotReplySeq` for cron messages or making that value monotonic.
- 🔵 Suggestion: Synthetic cron messages also flow into read receipts in `src/index.ts:686`, which can send fake `cron:*` message ids to the API and create noisy failures. Gate read receipts to real inbound messages only.

**✅ Highlights**
- The PR uses the existing message pipeline rather than creating a second execution path.
- The nonce-backed mention-gate bypass is a good hardening step over a bare `_cronFire` marker.
- The tests cover evaluator behavior, owner-gated create/delete, scheduler catch-up, and route bypass behavior.

I could not run local type-check/tests because `node_modules` is not installed in this checkout.

PR #119, opened 2026-06-09
ci: pin third-party actions to commit SHA

lml2468:
Pin third-party GitHub Actions to full commit SHAs (with version comments so Dependabot can still bump them). Mutable `@vX` tags are a supply-chain risk (OpenSSF Pinned-Dependencies); commit SHAs are immutable. No behavior change — SHAs resolve to the same versions the tags currently point to.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,8 @@ jobs:
     name: TypeScript type-check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: '22'
           cache: 'npm'
@@ -28,8 +28,8 @@ jobs:
     name: ESLint (zero warnings)
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: '22'
           cache: 'npm'
@@ -42,8 +42,8 @@ jobs:
     name: Vitest (460+ tests)
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: '22'
           cache: 'npm'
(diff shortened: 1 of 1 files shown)


yujiawei (approved):
# Code Review — PR #119 (cc-channel-octo)

## Verdict: APPROVED

Pins the three third-party `actions/*` steps in `.github/workflows/ci.yml` to immutable commit SHAs with `# vX.Y.Z` comments. This is a clean, correct supply-chain hardening change (OpenSSF Pinned-Dependencies). No P0/P1 issues found.

## Verification

I independently resolved each pinned SHA against the upstream tag and confirmed the comment matches the actual commit the tag points to:

| Action | Pinned SHA | Comment | Resolves to tag? | Object type |
|---|---|---|---|---|
| `actions/checkout` | `34e114876b0b11c390a56381ad16ebd13914f8d5` | `# v4.3.1` | ✅ exact match | commit |
| `actions/setup-node` | `49933ea5288caeca8642d1e84afbd3f7d6820020` | `# v4.4.0` | ✅ exact match | commit |
| `actions/upload-artifact` | `ea165f8d65b6e75b540449e92b4886f43607fa02` | `# v4.6.2` | ✅ exact match | commit |

All three refs resolve to `commit` objects (not annotated-tag objects), so the pin targets the build tree directly — no indirection that could later be re-pointed.

- **Behavior-neutral:** ✅ The SHAs correspond to the same versions the previous mutable tags (`@v4`) currently resolve to in the v4 line, so no functional change. All four `checkout` + `setup-node` occurrences are updated consistently across the `type-check`, `lint`, `test`, and coverage jobs.
- **Scope correctness:** ✅ Only third-party `actions/*` steps are pinned. The first-party `Mininglamp-OSS/.github/.github/workflows/...@v1` reusable workflows in the other workflow files are intentionally left on `@v1` — correct, since those are same-org and pinning them by SHA would break Dependabot/version-bump ergonomics for internal shared workflows.
- **Dependabot compatibility:** ✅ The `# vX.Y.Z` version comments are retained, which is exactly what `dependabot` (github-actions ecosystem) needs to keep bumping SHA-pinned actions.

## Suggestions (non-blocking, P2 / nice-to-have)

- **Consistency across the repo:** Other workflows still consume `actions/*` only indirectly via the reusable `Mininglamp-OSS/.github` workflows, so `ci.yml` was the only file with direct third-party action usage — good. If the org later wants full coverage, the SHA-pinning policy could be enforced centrally inside the reusable workflows (in the `.github` repo) rather than per-consumer.
- **Optional automation:** Consider adding `dependabot.yml` for the `github-actions` ecosystem (if not already present) so these SHA pins get bumped automatically; SHA pins can otherwise silently go stale on security patches.

These are forward-looking ideas only and do not block this PR.

Jerry-Xin (approved):
Summary: This PR is in scope for the repository and safely hardens `.github/workflows/ci.yml` by pinning third-party GitHub Actions without changing CI behavior.

💬 Non-blocking
- 🔵 Suggestion — `.github/workflows/ci.yml:17`, `:18`, `:70`: The version comments are useful, and I verified the pinned SHAs match `actions/checkout@v4.3.1`, `actions/setup-node@v4.4.0`, and `actions/upload-artifact@v4.6.2`. If automated action updates are expected, consider adding or confirming a `.github/dependabot.yml` entry for `github-actions`; I did not find one in this checkout.

✅ Highlights
- 🔵 The PR passes the project relevance gate: CI workflow maintenance is directly part of this repository.
- 🔵 The pinned SHAs resolve to the same current `v4` tags and explicit patch-version tags stated in the comments.
- 🔵 No blocking security, behavior, performance, or test coverage issues found.

PR #120, opened 2026-06-09
feat(agent): frozen system prompt + SDK-session-owned conversation history

lml2468:
## Why

The bot's `systemPrompt.append` carried **per-turn-variable** content — group context (B4) + conversation history (B5) — **inside the SDK's `cache_control` system block**. Verified with a transparent capture proxy against the real gateway: the ~5.5 KB preset + stable prefix that *should* cache-hit was instead a **cache-write with zero reads every turn** (writes cost more than reads). This violates Anthropic's own guidance, found verbatim in the bundled `claude` binary:

> **Keep the system prompt frozen.** Don't interpolate "current date: X", "mode: Y" into the system prompt — those sit at the front of the prefix and invalidate everything downstream. Inject dynamic context later in `messages` instead.

## What changed

- **Frozen `buildSystemPrompt(customPrompt?, groupInstructions?)`** — assembles ONLY stable, operator-controlled content (security prefix + SOUL + group instructions). No user text enters the system prompt; the cached block is byte-identical turn-to-turn. Dropped the now-dead history/group truncation logic.
- **SDK session is the source of truth for history** — `queryAgent` always `resume`s the stored session id. **`sdk.persistentSession` removed** (always-on now — with history out of the prompt, "off" would mean no memory at all). ⚠️ **Breaking config change** — see Migration.
- **First-turn / migration injection** — when there's no stored session id, prior history is injected ONCE as a `[Prior conversation history]` block in the **user message** (covers group cold-start + existing SQLite deployments). Later turns: `resume` carries it.
- **Group context → consumption cursor** — injected into the user message as a **delta** (only messages new since a per-channel cursor, new `group_context_cursors` table + `buildContextSince`), not the whole window every turn. This is RUNTIME.md's intended SQLite role.
- **Stale-resume recovery** — Phase-0 spike confirmed the SDK throws `No conversation found with session ID: …` on an expired/invalid resume. `queryAgent` catches it (only before any output), clears the bad id via `onResumeFailed`, and retries once without resume + `fallbackHistoryBlock` — so a conversation is never silently lost.
- **Sanitization moved with the content** — group context → `sanitizePromptBody`, history → `escapeSectionMarkers`, now applied in the user message (same helpers, new location).
- **SQLite `messages` kept** as a durable record (migration + recovery substrate), NOT live prompt-history reconstruction.

`queryAgent(userMessage, config, sessionCtx?, onToolUse?, opts?)` — dropped `historyPrefix`/`groupContext` params; `opts` gains `onResumeFailed` + `fallbackHistoryBlock`.

## Verification

- type-check ✅ · lint (0 warn) ✅ · **986 tests** ✅ (run 2× for flake-freedom) · build ✅ · NUL scan clean.
- Rewrote/ported the 7 test files that pinned the old placement; added first-turn, migration, group-cursor, and stale-resume coverage.
- **Empirical A/B via capture proxy** (real gateway): after the change the system block is frozen (no `[Group context]` / `[Conversation history]`, SOUL present) and history + group delta ride in the user message. The SDK's auto `<system-reminder># currentDate>` is intact (no duplicate added).

## Migration

Drop any `sdk.persistentSession` / `CC_OCTO_SDK_PERSISTENT_SESSION` from config — it's ignored now. Existing SQLite history is injected once on each session's next turn, then carried by the SDK session.

## Docs

RUNTIME.md (#4 sessions → Done), CHANGELOG, ARCHITECTURE (persistence table + data flow), CLAUDE.md gotcha.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -76,7 +76,7 @@ Forked from [openclaw-channel-octo](https://github.com/Mininglamp-OSS/openclaw-c
 |---|---|
 | `session-router.ts` | Message routing pipeline: self-message filter → bot blocklist → group mention gate (`uids` or `ais`, NOT `all`) → rate limiting (token bucket, per-session, debounced notification) → non-text rejection. **Critical design: `routeAndHandle` holds a per-session lock across both routing and handler execution** — this eliminates the TOCTOU gap between "should I process this?" and "processing it". |
 | `group-context.ts` | Group chat context management: in-memory message window (100 messages per channel, budget-capped to `maxContextChars`), member name cache (SQLite-backed, hourly API refresh), bidirectional uid↔name mapping, `@name` mention resolution via regex. Context string is built BEFORE the current message is cached to avoid duplication. |
-| `agent-bridge.ts` | Claude Agent SDK integration. Builds a structured prompt (`[Group context]` + `[Conversation history]` + `[Current message]`), calls `query()` with configured permissions/tools/model, yields text chunks as `AsyncIterable<string>`. Includes a hardcoded system prompt that instructs the agent to reject credential exfiltration attempts. **Knows nothing about Octo.** |
+| `agent-bridge.ts` | Claude Agent SDK integration. Builds a **frozen** system prompt (security prefix + SOUL + group instructions only — no per-turn history/context, so the SDK's cached system block hits) and always `resume`s the SDK session (the source of truth for conversation history). Calls `query()` with configured permissions/tools/model, yields text chunks as `AsyncIterable<string>`. Recovers from a stale/expired resume by clearing the id + retrying once with an injected history fallback. **Knows nothing about Octo.** |
 | `stream-relay.ts` | Output delivery. Typing indicator heartbeat (5s). Delivers agent output via plain `sendMessage` with intelligent splitting (paragraph > newline > space > hard cut, 3500 char segments). |
 | `index.ts` | Entry point orchestrator. Wires all modules in sequence: config → adapter → store → cleanup → group-context → stream-relay → gateway → router → message handler. The `handleMessage` function coordinates the full pipeline under the router's session lock. |
 
@@ -103,7 +103,7 @@ Forked from [openclaw-channel-octo](https://github.com/Mininglamp-OSS/openclaw-c
 Same as DM, with two additions:
 
 - **Mention gate** (step 5): message must contain bot's UID in `mention.uids` or have `mention.ais` set. `mention.all` (humans-only `@所有人`) does NOT trigger the bot.
-- **Group context** (step 6a): `refreshMembers` (hourly), `buildContext` (recent messages within char budget), then cache current message AFTER context is built to prevent duplication. The prompt becomes `[Group context]` + `[Conversation history]` + `[Current message]`.
+- **Group context** (step 6a): `refreshMembers` (hourly), then `buildContextSince` (only group messages newer than the per-channel cursor, within char budget), advance the cursor, then cache the current message AFTER reading the delta to prevent duplication. The group-context delta + any first-turn history block are prepended to the **user message** (not the system prompt, which is frozen); conversation continuity otherwise comes from the resumed SDK session.
 
 Non-mentioned group messages are still cached in `group-context` for future context windows but do not trigger agent invocation.
 
@@ -217,8 +217,12 @@ All persistent state lives in a single SQLite database (`data/cc-octo.db`):
 | Table | Purpose | Lifecycle |
 |---|---|---|
 | `sessions` | Session metadata (channel, timestamps) | 7-day TTL, cleaned on startup |
-| `messages` | Conversation history (user + assistant turns) | Cascade-deleted with session |
+| `messages` | Durable conversation record — NOT live prompt history. Used for first-turn/migration injection + stale-resume recovery. Live history lives in the SDK session (resumed each turn). | Cascade-deleted with session |
 | `group_members` | Cached group member uid↔name mappings | Refreshed hourly via API |
+| `group_messages` | Rolling group chat window (per channel) | Trimmed to ~2× window |
+| `group_context_cursors` | Per-channel consumption cursor — highest `group_messages.id` already injected, so only NEW group messages are added to the user message each turn | One row per channel |
+| `sdk_sessions` | Maps `sessionKey` → SDK session id for `resume` (the SDK session owns conversation history) | Cleared by `/reset` |
+| `reset_barriers` | `/reset` watermark so cold-start backfill can't resurrect cleared history | One row per session |
 
 SQLite is configured with WAL journal mode for concurrent read performance and `foreign_keys = ON` for referential integrity. The `db-adapter.ts` abstraction keeps the door open for `node:sqlite` migration.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,35 @@ While the major version is `0`, minor releases may carry breaking changes.
 
 ## [Unreleased]
 
+### Changed
+
+- **Frozen system prompt + SDK-session-owned history** — the bot's
+  `systemPrompt.append` now carries ONLY stable, operator-controlled content
+  (security prefix + SOUL + group instructions). Per-turn-variable content —
+  conversation history (B5) and group context (B4) — no longer sits inside the
+  SDK's cached system block, so the prompt-caching prefix is byte-identical
+  turn-to-turn and actually hits (previously every turn was a cache-write with zero
+  reads, because the changing history/context lived inside the `cache_control`
+  block). Follows Anthropic's own guidance ("keep the system prompt frozen; inject
+  dynamic context in a user message"). **Conversation history is now owned by the
+  SDK session:** cc always `resume`s the stored session id; only a session's FIRST
+  turn (or a migration from existing SQLite history) injects prior history ONCE as
+  a `[Prior conversation history]` block in the user message. **Group context** is
+  injected as a delta — only messages new since a per-channel consumption cursor
+  (`group_context_cursors`), not the whole window every turn. **`sdk.cron`-style
+  flag removed:** `sdk.persistentSession` is gone — SDK sessions are always on (with
+  the flag off and history out of the prompt, "off" would mean no memory at all).
+  **Stale-resume recovery:** an expired/invalid session id (the SDK throws "No
+  conversation found with session ID …") is caught, the bad id cleared, and the turn
+  retried once without resume — re-injecting history from SQLite so a conversation is
+  never silently lost. SQLite's role is now state/cursors/mappings + a durable record
+  (migration & recovery substrate), not live prompt-history reconstruction.
+  `buildSystemPrompt(customPrompt?, groupInstructions?)` and `queryAgent(userMessage,
+  config, sessionCtx?, onToolUse?, opts?)` lost their history/context params.
+  **Migration:** drop any `sdk.persistentSession` / `CC_OCTO_SDK_PERSISTENT_SESSION`
+  from config — it's ignored now; existing SQLite history is injected once on each
+  session's next turn, then carried by the SDK session.
+
 ### Removed
 
 - **Webhook transport** (#105) — removed; the Octo server does not POST to the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -54,6 +54,27 @@ cwd isolation), media-upload.ts / file-inline-wrap.ts (inbound media), db-adapte
 
 ## Gotchas
 
+- **Frozen system prompt + SDK-session-owned history** — `buildSystemPrompt`
+  (`agent-bridge.ts`) assembles ONLY stable operator content (security prefix +
+  SOUL + GROUP.md) so the SDK's cached system block is byte-identical turn-to-turn
+  (prompt-cache hits). **Do NOT put per-turn-variable text in the system prompt** —
+  it sits inside the `cache_control` block and invalidates the whole prefix every
+  turn (Anthropic's own guidance, found in the bundled `claude` binary). History
+  lives in the **SDK session** (`queryAgent` always `resume`s the stored id); only
+  a session's FIRST turn (or migration from existing SQLite history) injects prior
+  history ONCE into the **user message** as a `[Prior conversation history]` block.
+  Group context (B4) rides in the user message too, as a **delta** since a
+  per-channel cursor (`group_context_cursors` in `group-context.ts`
+  `buildContextSince` / `get|setContextCursor`), not the whole window. There is **no
+  `sdk.persistentSession` flag** — sessions are always on. Stale/expired resume (SDK
+  throws "No conversation found with session ID …") is caught in `queryAgent`
+  (`onResumeFailed` clears the id + retries once with `fallbackHistoryBlock`). SQLite
+  `messages` is kept as a durable record (migration + recovery substrate), NOT live
+  prompt history. Untrusted IM text moved out of the system prompt into the user
+  message is still escaped at the boundary (`sanitizePromptBody` for group context,
+  `escapeSectionMarkers` for history — same helpers, new location). The SDK still
+  auto-injects `<system-reminder># currentDate` (date only, no clock time) into the
+  first user message — don't duplicate it; for precise time the agent must run `date`.
 - **Pre-commit hook scope** — `.husky/pre-commit` runs `lint-staged` (ESLint on
   staged `*.ts` only) + `npm run type-check`. Non-`.ts` files (README, *.json)
   bypass lint-staged; full `npm test` is deferred to CI, so run it locally
--- a/README.md
+++ b/README.md
@@ -120,7 +120,6 @@ configurable.
 | `sdk.maxTurns` | *(SDK default)* | Max agentic turns per query |
 | `sdk.systemPrompt` | *(built-in)* | Custom system prompt (a `<baseDir>/<id>/SOUL.md` overrides this). |
 | `sdk.toolProgress` | `false` | When true, post `🔧 Running <tool>(<params>)…` notices as the agent invokes tools (params truncated; deduped, capped per turn) |
-| `sdk.persistentSession` | `false` | When true, persist agent workspace state across messages via the SDK v2 Session API (resume by stored session id). `/reset` clears it. |
 | `sdk.settingSources` | `['project']` | Filesystem settings sources the SDK loads. Default `['project']` so it discovers skills symlinked into the session sandbox's `.claude/skills/` (see [Agent skills](#agent-skills)). Memory stays isolated regardless (inline auto-memory dir pin). Add `'user'` only to deliberately load the operator's real `~/.claude`. |
 | `groupConfigDir` | *(unset)* | Directory of per-group instruction files (`<groupId>.md`). A match is injected into the system prompt as trusted custom instructions for that group. See [Per-group instructions](#per-group-instructions). |
 | `sdk.anthropicBaseUrl` | *(unset)* | Override the upstream Claude API endpoint. See [Self-hosted gateway](#self-hosted-gateway) below. |
@@ -472,7 +471,7 @@ src/
 - **Per-session workspace isolation** — Each session gets its own SHA-256 hex sandbox under the bot's `workspace/` (`<baseDir>/<botId>/workspace`), partitioned by the same key as conversation history — **per DM peer** and **per group channel** (a whole group shares one sandbox by design). Idle sandboxes (>7d) are auto-cleaned every 6h. Note: it separates sessions from each other but does not confine a session to its directory (absolute-path reads via Bash/Read remain possible) — see the Security Model section.
 - **Groups are a shared workspace** — All members of a group share one history, one sandbox, and one auto-memory store (the session key is the channel id). There is **no member-to-member isolation within a group**; DM sessions remain private per peer.
 - **Auto-memory is not TTL-reclaimed** — Long-term memory lives at `<baseDir>/<botId>/memory` (a sibling of `workspace/`) and is never swept by the cwd janitor, so it persists across `/reset` and grows unbounded on long-lived deploys.
-- **Stateless sessions by default** — Uses the v1 `query()` API; workspace state (open files, command history) does not persist across messages. Enable `sdk.persistentSession` to use the SDK v2 Session API, which resumes the prior agent session each turn so that state carries over.
+- **SDK session owns conversation history** — every turn resumes the stored SDK session (the source of truth for history + workspace state); the system prompt is frozen (no per-turn history/context, so the prompt cache hits). A session's first turn — or a migration from existing SQLite history — injects prior history once into the user message; later turns rely on resume. A stale/expired session id is recovered automatically (cleared + retried with history re-injected). SQLite keeps a durable record for migration/recovery, not live prompt history.
 
 ## Roadmap
 
--- a/RUNTIME.md
+++ b/RUNTIME.md
@@ -34,7 +34,7 @@ skills, sessions) for free — at the cost of living within the SDK's model (see
 | Persona / identity (SOUL.md) | cc composes → SDK `systemPrompt` | ✅ Done |
 | Behavior rules (CLAUDE.md) | SDK `project` settings discovery | ✅ Done |
 | Long-term memory | SDK auto-memory | ✅ Done |
-| Conversation continuity (sessions) | SDK session + `resume` | ◑ Partial (opt-in; see #2) |
+| Conversation continuity (sessions) | SDK session + `resume` (always-on) | ✅ Done (see #4) |
 | Skills / external tools | SDK skills + Bash | ✅ Done |
 | Per-bot isolation & identity | cc config + dirs | ✅ Done |
 | Scheduled tasks | cc cron tool + gateway scheduler | ✅ Done |
@@ -46,8 +46,9 @@ skills, sessions) for free — at the cost of living within the SDK's model (see
 
 Each bot's persona lives in `<baseDir>/<botId>/SOUL.md`. cc loads it
 (`config.ts` `loadSoul`) and it **overrides** `sdk.systemPrompt`. The composed
-system prompt (security prefix + SOUL + group instructions + context/history)
-rides in the SDK preset prompt's `append` (`agent-bridge.ts` `buildSystemPrompt`).
+system prompt — security prefix + SOUL + group instructions, all
+operator-controlled and **frozen** (no per-turn history/context; see #4) — rides
+in the SDK preset prompt's `append` (`agent-bridge.ts` `buildSystemPrompt`).
 openclaw-style "give the bot a soul" — done, per bot.
 
 ## 2. Behavior rules — CLAUDE.md ✅ (with a documented caveat)
@@ -74,35 +75,52 @@ precedence over any `projectSettings` value, so memory stays contained even with
 (group = shared per channel, DM = per peer) and live **outside** `cwdBase` so the
 7-day cwd TTL never reclaims it.
 
-## 4. Conversation continuity — sessions ◑
+## 4. Conversation continuity — sessions ✅
 
 The SDK persists each conversation as a session (`~/.claude/projects/.../*.jsonl`)
-and continues it via `resume`. cc supports this through `persistentSession`
-(opt-in) — it stores the SDK session id and `resume`s next turn.
+and continues it via `resume`. cc **always** resumes: it stores the SDK session id
+per `sessionKey` and resumes it next turn. The SDK session is the **source of
+truth** for conversation history.
 
-**Today (default OFF):** cc rebuilds `[Conversation history]` from its own SQLite
-`messages` table and injects it into the prompt each turn. This predates heavy
-session use.
+**Frozen system prompt.** Per Anthropic's own guidance (found in the bundled
+`claude` binary: *"keep the system prompt frozen; inject dynamic context in a user
+message"*), cc's `systemPrompt.append` now carries ONLY stable, operator-controlled
+content — security prefix + SOUL + group instructions. It no longer contains
+`[Conversation history]` (B5) or `[Group context]` (B4), so the SDK's cached system
+block is byte-identical turn-to-turn and the prompt-caching prefix actually hits.
+(Previously the per-turn-variable history/context sat inside the cached block,
+forcing a cache-write with zero reads every turn.)
+
+**Where dynamic content goes now:**
+- **History (B5):** lives in the SDK session. Only the FIRST turn of a session
+  (no stored id) — or a migration from an existing SQLite deployment — injects the
+  prior history ONCE as a `[Prior conversation history]` block prepended to the
+  user message. After that, `resume` carries it.
+- **Group context (B4):** injected into the user message as a delta — only the
+  group messages NEW since the channel's consumption cursor
+  (`group_context_cursors`), not the whole window every turn.
 
 **Verified by spike (#2) — group chat over SDK session works:**
 - Multi-speaker attribution survives: encode the speaker in the turn content
   (`[user Alice]: …`); the SDK persists it verbatim and resume replays it. A
   later turn correctly answered "who said what".
-- `resume` continuity across speakers is fine; the session id is stable across
-  turns (cc's `onSessionId` re-store is harmless but more than needed).
+- `resume` continuity across speakers is fine; the session id is stable across turns.
 
-→ **DM and group are NOT fundamentally different** at the session level; both can
-use SDK persistent sessions (only the `sessionKey` granularity differs:
-DM = `spaceId:peer`, group = `channel_id`). The SQLite-history approach is
-largely a pre-session legacy, not a group requirement.
+→ **DM and group are NOT fundamentally different** at the session level; both use
+SDK sessions (only the `sessionKey` granularity differs: DM = `spaceId:peer`,
+group = `channel_id`).
 
-**Intended role split (evolution direction):**
-- **Conversation history → SDK session** (persistent + resume): the source of truth.
-- **SQLite → state, cursors, mappings**: group background-message consumption
-  offset, session-id maps, `/reset` control — *not* history reconstruction.
+**Stale-resume recovery.** If a stored session id is invalid/expired the SDK throws
+`No conversation found with session ID: …` (verified by spike). `queryAgent` catches
+this (only when it fails before any output), clears the bad id, and retries once
+WITHOUT resume, re-injecting the history block from SQLite — so a turn never silently
+loses the conversation.
 
-This is a future refactor, not yet implemented — tracked as the evolution from
-the current opt-in `persistentSession` to session-as-default.
+**Role split (implemented):**
+- **Conversation history → SDK session** (resume): the source of truth.
+- **SQLite → state, cursors, mappings**: durable conversation record (migration +
+  stale-resume recovery substrate), group consumption cursor, session-id maps,
+  `/reset` control — NOT live prompt-history reconstruction.
 
 ## 5. Skills & external tools ✅
 
@@ -185,15 +203,16 @@ loop would build on #7.
 
 ## Summary
 
-cc replicates the **identity + memory + skills + (optional) session + scheduled
-tasks** parts of an openclaw-style runtime by configuring the Claude Agent SDK per
-bot — with little runtime code of its own. The remaining gap is **full autonomy**:
-the agent can schedule its own future runs (cron) but has no always-on
-free-running loop — every action still originates from an inbound message or a
-registered cron fire. The **session layer** is mid-evolution (SQLite-history today
-→ SDK-session-as-truth, with SQLite demoted to state/cursors). The deliberate
-trade for staying thin is living within the SDK's model — notably **lossy
-compaction**: on a long conversation the SDK summarizes older turns and *can drop
-early details, including a speaker's specific facts* (verified in the #2 spike).
-Business-critical facts that must survive compaction belong in auto-memory or
-SQLite, not only in the rolling session.
+cc replicates the **identity + memory + skills + session + scheduled tasks** parts
+of an openclaw-style runtime by configuring the Claude Agent SDK per bot — with
+little runtime code of its own. The remaining gap is **full autonomy**: the agent
+can schedule its own future runs (cron) but has no always-on free-running loop —
+every action still originates from an inbound message or a registered cron fire.
+The **session layer** is now SDK-session-as-truth: the system prompt is frozen
+(cacheable), history lives in the SDK session (resumed every turn), and SQLite is
+demoted to state/cursors/mappings + a durable record for migration & stale-resume
+recovery. The deliberate trade for staying thin is living within the SDK's model —
+notably **lossy compaction**: on a long conversation the SDK summarizes older turns
+and *can drop early details, including a speaker's specific facts* (verified in the
+#2 spike). Business-critical facts that must survive compaction belong in
+auto-memory or SQLite, not only in the rolling session.
--- a/config.example.json
+++ b/config.example.json
@@ -15,7 +15,6 @@
     "allowedTools": "*",
     "permissionMode": "bypassPermissions",
     "_comment_tool_progress": "v0.3: set toolProgress=true to post '🔧 Running <tool>(<params>)…' notices (params truncated). Off by default.",
-    "_comment_persistent_session": "v0.3: set persistentSession=true to keep agent workspace state across messages via the SDK v2 Session API. Off by default.",
     "_comment_setting_sources": "#100: which host filesystem settings the SDK loads (user=~/.claude, project=.claude, local=.claude/*.local). Default [\"project\"] so the SDK discovers skills symlinked into each session sandbox's .claude/skills/. Memory stays isolated regardless (the auto-memory dir is pinned via inline settings, ranked above projectSettings). Add \"user\" only to deliberately load the operator's real ~/.claude.",
     "settingSources": ["project"]
   },
--- a/src/__tests__/agent-bridge-config.test.ts
+++ b/src/__tests__/agent-bridge-config.test.ts
@@ -51,7 +51,7 @@ function createMockStream() {
 }
 
 async function drain(config: Config): Promise<void> {
-  for await (const chunk of queryAgent('hello', '', '', config)) {
+  for await (const chunk of queryAgent('hello', config)) {
     void chunk;
     // Drain generator so queryAgent invokes the SDK mock.
   }
--- a/src/__tests__/agent-bridge-query.test.ts
+++ b/src/__tests__/agent-bridge-query.test.ts
@@ -70,7 +70,7 @@ describe('queryAgent', () => {
     mockQuery.mockReturnValue(stream);
 
     const chunks: string[] = [];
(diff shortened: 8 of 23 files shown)


Jerry-Xin (changes requested):
Summary: The PR is in scope, but two runtime regressions can lose or duplicate conversation input, so it should not merge as-is.

**🔴 Blocking**

🔴 Critical — `src/index.ts:636` / `src/index.ts:642`: first-turn and migration history is prepended before `userBody`, then `truncateUtf8ByBytes` keeps the prefix and cuts from the end. With the default `historyLimit` of 40 and 32 KB accepted messages, `firstTurnHistory` can exceed the 96 KB cap by itself, causing the actual new user request to be dropped. This breaks group cold-start, SQLite migration, and any first turn with large prior history. The cap needs to budget/truncate context from the front or separately so the current `userBody` is always preserved. Add a test with oversized prior history proving the current message still reaches `queryAgent`.

🔴 Critical — `src/index.ts:386` / `src/index.ts:392` / `src/index.ts:399`: processed group messages are pushed into `group_messages` after the cursor is advanced, and if there was no delta the cursor is not advanced at all. On the next mentioned turn, the previous mentioned message is injected as `[Recent group messages]` even though the resumed SDK session already contains it as the prior user turn. This duplicates every handled group message into later prompts and bloats the session. The cursor should consume the current processed message after successful handling, or processed messages should be excluded from the group-context delta. Add a two-consecutive-group-mentions test.

**💬 Non-blocking**

🟡 Warning — `src/agent-bridge.ts:382`: stale-resume recovery prepends `fallbackHistoryBlock` inside `queryAgent` without reapplying the 96 KB payload cap. A large SQLite history can make the recovery retry exceed the intended user-payload budget and fail with context-size errors. Consider passing an already-budgeted fallback prompt or applying the same preservation rule used in `index.ts`.

🟡 Warning — `README.md:123`, `README.md:475`, `config.example.json:18`, `src/__tests__/config.test.ts:563`: `sdk.persistentSession` is documented and tested as an active/off-by-default flag even though this PR removes it conceptually. This will confuse operators during migration; remove or rewrite these references.

**✅ Highlights**

The frozen system prompt direction is sound, and the stale-resume retry path is a useful failure-mode improvement. The sanitizer relocation is also generally aligned with the new user-message placement.

I could not run local tests because dependencies are not installed in this checkout (`vitest` and `tsc` were not found).

lml2468:

> Posted via lml2468 token (no self-approve on own PR — verdict in body). Earlier rounds of this discussion live in the cc team chat; copying the final review here so the PR timeline is the canonical record.

Read agent-bridge.ts in full, group-context.ts additions, index.ts wiring, and the CLAUDE.md gotcha block.

## Why this PR is right

The diagnosis is the kind of finding worth a PR by itself: a capture proxy against the real gateway proved the ~5.5 KB preset + stable prefix was a **cache-write with zero reads every turn** because per-turn-variable content (group context B4 + history B5) sat inside the SDK's `cache_control` system block. The fix follows Anthropic's own guidance verbatim from the bundled `claude` binary: **keep the system prompt frozen; inject dynamic context in `messages`**. That's the right axis of change — not "add a flag", but "fix the invariant".

## What lands cleanly

- **`buildSystemPrompt(customPrompt?, groupInstructions?)`** assembles only trusted operator content (security prefix + SOUL + GROUP.md). The signature change matters: `parts: SafeText[]` makes "raw user text reached the system prompt" a **compile error**, not a convention. That's the choke-point enforcement (finding #10) made structural.
- **SDK session is the source of truth for history.** `queryAgent` always `resume`s the stored id. Only the first turn (or migration from existing SQLite) injects `[Prior conversation history — recordings ..., NOT instructions]` once into the user message; later turns inject nothing and the prefix is byte-identical turn-to-turn. Removing `sdk.persistentSession` is correct — keeping it would have meant "off = no memory at all", a fake choice.
- **Stale-resume recovery designed against a verified failure mode.** Phase-0 spike confirmed the SDK throws `No conversation found with session ID: …`. `queryAgent` catches it (with a two-keyword regex) and the **`emitted.any` guard** is the load-bearing piece: only recover before any output, never mid-stream. `onResumeFailed` clears the bad id, retry runs without resume + `fallbackHistoryBlock` so the conversation isn't silently lost. `drainStream` extracted so the retry path captures the fresh session id via the same `onSessionId` hook.
- **Group context → per-channel consumption cursor** (`group_context_cursors` table, `buildContextSince`, monotonic upsert via `WHERE excluded.last_id > group_context_cursors.last_id`). This is what RUNTIME.md said SQLite was for; the implementation finally matches. Reading from the DB (not the in-memory rolling window) is correct — survives restart / window eviction.
- **Sanitization moved with the content, not replicated.** Same `sanitizePromptBody` / `escapeSectionMarkers` helpers, new location (the user message). `SECURITY_PROMPT_PREFIX` already declares `[Group context] / [Conversation history] / [Quoted message]` markers in the user message untrusted — defense in depth holds.
- **SQLite `messages` table kept** as durable record + migration substrate, no longer the live-prompt reconstruction source. ARCHITECTURE diagram updated to match.
- **Breaking config call-out is honest** — `sdk.persistentSession` / `CC_OCTO_SDK_PERSISTENT_SESSION` ignored, CHANGELOG flags it ⚠️ Breaking, migration paragraph spells out the SQLite-history first-turn injection.
- **Test rewrite is faithful.** 7 ported tests pinned to old placement now assert the new user-message position; new coverage for first-turn / migration / group-cursor / stale-resume. 986 tests, run 2× for flake.

## Non-blocking follow-ups (file when convenient, not gating)

1. **`isResumeError` regex is soft-bound to SDK error strings.** Today: `/No conversation found with session ID|--resume requires a valid session/i`. The spike verified the current text, but the SDK changing "session ID" to "session_id" or adding punctuation silently breaks recovery → the user sees an opaque error. Two options, either is fine: (a) a regression test using a mock SDK that throws the exact verified strings, or (b) inline the full original error string in the `isResumeError` comment so a future SDK bump triggers grep-driven attention.
2. **Tiny ordering nuance in stale-resume recovery.** `onResumeFailed()` clears the dead id; then the retry's `onSessionId` writes the fresh id. Final store state is correct, but there is a microscopic window where the store has no id at all — a concurrent turn arriving in that window would re-run first-turn injection. Not a bug today (single-turn-per-session pipeline), but worth a CAS-style `clearSdkSessionId(sessionKey, expectedId)` if this code ever serves concurrent turns per session.

CI: `check-sprint` red is board governance (issue assignment), not a code signal.

LGTM. Ship it.

— 齐静春 📖

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#120 Review Report — Frozen System Prompt + SDK-Session-Owned History

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Frozen system prompt (no per-turn variable content) | ✅ | `buildSystemPrompt()` in `agent-bridge.ts` takes only `customPrompt?` + `groupInstructions?` — both operator-controlled. No user text path exists. |
| `buildSystemPrompt` signature change propagated | ✅ | All 7 test files updated; `queryAgent` calls in `index.ts:735` use new 5-arg signature. |
| SDK session always-on (resume every turn) | ✅ | `sessionOpts` always includes `onSessionId`; `resume` conditionally spread when truthy (`index.ts:681-693`). |
| `store.clearSdkSessionId` exists | ✅ | `session-store.ts:297` — `DELETE FROM sdk_sessions WHERE session_id = ?`. |
| `sanitizePromptBody` + `escapeSectionMarkers` imported in index.ts | ✅ | `index.ts:18` — both imported from `prompt-safety.js`. |
| Group context delta via consumption cursor | ✅ | `group-context.ts` — `group_context_cursors` table, `buildContextSince`, `get|setContextCursor`. Monotonic upsert (`WHERE excluded.last_id > group_context_cursors.last_id`). |
| History block built BEFORE `appendUser` | ✅ | `index.ts:567` (`buildSegmentedHistoryPrefix`) → `index.ts:603` (`appendUser`) → `index.ts:621` (`historyBlock` from `historyPrefix`). Current message excluded. |
| Stale-resume recovery (catch → clear → retry) | ✅ | `agent-bridge.ts` — `isResumeError` regex, `drainStream` extracted, `emitted.any` gate prevents retry after partial output. `finally { s.close() }` in `drainStream`. |
| Sanitization correctly relocated | ✅ | Group context → `sanitizePromptBody` (escapes role labels + section markers). History → `escapeSectionMarkers` only (role labels already escaped by `renderTurn`). Matches old `safeBody` / `safeSectioned` split. |
| Test coverage | ✅ | 986 tests. New: stale-resume (3 tests), consumption cursor (6 tests), migration injection, frozen-prompt assertions. |

## 2. Findings

### F1 — `buildContextSince` cursor stall on oversized messages (P2, new)

**Diff-scope:** new (introduced by this PR's cursor logic).

**Path:** `group-context.ts:335-340` — when `selected.length === 0` (every message in the delta exceeds `maxContextChars` individually), returns `{ text: '', lastId }`. Caller at `index.ts:385` checks `if (delta.text)` — empty string is falsy → cursor NOT advanced.

**Impact:** The same oversized messages are re-fetched every turn until they age out of the `maxWindowSize` (100) rolling window. The group-context delta for that channel is silently stalled for potentially many turns. No user-visible error, but the model loses group awareness for that channel temporarily.

**Severity:** P2 — edge case (single message > `maxContextChars`, typically 6000 chars), temporary (self-heals via window trim), no data corruption. Not P1 because it doesn't make a working path unavailable — the old `buildContext` had a similar budget limitation.

**Suggestion:** Either advance the cursor to `lastId` even when `selected` is empty (skip the unrenderable messages), or log a warning when this edge case hits so operators can investigate.

### F2 — Misleading comment in `buildContextSince` (P2/nit, new)

**Diff-scope:** new.

**Path:** `group-context.ts:333` — comment says "advance the cursor to the newest included id regardless so we never re-show a line." But the caller only advances when `delta.text` is non-empty. The `lastId` in the `{ text: '', lastId }` return (when `selected.length === 0`) is computed but unused by the caller.

**Suggestion:** Clarify the comment: "advance the cursor to the newest included id when any message fits the budget; when none fit, return the lastId for caller discretion."

### F3 — Stale-resume fallback may inject divergent history (P2, new)

**Diff-scope:** new (the fallback path is new).

**Path:** `agent-bridge.ts:280-285` — on stale-resume recovery, `fallbackHistoryBlock` is the full `historyBlock` built from SQLite's `buildSegmentedHistoryPrefix`. The SDK session (now gone) may have compacted/summarized older turns differently from what SQLite stores. The model sees the raw SQLite version, which may include details the SDK session had already summarized away, or miss details the SDK session had from tool outputs that SQLite doesn't capture.

**Severity:** P2 — inherent limitation of dual-storage, documented in the PR's RUNTIME.md update ("lossy compaction" caveat). The fallback fires rarely (only on stale session ids) and the model gets *more* context, not less.

### F4 — `sessionOpts` inline type annotation (nit, new)

**Path:** `index.ts:681` — the type annotation is ~200 chars on one line. Consider extracting a `SessionOpts` type alias for readability.

## 3. Data Flow Trace

| Consumed Data | Upstream Source | Verified Flow |
|---|---|---|
| `resume` (SDK session id) | `store.getSdkSessionId(sessionKey)` → `sdk_sessions` table | ✅ Returns `undefined` on first turn; populated by `onSessionId` callback from prior turn |
| `historyPrefix` | `store.buildSegmentedHistoryPrefix(sessionKey, limit)` → SQLite `messages` table, filtered by `reset_barriers` | ✅ Built at line 567, before `appendUser` at line 603. Current message excluded. |
| `groupContextBlock` | `groupContext.buildContextSince(channelId, cursor)` → `group_messages` table, `id > cursor` | ✅ Delta only. Cursor advanced only when text is non-empty. Current message excluded (cached after delta read). |
| `userContentForLLM` | `firstTurnHistory + groupContextBlock + userBody` → capped at 96KB via `truncateUtf8ByBytes` | ✅ Assembly order correct. Body is last so truncation eats context first. |
| `fallbackHistoryBlock` | Same `historyBlock` as `firstTurnHistory` (built from `historyPrefix`) | ✅ Available on every turn where `resume` is truthy, not just first turn. |
| `onResumeFailed` | `() => store.clearSdkSessionId(sessionKey)` | ✅ Deletes from `sdk_sessions`. Next turn starts fresh. |
| `sanitizePromptBody(delta.text)` | `escapeSectionMarkers(escapeRoleLabels(text))` | ✅ Correct for user-authored group chat lines (`<name>：<body>` format). |
| `escapeSectionMarkers(historyPrefix)` | Escapes `[Group context]`, `[Conversation history]`, etc. | ✅ Correct for rendered history turns (role labels already escaped by `renderTurn`). |

## 4. Distributed / Multi-Instance Side Effects (R3 Checklist)

- **Multi-client fan-out:** N/A — this PR doesn't change event emission or notification paths.
- **Observer ≠ actor:** N/A — no emit attribution changes.
- **Scheduled/background triggers:** N/A — the cron tool server path is unchanged.

## 5. Additional Observations

- The `drainStream` extraction is clean and the `emitted.any` gate correctly prevents retry after partial output (avoids double-reply).
- The `isResumeError` regex covers both the documented SDK error message and a `--resume` variant.
- The `group_context_cursors` table has no FK to `group_messages` — intentional, since cursor values may reference trimmed rows. The monotonic upsert (`WHERE excluded.last_id > group_context_cursors.last_id`) prevents backward cursor movement.
- The `sdk.persistentSession` config removal is a breaking change, properly documented in CHANGELOG and CLAUDE.md.

---

[SHADOW/qwen3.7-max] verdict: **APPROVE**

No P0 or P1 findings. All findings are P2 or nit. The PR is well-designed (frozen prompt follows Anthropic guidance), well-implemented (data flows verified end-to-end), and well-tested (986 tests, new coverage for cursor, stale-resume, migration, and frozen-prompt assertions). The four P2 findings are edge-case design notes and a comment clarification, not blockers.

lml2468:
**齐静春 📖 review correction — verdict: CHANGES_REQUESTED (revising my earlier APPROVED on this PR)**

> Posted via lml2468 token. Earlier comment on this PR (#issuecomment-4657302800) gave APPROVED based on reading buildSystemPrompt + buildContextSince + the stale-resume recovery in isolation. Jerry-Xin / Steve / mochashanyao surfaced two real blockers I'd missed; this comment is the consolidated re-review with the verification I should have done the first time.

## Self-correction up front

Both blockers below come from the same failure mode on my side: **I read the docstring and didn't grep the implementation**. This is the third time in this review cycle that "信注释不验代码" has bitten me (the prior two: PR #116 confused-deputy on `_cronFire` flow, daemon-cli PR #31 `ApplyManagedBotsDelta` workspace_id placeholder). Logging it loudly so it's visible on the PR timeline and so I treat "comment ↔ implementation disagreement" as a hard-rule grep next time, not a vibe check.

## Must-fix blockers

### 1. First-turn history truncation eats the user message, not the oldest context

`src/index.ts` (around the user-message assembly):
```ts
let userContentForLLM = firstTurnHistory + groupContextBlock + userBody;
// S2: hard cap on total user-role payload. Byte-safe truncation
// (truncateUtf8ByBytes) so a CJK-heavy payload is actually capped and never
// emits a replacement char.
const MAX_USER_LLM_BYTES = 98_304; // 96 KB
const { truncated, wasTruncated } = truncateUtf8ByBytes(userContentForLLM, MAX_USER_LLM_BYTES);
if (wasTruncated) {
  userContentForLLM = truncated + '\n[… user input truncated to 96KB cap]';
}
```

Comment claims: *"current body always survives (it's last, so truncation eats oldest context first only if everything together exceeds the budget)"*.

Reality: `truncateUtf8ByBytes` in `src/file-inline-wrap.ts` is:
```ts
const buf = Buffer.from(input, 'utf-8');
if (buf.length <= maxBytes) return { truncated: input, originalBytes: buf.length, wasTruncated: false };
const baseTrimmed = buf.subarray(0, maxBytes);
// (then scan back to a UTF-8 boundary)
```

`subarray(0, maxBytes)` keeps the **head** and drops the **tail**. `userBody` is the last segment of the concatenation, so when the assembly overflows, the user's current message is exactly what's deleted, while old SQLite history and group-context backlog are preserved. That's the opposite of what's documented and the opposite of what the user expects.

Two fix shapes, either is fine:
- **(A, simpler)** Reorder the assembly so the user message is first: `userBody + groupContextBlock + firstTurnHistory`. Tail-truncate then eats the oldest history first, matching the docstring's intent.
- **(B, more bookkeeping)** Cap each segment with its own sub-budget (`history ≤ 64KB`, `group ≤ 24KB`, `userBody guaranteed`) before concatenation, then concatenate. Friendlier diagnostic messages too.

### 2. `buildContextSince` cursor stalls on an over-budget single message

`src/group-context.ts` (`buildContextSince`):
```ts
const lastId = rows[rows.length - 1].id;  // newest fetched, not newest included
const selected: string[] = [];
let used = 0;
for (let i = rows.length - 1; i >= 0; i--) {
  const line = `${rows[i].from_name}：${rows[i].content}`;
  const cost = line.length + (selected.length > 0 ? 1 : 0);
  if (used + cost > budget) break;
  selected.push(line);
  used += cost;
}
if (selected.length === 0) return { text: '', lastId };
```

Then `src/index.ts` consumes it:
```ts
const delta = groupContext.buildContextSince(channelId, cursor);
if (delta.text) {
  groupContextBlock = sanitizePromptBody(delta.text) + '\n';
  groupContext.setContextCursor(channelId, delta.lastId);  // ← inside the gate
}
```

Failure case: a single message is large enough that even alone it exceeds `budget`. Then `selected` ends up empty → `text: ''` → the caller's `if (delta.text)` is false → `setContextCursor` is **never called** → the next turn fetches the same row, gets the same empty-text result, the cursor never advances. Until that row is pruned out of `group_messages` (which only happens when the channel rolls past `maxWindowSize * 2`), the channel is wedged.

(Secondary: the docstring on `buildContextSince` says "advance the cursor to the newest **included** id" but the code uses `rows[rows.length - 1].id` which is the newest **fetched** id. Another comment-vs-implementation disagreement; for the success path the discrepancy is benign since included = fetched, but the wording is misleading.)

Two fix shapes:
- **(a)** Move `setContextCursor` outside the `if (delta.text)` gate — call it on every non-empty `rows` result, regardless of whether anything fit. Advances the cursor past wedged messages so the channel can make progress.
- **(b, preferred)** When a single message overflows budget, include a truncated form of it (e.g. first N bytes + `[… message truncated]`) and advance the cursor past it. Something is shown, the cursor moves, no permanent stall.

## Should-fix in this PR (cheap follow-ups)

### 3. Stale-resume retry has no fallbackHistoryBlock cap

`queryAgent` recovery path:
```ts
const retryPrompt = (opts.fallbackHistoryBlock ?? '') + userMessage;
```

No size guard. Today the caller passes `historyBlock`, which has been through `truncateUtf8ByBytes` exactly once — but in the wrong direction (see #1). After #1 is fixed the cap order will be correct, but the retry path should still cap `fallbackHistoryBlock` to a sensible fraction of `MAX_USER_LLM_BYTES` (e.g. 48 KB) so even a pathological history doesn't blow the retry budget. Defense in depth, not gating.

### 4. Grep for stale `sdk.persistentSession` references

CHANGELOG correctly flags the breaking config removal, but a sweep of README / docs / test fixtures / example configs for `sdk.persistentSession` and `CC_OCTO_SDK_PERSISTENT_SESSION` is worth doing before merge to avoid the "deleted the flag but the docs still tell users to set it" trap.

## What still lands well (preserved from earlier review)

The high-level design is right — diagnosis via capture proxy was excellent (cache-write zero-reads), `SafeText[]` choke-point makes "raw user text into system prompt" a compile error rather than a convention, SDK-session as the source of truth is the correct ownership shift, group_context_cursors mirrors the existing sdk_sessions / reset_barriers single-row pattern, and the stale-resume `emitted.any` guard correctly prevents mid-stream retry duplication. The Anthropic-bundled guidance citation is real, the spike validation is real, and removing `sdk.persistentSession` (instead of papering over with a flag) is the courageous choice.

The defects are in the assembly + cursor mechanics, not in the architecture.

— 齐静春 📖

lml2468:
Round-2 fixes pushed (280b87b) — all four findings addressed.

**🔴 #1 — current message could be evicted by oversized history.** Fixed: new shared `assembleUserMessage(context, body, maxBytes)` (in `file-inline-wrap.ts`) reserves the **body whole** (it's the actual new request), gives the remaining budget to context, and **front-truncates** context (drops oldest, marker `[… earlier context truncated]`) — never the end. If the body alone meets/exceeds the cap, context is dropped and the body is byte-capped last. `index.ts` now assembles via this helper. Regression test added: a first turn with ~360 KB prior history → the current message still reaches `queryAgent` whole and within budget.

**🔴 #2 — handled group messages re-injected.** Fixed: the cursor now **always advances to `getMaxMessageId()` after caching the current message** (not just when there was a delta, and past the current turn). The current mentioned message — which the resumed SDK session already holds as the prior user turn — is consumed, so a later mention won't re-inject it. Regression test added: two consecutive group mentions → turn 2's user message does NOT contain turn 1's message.

**🟡 #3 — stale-resume retry skipped the cap.** Fixed: `queryAgent` budgets the retry via `assembleUserMessage(fallbackHistoryBlock, userMessage, 96KB)`, so a large SQLite history can't push the recovery retry over the payload limit. Test added.

**🟡 #4 — stale `persistentSession` docs/tests.** Removed from `README.md` (×2), `config.example.json`, and `config.test.ts` (env-clear list + the toggle test block).

Gates: type-check ✅ · lint (0 warn) ✅ · **992 tests** ✅ (+6) · build ✅ · NUL scan clean. (The reviewer noted deps weren't installed in their checkout — `npm ci && npm test` runs clean here and in CI.)

lml2468:
**齐静春 📖 re-review (4th round, head `280b87bd3241`) — verdict: APPROVED**

> Posted via lml2468 token. Earlier comments on this PR are #issuecomment-4657302800 (initial APPROVED, defective — missed the truncation direction and the cursor-gate stall) and #issuecomment-4657326529 (correction → CHANGES_REQUESTED with 4 items). This comment verifies the new commit.

Diffed `a2943b33766f → 280b87bd3241` end-to-end and re-grepped the implementation behind every claim. The four items from my correction comment are all addressed, and the fixes are structural rather than band-aids.

## Must-fix #1 — first-turn history truncation eats userBody → ✅ fixed

New helper `assembleUserMessage(context, body, maxBytes)` in `src/file-inline-wrap.ts`:

```ts
const bodyBytes = Buffer.byteLength(body, 'utf-8');
if (bodyBytes >= maxBytes) {
  // Pathological: drop context entirely, cap the body.
  const { truncated, wasTruncated } = truncateUtf8ByBytes(body, maxBytes);
  return wasTruncated ? truncated + '\n[… user input truncated to cap]' : body;
}
const contextBudget = maxBytes - bodyBytes;
const ctxBytes = Buffer.byteLength(context, 'utf-8');
if (ctxBytes <= contextBudget) return context + body;
// Front-truncate context, keep the most-recent tail.
const ctxBuf = Buffer.from(context, 'utf-8');
const tail = ctxBuf.subarray(ctxBuf.length - contextBudget);
const decoded = new TextDecoder('utf-8').decode(tail).replace(/^�+/, '');
return '[… earlier context truncated]\n' + decoded + body;
```

This is the right shape:
- The body's bytes are **reserved first**, so the new user request always survives whole.
- Context is **front-truncated** (drop oldest history, keep most-recent tail) with a `[… earlier context truncated]` marker so the model knows the prefix was cut.
- UTF-8 boundary is handled by `replace(/^�+/, '')` — no U+FFFD emitted on a multi-byte boundary cut (validated by `'历史记录'.repeat(20_000)` test case).
- Pathological body-alone-exceeds-budget path drops context entirely and still preserves the request.

`src/index.ts` now calls `assembleUserMessage(firstTurnHistory + groupContextBlock, userBody, MAX_USER_LLM_BYTES)`. E2E test (`first-turn: oversized prior history does NOT evict the current message`) verifies that with ~360 KB of seeded history a current message of 'THE CURRENT QUESTION' survives whole at the end, marker is present, and total payload ≤ 96 KB + 64 B marker slack.

## Must-fix #2 — `buildContextSince` cursor stalls on a single oversized message → ✅ fixed

`src/index.ts` `setContextCursor` was inside the `if (delta.text)` gate — that's gone. The cursor now advances **on every bot-handled turn**, past everything currently in the channel (the just-pushed current message included), via `groupContext.setContextCursor(channelId, groupContext.getMaxMessageId(channelId))`. An empty `delta.text` (because a single backlog message exceeded `budget`) no longer wedges the channel.

E2E test (`two consecutive group mentions: the first handled message is NOT re-injected on the second`) verifies that turn 2's user message does not contain turn 1's text — the resumed SDK session already holds it, the cursor advanced past it.

Semantic note worth flagging in the PR description (not blocking): the new policy "cursor always advances to `getMaxMessageId`" means a **non-mention** middle message that exceeds `maxContextChars` alone now silently skips appearing in any future `[Recent group messages]` block (cursor jumped past it). That's a deliberate trade-off — for the `mention` path the message lives in the SDK session anyway; for `non-mention` paths the message is genuinely lost from the context view. mochashanyao's review proposed an alternative (truncate the oversized line in place) which would be a stricter fix; the current code's behavior is "always make progress, never wedge", which I think is the right call for this PR. If a future scenario calls for truncating-in-place, a one-liner in `buildContextSince`.

## Should-fix #3 — stale-resume retry has no fallback cap → ✅ fixed

`queryAgent` recovery path now uses the same `assembleUserMessage` with `MAX_USER_LLM_BYTES`:
```ts
const retryPrompt = assembleUserMessage(
  opts.fallbackHistoryBlock ?? '',
  userMessage,
  MAX_USER_LLM_BYTES,
);
```
Test `caps the stale-resume retry prompt so a huge fallback history cannot blow the budget` injects a 300 KB fallback, asserts retry prompt ends with `THE REQUEST` whole, contains the truncation marker, and is ≤ 96 KB + 64 B slack. Same byte budget as the live path — consistent.

## Should-fix #4 — sweep stale `sdk.persistentSession` references → ✅ fixed

Verified `gh api .../contents/<path>?ref=280b87bd3241 | grep persistentSession` is empty for `RUNTIME.md`, `src/config.ts`, `src/index.ts`, `src/__tests__/e2e.test.ts`. README's config table line is removed; `config.example.json` `_comment_persistent_session` is removed; `src/__tests__/config.test.ts` `v0.3: persistent session toggle` describe block is removed. `CC_OCTO_SDK_PERSISTENT_SESSION` is removed from the `CC_VARS` allowlist in the env-binding test. `CHANGELOG.md` retains the breaking-change call-out, which is correct (historical record).

## What I appreciate beyond the fixes

- The new code extracts the policy ("body always survives, context is front-truncated") as a **named function with a docstring and 5 explicit test cases**, rather than threading the logic inline at two call sites. Future callers (and there will be future callers) won't reinvent the wrong shape.
- The cursor-advance comment in `src/index.ts:403-410` explains **why** we advance even when delta was empty ("the current mentioned message is the user turn the resumed SDK session already holds"), which is the load-bearing observation a future refactor needs to know.
- Tests are layered correctly: unit (`file-inline-wrap.test.ts` `assembleUserMessage` block — 5 cases), agent-bridge integration (`queryAgent` retry cap), full e2e (`E2E smoke tests` two new cases). Each fix has a regression at the right granularity.

## Self-correction reflection (continuing from my previous comments)

I'm logging this so the PR timeline keeps the failure record visible: my initial APPROVED comment on this PR was wrong because I read the truncation comment ("current body always survives") instead of grepping `truncateUtf8ByBytes`'s body, which clearly does `buf.subarray(0, maxBytes)` — head-keep, tail-drop. That was the third "信注释不验代码" miss in this review cycle (the other two: PR #116 `_cronFire` flow, daemon-cli #31 `ApplyManagedBotsDelta` workspace_id placeholder). Three is enough — internalizing "comment ↔ implementation disagreement → always grep" as a hard rule going forward.

LGTM. Ship it.

— 齐静春 📖

Jerry-Xin (changes requested):
Project relevance gate: pass. The PR is in scope, but it introduces a blocking lifecycle bug in conversation retention.

🔴 Blocking

🔴 Critical — Expired sessions can still resume old SDK history after the documented 7-day TTL.  
`src/session-store.ts:249` deletes only rows from `sessions`, and `sdk_sessions` has no foreign key/cascade relationship to `sessions` (`src/session-store.ts:68`). After cleanup removes the SQLite session/messages, the stale SDK mapping remains. On the next message, `handleMessage` recreates the session and then reads the old mapping at `src/index.ts:618`, so it resumes the old SDK conversation instead of treating this as a fresh first turn. This bypasses the existing 7-day conversation expiry and can silently resurrect expired history. Since SDK sessions are now always on, this is no longer an opt-in edge case.

Fix by expiring `sdk_sessions` in `cleanExpired()` as well, preferably by `updated_at < cutoff` or by deleting mappings for sessions being expired, and add a regression test showing that an expired session loses both SQLite history and the stored SDK session id.

💬 Non-blocking

🟡 Warning — Stale-resume retry considers only emitted text as “output.”  
In `src/agent-bridge.ts:334`, `emitted.any` is set only for text blocks. If a stream emits a tool-use block or other side-effect-bearing assistant content and then throws a resume-shaped error, `src/agent-bridge.ts:377` may retry and potentially duplicate work. This is probably rare for invalid resumes, but safer would be to mark the stream as emitted once any assistant content/tool-use is observed.

✅ Highlights

The system prompt freeze is implemented cleanly in `src/agent-bridge.ts:121`, and moving dynamic history/context into the user message is well covered. The `assembleUserMessage` helper in `src/file-inline-wrap.ts:190` is a good guard against oversized history evicting the current request.

Verification note: I attempted the focused tests and type check, but this checkout does not have local dependencies installed (`vitest` and `tsc` were not found).

lml2468:
**齐静春 📖 re-review (5th round, head `280b87bd3241`) — verdict: CHANGES_REQUESTED (revising my just-prior APPROVED on this head)**

> Posted via lml2468 token. Earlier comments on this PR: #issuecomment-4657302800 (initial APPROVED, defective), #issuecomment-4657326529 (correction → CHANGES_REQUESTED, 4 items), #issuecomment-4657385719 (4th-round APPROVED — verified the 4 fixes). Jerry-Xin's latest CR caught a class of issue I missed in all three. Re-verifying and revising.

## Self-correction up front (4th occurrence in this PR cycle)

The pattern I keep falling into is the same: I described a mechanism (the stale-resume `emitted.any` guard, the SDK-session-as-source-of-truth design) without grepping the **lifecycle and invariant** layer below it. The bugs Jerry-Xin flagged aren't in the logic I checked; they're in the lifecycle plumbing I assumed worked because the design diagram says it should.

Reminder to self: a design statement like "SDK session is the source of truth for history" implies invariants about **when and how that source of truth is invalidated**. Skipping the invalidation grep is the same failure mode as skipping the implementation grep — comments and design docs are evidence, not conclusions.

## Must-fix #1+#2 (Jerry-Xin #1 + #2 are the same root issue) — `sdk_sessions` row outlives the `sessions` row that justifies it

Grep-verified:
- `src/session-store.ts`: `sessions` table has 7-day TTL via `cleanExpired() { DELETE FROM sessions WHERE updated_at < ? }`.
- `src/session-store.ts`: `sdk_sessions` table has **no TTL**. `deleteSdkSession` only fires from explicit `clearSdkSessionId()`, which is called only from `/reset` and `onResumeFailed`.
- `src/session-store.ts`: `deleteSession(sessionId)` also only deletes from `sessions`. Doesn't cascade to `sdk_sessions` or `reset_barriers`.
- `src/index.ts:161`: `cleanExpired()` runs once at boot. It deletes from `sessions` only.

Failure scenario:
1. User chats with the bot, then goes silent for ≥ 7 days.
2. Boot-time `cleanExpired()` deletes the `sessions` row.
3. User returns. `getOrCreate(sessionKey, ...)` finds no row → INSERTs a fresh session row (fresh `updated_at`).
4. But `store.getSdkSessionId(sessionKey)` still returns the **old** `sdk_session_id` from the un-pruned `sdk_sessions` row.
5. `queryAgent` is asked to `resume: <old-sdk-id>`. Two sub-cases:
   - **SDK-side session has also expired** → throws `No conversation found with session ID: …` → `isResumeError` recovers gracefully (fresh start + `fallbackHistoryBlock`). Mostly fine, modulo the `fallbackHistoryBlock` containing 7-day-old SQLite history that wasn't pruned either.
   - **SDK-side session is still alive** (Anthropic's TTL is independent of ours) → resume succeeds → the user gets the **7-day-old conversation grafted onto what they think is a fresh chat**. From the user's standpoint a "new conversation" is contaminated with a buried prior context they may have forgotten. This is exactly the failure mode the design diagram says cannot happen because "SDK session is the source of truth" — but the source of truth's lifecycle isn't bound to the session lifecycle.

This applies to `reset_barriers` similarly (though `reset_barriers` is *designed* to outlive `deleteSession` per its docstring, so the lifecycle there is intentional — call it out and exempt).

Fix shape:
- `cleanExpired` should also `DELETE FROM sdk_sessions WHERE updated_at < ?` (it has its own `updated_at` column, already populated by `setSdkSessionId`). Use the same 7-day cutoff or a coordinated longer TTL — whatever, but it must be bounded.
- `deleteSession` should `DELETE FROM sdk_sessions WHERE session_id = ?` in the same path. Otherwise an explicit cleanup leaves the mapping behind, which violates least-surprise even more sharply than the TTL drift.
- Don't touch `reset_barriers` — its survival across `deleteSession` is intentional (G4 cold-start backfill consults it to avoid resurrecting pre-reset history).
- Regression test: seed a session + sdk_sessions row, advance `updated_at` by 8 days, call `cleanExpired`, assert both rows are gone.

## Must-fix #3 (Jerry-Xin #3) — stale-resume retry can re-run tools

`src/agent-bridge.ts` `drainStream`:
```ts
if (block.type === 'text' && block.text) {
  emitted.any = true;
  yield block.text;
} else if (block.type === 'tool_use' && onToolUse) {
  // ... onToolUse() called, but emitted.any NOT set
}
```

`emitted.any` is only set for text blocks, not for tool_use. Recovery condition:
```ts
if (opts?.resume && !emitted.any && isResumeError(err)) { /* retry */ }
```

Failure scenario: a turn starts, the model immediately issues a tool call (Bash, Write, an MCP server call), and only **after** the tool completes does the SDK throw a `No conversation found with session ID: …` error (this can happen when a tool_use turn carries forward into a follow-up turn whose session id has been invalidated). The recovery path sees `emitted.any === false`, retries the whole prompt, and the model re-issues the same tool call — running side-effecting work twice. For a Read that's annoying; for a Bash or a Write that's a real correctness bug.

I described this `emitted.any` guard as "the load-bearing piece" in my 4th-round comment without grepping which block types actually set it. Same failure mode as the other three: stated mechanism vs implemented mechanism mismatch.

Fix: set `emitted.any = true` in the `tool_use` branch as well. Use a separate flag if you want to keep "yielded text vs ran side effects" distinct, but the recovery gate must consider side effects to be irreversible just like text output.

Regression test: mock SDK stream yields a `tool_use` block then throws `No conversation found...`; assert no retry happens (or assert `onToolUse` was called exactly once even with `resume` set).

## Non-blocking

- Once #1+#2 is fixed, consider whether `reset_barriers` should be `cleanExpired`-pruned with a much longer TTL (e.g. 30 days). A barrier for a session that was last reset 6 months ago is a permanent row. Not urgent.
- The `fallbackHistoryBlock` carries SQLite history that may itself be 7-day-old after #1+#2 is fixed (because SQLite messages have their own retention path, separate from session TTL). Worth a follow-up audit, but not a blocker for this PR — the recovery message presents it explicitly as "Prior conversation history — recordings of earlier messages", so even if stale it's labeled.

## Reverting my 4th-round verdict

My 4th-round APPROVED comment (#issuecomment-4657385719) verified the four fixes I had asked for. Those fixes are still correctly implemented. But it gave a global APPROVED verdict on the head without doing the lifecycle invariant pass that would have caught the issues Jerry-Xin found. Treating that comment as superseded by this one.

— 齐静春 📖

lml2468:
Round-3 fixes pushed (28401ab) — both findings from the latest review addressed.

**🔴 Blocking — expired sessions could resurrect history past the 7-day TTL.** You're right: `sdk_sessions` has no FK cascade to `sessions`, so `cleanExpired()` left the stale SDK mapping behind → next message recreated the session and `resume`d the expired conversation. Now live (not opt-in) since sessions are always-on. Fixed: `cleanExpired()` also deletes `sdk_sessions WHERE updated_at < cutoff` (`updated_at` is bumped every turn via `setSdkSessionId`, so it tracks activity exactly like `sessions.updated_at`). Regression test added proving an expired session loses **both** its SQLite history and its SDK mapping — verified it fails without the fix.

**🟡 Non-blocking — retry considered only text as output.** Fixed: `emitted.any` is now set on **any** assistant content block (text OR tool_use), so a tool-use side effect followed by a resume-shaped error can't trigger a from-scratch retry that re-runs the tool. Recovery only fires before any content is observed. Regression test added.

Gates: type-check ✅ · lint (0 warn) ✅ · **994 tests** ✅ (+2) · build ✅.

lml2468:
**齐静春 📖 re-review (6th round, head `28401abdddd5`) — verdict: APPROVED**

> Posted via lml2468 token (self-vote on own-author PR is blocked for both APPROVE and CHANGES_REQUESTED — verdict in body). Earlier comments: #issuecomment-4657302800 (initial APPROVED, defective), #issuecomment-4657326529 (R2 CHANGES_REQUESTED, 4 items, all fixed by 280b87bd3241), #issuecomment-4657385719 (R4 APPROVED, defective — missed lifecycle invariant), #issuecomment-4657449814 (R5 CHANGES_REQUESTED, 3 items).

Diffed `280b87bd3241 → 28401abdddd5` and grep-verified each fix against the implementation (not the comment).

## Must-fix #1+#2 — sdk_sessions lifecycle bound to sessions TTL → ✅ fixed

`src/session-store.ts`:
```ts
this.deleteExpiredSdkSessions = this.adapter.prepare(
  'DELETE FROM sdk_sessions WHERE updated_at < ?',
);
// ...
cleanExpired(): number {
  const cutoff = Date.now() - SEVEN_DAYS_MS;
  const result = this.deleteExpired.run(cutoff);
  // ... [comment explaining the invariant] ...
  this.deleteExpiredSdkSessions.run(cutoff);
  return result.changes;
}
```

- Uses `sdk_sessions.updated_at`, which `setSdkSessionId()` already bumps every turn. So "last SDK activity" = "last session activity" within the same cutoff. Correct semantics.
- The comment now names the invariant explicitly ("sdk_sessions is a separate table (no FK cascade to sessions)") and explains why it must drop on the same TTL ("the next message would recreate the session and `resume` the expired SDK conversation, silently resurrecting history past the TTL"). That's the kind of comment future maintainers will not strip.

Regression test (`session-store.test.ts` `cleanExpired also expires the SDK session mapping`):
- Seeds an old session + sdk_sessions row, manually rewinds `updated_at` 8 days, seeds a fresh session + mapping
- Asserts after `cleanExpired()`: old history empty, **`getSdkSessionId('old-session') === undefined`**, fresh mapping untouched
- This pins the exact failure mode (next message would resume the expired SDK conversation) — a future refactor can't silently regress

Note on `deleteSession` cascade: I had also asked for `deleteSession()` to drop the sdk_sessions row. Grep shows `deleteSession()` has exactly one caller — `commands.ts` `/reset` — which already calls `store.clearSdkSessionId(sessionKey)` immediately after `store.deleteSession(sessionKey)`. So today it's not a bug, both are cleared on the only path that uses them. Worth noting as a small API-hygiene tradeoff (the next caller of `deleteSession` may forget the partner call) — but not a blocker for this PR. Could be a one-line cleanup in a follow-up to fold the cascade into `deleteSession` itself.

## Must-fix #3 — `emitted.any` set on any assistant content (tool_use included) → ✅ fixed

`src/agent-bridge.ts`:
```ts
const content = message.message?.content ?? [];
// Mark the stream as having produced output as soon as ANY assistant
// content block is seen — text OR tool_use. A tool_use is a side effect
// (the agent already acted); if the stream then throws a resume-shaped
// error, we must NOT retry from scratch and risk duplicating that work
// (PR #120 review). Recovery is only safe before any content is emitted.
if (content.length > 0) emitted.any = true;
for (const block of content) {
  if (block.type === 'text' && block.text) {
    yield block.text;
  } else if (block.type === 'tool_use' && onToolUse) {
    // ...
  }
}
```

The set moved out of the per-block loop and into the outer block once any content arrives. Covers tool_use, text, and any future block type — the principle "any assistant content = side effect was observed" is encoded structurally rather than per-type.

Regression test (`agent-bridge-query.test.ts` `does NOT retry after a tool_use block then a resume error`):
- Mock stream yields `tool_use{name: 'Bash'}` then throws `No conversation found with session ID: x`
- Asserts the error propagates (no recovery) AND `mockQuery` called exactly once (no retry → no duplicated Bash)
- Pins the exact failure mode I asked about: a Bash/Write tool that ran once doesn't get to run twice via recovery retry

## What still lands well (unchanged from earlier reviews)

The high-level design (frozen system prompt, `SafeText[]` choke-point, SDK session as source of truth, `assembleUserMessage` body-priority byte-cap, group-context monotonic cursor) all remain correct. The fixes across R2–R6 patched real defects in the implementation layer without touching the architecture — which is what I want to see when this many issues are uncovered: each one is a local correctness fix, none required design walk-back.

## Self-correction reflection (final pass)

This PR's review cycle on my side: APPROVED → corrected to CHANGES_REQUESTED → APPROVED → corrected to CHANGES_REQUESTED. Both reversals came from outside reviewers (Jerry-Xin + Steve + mochashanyao + yujiawei across the cycle) catching invariants I'd skipped. The shape of my mistake was always the same: described a mechanism using its comment / design statement, didn't grep the implementation or the lifecycle layer below it.

Hard rules I'm internalizing going forward, written down so they outlast this PR:
1. Comment ↔ implementation disagreement → always grep the implementation. The comment is evidence, not the conclusion.
2. A design statement like "X is the source of truth" → grep the **lifecycle** of X (creation / TTL / explicit deletion / cascade). The statement implies invariants the implementation may not preserve.
3. Tool behavior unknown → directly test it on a low-stakes target, do not induce from a single sample across a different case.

This is the 5th and final correction on this PR. LGTM — ship it.

— 齐静春 📖

Jerry-Xin (approved):
The PR is in scope and the main design change is relevant to this repository’s agent/session runtime. I found no blocking issues, but there are a couple of edge cases worth tightening.

**💬 Non-blocking**

🟡 Warning: `src/group-context.ts:101` selects the oldest unseen rows with `ORDER BY id ASC LIMIT ?`, and `src/index.ts:410` then advances the cursor to the channel max after the turn. In a busy group with more than 100 unseen messages, the bot receives only the oldest batch and permanently skips the newer, more relevant messages. Consider selecting newest unseen rows first, then reversing for chronological rendering.

🟡 Warning: `src/agent-bridge.ts:304` creates the SDK stream before entering the `try` that handles stale resume recovery. If `sdkQuery(...)` throws synchronously for an invalid resume id, `onResumeFailed` and the retry path will not run. Moving `runStream(opts?.resume, userMessage)` inside the `try` would make recovery cover both synchronous and async-iteration failures.

🔵 Suggestion: `src/agent-bridge.ts:5-7` and `src/agent-bridge.ts:161-162` still describe history/group context as system prompt content. The implementation no longer does that, so updating those comments would reduce future confusion.

**✅ Highlights**

The frozen system prompt boundary is much cleaner than the prior dynamic prompt assembly, and the stale-resume retry path correctly avoids retrying after text/tool output. The new tests cover migration, retry, cursor behavior, and prompt-size preservation well.

I attempted `npm run type-check` and `npm run lint`, but this checkout does not have dependencies installed (`tsc` and `eslint` were not found).

PR #122, opened 2026-06-09
fix(session): correct group-context cursor data loss + resume/cap edge cases

lml2468:
Follow-up review of #120 (frozen system prompt + SDK-session-owned conversation history) surfaced six defects in the new session-management / message-assembly paths. All fixed here.

## Findings & fixes

| # | Finding | Fix |
|---|---------|-----|
| **1+2** | **Group-context delta dropped recent messages.** `selectMessagesSince` fetched the *oldest* 100 unseen rows (`ORDER BY id ASC`), so the char-budget loop surfaced ancient backlog while genuinely-recent messages (beyond row 100, or beyond the ~6KB budget) were never fetched — then the cursor advanced past them, losing them permanently. `deleteOldMessages` (retains 200) could also delete un-injected rows. | Fetch **newest-first** (`ORDER BY id DESC`), matching the in-memory `buildContext` rolling-window. Budget now keeps the most-recent messages; `LIMIT 100 ≤ retained 200` so no fetchable row is trimmed prematurely; `lastId` reflects the true channel max. |
| **3** | **First group turn double-injected the same messages** — G4 SQLite backfill (history) and the group-context delta both carried them into one user message. | Prefer history on the first turn; use the delta only when there is no history. Cursor is already advanced, so nothing is stranded. |
| **4** | **Stale SDK session id leaked** when a resume error surfaced *after* output (`emitted.any`): `queryAgent` rethrew without clearing the dead id, forcing the next turn to fail too. | Clear the id via `onResumeFailed` for any resume-shaped error before rethrowing (still no retry — avoids duplicating a tool side effect — but the next turn recovers cleanly). |
| **5** | **`assembleUserMessage` overshot `maxBytes`** by the truncation marker's length (prepended without reserving its bytes). | Reserve the marker bytes from the context budget so the cap holds strictly. |
| **6** | **Stale `drainStream` docstring** referenced a nonexistent `suppressSessionId` param. | Rewritten to match the real `(s, emitted)` signature. |

One sweep candidate was dropped as a false positive: the group delta is **not** lost on stale-resume recovery — on a non-first turn it is part of `userMessage`, which the retry re-uses verbatim.

## Tests
- +4 regression cases: cursor newest-first within a tight budget, backlog larger than the fetch limit, stale-id clear on both no-retry paths (after-text and after-tool_use).
- Tightened cap assertions to strict `≤ 98_304` (previously `+64` slack).
- Full suite green: **996 passed**. `lint` + `type-check` + `build` clean.

## Files
- `src/group-context.ts` — DESC fetch + `buildContextSince` rework
- `src/index.ts` — first-turn history-vs-delta selection
- `src/agent-bridge.ts` — clear stale id on no-retry resume error + docstring
- `src/file-inline-wrap.ts` — reserve marker bytes in `assembleUserMessage`
- tests: `group-context`, `agent-bridge-query`, `file-inline-wrap`

### DIFF
--- a/src/__tests__/agent-bridge-query.test.ts
+++ b/src/__tests__/agent-bridge-query.test.ts
@@ -531,8 +531,8 @@ describe('queryAgent', () => {
     // The current request survives WHOLE at the end; the fallback was front-truncated.
     expect(retryPrompt.endsWith('THE REQUEST')).toBe(true);
     expect(retryPrompt).toContain('[… earlier context truncated]');
-    // Retry prompt stays within the 96 KB payload budget (+ small marker slack).
-    expect(Buffer.byteLength(retryPrompt, 'utf-8')).toBeLessThanOrEqual(98_304 + 64);
+    // Retry prompt stays within the 96 KB payload budget (marker bytes reserved).
+    expect(Buffer.byteLength(retryPrompt, 'utf-8')).toBeLessThanOrEqual(98_304);
   });
 
   it('does NOT recover (rethrows) when the error is unrelated to resume', async () => {
@@ -551,9 +551,10 @@ describe('queryAgent', () => {
     expect(mockQuery).toHaveBeenCalledTimes(1); // no retry
   });
 
-  it('does NOT retry if the stale error arrives AFTER output (no double reply)', async () => {
+  it('does NOT retry if the stale error arrives AFTER output (no double reply) but DOES clear the stale id', async () => {
     // Emit a chunk, THEN throw a resume-shaped error: recovery must not fire
-    // (we already streamed a partial reply; a retry would duplicate it).
+    // (we already streamed a partial reply; a retry would duplicate it). But the
+    // stale id is still cleared so the NEXT turn can recover (PR #120 review #4).
     const partial = createMockStream([]);
     partial[Symbol.asyncIterator] = async function* () {
       yield { type: 'assistant', session_id: 's', message: { content: [{ type: 'text', text: 'partial' }] } };
@@ -561,31 +562,36 @@ describe('queryAgent', () => {
     };
     mockQuery.mockReturnValue(partial);
     const chunks: string[] = [];
+    let resumeFailed = false;
     await expect(async () => {
       for await (const chunk of queryAgent('hi', makeConfig(), undefined, undefined, {
-        resume: 'sid', onResumeFailed: () => {},
+        resume: 'sid', onResumeFailed: () => { resumeFailed = true; },
       })) { chunks.push(chunk); }
     }).rejects.toThrow('No conversation found');
     expect(chunks).toEqual(['partial']);
     expect(mockQuery).toHaveBeenCalledTimes(1); // no retry after partial output
+    expect(resumeFailed).toBe(true); // but the stale id was cleared
   });
 
-  it('does NOT retry after a tool_use block then a resume error (no duplicated side effect)', async () => {
+  it('does NOT retry after a tool_use block then a resume error (no duplicated side effect) but DOES clear the stale id', async () => {
     // A tool_use is a side effect — if the stream throws a resume-shaped error
     // after it, retrying from scratch would re-run the tool. emitted.any must be
     // set on ANY assistant content, not just text (PR #120 review, non-blocking).
+    // The stale id is still cleared so the next turn recovers (PR #120 review #4).
     const partial = createMockStream([]);
     partial[Symbol.asyncIterator] = async function* () {
       yield { type: 'assistant', session_id: 's', message: { content: [{ type: 'tool_use', name: 'Bash', input: {} }] } };
       throw new Error('No conversation found with session ID: x');
     };
     mockQuery.mockReturnValue(partial);
+    let resumeFailed = false;
     await expect(async () => {
       for await (const _ of queryAgent('hi', makeConfig(), undefined, undefined, {
-        resume: 'sid', onResumeFailed: () => {},
+        resume: 'sid', onResumeFailed: () => { resumeFailed = true; },
       })) { void _; }
     }).rejects.toThrow('No conversation found');
     expect(mockQuery).toHaveBeenCalledTimes(1); // no retry after a tool_use side effect
+    expect(resumeFailed).toBe(true); // but the stale id was cleared
   });
 });
 
--- a/src/__tests__/file-inline-wrap.test.ts
+++ b/src/__tests__/file-inline-wrap.test.ts
@@ -289,8 +289,9 @@ describe('assembleUserMessage (PR #120: current message must always reach the mo
     expect(out.endsWith(body)).toBe(true);
     // Context was front-truncated with a marker.
     expect(out).toContain('[… earlier context truncated]');
-    // Within budget (+ small marker slack).
-    expect(Buffer.byteLength(out, 'utf-8')).toBeLessThanOrEqual(98_304 + 64);
+    // Within budget — the marker bytes are reserved, so the cap holds strictly
+    // (PR #120 review #5: previously overshot by the marker length).
+    expect(Buffer.byteLength(out, 'utf-8')).toBeLessThanOrEqual(98_304);
   });
 
   it('drops context entirely when the body alone meets/exceeds the budget', () => {
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -341,6 +341,37 @@ describe('GroupContext consumption cursor', () => {
     const ctx2 = new GroupContext(adapter, 6000);
     expect(ctx2.getContextCursor('ch1')).toBe(42);
   });
+
+  it('buildContextSince keeps the NEWEST messages (not oldest) when the budget is tight (PR #120 review)', () => {
+    // A small budget that fits ~2 short lines. With the old ASC fetch + newest-
+    // within-budget loop the cursor would still advance past everything, dropping
+    // recent lines; the newest-first fetch makes the budget keep the latest ones.
+    const small = new GroupContext(adapter, 40);
+    for (let i = 1; i <= 6; i++) {
+      small.pushMessage('ch1', `u${i}`, 'A', `m${i}`, i);
+    }
+    const out = small.buildContextSince('ch1', 0);
+    // The most-recent message must be present; the oldest must be the one dropped.
+    expect(out.text).toContain('A：m6');
+    expect(out.text).not.toContain('A：m1');
+    // Cursor advances past the whole delta (highest existing id), so dropped-oldest
+    // lines are never re-shown on a later turn.
+    expect(out.lastId).toBe(small.getMaxMessageId('ch1'));
+  });
+
+  it('buildContextSince advances lastId past a backlog larger than the fetch limit', () => {
+    // maxWindowSize=100: with >100 unseen, the newest-first fetch returns the
+    // latest 100; lastId is the highest existing id so the cursor jumps past the
+    // entire backlog (the oldest beyond-window messages are intentionally skipped).
+    for (let i = 1; i <= 130; i++) {
+      ctx.pushMessage('ch1', 'u', 'A', `m${i}`, i);
+    }
+    const out = ctx.buildContextSince('ch1', 0);
+    // Newest message reaches the model.
+    expect(out.text).toContain('A：m130');
+    // lastId reflects the true max in the channel, not just the last fetched row.
+    expect(out.lastId).toBe(ctx.getMaxMessageId('ch1'));
+  });
 });
 
 // ─── G23: robot flag tracking ───────────────────────────────────────────────────────────────
--- a/src/agent-bridge.ts
+++ b/src/agent-bridge.ts
@@ -303,10 +303,11 @@ export async function* queryAgent(
 
   const stream = runStream(opts?.resume, userMessage);
 
-  // Drain one SDK stream, yielding assistant text. `suppressSessionId` skips the
-  // onSessionId report (used on the FIRST attempt only matters; on a recovery
-  // retry we DO want to capture+persist the fresh id). Tracks whether any text
-  // was emitted so the caller can decide if a mid-stream failure is recoverable.
+  // Drain one SDK stream, yielding assistant text. Reports the SDK session id once
+  // (the first message that carries one) via opts.onSessionId — on a recovery retry
+  // this captures+persists the fresh id. Sets `emitted.any` true once ANY assistant
+  // content block (text OR tool_use) is seen, so the caller can tell whether a
+  // mid-stream failure is still recoverable (recovery is only safe before output).
   async function* drainStream(
     s: ReturnType<typeof runStream>,
     emitted: { any: boolean },
@@ -401,6 +402,18 @@ export async function* queryAgent(
       const retryEmitted = { any: false };
       yield* drainStream(runStream(undefined, retryPrompt), retryEmitted);
       return;
+    }
+    // Resume error that surfaced AFTER output (emitted.any) — we must NOT retry
+    // (a tool_use side effect may already have run; a retry could duplicate it).
+    // But the stored id is still stale, so clear it here too: otherwise the next
+    // turn would resume the same dead id and fail again. Clearing lets the next
+    // turn start fresh and recover via fallbackHistoryBlock (PR #120 review #4).
+    if (opts?.resume && isResumeError(err)) {
+      try {
+        opts.onResumeFailed?.();
+      } catch (cbErr) {
+        console.error(`[cc-channel-octo] onResumeFailed callback threw: ${String(cbErr)}`);
+      }
     }
     throw err;
   }
--- a/src/file-inline-wrap.ts
+++ b/src/file-inline-wrap.ts
@@ -204,13 +204,22 @@ export function assembleUserMessage(context: string, body: string, maxBytes: num
   if (ctxBytes <= contextBudget) {
     return context + body;
   }
-  // Truncate context from the FRONT (keep the most-recent tail). Slice the buffer
-  // to the last `contextBudget` bytes; a leading partial UTF-8 sequence decodes to
+  // Truncate context from the FRONT (keep the most-recent tail). A truncation
+  // marker is prepended, so reserve its byte size from the budget too — otherwise
+  // the result would exceed maxBytes by the marker length (PR #120 review). Slice
+  // the buffer to the remaining bytes; a leading partial UTF-8 sequence decodes to
   // a replacement char which we strip so we never emit U+FFFD.
+  const marker = '[… earlier context truncated]\n';
+  const markerBytes = Buffer.byteLength(marker, 'utf-8');
+  const tailBudget = contextBudget - markerBytes;
+  if (tailBudget <= 0) {
+    // No room for any context once the marker is accounted for — drop it entirely.
+    return body;
+  }
   const ctxBuf = Buffer.from(context, 'utf-8');
-  const tail = ctxBuf.subarray(ctxBuf.length - contextBudget);
+  const tail = ctxBuf.subarray(ctxBuf.length - tailBudget);
   const decoded = new TextDecoder('utf-8').decode(tail).replace(/^�+/, '');
-  return '[… earlier context truncated]\n' + decoded + body;
+  return marker + decoded + body;
 }
 
 /** Exported for tests. */
--- a/src/group-context.ts
+++ b/src/group-context.ts
@@ -96,10 +96,14 @@ export class GroupContext {
     this.deleteOldMessages = this.adapter.prepare(
       'DELETE FROM group_messages WHERE channel_id = ? AND id NOT IN (SELECT id FROM group_messages WHERE channel_id = ? ORDER BY id DESC LIMIT ?)',
     );
-    // Cursor delta: messages strictly newer than the cursor, oldest-first, with
-    // their rowid so the caller can advance the cursor to the last included id.
+    // Cursor delta: messages strictly newer than the cursor, NEWEST-first so a
+    // backlog larger than the fetch limit keeps the most-recent messages (the
+    // relevant ones) rather than ancient ones. LIMIT (maxWindowSize=100) is <=
+    // deleteOldMessages' retained 200, so every fetchable row survives trimming.
+    // buildContextSince re-sorts the budget-selected slice into chronological
+    // order for display. Mirrors the in-memory buildContext rolling-window.
     this.selectMessagesSince = this.adapter.prepare(
-      'SELECT id, from_name, content FROM group_messages WHERE channel_id = ? AND id > ? ORDER BY id ASC LIMIT ?',
+      'SELECT id, from_name, content FROM group_messages WHERE channel_id = ? AND id > ? ORDER BY id DESC LIMIT ?',
     );
     this.selectMaxId = this.adapter.prepare(
       'SELECT MAX(id) AS maxId FROM group_messages WHERE channel_id = ?',
@@ -303,13 +307,17 @@ export class GroupContext {
 
   /**
    * Build a group-context block from messages NEWER than `sinceId` (the delta the
-   * model hasn't seen yet), oldest-first, within `maxContextChars`. Returns the
-   * block text (same `[Recent group messages]` format as buildContext) and the
-   * highest id included so the caller can advance the cursor. Empty text + the
+   * model hasn't seen yet), within `maxContextChars`. Returns the block text (same
+   * `[Recent group messages]` format as buildContext) and the highest id that
+   * EXISTS in the channel above `sinceId` (so the caller advances the cursor past
+   * the whole delta, including any oldest lines the char budget dropped — those
+   * are the least-relevant and are intentionally not re-shown). Empty text + the
    * unchanged cursor when there's nothing new.
    *
-   * Unlike buildContext (in-memory rolling window), this reads from the DB so the
-   * cursor delta is exact even across restarts / window eviction.
+   * Rows are fetched newest-first (so a backlog larger than the budget keeps the
+   * most-recent messages); the selected slice is reversed back to chronological
+   * order for display. Unlike buildContext (in-memory rolling window), this reads
+   * from the DB so the cursor delta is exact even across restarts / window eviction.
    */
   buildContextSince(channelId: string, sinceId: number): { text: string; lastId: number } {
     let rows: Array<{ id: number; from_name: string; content: string }>;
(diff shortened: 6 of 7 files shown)


lml2468:

Six-finding follow-up to #120. Grep-verified each fix at the implementation layer + matched regression coverage. Three of the six surface defects in code I personally reviewed five rounds last cycle and missed — that's a sober reminder that the "信注释不验代码" pattern needs to extend to SQL `ORDER BY` direction and downstream effect of cursor-advance under "cursor stalled vs cursor stranded data" framing.

## Findings ↔ Fixes verification

### #1+#2 — Group-context delta dropped recent messages (worse than the stall I caught)

`selectMessagesSince` was `ORDER BY id ASC LIMIT 100` while my R5 only caught the empty-text cursor-stall (oversized single message wedging). I missed the upstream bug: with >100 unseen rows the ASC fetch surfaced ancient backlog (rows 1-100) → budget loop kept the newest-among-fetched but **never saw rows 101+**, then the cursor advanced past them, losing them permanently. Compounding: `deleteOldMessages` retains 200, so the un-injected window-101..200 rows could also be evicted before any future fetch.

Fix at `src/group-context.ts:106`:
```sql
ORDER BY id DESC LIMIT ?   -- (maxWindowSize = 100, ≤ retained 200)
```
+ `buildContextSince` rework: `rows[0]` is newest → `lastId = rows[0].id`; iterate `0..n-1` newest-first within budget; `selected.reverse()` for chronological display. Mirrors the in-memory `buildContext` policy exactly.

Regression at `__tests__/group-context.test.ts:345`: tight 40-char budget across 6 messages → asserts `m6` survives, `m1` dropped. And `:362`: 130-message backlog → asserts `m130` reaches the model + `lastId === getMaxMessageId(channel)` so the cursor never strands beyond-window messages permanently.

I should have caught this in R5. The framing trap was treating "cursor not advancing" as the worst-case → the fix preserved-but-extended that framing. The real worst case was "cursor advancing past data you never read".

### #3 — First-turn double-injection of group chatter

G4 backfill (SQLite history seed for cold-start) and the group-context delta both read from `group_messages` on a brand-new group session, so the user message got the same lines twice. Fix at `src/index.ts:651`:
```ts
const injectedContext = firstTurnHistory ? firstTurnHistory : groupContextBlock;
```
Correct: on first turn, history covers the recent chatter; later turns rely on resume + delta. Cursor was already advanced unconditionally before this point, so dropping the delta here does NOT strand messages — history covers them.

### #4 — Stale SDK session id leaked when resume error surfaces AFTER output

My R6 fix made `emitted.any` cover `tool_use`, correctly preventing retry-after-side-effect. But the `if (opts?.resume && !emitted.any && isResumeError(err))` branch never cleared `onResumeFailed` on the **after-output** path — the next turn would resume the same dead id and fail again.

Fix at `src/agent-bridge.ts:411`:
```ts
  try { opts.onResumeFailed?.(); } catch (cbErr) { ... }
}
throw err;
```
No retry (side effect protection preserved) but the id is now cleared so the next turn starts fresh and runs through `fallbackHistoryBlock`. Regression at `__tests__/agent-bridge-query.test.ts:576` asserts both `mockQuery.toHaveBeenCalledTimes(1)` AND `resumeFailed === true` — the right invariants paired.

### #5 — `assembleUserMessage` overshot cap by marker length

Cap was `maxBytes`, but I prepended `[… earlier context truncated]\n` (30 bytes) without reserving its bytes from the budget → result could be `maxBytes + 30`. Fix at `src/file-inline-wrap.ts:212`:
```ts
const markerBytes = Buffer.byteLength(marker, 'utf-8');
const tailBudget = contextBudget - markerBytes;
if (tailBudget <= 0) return body;  // no room for any context with marker, drop entirely
```
Test cap assertions tightened from `≤ 98_304 + 64` to strict `≤ 98_304`. The "+64 slack" in the old tests was hiding this exact overshoot.

### #6 — Stale `drainStream` docstring

Referenced nonexistent `suppressSessionId` parameter. Rewritten to match real `(s, emitted)` signature. Documentation correctness matters — this was a leftover from earlier iteration. Cheap fix, right hygiene.

## Test discipline observations

- **+4 regression cases** all target the specific failure modes (newest-first within budget; backlog beyond fetch limit; stale-id clear on both no-retry paths).
- **Tightened cap to strict ≤** instead of `+64` slack — slack assertions hide overshoot bugs. Good move to remove.
- **996 tests passing** (up from 986), so the existing surface is untouched.

## On the "false positive" the author dropped from sweep

Author called out one sweep candidate as not-actually-a-bug: "group delta lost on stale-resume recovery". I agree with the assessment — on non-first turn the delta is part of `userMessage` argument and the retry uses `userMessage` verbatim, so the delta survives the retry path. Correctly identified as not a defect.

## Lingering self-correction reflection

This PR finds **three defects in code I personally reviewed five times across cycle #120**. The pattern:
- R5 caught the cursor-stall symptom but missed the upstream cause (ASC fetch direction). I focused on "cursor not advancing" and didn't grep `ORDER BY` direction or trace "cursor advancing past unseen rows".
- R6 caught `emitted.any` not covering `tool_use` but missed the symmetric stale-id leak on the after-output path. I patched the retry gate without grepping all branches of the same error condition.
- R3 caught the byte-cap inversion (tail-truncation eats body) but missed the marker-overshoot. I tested for "body survives" but didn't tighten the cap assertion to strict-equality.

The common failure mode is **fixing the symptom one step shallow of the root**. Hard-rule addition: **after fixing one defect, re-grep the surrounding code for the next plausible variant of the same root cause**. The marker overshoot was inside the function I rewrote; the stale-id leak was inside the `try/catch` I extended; the cursor-advance loss was downstream of the SQL I never grepped.

This is rule #4 ("trace caller-callee full path") generalized to **"trace the family of defects, not just the one named"**. Adding to memory.

LGTM. Ship it.

— 齐静春 📖

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#122 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo #122
**Head SHA:** `198f557017ad1f723fc51c066185b3f3f8e5cb5e`
**Scope:** 7 files, +111/-32 — group-context cursor fix, resume stale-id clear, byte-cap tightening, first-turn dedup

---

## 1. Verification Summary

| # | Change | Verdict | Evidence |
|---|--------|---------|----------|
| 1 | `selectMessagesSince` ASC→DESC | ✅ Correct | `group-context.ts:107` — DESC LIMIT 100 fetches newest; `rows[0].id` = true max |
| 2 | `buildContextSince` lastId + budget rework | ✅ Correct | `group-context.ts:336-356` — lastId from `rows[0]` (newest); loop walks newest→oldest; reverse for display |
| 3 | First-turn history-vs-delta dedup | ✅ Correct (P2 note below) | `index.ts:653` — `firstTurnHistory ? firstTurnHistory : groupContextBlock` avoids cold-start duplication |
| 4 | Stale session id clear after output | ✅ Correct | `agent-bridge.ts:410-418` — `onResumeFailed()` called when `emitted.any && isResumeError`; recovery block's `return` prevents double-call |
| 5 | Marker bytes reserved in `assembleUserMessage` | ✅ Correct | `file-inline-wrap.ts:213-222` — `tailBudget = contextBudget - markerBytes`; total ≤ maxBytes holds strictly |
| 6 | `drainStream` docstring fix | ✅ Correct | `agent-bridge.ts:306-310` — matches actual `(s, emitted)` signature |

---

## 2. Findings

### P2-1: First-turn migration path may lose non-bot-directed group context

**Diff-scope:** NEW — introduced by this PR (before: both `firstTurnHistory + groupContextBlock` were concatenated; after: only one is selected).

**Path:** `index.ts:653` — `const injectedContext = firstTurnHistory ? firstTurnHistory : groupContextBlock;`

**Scenario:** On a migration first turn (existing deployment with SQLite conversation history but no SDK session id):
- `firstTurnHistory` = prior bot-directed conversation turns from `SessionStore.buildSegmentedHistoryPrefix`
- `groupContextBlock` = ALL unseen group chatter from `group_messages` (cursor=0, never set)
- The session store only holds bot-directed turns; `group_messages` holds all group chatter
- Selecting `firstTurnHistory` drops non-bot-directed group messages from the first turn's context

**Impact:** Minor context loss on a one-time migration path. The conversation still works — the model just has less ambient group awareness on the very first turn. Cursor advances to `getMaxMessageId` regardless (line 412), so dropped messages are never re-shown. Subsequent turns use the delta normally.

**Why P2 not P1:** No working path becomes unavailable (the model still has conversation history + the user's message). No user-visible wrong data. The loss is strictly ambient context on a one-time path. The PR's rationale (avoiding cold-start duplication) is sound.

**Suggestion (non-blocking):** Consider a future refinement: when `firstTurnHistory` is non-empty AND `groupContextBlock` contains messages not covered by the history, include both (or at least a summary). Low priority — the migration window is brief.

---

## 3. Data-Flow Tracing

### 3.1 `buildContextSince` — DESC fetch → cursor advance

```
selectMessagesSince(channelId, sinceId, 100)
  → ORDER BY id DESC LIMIT 100
  → rows[0] = newest message in channel above sinceId
  → lastId = rows[0].id  ← correct: highest id among fetched = channel max when ≤100 unseen

Budget loop: i=0 (newest) → i=rows.length-1 (oldest fetched)
  → selected = [newest, 2nd-newest, ...] within char budget
  → selected.reverse() → chronological order for display

Cursor: caller sets to getMaxMessageId(channelId) at index.ts:412
  → rows[0].id === getMaxMessageId when ≤100 unseen ✓
  → When >100 unseen: rows[0].id = absolute newest (DESC returns top 100) = getMaxMessageId ✓
  → Messages beyond the 100-window are intentionally skipped; cursor jumps past them ✓
```

### 3.2 `deleteOldMessages` vs fetch window safety

```
deleteOldMessages retains: maxWindowSize × 2 = 200 rows (newest)
selectMessagesSince fetches: maxWindowSize = 100 rows (newest)
→ Every fetchable row (top 100) is within retained 200 ✓
→ No fetchable row can be trimmed before it's read
```

### 3.3 Stale session id clear — no double-call

```
catch (err) {
  if (resume && !emitted.any && isResumeError) {
    onResumeFailed()  ← path A: pre-output recovery
    yield* drainStream(retry)
    return             ← exits; path B never reached
  }
  if (resume && isResumeError) {
    onResumeFailed()  ← path B: post-output clear (NEW)
  }
  throw err
}
```
Paths A and B are mutually exclusive (A's `return` prevents fall-through). ✓

### 3.4 `assembleUserMessage` byte cap

```
Before: marker(30B) + tail(contextBudget B) + body(bodyBytes B) = maxBytes + 30 ← overshoot
After:  marker(30B) + tail(contextBudget-30 B) + body(bodyBytes B) = maxBytes ← exact
Edge:   tailBudget ≤ 0 → return body only ✓
```

---

## 4. Test Coverage Assessment

| Test | Coverage |
|------|----------|
| `buildContextSince keeps NEWEST messages` | ✅ Verifies m6 present, m1 dropped under tight budget; lastId = max |
| `buildContextSince advances past backlog > limit` | ✅ 130 messages, lastId = getMaxMessageId, m130 reaches model |
| `does NOT retry after output but clears stale id` | ✅ `resumeFailed=true` after text emission + resume error |
| `does NOT retry after tool_use but clears stale id` | ✅ `resumeFailed=true` after tool_use + resume error |
| `assembleUserMessage` cap tightened | ✅ Assertion `≤ 98_304` (was `+64` slack) |

All regression tests directly validate the fix paths. No hand-fed data bypassing real integration paths detected.

---

## 5. Additional Observations

- **Cursor monotonicity:** `upsertCursor` uses `WHERE excluded.last_id > group_context_cursors.last_id` — cursor can never move backward. Safe against any race or re-ordering. ✓
- **`onResumeFailed` callback isolation:** Both call sites wrap in try/catch so a throwing callback doesn't mask the original error. ✓
- **UTF-8 boundary safety:** `TextDecoder` + `replace(/^+/, '')` pattern unchanged; partial sequences at the truncation point are stripped. ✓

---

## 6. Verdict

All six fixes are correctly implemented with clean data-flow paths. The DESC fetch rework, cursor advance semantics, stale-id clearing, and byte-cap tightening all trace correctly to their upstream data sources. Test coverage validates each fix path without hand-fed shortcuts.

One P2 note (migration first-turn context loss) is a minor tradeoff, not a blocker.

**[Octo-Q] verdict: APPROVE** — no P0/P1 findings. One P2 note for future consideration.

yujiawei (approved):
# Code Review — PR #122 (cc-channel-octo)

**Verdict: APPROVED.** All six targeted defects from the #120 follow-up are correctly fixed, the data flow holds together across files, and the gates are green: `type-check`, `lint --max-warnings 0`, and `build` all clean; full suite **996 passed** on head `198f557`. No P0/P1 introduced by this PR. One genuine P2 regression and a few nits are noted below as non-blocking.

## 1. Verification of the six fixes

- ✅ **#1+2 — group-context delta DESC fetch + correct cursor.** `selectMessagesSince` now `ORDER BY id DESC LIMIT 100` (group-context.ts:106). `lastId = rows[0].id` (the true channel max above `sinceId`, group-context.ts:338) so the cursor jumps past the whole delta even when the char budget drops the oldest lines — those are the least-relevant and are intentionally not re-shown. `LIMIT 100 ≤ deleteOldMessages` retained `200` (group-context.ts:97), so no fetchable row is trimmed before it can be surfaced. Budget loop walks newest→oldest then `reverse()` for chronological display (group-context.ts:348-356). Backlog-larger-than-limit and tight-budget cases are both covered by new tests.
- ✅ **#3 — first-turn double-injection removed.** `injectedContext = firstTurnHistory ? firstTurnHistory : groupContextBlock` (index.ts:651) replaces the old concatenation. For the common cold-start group, G4 backfill (index.ts:572) and the delta read the same messages, so preferring history correctly eliminates the duplicate. The cursor is advanced unconditionally first (index.ts:410), so dropping the delta does not strand the current request — `userBody` always carries it. (See §2 for the one narrow case where this heuristic is too aggressive.)
- ✅ **#4 — stale SDK id cleared on the after-output path.** agent-bridge.ts:406-417 now calls `onResumeFailed()` for a resume-shaped error even when `emitted.any` is true, then rethrows without retrying (a `tool_use` side effect may already have run). Confirmed there is no double-clear: the `!emitted.any` branch returns at line 404, so the line-411 block only runs when output was already emitted. Verified `/reset` and the recovery path agree — `clearSdkSessionId` is the same sink (commands.ts:127).
- ✅ **#5 — `assembleUserMessage` marker bytes reserved.** file-inline-wrap.ts:212-218 subtracts the marker length from the context budget; the `tailBudget <= 0` guard drops context entirely rather than overshooting. The PR's touched path now holds `≤ maxBytes` strictly (verified by repro: context-truncation of a 300-byte context with a 4-byte body at cap 100 → exactly 100 bytes). Test slack tightened from `+64` to strict `≤ 98_304`.
- ✅ **#6 — `drainStream` docstring** matches the real `(s, emitted)` signature (agent-bridge.ts:306-310).

## 2. Findings

### P2 — Migration first turn can drop ambient group context for one turn (non-blocking)

`src/index.ts:403-410` + `651`

The new `firstTurnHistory ? firstTurnHistory : groupContextBlock` choice assumes the injected history covers recent group chatter. That holds for a cold-start group (G4 backfill seeds the same rows the delta reads). It does **not** hold on a **migration first turn**:

1. New code is deployed onto an existing group that has prior conversation in the store but no SDK session id (`isFirstTurn = true`).
2. Other members post ambient (non-mentioned) messages — cached into `group_messages` via `pushMessage` (index.ts:840) but never processed, so no SDK session is created.
3. First mention arrives. `historyPrefix` is non-empty (migrated store history) → **G4 backfill is skipped** (index.ts:572-578).
4. Cursor is advanced unconditionally to the channel max (index.ts:410).
5. `injectedContext` picks `firstTurnHistory` and **drops `groupContextBlock`** (index.ts:651) — but `firstTurnHistory` comes from `store.buildSegmentedHistoryPrefix`, which does **not** contain those ambient messages.

Net effect: the ambient lines reach the model on **zero** turns (the cursor has already moved past them). Impact is bounded — the user's actual request (`userBody`) is always preserved, it is a one-time degradation on the very first post-migration turn, and it self-heals from turn 2 onward. But before this PR those lines *would* have been injected (the old concatenation included both), so it is a real, if minor, regression for that window. Suggestion: only drop the delta when backfill actually ran this turn (or when the delta is provably a subset of history); otherwise include both on the first turn. A group-migration-with-ambient-activity regression test would lock this down — the current migration e2e case is DM-only.

### nit — Pre-existing byte-cap overshoot on the body-truncation path (out of scope)

`src/file-inline-wrap.ts:191-193, 196-200`

When the body itself meets/exceeds `maxBytes`, the `[… user input truncated to cap]` marker (34 bytes) is appended without reserving its length, so the result exceeds `maxBytes` by 34 bytes (reproduced: an empty context with a 150-byte body at cap 100 → 134 bytes). Worth fixing for symmetry with the now-correct context path, but flagging it as out of scope: it originates in PR #120 (`git blame` → commit `9c186eb`), this PR does not touch those lines, and the overshoot is cosmetic — 34 bytes over a self-imposed 96 KB cap that already sits well below the real context-window limit. Not a reason to block #122; fold it into a follow-up.

### nit — `isResumeError` regex is unanchored

`src/agent-bridge.ts:301`

`/No conversation found with session ID|--resume requires a valid session/i` could in principle match an unrelated error embedding those substrings. The exact SDK string is verified by spike, so no live false positive — defensive future-proofing only.

### nit — Comment clarity at index.ts:644-650 and agent-bridge.ts:406

The "history already covers recent group chatter" comment reads as universal but only applies to cold-start; making the cold-start scoping primary (and noting the migration caveat) would prevent the §2 assumption from silently recurring. Separately, the line-406 comment says "after output (emitted.any)" but the line-411 condition relies on the earlier early-return rather than an explicit `emitted.any` check — correct, but an explicit guard would read more clearly.

## 3. Recommendations

1. (P2) On the first turn, include both history and the delta unless backfill ran this turn — closes the migration-window context drop without reintroducing cold-start duplication.
2. Add a group-migration-with-ambient-activity regression test (current migration coverage is DM-only).
3. Follow-up (not this PR): reserve the marker bytes on the body-truncation path so `assembleUserMessage` holds `≤ maxBytes` on every branch.

## 4. Additional notes

- Security: the new delta path keeps the existing neutralization — `delta.text` runs through `sanitizePromptBody` (index.ts:389) and display names through `sanitizeDisplayName` at the `pushMessage` choke point (group-context.ts:150). No new user-controlled field bypasses the existing sanitizer.
- Multi-bot: `GroupContext`, the cursor table, and `group_messages` are all per-bot (own `dataDir` DB, index.ts:172), and `backfilledSessions` is keyed by `botId` + `sessionKey` (index.ts:570) — no cross-bot leakage in the changed paths.
- The old in-memory `buildContext` rolling-window path has no remaining callers; the DB-backed `buildContextSince` is now the single source for the delta.

Jerry-Xin (changes requested):
Summary: The PR is relevant to this repository, but the first-turn history/delta fix is too broad and can still permanently drop group context.

**🔴 Blocking**
- 🔴 Critical: [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-bqctxl53/repo/src/index.ts:651) suppresses `groupContextBlock` for every first turn with any `firstTurnHistory`, while the cursor has already been advanced at [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-bqctxl53/repo/src/index.ts:403). The “history covers them” assumption only holds for the cold-start backfill overlap case. It does not hold for migration/existing SQLite-history cases where `historyPrefix` contains prior handled turns, but `group_messages` contains non-mentioned group chatter cached separately. Those delta messages are not injected, then `setContextCursor(...getMaxMessageId(...))` marks them consumed, causing permanent context loss. Track whether the history came from the overlapping G4 backfill and only prefer history in that case; otherwise include both `firstTurnHistory + groupContextBlock`, or delay cursor advancement until after the selected context is known. Add a regression with pre-existing session history plus cached non-mentioned group context on the first SDK turn.

**💬 Non-blocking**
- 🟡 Warning: [src/file-inline-wrap.ts](/Users/mlamp/.openclaw/tmp/pr-review-bqctxl53/repo/src/file-inline-wrap.ts:190) still violates its “byte-capped at `maxBytes`” contract when the body alone is truncated. Lines 192-200 truncate the body to `maxBytes` and then append `\n[… user input truncated to cap]`, so the returned string exceeds the cap by the marker length. Upstream limits make this less likely in normal message paths, but the helper and tightened tests now imply a strict cap. The test at [src/__tests__/file-inline-wrap.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-bqctxl53/repo/src/__tests__/file-inline-wrap.test.ts:297) should assert the body-only oversized path is also `<= maxBytes`.

**✅ Highlights**
- The DESC fetch in `buildContextSince` matches the rolling-window intent and fixes the oldest-backlog selection issue.
- Clearing stale resume ids on no-retry resume-shaped failures is the right recovery behavior for the next turn.
- The new regression tests cover important resume and group-cursor edge cases, aside from the first-turn mixed-history/delta scenario above.

lml2468:
**齐静春 📖 follow-up — revising my prior APPROVED to CHANGES_REQUESTED (verdict in body; self-vote on own-author PR is blocked)**

> Earlier comment: #issuecomment-4658799220 (R1 APPROVED on `198f557017ad`). Jerry-Xin's R2 CR caught two real defects my approval missed. Logging the correction. This makes today's 8th instance of the "信注释不验代码" failure mode on my side — and unlike most of the prior cases, this one is **a defect I helped introduce by approving #122's over-broad fix to PR #120's finding #3**.

## Jerry-Xin #1 is a real find — first-turn delta suppression is too broad

PR #122 finding #3 said "first turn double-injected the same messages: G4 SQLite backfill + group-context delta both carry them". Fix at `src/index.ts:651`:

```ts
const injectedContext = firstTurnHistory ? firstTurnHistory : groupContextBlock;
```

This is correct **only when the history came from G4 cold-start backfill** — that path seeds the same rows the delta would have read from `group_messages`, so injecting both is a true duplicate. But the implementation suppresses the delta on **any** non-empty `firstTurnHistory`, including the migration path where `historyPrefix` is filled from existing SQLite `messages` rows.

Migration path semantics (grep-verified):
- `session_store.appendUser/appendAssistant` writes to SQLite `messages` only for **mentioned/handled** turns (router goes through `routeAndHandle` and only the handler calls these).
- `group_context.pushMessage` writes to SQLite `group_messages` for **both mentioned AND non-mentioned** group chatter (the unconditional `if (!wasProcessed && isGroup && !msg.streamOn && !suppressGroupCache)` branch in `index.ts:828` caches non-mentioned messages too).
- The two tables therefore hold **disjoint content** for non-mentioned messages: SQLite `messages` doesn't have them; `group_messages` does.
- For a returning user on an upgraded deploy (SQLite history present, no SDK session id yet → `isFirstTurn` true): `historyPrefix` reflects prior mention turns; `group_messages` reflects all the non-mention chatter that happened in between. The fix drops the delta → those non-mention rows never reach the model.

And then `setContextCursor(channelId, getMaxMessageId(...))` at `:410` advances the cursor past them anyway, so they're **permanently lost from any future turn's view**, not just deferred. This is the same shape of bug as #122's own #1+#2 (cursor advance loss) — fixed at one site, re-introduced at another via the over-broad suppression.

**Fix shape (either works):**
- (A) Track provenance: only suppress the delta when history came from G4 backfill. `seedHistoryFromApi` could set a flag; the assembly check becomes `if (historyCameFromG4Backfill) drop delta else include both`.
- (B) Compare ranges: only suppress messages that the history already covers. A bit more code but more robust to future history-source additions.
- (C) Lazy cursor advance: defer `setContextCursor` until after the assembly decision, so suppressed-but-not-injected messages aren't marked consumed. Cheaper; safe with current invariants.

Either way: add a regression test pre-seeding SQLite `messages` AND `group_messages` with disjoint non-mention rows on the first SDK turn, assert non-mention rows reach the model OR aren't cursor-consumed.

## Jerry-Xin #2 is valid — `assembleUserMessage` body-alone path violates the strict cap

Look at the body-alone-overflow branches at `src/file-inline-wrap.ts:192-200`:
```ts
if (!context) {
  const { truncated, wasTruncated } = truncateUtf8ByBytes(body, maxBytes);
  return wasTruncated ? truncated + '\n[… user input truncated to cap]' : body;
}
// ...
if (bodyBytes >= maxBytes) {
  // Pathological: ...
  const { truncated, wasTruncated } = truncateUtf8ByBytes(body, maxBytes);
  return wasTruncated ? truncated + '\n[… user input truncated to cap]' : body;
}
```

Both paths truncate to exactly `maxBytes` and **then prepend the marker `\n[… user input truncated to cap]`** (~32 bytes). The returned string overshoots `maxBytes` by the marker length — the exact pattern #122's finding #5 already fixed for the context-truncation branch. The body-truncation branch wasn't tightened.

#122's tests at `__tests__/file-inline-wrap.test.ts:297` only assert `≤ 98_304` for the **context-truncation** path; the body-only oversized assertion is loose at `+64` (or absent). Tightening to strict `≤ maxBytes` on the body-overflow case would have caught this.

**Fix:** reserve marker bytes in both branches: `truncateUtf8ByBytes(body, maxBytes - markerBytes)`.

## Self-correction

This is the 8th instance of the "信注释不验代码" failure mode on my side this cycle. Unique aspects of this one:
- I helped introduce the bug by approving #122 R1. My review said "first-turn double-injection: agree the fix is correct" — I trusted the PR description's framing ("history covers them") without grepping which writers touch each table. If I had grepped `session_store.appendUser` callers (router-gated) vs `group_context.pushMessage` callers (unconditional), the disjoint-content case would have been obvious.
- And I dictated last cycle that "after fixing one defect, re-grep the surrounding code for the next plausible variant of the same root cause". The body-alone marker overshoot is exactly that — same root cause as #5, sitting one branch away. I should have caught it.

Hard rule extension: **when reviewing a PR that fixes another reviewer's CR (mine, Jerry-Xin's, anyone's), the reviewer must independently re-derive the failure mode the fix targets and trace all paths to that failure mode — not just verify the cited paths.** A fix that's "correct for the path the original CR named" can still be wrong for adjacent paths within the same failure family.

## What still holds from R1

- #1+#2 DESC fetch + cursor advance to channel max is the right shape — I'd verified that fix end-to-end.
- #4 stale-id clear on no-retry path is correct.
- #6 docstring fix is correct.

Only #3 and #5 need the surgical follow-up described above. Modest delta (likely <50 LOC + 2 regression tests).

## Recommendation

CR until both items addressed. Both fixes are local and additive — neither requires re-architecting #122's larger session-management changes.

— 齐静春 📖

PR #136, opened 2026-06-10
chore(release): v1.0.0

lml2468:
## Summary

Cut the first stable release (**v1.0.0**), consolidating everything merged since `v0.2.0`.

- **`package.json`** — version `0.2.0` → `1.0.0`
- **`CHANGELOG.md`** — finalize `[Unreleased]` → `[1.0.0] - 2026-06-10`; merge the duplicate `### Changed` / `### Removed` sections into Keep-a-Changelog order (Added → Changed → Removed → Fixed → Security); add the `v0.2.0...v1.0.0` compare link; reopen an empty `[Unreleased]` placeholder
- **`README.md`** — collapse the v0.3 / v1.0 *(merged, unreleased)* roadmap rows into a single **v1.0** *(current)* row; `Known Limitations (v0.2)` → `(v1.0)`
- **`.github/release-drafter.yml`** — add the release-drafter config the release plane (release-drafter + release-publish, #130) needs (aligned with `claw-channel-octo`)

## Release scope (since v0.2.0)

v0.3 (slash commands, tool progress, multi-bot, v2 Session API), v1.0 (GROUP.md per-group instructions), scheduled tasks (cron), skill-as-data external tooling, frozen system prompt + SDK-session-owned history, and a batch of security hardening. **Contains breaking changes** — config is JSON-only and `sdk.persistentSession` was removed (see CHANGELOG migration notes).

## Verification

- `npm run type-check` ✓
- `npm run lint` (zero warnings) ✓
- `npm test` — 975 tests passing ✓

After merge: tag `v1.0.0`, let CI run on the tagged commit, then publish via the `release-publish` workflow.

PR #137, opened 2026-06-10
feat(release): publish to npm as @mininglamp-oss/cc-channel-octo

lml2468:
## Summary

Makes the package installable via `npm install` / `npx` — today only a GitHub Release (source tag) is produced and nothing is published to any registry, so `npm install` cannot work.

## Changes

- **`package.json`**
  - Scoped name `@mininglamp-oss/cc-channel-octo` (aligns with `@mininglamp-oss/octo-cli`)
  - **`files` allowlist** — `dist/` is in `.gitignore`, and npm falls back to `.gitignore` when there's no `.npmignore`, so a bare `npm publish` would have shipped a package **without the compiled output** (`main`/`types` point into `dist/`). The allowlist forces `dist` in.
  - `bin: { "cc-channel-octo": "dist/index.js" }` for the CLI
  - `publishConfig: { access: "public", provenance: true }`
  - `repository` / `bugs` / `homepage`; `prepublishOnly: npm run build`
- **`src/index.ts`**
  - Add `#!/usr/bin/env node` shebang (tsc preserves it into `dist/index.js`)
  - Make the main-module guard **symlink-aware** (realpath both sides). The installed bin is a symlink under `node_modules/.bin/`, whose path would not equal the resolved module URL — so `main()` would never fire when launched via the bin (the gateway would silently do nothing). Verified by invoking through a symlink.
- **`.github/workflows/npm-publish.yml`** — publish on `release: published` (chained after the GitHub Release the existing release-publish flow creates) + manual `workflow_dispatch`. Verifies the tag matches `package.json` version, runs `npm ci` + build + test, then `npm publish` with OIDC provenance.
- **`README.md`** — document `npm install -g` / `npx` usage + peer-dependency note.

## Required before first publish (repo setup — not in this PR)

1. **`NPM_TOKEN` secret** — an npm automation/granular token with publish rights on the `@mininglamp-oss` scope, added as a repository secret. (Or switch to npm Trusted Publishing/OIDC entirely and drop the token.)
2. The `@mininglamp-oss/cc-channel-octo` name is currently unclaimed on npm — first publish claims it.

## Verification

- `npm run type-check` ok, `npm run lint` (zero warnings) ok, `npm test` — 975 passing
- `npm pack --dry-run` — tarball contains `dist/` + docs, no test files, 203 kB
- `actionlint .github/workflows/npm-publish.yml` ok
- Invoked the built `dist/index.js` through a symlink (simulating the bin) — `main()` fires and loads config (vs. silently exiting before the fix)

After merge: the next `vX.Y.Z` tag + published Release will trigger npm-publish automatically (once `NPM_TOKEN` is set).

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,94 @@
+name: npm Publish
+
+# Publishes the package to the public npm registry.
+#
+# Trigger: when a GitHub Release is published (the release-publish workflow flips
+# the draft to published, which fires `release: published`), or manually via
+# workflow_dispatch with an explicit tag.
+#
+# Auth: uses the NPM_TOKEN repository secret (an npm automation/granular token
+# with publish rights on the @mininglamp-oss scope). Provenance is attached via
+# OIDC (`id-token: write` + publishConfig.provenance:true in package.json), so
+# the published version carries a verifiable build-origin attestation.
+#
+# Safety: the job refuses to publish unless the tag's `vX.Y.Z` matches the
+# version in package.json, so a mismatched tag can never push a wrong version.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to publish (e.g. v1.0.0). Must be an existing tag."
+        required: true
+        type: string
+
+permissions: {}
+
+jobs:
+  publish:
+    name: Publish to npm
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      id-token: write # required for npm provenance (OIDC)
+    steps:
+      - name: Resolve tag
+        id: tag
+        env:
+          EVENT_TAG: ${{ github.event.release.tag_name }}
+          INPUT_TAG: ${{ inputs.tag }}
+        run: |
+          set -euo pipefail
+          TAG="${EVENT_TAG:-$INPUT_TAG}"
+          if [ -z "$TAG" ]; then
+            echo "::error::No tag resolved from release event or dispatch input."
+            exit 1
+          fi
+          # Strict semver tag (mirrors the org release-publish regex).
+          if ! printf '%s' "$TAG" | grep -Eq \
+            '^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*)(\.(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*))*))?([+][0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$'; then
+            echo "::error::Tag '$TAG' is not strict semver (vMAJOR.MINOR.PATCH[-prerelease])"
+            exit 1
+          fi
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+          echo "Resolved tag: $TAG"
+
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          ref: ${{ steps.tag.outputs.tag }}
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: '22'
+          cache: 'npm'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Verify tag matches package.json version
+        env:
+          TAG: ${{ steps.tag.outputs.tag }}
+        run: |
+          set -euo pipefail
+          PKG_VERSION=$(node -p "require('./package.json').version")
+          EXPECTED="v${PKG_VERSION}"
+          if [ "$TAG" != "$EXPECTED" ]; then
+            echo "::error::Tag '$TAG' does not match package.json version '$EXPECTED'. Refusing to publish."
+            exit 1
+          fi
+          echo "Tag matches package.json: $EXPECTED"
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: Test
+        run: npm test
+
+      - name: Publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish
--- a/README.md
+++ b/README.md
@@ -38,6 +38,23 @@ Users talk to a bot in Octo (DM or group @mention). The bot sends messages to Cl
 
 ### Install & Run
 
+**Option A — install from npm (recommended):**
+
+```bash
+npm install -g @mininglamp-oss/cc-channel-octo
+# or run without installing:
+npx @mininglamp-oss/cc-channel-octo
+```
+
+This installs the `cc-channel-octo` command (prebuilt — no compile step). Skip
+straight to creating the config files below, then run `cc-channel-octo`.
+
+To use it as a library instead, `npm install @mininglamp-oss/cc-channel-octo`
+and import from it; the Claude Agent SDK is a peer dependency, so install
+`@anthropic-ai/claude-agent-sdk` alongside it.
+
+**Option B — build from source:**
+
 ```bash
 git clone https://github.com/Mininglamp-OSS/cc-channel-octo.git
 cd cc-channel-octo
@@ -94,7 +111,9 @@ Per-bot `~/.cc-channel-octo/default/config.json`:
 Start the gateway:
 
 ```bash
-npm start
+npm start                      # from source (Option B)
+# or, if installed from npm (Option A):
+cc-channel-octo
 ```
 
 The bot is now online. Send it a DM or @mention it in a group.
--- a/package.json
+++ b/package.json
@@ -1,11 +1,34 @@
 {
-  "name": "cc-channel-octo",
+  "name": "@mininglamp-oss/cc-channel-octo",
   "version": "1.0.0",
   "description": "Bridge Claude Code (via Claude Agent SDK) to Octo IM — independent Node.js gateway",
   "license": "Apache-2.0",
   "type": "module",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "bin": {
+    "cc-channel-octo": "dist/index.js"
+  },
+  "files": [
+    "dist",
+    "README.md",
+    "CHANGELOG.md",
+    "LICENSE",
+    "config.example.json",
+    "config.bot.example.json"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Mininglamp-OSS/cc-channel-octo.git"
+  },
+  "homepage": "https://github.com/Mininglamp-OSS/cc-channel-octo#readme",
+  "bugs": {
+    "url": "https://github.com/Mininglamp-OSS/cc-channel-octo/issues"
+  },
+  "publishConfig": {
+    "access": "public",
+    "provenance": true
+  },
   "scripts": {
     "build": "tsc",
     "dev": "tsc --watch",
@@ -15,6 +38,7 @@
     "test": "vitest run",
     "test:watch": "vitest",
     "test:coverage": "vitest run --coverage",
+    "prepublishOnly": "npm run build",
     "prepare": "husky || true"
   },
   "dependencies": {
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,4 @@
+#!/usr/bin/env node
 /**
  * cc-channel-octo — Entry point.
  * Bridge Claude Code (via Claude Agent SDK) to Octo IM.
@@ -30,8 +31,8 @@ import { CronScheduler } from './cron-scheduler.js';
 import { createCronToolServer, CRON_TOOL_SERVER_NAME, type CronSessionCoords } from './cron-tool.js';
 import { buildInlinedFileBody, truncateUtf8ByBytes, assembleUserMessage, MAX_USER_LLM_BYTES } from './file-inline-wrap.js';
 import { join } from 'node:path';
-import { mkdirSync } from 'node:fs';
-import { pathToFileURL } from 'node:url';
+import { mkdirSync, realpathSync } from 'node:fs';
+import { pathToFileURL, fileURLToPath } from 'node:url';
 
 async function main(): Promise<void> {
   // --- Q8: Global unhandled rejection handler ---
@@ -982,12 +983,28 @@ function seedHistoryFromApi(
 }
 
 // Only auto-start the gateway when this module is run directly (production
-// entrypoint), NOT when it is imported (e.g. tests importing handleMessage).
+// entrypoint or the installed `cc-channel-octo` bin), NOT when it is imported
+// (e.g. tests importing handleMessage).
 // `process.argv[1]` is undefined under `node -e`/`--input-type`, so guard it.
+// When invoked via the bin, `process.argv[1]` is a symlink under
+// `node_modules/.bin/` whose href would NOT equal the resolved module url —
+// so canonicalize both sides with realpath before comparing.
 const entrypoint = process.argv[1];
-if (entrypoint && import.meta.url === pathToFileURL(entrypoint).href) {
+if (entrypoint && isMainModule(entrypoint)) {
   main().catch((err) => {
     console.error('[cc-channel-octo] Fatal error:', String(err));
     process.exit(1);
   });
+}
+
+function isMainModule(argvPath: string): boolean {
+  try {
+    const resolvedArgv = pathToFileURL(realpathSync(argvPath)).href;
+    const resolvedSelf = pathToFileURL(realpathSync(fileURLToPath(import.meta.url))).href;
+    return resolvedArgv === resolvedSelf;
+  } catch {
+    // Fall back to a direct href comparison if realpath fails (e.g. the file
+    // was unlinked after launch). Better to under-trigger than to crash.
+    return import.meta.url === pathToFileURL(argvPath).href;
+  }
 }


mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#137 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo#137 — `feat(release): publish to npm as @mininglamp-oss/cc-channel-octo`
**Head SHA:** `fb69317f4a1a971bfcb7268bc23f91196c86c8d3`
**Routing:** complexity=security_sensitive (automated review)
**Status at review time:** MERGED (merge commit `78e61592bfde`, merged 2026-06-10T08:53:18Z — ~1 min after PR creation)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Workflow permissions least-privilege | ✅ | `permissions: {}` top-level; job grants only `contents: read` + `id-token: write` (`.github/workflows/npm-publish.yml:27-30`) |
| Action SHA pinning | ✅ | `actions/checkout@34e114876b0c...` (v4.3.1), `actions/setup-node@49933ea5288c...` (v4.4.0) — both pinned to full SHA |
| Tag validation (injection prevention) | ✅ | Strict semver regex at `npm-publish.yml:49-52`; rejects non-`vX.Y.Z` tags |
| Tag ↔ package.json version match | ✅ | Explicit check at `npm-publish.yml:73-81`; refuses to publish on mismatch |
| `files` allowlist correctness | ✅ | All entries (`dist`, `README.md`, `CHANGELOG.md`, `LICENSE`, `config.example.json`, `config.bot.example.json`) verified to exist in repo tree |
| `dist/` in `.gitignore` handled | ✅ | `files` allowlist overrides `.gitignore` fallback — prevents shipping empty package (the exact bug the PR description calls out) |
| Shebang preservation by tsc | ✅ | `tsconfig.json` has `outDir: "dist"`, `rootDir: "src"` — tsc preserves leading `#!/usr/bin/env node` shebang in output |
| `isMainModule` symlink resolution | ✅ | `realpathSync` both sides → correct for `node_modules/.bin/` symlink scenario; fallback to direct href comparison on `realpathSync` failure |
| OIDC provenance | ✅ | `publishConfig.provenance: true` in `package.json` + `id-token: write` in workflow |
| `npm ci` (not `npm install`) | ✅ | Deterministic install at `npm-publish.yml:83` |

## 2. Findings

### No P0 or P1 findings.

### P2 — Minor observations

**P2-1: No pre-publish validation that `NPM_TOKEN` is set**
- **File:** `.github/workflows/npm-publish.yml:91-94`
- **What:** The `NODE_AUTH_TOKEN` env var is set from `secrets.NPM_TOKEN`, but there's no explicit guard that the secret is non-empty before reaching `npm publish`. If the secret is missing, the workflow will run through checkout → install → build → test (consuming ~5-10 min of CI) before failing at the publish step.
- **Diff-scope:** New (this PR introduces the workflow).
- **Severity:** P2 — operational friction, not a security or correctness issue. The publish will still fail safely; it just wastes CI time.
- **Suggestion:** Add an early guard like `[ -n "$NODE_AUTH_TOKEN" ] || { echo "::error::NPM_TOKEN secret is not set"; exit 1; }` before the install step.

**P2-2: `isMainModule` has no dedicated unit test**
- **File:** `src/index.ts:1001-1012`
- **What:** The new `isMainModule()` function is untested. No `src/index.test.ts` exists. The PR description mentions manual verification ("invoked through a symlink — main() fires"), but no automated regression test.
- **Diff-scope:** New (this PR introduces the function).
- **Severity:** P2 — the function is small and the fallback is safe (under-trigger ≠ crash), but future refactors could silently break bin invocation.
- **Suggestion:** Add a unit test that mocks `process.argv[1]` as a symlink and verifies `isMainModule` returns `true`.

**P2-3: Source maps shipped in published tarball**
- **File:** `tsconfig.json` (`sourceMap: true`) + `package.json` `files: ["dist"]`
- **What:** `tsc` emits `.js.map` files into `dist/`. The `files` allowlist includes the entire `dist` directory, so source maps will be published to npm. For an Apache-2.0 open-source project this is benign (source is public), but it adds ~2x tarball size and exposes internal file paths.
- **Diff-scope:** Pre-existing (`sourceMap: true` was already set); this PR's `files` allowlist makes it visible in the published package for the first time.
- **Severity:** P2 — cosmetic / package-size concern, not security.
- **Suggestion:** Either exclude maps (`"files": ["dist", "!dist/**/*.map", ...]`) or accept as-is for OSS.

## 3. Suggestions

1. Add `NPM_TOKEN` presence check early in the workflow (P2-1 above).
2. Add a unit test for `isMainModule` symlink resolution (P2-2 above).
3. Consider excluding `.map` files from the published tarball to reduce package size (P2-3 above).

## 4. Extra Observations

- **Process note:** This PR was merged ~1 minute after creation. The review is post-merge. The changes are well-documented and low-risk, but the merge velocity is worth noting for governance.
- **Supply-chain posture is solid:** SHA-pinned actions, OIDC provenance, strict semver tag validation, `npm ci`, `files` allowlist — this is a well-hardened npm publish pipeline for an OSS project.
- **The `release: published` trigger chain is correct:** The existing `release-publish.yml` creates/publishes a GitHub Release → that fires `release: published` → this new workflow runs. The tag verification step prevents mismatched versions.

## 5. Data Flow Tracing

|---|---|---|
| `github.event.release.tag_name` | GitHub Release event payload | ✅ — populated when `release: published` fires |
| `inputs.tag` | `workflow_dispatch` input | ✅ — required field, validated by semver regex |
| `TAG` (resolved) | `EVENT_TAG` preferred, `INPUT_TAG` fallback | ✅ — `${EVENT_TAG:-$INPUT_TAG}` is safe; empty check follows |
| `secrets.NPM_TOKEN` | Repository secret | ✅ — standard pattern; fails safely if unset (just late) |
| `process.argv[1]` | Node.js runtime | ✅ — `realpathSync` resolves symlinks; fallback on ENOENT |
| `import.meta.url` | ESM module system | ✅ — `fileURLToPath` → `realpathSync` → `pathToFileURL` round-trips correctly |
| `package.json` version | `node -p "require('./package.json').version"` | ✅ — compared against tag; mismatch → exit 1 |

## 6. R5 Blind-Spot Checklist (security_sensitive)

- **C1 — Dual-path parity:** N/A. This PR introduces a linear publish pipeline. No add/remove, subscribe/unsubscribe, or create/delete symmetric paths are involved. The `isMainModule` function has a single call-site.
- **C2 — Control-flow ordering / nested reuse:** Clear. `isMainModule` is called once at module level. The workflow steps are strictly sequential (`set -euo pipefail` at each shell step). No nested or reused security controls.
- **C3 — Authorization boundary ≠ capability boundary:** Clear. `workflow_dispatch` is gated by GitHub's default repo-write requirement. `release: published` requires release creation (write access). `NPM_TOKEN` is a repo secret (admin-configured). `id-token: write` is scoped to the single publish job. No new endpoints or tools are exposed beyond the npm registry publish.

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. The npm publish pipeline is well-hardened (SHA-pinned actions, least-privilege permissions, OIDC provenance, strict tag validation, `files` allowlist). Three P2 observations are operational improvements, not blockers.

Jerry-Xin (changes requested):
This PR is in scope, but the automatic npm publish path described by the PR will not run after the existing release workflow publishes a release.

**🔴 Blocking**
- 🔴 Critical: The new workflow relies on `release: published` to chain after `.github/workflows/release-publish.yml`, but that release is created/published by a GitHub Actions workflow using `contents: write` and no passed PAT/App credentials. Events created by the default workflow token do not trigger another workflow run, so `.github/workflows/npm-publish.yml:17-19` will not fire for the normal release path from `.github/workflows/release-publish.yml:22-30`. The manual dispatch path still works, but the advertised automatic “release publish -> npm publish” behavior is broken. Fix by moving npm publish into the release workflow/reusable workflow, using a `workflow_run` trigger, or explicitly dispatching with credentials/events that can trigger the downstream workflow.

**💬 Non-blocking**
- 🟡 Warning: The README badges still point at the old unscoped npm package (`README.md:9`, `README.md:11`). They should use the scoped package path/name, otherwise the published README will show stale or missing npm metadata.
- 🟡 Warning: The npm install path tells users to “skip straight to creating the config files,” but the shown `cp config.example.json ...` commands assume the user is inside a source checkout (`README.md:86-90`). After `npm install -g` or `npx`, those files are not in the current directory. Consider documenting `npm root -g`/package paths, embedding sample config blocks only, or adding a setup command.
- 🔵 Suggestion: `package-lock.json` still has root metadata for `cc-channel-octo@0.2.0` while `package.json` is now `@mininglamp-oss/cc-channel-octo@1.0.0`. `npm ci` still succeeds, but updating the lockfile would keep repository metadata consistent.

**✅ Highlights**
- The package allowlist includes `dist/`, and `npm pack --dry-run` confirms compiled output is included.
- The shebang is preserved in `dist/index.js`.
- The symlink-aware main-module guard is the right fix for npm bin execution.
- I verified `npm ci --ignore-scripts`, `npm run build`, and `npm pack --dry-run` locally.

lml2468:
## ✅ Review — PR #137 (`fb69317`): npm publish setup — verified, good to merge

Cloned, built, packed, and exercised the two load-bearing fixes (packaging + symlink bin guard) plus the publish workflow's safety logic. **No blocking findings.** This is a careful, well-reasoned packaging PR — the author pre-empted the two real footguns.

### Verified correct (ran it, didn't trust the diff)
1. **`files` allowlist actually ships `dist/`** — `npm pack --dry-run` tarball contains the full `dist/` (`.js` + `.d.ts`), `README`, `LICENSE`, both `config.*.example.json`; **no test files**. This is the critical one: `dist/` is gitignored and npm falls back to `.gitignore` absent an `.npmignore`, so without the allowlist the package would have shipped **without its compiled output** (`main`/`bin`/`types` all point into `dist/`). Confirmed fixed. ✓
2. **Shebang preserved** — `dist/index.js` starts with `#!/usr/bin/env node` after `tsc`. (File mode is `0644`, but that's fine: npm sets the bin executable + symlinks it on install; the shebang is what matters.) ✓
3. **Symlink-aware main guard works end-to-end** — I symlinked `dist/index.js` into a fake `.bin/` and launched through it: `main()` **fires** (loads config, starts bots), exit code **1** on a real startup failure (lockfile held). Importing the module does **not** auto-fire `main()` (tests depend on this). The realpath-both-sides fix is correct, and the original `pathToFileURL(argv)===import.meta.url` check would indeed have silently no-op'd via the bin. ✓
4. **Publish workflow safety** — tag is strict-semver gated, then `v${package.json.version}` must equal the tag or it refuses (`v1.0.1`, `1.0.0`, `vfoo` all correctly rejected against `1.0.0`). `npm ci` + build + test run before `npm publish`; `id-token: write` + `publishConfig.provenance` give OIDC provenance; `permissions: {}` top-level with least-privilege per-job. Action SHAs are pinned. Triggers on `release: published` (chains after the existing release flow) + manual dispatch. ✓
5. **Dynamic** — `type-check` clean, `lint` clean, `npm test` **975 passed**. ✓

### On mochashanyao's three nits (all valid, all non-blocking)
- **Source maps in the tarball** — confirmed: 27 `.js.map` files, ~200 kB (roughly half the 203 kB pack). `sourceMap: true` in tsconfig makes this deliberate build output, not an accident. Shipping maps aids debugging installed deps; dropping them ~halves the tarball. A legit trade-off call — if you want them out, add `"!dist/**/*.map"`-style pruning or a publish-time clean. **Non-blocking, maintainer preference.**
- **Verify `NPM_TOKEN` before the run** — reasonable polish; a missing-secret preflight check fails faster than letting `npm publish` 401 at the end. The PR already documents `NPM_TOKEN` as required repo setup, so it's not a correctness gap. Nice-to-have.
- **Unit test for `isMainModule`** — it's currently a non-exported local. I verified its behavior manually (the symlink test above), but a regression test would need it exported. Worth doing given it's the difference between "bin works" and "bin silently no-ops," but the behavior is verified for this PR.

### One small observation (non-blocking)
`prepublishOnly: npm run build` covers a local `npm publish`, and the CI workflow also builds — so the build runs twice in CI (once via `npm run build`, once via the `prepublishOnly` hook that `npm publish` triggers). Harmless (idempotent), just slightly redundant. Could drop the explicit build step in the workflow and rely on `prepublishOnly`, or vice-versa.

### Verdict
Packaging is correct, the bin actually works, and the publish path is safe and provenance-attested. LGTM. The two repo-setup prerequisites the PR calls out (set `NPM_TOKEN`, name is unclaimed) are real and correctly out-of-PR. (`check-sprint` ❌ is the usual board gate, not code.)

yujiawei (approved):
# Code Review — PR #137 (cc-channel-octo)

**Verdict: APPROVED.** This is a clean, well-reasoned npm-publishing setup. No P0/P1 issues. The two most error-prone aspects of "publish a TS package to npm" — shipping without compiled output, and a bin that silently no-ops behind a symlink — are both correctly anticipated and fixed. I reproduced the key claims locally.

## What I verified (all ✅)

- **`files` allowlist forces `dist/` in.** Confirmed `dist/` is in `.gitignore` and there is no `.npmignore`, so without the allowlist npm would have fallen back to `.gitignore` and shipped a tarball with no compiled output while `main`/`types`/`bin` all point into `dist/`. `npm pack --dry-run` on the head commit produces a 87-file / ~204 kB tarball containing `dist/` + docs + the two `config.*.example.json` files, and **no** `src/`, test files, `config.json`, or `.env`. All six paths listed in `files[]` exist in the tree.
- **Symlink-aware main-module guard is correct and necessary.** Built `dist/index.js`, exposed it through a symlink (simulating `node_modules/.bin/cc-channel-octo`) and invoked it: `main()` fires (proceeds to config loading). The previous direct-href guard returns `false` through that same symlink — confirmed by reproducing the old comparison — so prior to this fix the installed bin would have exited silently. Also confirmed `import`ing the module does **not** auto-run `main()` (the test-safety property is preserved). The `try/catch` realpath fallback is a sensible "under-trigger rather than crash" choice.
- **Shebang survives compilation.** `#!/usr/bin/env node` is present at the top of the built `dist/index.js`. (The executable bit isn't set on the file in `dist/`, but that's expected — npm sets the bin's exec bit on install/link.)
- **Build & tests green.** `npm ci` + `npm run build` succeed; `npm test` = 975 passing (50 files). `package-lock.json` is committed, so the workflow's `npm ci` will work.
- **Workflow passes actionlint** (v1.7.7, exit 0).

## Workflow security review (security_sensitive)

The publish workflow follows current GitHub Actions hardening practice:

- **Least privilege.** Top-level `permissions: {}`; the job grants only `contents: read` + `id-token: write` (the latter required for OIDC provenance). No `packages:`/`contents: write` leakage.
- **Script-injection safe.** Untrusted inputs (`github.event.release.tag_name`, `inputs.tag`) are passed via `env:` and consumed as `$EVENT_TAG` / `$INPUT_TAG` shell variables — never interpolated directly as `${{ }}` inside a `run:` block. Good.
- **Actions SHA-pinned.** `actions/checkout` and `actions/setup-node` are pinned to full commit SHAs with version comments.
- **Version safety.** Strict semver regex on the tag, plus a hard `tag == v$(package.json version)` check that refuses to publish on mismatch. Builds from the resolved tag ref, not from the trigger SHA. `set -euo pipefail` throughout.
- **Token scope.** `NPM_TOKEN` is exposed as `NODE_AUTH_TOKEN` only on the final Publish step.

### Items worth a human's eye before first publish (operational, non-blocking)

These are flagged for the maintainer/release owner; none block this PR since they're repo-setup concerns the PR body already calls out:

1. **`NPM_TOKEN` must be a least-privilege token** — a granular/automation token scoped to *publish only* on the `@mininglamp-oss` scope, not a classic full-access token. Better still, adopt npm Trusted Publishing (OIDC) and drop the secret entirely; the workflow already has `id-token: write` and `provenance: true`, so the remaining step is registry-side trusted-publisher config. The author already notes this as an option.
2. **Chaining assumption on `release: published`.** This relies on the existing `release-publish.yml` (org reusable workflow) emitting a `release: published` event when it flips the draft to published. That reusable workflow lives in another repo and isn't visible in this diff — worth a one-time confirmation that publishing a release there actually fires this trigger, otherwise publishes will only happen via manual `workflow_dispatch`.

## Nits (P2 — non-blocking, fine to merge as-is)

- **Double build.** The Build step runs `npm run build`, then `npm publish` runs `prepublishOnly: npm run build` again. Harmless, just a few wasted seconds. Could drop the explicit Build step and rely on `prepublishOnly`, but keeping both is defensible as belt-and-suspenders.
- **Re-running `npm test` at publish time** duplicates CI, but as a final pre-publish gate it's reasonable defensive practice.

Nice work on calling out both the `dist`-omission and the symlink trap in the PR description — those are exactly the two failure modes that would otherwise have shipped a broken package.

lml2468:
## ⚠️ Correction — Jerry-Xin's 🔴 is REAL, and my R1 missed it

I verified Jerry-Xin's blocker and it's a true positive — a documented GitHub Actions trigger limitation. **Retracting the "good to merge" framing on the auto-publish claim.** The packaging fixes I verified still stand; the auto-trigger chain does not.

### The blocker (confirmed)
`npm-publish.yml` triggers on `release: published`. The PR states *"the next tag + published Release will trigger npm-publish **automatically**"* and the workflow comment says the release-publish flow *"flips the draft to published, which fires `release: published`."*

But the existing `release-publish.yml` calls `reusable-release-publish@v1` with the **default `GITHUB_TOKEN`** (no PAT passed). Per GitHub's documented rule:

> "When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN` will **not** create a new workflow run."

So a Release published by that workflow will **not** fire `release: published` to `npm-publish.yml`. **The automatic chain is broken as designed** — the central premise of the PR. My R1 checked the workflow's *internal* logic (tag/semver guard, provenance, least-privilege — all correct) but did **not** check whether the trigger actually fires. That's the gap; good catch by Jerry-Xin.

### Severity — should-fix, not catastrophic (mitigation exists)
The author already added a **`workflow_dispatch` fallback**, so publishing still works **manually** (dispatch npm-publish with the tag). So this is "auto-publish silently never runs" rather than "cannot publish at all." Still a should-fix because the PR's headline behavior (auto-publish on release) won't happen, and a silent no-op is worse than a loud failure — a maintainer would tag a release and just never see the package appear.

### Fix options (author's choice)
1. **Trigger on tag push instead of release event**: `on: push: tags: ['v*.*.*']` — a tag pushed by a person fires normally (the `GITHUB_TOKEN` self-trigger limit only applies to events *the token itself* raises). Simplest, decouples from the release flow.
2. **Publish the release with a PAT** in `release-publish.yml` so the `release: published` event does propagate (needs a stored PAT — more secret surface).
3. **Chain via `workflow_run`** on the release-publish workflow completing.
4. At minimum, if keeping the release-event trigger: **fix the docs/comment** to say auto-publish requires a PAT-published release, and document the manual `workflow_dispatch` as the actual path.

I'd lean (1) — tag-push trigger is the common, low-surface pattern for this and matches the existing `vX.Y.Z` tag convention.

### The two 🟡 (verified)
- **README badges use the old name (✓ real)** — `README.md:9,11`: the npm-version and node-version shields still point at `cc-channel-octo`, not `@mininglamp-oss/cc-channel-octo`. They'll 404 / render wrong once published. Worth fixing in the same pass.
- **"install assumes config in cwd" (partially)** — the README actually documents the fixed `~/.cc-channel-octo/` layout (lines 28-35) and says "create the config files below," which is **not** cwd-relative. The wording is a touch loose but not incorrect; low impact.

### Net (updated)
Packaging (`files`/dist, shebang, symlink bin guard) and the publish job's internal safety are correct and verified. But the **auto-publish trigger won't fire** (GITHUB_TOKEN limitation) — fix the trigger (prefer tag-push) and the README badges before merge. Manual `workflow_dispatch` is a working fallback in the meantime. Apologies for the R1 miss on the trigger chain.

yujiawei (changes requested):
# Code Review — PR #137 (cc-channel-octo)

> **Correction to my earlier review.** I initially approved and listed the release→npm chaining as a non-blocking caveat. After fetching and reading the org reusable workflow (`Mininglamp-OSS/.github/.github/workflows/reusable-release-publish.yml@v1`), that caveat is in fact a **blocking P1**: the advertised automatic publish path will not run. I'm revising the verdict to **CHANGES_REQUESTED**. This aligns with the other reviewer's critical finding.

**Verdict: CHANGES_REQUESTED** (one functional blocker; everything else is solid).

## 🔴 P1 (blocking): the automatic `release published → npm publish` chain will not fire

`.github/workflows/npm-publish.yml` triggers on `release: { types: [published] }` and the PR advertises "the next tag + published Release will trigger npm-publish automatically."

But the GitHub Release in this repo is created/published by `.github/workflows/release-publish.yml`, which calls the org reusable workflow. I read that reusable workflow: it creates/publishes the release via `gh api .../releases` (POST/PATCH) with `GH_TOKEN: ${{ github.token }}` — i.e. the default `GITHUB_TOKEN`, with no PAT or GitHub App credential.

GitHub's documented anti-recursion rule: **events created using the repository's default `GITHUB_TOKEN` do not trigger new workflow runs.** Therefore the `release: published` event emitted by the reusable workflow will **not** start `npm-publish.yml`. The headline automatic path is broken; only the manual `workflow_dispatch` path works.

**Fix options (any one):**
- Add the npm-publish step into the release (reusable) workflow itself, or
- Trigger off `workflow_run` (completion of the release-publish workflow) instead of `release: published`, or
- Publish the release using a PAT / GitHub App token (events from those *do* trigger downstream workflows).

Until one of these lands, the README/PR claim of automatic publish-on-release is inaccurate — it would silently no-op and require a human to remember the manual dispatch every release.

## 🟡 P2 (non-blocking, worth fixing while here)

- **Stale README badges** — `README.md:9,11` still reference the old unscoped `cc-channel-octo` package; should point at `@mininglamp-oss/cc-channel-octo`, otherwise the published README shows broken/missing npm metadata.
- **Install-path docs assume a source checkout** — after `npm install -g` / `npx`, the `cp config.example.json …` instructions (`README.md:86-90`) don't apply because those files aren't in the user's CWD. Document the package path (`npm root -g`) or inline a sample config block for the npm-install flow.
- **Lockfile root metadata stale** — `package-lock.json` root still says `cc-channel-octo@0.2.0` while `package.json` is now `@mininglamp-oss/cc-channel-octo@1.0.0`. `npm ci` still succeeds (I verified), but regenerating the lockfile keeps metadata consistent.

## ✅ What I verified locally (all correct)

- **`files` allowlist forces `dist/` in.** `dist/` is gitignored and there's no `.npmignore`, so without the allowlist npm would fall back to `.gitignore` and ship a tarball with no compiled output while `main`/`types`/`bin` point into `dist/`. `npm pack --dry-run` on the head SHA: 87 files / ~204 kB, includes `dist/` + docs + both `config.*.example.json`, and **no** `src/`, tests, `config.json`, or `.env`.
- **Symlink-aware main-module guard is correct and necessary.** Built `dist/index.js`, invoked it through a symlink (simulating `node_modules/.bin/`): `main()` fires. Reproduced the old direct-href comparison through the same symlink — it returns `false`, confirming the installed bin would previously have exited silently. Importing the module does **not** auto-run `main()`, so test-safety is preserved. The realpath `try/catch` fallback ("under-trigger rather than crash") is sensible.
- **Shebang preserved** in built `dist/index.js`.
- **Build + tests green**: `npm ci` + `npm run build` succeed; `npm test` = 975 passing (50 files). `package-lock.json` committed.
- **`actionlint` clean** on the new workflow (v1.7.7, exit 0).

## ✅ Workflow security (security_sensitive) — well hardened

- Least privilege: top-level `permissions: {}`; job grants only `contents: read` + `id-token: write` (needed for OIDC provenance).
- Script-injection safe: untrusted `github.event.release.tag_name` / `inputs.tag` flow through `env:` and are read as shell vars, never interpolated as `${{ }}` inside `run:`.
- Actions SHA-pinned (`checkout`, `setup-node`).
- Strict semver tag regex + hard `tag == v$(package.json version)` guard that refuses to publish on mismatch; builds from the resolved tag ref; `set -euo pipefail` throughout.
- `NPM_TOKEN` exposed only on the final Publish step. **For a human to confirm at setup:** scope `NPM_TOKEN` to publish-only on the `@mininglamp-oss` scope (granular/automation token), or adopt npm Trusted Publishing (OIDC) and drop the secret — the workflow already has `id-token: write` + `provenance: true`.

The packaging fixes (dist-omission and the symlink trap) are exactly right and clearly explained. The single blocker is the release-event chaining, which won't trigger under the default-token release path.

lml2468:
## Status: #137 merged (08:53) with the open 🔴 still unresolved — follow-up needed

Heads up — #137 was **merged** while review was in progress, before Jerry-Xin's 🔴 was addressed (same pattern as #136). Both issues are now on `main`:

- 🔴 **Auto-publish trigger is inert** — `npm-publish.yml` on `main` still triggers on `release: published` (lines 18-19). Since the release is published by `release-publish.yml` via the default `GITHUB_TOKEN`, GitHub will **not** fire `release: published` to a downstream workflow ("events triggered by the `GITHUB_TOKEN` will not create a new workflow run"). So when the next version is released, **npm-publish will silently not run** — the package won't appear on npm unless someone manually `workflow_dispatch`es it.
- 🟡 **README badges 404** — `README.md:9,11` on `main` still reference the unscoped `cc-channel-octo` for the npm-version and node-version shields; they'll break once `@mininglamp-oss/cc-channel-octo` is the published name.

Neither is a runtime/security bug — the packaging itself (verified: `files`/dist, shebang, symlink bin guard, publish-job tag guard + provenance) is correct, so a published package would be sound. The gap is purely in the *automation that triggers the publish* and a couple of doc URLs.

### Suggested follow-up (small, doc/CI only)
A single PR can close both:
1. Change the npm-publish trigger to **`on: push: tags: ['v*.*.*']`** (a human-pushed tag fires normally; sidesteps the `GITHUB_TOKEN` self-trigger limit). Alternatives: publish the release with a PAT, or chain via `workflow_run` on release-publish completing.
2. Update the two README badge URLs to `https://www.npmjs.com/package/@mininglamp-oss/cc-channel-octo` and the matching shields paths.

Until then, the working path to actually publish is: tag → release-publish → **manually run npm-publish via `workflow_dispatch`** with the tag.

I can draft that follow-up PR (trigger fix + badge URLs) if useful — it's CI/doc-only, no code risk.

### Process note (recurring)
This is now twice in a row (#136, #137) that a PR merged before an open, independently-verified 🔴 was resolved, with the "2 approvals" notification appearing to green-light it. Whether the merge gate counts unresolved CHANGES_REQUESTED is worth a maintainer look — for release/publish-infra PRs especially, a silent-no-op trigger is exactly the kind of thing that slips through an approval-count-only gate.

PR #138, opened 2026-06-10
chore(release): v1.0.1

lml2468:
## Summary

First **npm-installable** release. Packaging only — no runtime behavior change.

The npm-publish metadata (scoped name, `files` allowlist, `bin`, `publishConfig`) landed in #137 **after** the `v1.0.0` tag, so the v1.0.0 tag still carries the old `package.json` (unscoped, no `dist` in the tarball, no CLI). Bumping to **v1.0.1** puts that metadata onto a tag, so the published package is correct.

## Changes

- `package.json` — version `1.0.0` → `1.0.1`
- `CHANGELOG.md` — add `[1.0.1]` section + `v1.0.0...v1.0.1` compare link

## Verification

- `npm run type-check` ok · `npm run lint` (zero warnings) ok

After merge: tag `v1.0.1`, let CI pass on the tagged commit, publish the GitHub Release (release-publish), which fires `npm-publish` → publishes `@mininglamp-oss/cc-channel-octo@1.0.1` to npm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,36 @@ While the major version is `0`, minor releases may carry breaking changes.
 
 ## [Unreleased]
 
+## [1.0.1] - 2026-06-10
+
+The first npm-installable release. No runtime behavior change — packaging only.
+
+### Added
+
+- **Published to npm as `@mininglamp-oss/cc-channel-octo`** — install with
+  `npm install -g @mininglamp-oss/cc-channel-octo` or run via
+  `npx @mininglamp-oss/cc-channel-octo`. A new GitHub Actions workflow
+  (`npm-publish.yml`) publishes on a released tag (and via manual dispatch),
+  with OIDC build provenance; it refuses to publish unless the tag matches
+  `package.json`.
+- **`cc-channel-octo` CLI bin** — the package now exposes a `bin`, so a global
+  install / `npx` starts the gateway directly (no clone + build needed).
+
+### Fixed
+
+- **Gateway auto-start when launched via the installed bin** — the main-module
+  guard compared `import.meta.url` against `process.argv[1]` verbatim, but the
+  installed bin is a symlink under `node_modules/.bin/`, so the paths never
+  matched and `main()` never fired (the command exited silently). The guard now
+  canonicalizes both sides with `realpath`.
+
+### Changed
+
+- **`package.json` packaging metadata** — scoped name, a `files` allowlist
+  (so the compiled `dist/` is actually shipped despite being `.gitignore`d),
+  `publishConfig` (public + provenance), `repository` / `bugs` / `homepage`,
+  and a `prepublishOnly` build hook.
+
 ## [1.0.0] - 2026-06-10
 
 First stable release. Consolidates everything merged since `0.2.0` — the v0.3
@@ -313,6 +343,7 @@ hardening across the SSRF, prompt-injection, and protocol-DoS surfaces.
 Initial tagged baseline: text messaging, streaming output, SQLite session
 persistence, rate limiting, and the core security model.
 
+[1.0.1]: https://github.com/Mininglamp-OSS/cc-channel-octo/compare/v1.0.0...v1.0.1
 [1.0.0]: https://github.com/Mininglamp-OSS/cc-channel-octo/compare/v0.2.0...v1.0.0
 [0.2.0]: https://github.com/Mininglamp-OSS/cc-channel-octo/compare/v0.1.0...v0.2.0
 [0.1.0]: https://github.com/Mininglamp-OSS/cc-channel-octo/releases/tag/v0.1.0
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mininglamp-oss/cc-channel-octo",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Bridge Claude Code (via Claude Agent SDK) to Octo IM — independent Node.js gateway",
   "license": "Apache-2.0",
   "type": "module",


lml2468:
## 🟡 Review — PR #138 (`16469571`): v1.0.1 — correct in what it does, but won't achieve its stated goal as-is

The version bump + `[1.0.1]` changelog are accurate, and putting #137's packaging metadata onto a tag is the right call. **But this "first npm-installable release" depends on the publish chain that three reviewers just confirmed is broken — so as written, it will not publish to npm.** Two should-fixes before tagging.

### 🔴 Should-fix #1 — the release won't actually publish (broken trigger, carried over)
The PR body says: *"publish the GitHub Release (release-publish), which **fires `npm-publish`** → publishes `@mininglamp-oss/cc-channel-octo@1.0.1` to npm."* That chain does not work — confirmed on `main` and re-confirmed here: `npm-publish.yml` triggers on `release: published`, but the release is published by `release-publish.yml` via the default `GITHUB_TOKEN`, and **"events triggered by the `GITHUB_TOKEN` will not create a new workflow run."** This is the exact 🔴 Jerry-Xin / yujiawei / I flagged on #137 (now merged unfixed).

**Consequence specific to this PR**: you can tag v1.0.1 and publish the Release, and npm-publish will **silently not run** — v1.0.1 never reaches npm, so the "first npm-installable release" isn't, until someone manually `workflow_dispatch`es npm-publish. **Recommend fixing the trigger (e.g. `on: push: tags: ['v*.*.*']`) in this PR** so v1.0.1 actually publishes — otherwise the release's whole purpose silently no-ops. If you'd rather fix the trigger separately, at least correct the PR body / release runbook to say npm-publish must be dispatched manually.

### 🟡 Should-fix #2 — `package-lock.json` still `0.2.0` (now two minors stale)
`package.json` is `1.0.1`; `package-lock.json` is still `"0.2.0"` in both root fields. This is the same lockfile-sync miss from #136, uncorrected — and a v1.0.1 whose explicit job is "make the published package correct" is exactly where it should be fixed. `npm ci` tolerates it (root version isn't hash-checked) so CI stays green, but the published tag's tree self-reports the wrong version. **Fix:** `npm version 1.0.1 --no-git-tag-version` (or hand-sync both fields) and include the lockfile here.

### ✅ Correct in this PR
- **Version bump** `1.0.0 → 1.0.1`. ✓
- **`[1.0.1]` changelog** accurately describes #137 (npm publish, `bin`, symlink-guard fix, packaging metadata) in clean Keep-a-Changelog order (Added → Fixed → Changed); compare link `v1.0.0...v1.0.1` added correctly. ✓
- `type-check` clean, `lint` clean. ✓

### 🟢 Pre-existing carryovers (not introduced here; fold in if you're touching docs anyway)
- README badges (`README.md:9,11`) still use the unscoped `cc-channel-octo` → will 404 once published under the scope. This release is literally about npm-installability, so fixing the npm badge here is apt.
- CHANGELOG SemVer header (`:7`) still says *"While the major version is `0`…"* — stale since 1.0.0.
- The 1.0.0 section still both "Adds" and "Removes" `sdk.persistentSession` (the contradiction from #136). Out of scope to rewrite released history, but noting it persists.

### Verdict
What the PR *does* is correct, but it **won't accomplish its stated goal** (publish to npm) because of the carried-over broken trigger, and it re-misses the lockfile sync. I'd fix both in this PR before tagging v1.0.1 — especially the trigger, since a silent non-publish is the worst failure mode for a release whose sole purpose is to publish. (`check-sprint` ❌ is the usual board gate.)

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#138 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo#138 — `chore(release): v1.0.1`
**Head SHA:** `16469571ad7ff25e766e44746b8f5f76de84b5a3`
**Diff scope:** 2 files, +32/−1 (packaging-only release)

---

## 1. Verification

| Item | Status | Evidence |
|------|--------|----------|
| Version bump `1.0.0` → `1.0.1` | ✅ | `package.json:3` |
| CHANGELOG `[1.0.1]` section added | ✅ | `CHANGELOG.md:11–40` |
| Compare link `v1.0.0...v1.0.1` | ✅ | `CHANGELOG.md:346` — correct format, placed above `[1.0.0]` link |
| CHANGELOG date matches release date | ✅ | `2026-06-10` |
| npm-publish workflow exists | ✅ | `.github/workflows/npm-publish.yml` present at head SHA |
| `bin` field in package.json | ✅ | `"cc-channel-octo": "dist/index.js"` |
| `files` allowlist includes `dist` | ✅ | `["dist", "README.md", "CHANGELOG.md", "LICENSE", ...]` |
| `publishConfig` (public + provenance) | ✅ | `{"access": "public", "provenance": true}` |
| `prepublishOnly` build hook | ✅ | `"prepublishOnly": "npm run build"` |
| `realpath` main-module guard (CHANGELOG claim) | ✅ | `src/index.ts` — `isMainModule()` uses `realpathSync` on both `argvPath` and `import.meta.url` |

## 2. Findings

**None.** This PR is a pure version bump and CHANGELOG entry. No runtime code is touched.

### Diff-scope check (R2)

- `package.json`: single-line version change. **New**, benign.
- `CHANGELOG.md`: additive only — new section + compare link. **New**, documentation.
- No pre-existing behavior is altered or amplified.

## 3. Data-flow trace

No data consumed by runtime code is changed in this diff. The only "consumed" values are:
- `package.json#version` — read by npm tooling at publish time. Source: this diff. ✅
- CHANGELOG compare link — consumed by humans/docs tooling. Format matches existing pattern. ✅

## 4. R5 Blind-spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric paths touched.
- **C2 (control-flow ordering / nesting):** N/A — no control flow changed.
- **C3 (authorization boundary):** N/A — no auth/capability surface touched.

## 5. Verdict

Clean release PR. Version bump is consistent with PR title and body. CHANGELOG accurately describes changes that landed in prior PRs (#137). No runtime code modified. No findings at any severity level.

**[Octo-Q] verdict: APPROVE** — packaging-only release, zero risk to runtime behavior.

lml2468:
## Status: #138 merged + v1.0.1 tagged — merged *before any review*, both should-fixes carried over

For the record, with timestamps:

- **#138 merged: 09:04:00**
- mochashanyao APPROVED: **09:06:00** — i.e. the approval landed **2 minutes after** the merge.
- Jerry-Xin's 🔴 (per Steve) and my review were also in-flight / post-merge.

So #138 was merged with **zero approvals at merge time** and an open 🔴 in progress. This is a step beyond #136/#137 (which at least had approvals, just ones that missed the blocker) — here the merge didn't wait for review at all. `v1.0.1` is now tagged.

### What shipped (both should-fixes I raised are now on `main` / the tag)
1. 🔴 **The release still won't auto-publish.** `npm-publish.yml` on `main` still triggers on `release: published`, published via `GITHUB_TOKEN` → won't cascade. So tagging v1.0.1 + publishing the Release will **silently not run npm-publish** — `@mininglamp-oss/cc-channel-octo@1.0.1` won't reach npm unless someone manually `workflow_dispatch`es it. The stated purpose of v1.0.1 ("first npm-installable release") is not met by the automation as merged.
2. 🔴 **`package-lock.json` on `main` is still `0.2.0`** — now two minors behind `package.json` (1.0.1), shipped on the v1.0.1 tag.

Neither is a runtime bug; the package contents (if published) are correct. The gap is the publish *automation* + lockfile metadata.

### Recommended follow-up (one small CI/doc PR)
1. `npm-publish` trigger → **`on: push: tags: ['v*.*.*']`** (or publish the Release with a PAT, or `workflow_run`-chain on release-publish). This is the blocker — without it no tagged release auto-publishes.
2. `npm version <next> --no-git-tag-version` to sync `package-lock.json`.
3. README badges (`:9,11`) → scoped name.
4. (Optional) SemVer header line; leave released CHANGELOG history as-is.

**To actually publish v1.0.1 right now**: manually run the `npm-publish` workflow via `workflow_dispatch` with tag `v1.0.1` (the publish job itself — tag guard, build, test, provenance — is verified correct; only its *trigger* is broken).

I can draft that follow-up PR (trigger + lockfile + badges) — CI/doc-only, no code risk. Say the word.

### Process (now 3 in a row)
#136, #137, #138 all merged before an open, independently-verified 🔴 was resolved — and #138 merged before *any* review landed. Strongly recommend a maintainer check whether the merge gate (a) waits for required reviews and (b) blocks on unresolved CHANGES_REQUESTED. Release/publish-infra PRs are exactly where a silent-no-op slips through an approval-count-only or no-wait gate.

yujiawei (approved):
# Code Review — PR #138 (cc-channel-octo)

**Verdict: APPROVED.** This is the `v1.0.1` release tag commit. Its own diff against `main` is minimal and correct (version bump `1.0.0 → 1.0.1` + a well-written CHANGELOG entry). The substantive packaging work (`bin`, `files` allowlist, `publishConfig`, the `npm-publish.yml` workflow, and the main-module realpath fix) landed in the earlier #137 commit, which is part of this release surface; I reviewed it as well. No P0 or functional/blocking P1 was found. Two hardening/hygiene items are worth addressing in a follow-up but do not block this release.

## 1. Verification

| Area | Result | Evidence |
|---|---|---|
| Version bump consistency | ✅ | `package.json` version `1.0.1`; CHANGELOG `[1.0.1]` section + compare link `v1.0.0...v1.0.1`; the publish workflow's "Verify tag matches package.json version" step guards `tag == v$PKG_VERSION`. |
| `files` allowlist ships `dist/` | ✅ | `dist/` is `.gitignore`d, but a `files` array makes npm ignore `.gitignore` for inclusion, so the compiled output is packed. `tsconfig` `outDir: dist` / `declaration: true` matches `main`/`types`. `LICENSE`, `config.example.json`, `config.bot.example.json` all present. |
| `bin` + main-module fix | ✅ | `bin: { cc-channel-octo: dist/index.js }` + shebang. `isMainModule()` canonicalizes both `process.argv[1]` and `import.meta.url` with `realpathSync` before comparing — correct fix for the symlinked `node_modules/.bin/` case where the verbatim comparison never matched and `main()` silently never fired. The `catch` falls back to direct href comparison, and `main().catch` now `process.exit(1)` on error (strict improvement for a CLI). |
| Engine floor vs `import.meta.main` | ✅ | The manual `realpath` helper is the *correct* portable choice: `import.meta.main` only exists in Node ≥ 22.18, but `engines` is `node>=22` (includes 22.0–22.17 where it is `undefined`). Do not "simplify" to it unless the floor is raised. |
| `npm ci` with current lockfile | ✅ (tested) | See "Non-blocking observations" #3 — empirically verified `npm ci` succeeds despite a stale lockfile name/version. |

## 2. Findings

No **P0**. No blocking **P1**.

### P1 (hardening — non-blocking for this release; inherited from the packaging commit)

**Independently-dispatchable publisher holds `NPM_TOKEN` with no human/consistency gate.**
`npm-publish.yml` is triggered by both `release: published` and `workflow_dispatch` (input `tag`), and `actions/checkout` uses `ref: <resolved tag>`. The only content gate is "Verify tag matches package.json version", which compares the dispatched tag against `package.json.version` **inside the checked-out commit** — both operands come from the same attacker-controllable tree, so the check is self-referential. An actor with `actions:write` (dispatch) or `contents:write` (push a tag at a malicious commit + publish a Release to fire `release: published`) can make the job publish an arbitrary tree to the public registry `@latest`. With `publishConfig.provenance: true` + `id-token: write`, Sigstore provenance would faithfully attest the malicious commit, so `npm audit signatures` passes — a *false* sense of security (provenance proves where/how it was built, not that the code is legitimate).

This requires a trusted write-access actor or a leaked write credential, so it is a privilege-separation / hardening gap, not an unauthenticated or fork-exploitable RCE — hence non-blocking for a v1.0.1 patch. But the blast radius (one leaked write credential → `@latest` poisoning masked by valid provenance) is exactly why a publish gate should exist. Any one of these closes most of the gap:
1. Bind `NPM_TOKEN` to a protected `environment:` with **required reviewers**, so a human approves before the token enters scope.
2. Before publishing, assert the tag corresponds to a real non-draft GitHub Release built from a default-branch commit (e.g. `gh release view`) rather than trusting a self-consistent `package.json`.
3. Migrate to **npm Trusted Publishing** (OIDC-based publish, GA since 2025-07-31) and drop the long-lived `NPM_TOKEN` entirely — the OIDC plumbing (`id-token: write`) is already present, and this removes the leaked-token branch above completely.

*Location:* `.github/workflows/npm-publish.yml` (triggers, "Resolve tag", checkout `ref`, "Verify tag matches package.json version", "Publish" step); `package.json` `publishConfig.provenance`.

## 3. Non-blocking observations (nits / follow-ups)

1. **`release: published` trigger likely won't auto-fire.** The org `release-publish.yml` → `reusable-release-publish.yml@v1` chain creates/publishes the GitHub Release with `GH_TOKEN: ${{ github.token }}` (the default `GITHUB_TOKEN`). Per GitHub's anti-recursion rule, events created by the default token do **not** trigger new workflow runs — so the publisher's `release: published` trigger silently won't fire, and publishing will rely on the `workflow_dispatch` path. Consider documenting that operators must dispatch manually, or have the upstream Release be created with a PAT/App token so the chain actually fires.

2. **CHANGELOG summary overstates "packaging only."** The `[1.0.1]` header says *"No runtime behavior change — packaging only,"* but the same section's **Fixed** bullet documents the main-module guard change, which IS a runtime behavior change (it changes whether the gateway starts). The Fixed bullet itself is accurate, so a careful reader is informed — consider softening the one-line summary to e.g. *"packaging + a startup bugfix; no change to message-handling behavior."* Also, the preamble still reads *"While the major version is `0`, minor releases may carry breaking changes"* — stale now that the project is at `1.x`.

3. **Stale `package-lock.json` (hygiene, not a build break).** At the head commit the lockfile root is still `name: cc-channel-octo`, `version: 0.2.0`, while `package.json` is `@mininglamp-oss/cc-channel-octo` / `1.0.1`. I **tested this directly** on the workflow's exact toolchain (Node 22.22 / npm 10.9.4): `npm ci` **succeeds** despite the root name/version mismatch, because `npm ci`'s sync check validates the *dependency tree* (which matches exactly here), not the root `name`/`version`. So this does **not** break the publish pipeline — but it's worth a one-line `npm install --package-lock-only` + commit to keep the lockfile honest.

4. **`better-sqlite3` is a native module.** Global install / `npx` works compiler-free on Node 22 LTS over glibc Linux / macOS / Windows on x64/arm64 (prebuilt binaries), but Alpine/musl or uncommon arches fall back to source compilation (needs `python3` + `make` + `g++`). Consider a short supported-platform / prereq note in the README so first-run install failures are diagnosable.

5. **`NPM_TOKEN` scope (unverifiable from the diff).** If a long-lived token is kept rather than moving to Trusted Publishing (see P1 fix #3), scope it as a granular token limited to this single package with the shortest viable expiry.

6. **Redundant build.** The workflow runs `npm run build` explicitly and `prepublishOnly` runs it again on `npm publish`. Harmless and defensive (guarantees `dist/` for a manual publish) — flagging only so it isn't mistaken for a bug.

## Summary

The release diff is correct and the packaging is sound; install/run paths (`npm i -g`, `npx`, `node dist/index.js`) work on the supported platform matrix, and the main-module fix is the right portable solution for the declared engine floor. Approving. The one P1 (publish-workflow hardening) and the `release: published` trigger note are real but inherited from the packaging commit and not blocking for a v1.0.1 patch — recommend a fast follow-up to add a protected environment / Trusted Publishing and to confirm the release trigger fires.

Jerry-Xin (approved):
Release bump is in scope for this repository, and the publish path/package contents look sane.

🔴 Blocking: None.

💬 Non-blocking:
- 🟡 Warning: `package-lock.json:2` and `package-lock.json:8` still identify the root package as unscoped `cc-channel-octo` at `0.2.0`, while `package.json:2` and `package.json:3` are `@mininglamp-oss/cc-channel-octo@1.0.1`. `npm ci` tolerates this and the lockfile is not packed, but the repository release metadata is inconsistent.
- 🔵 Suggestion: `README.md:9` and `README.md:11` still point npm badges at the old unscoped package name. The install instructions are correct, but the published README will show/package-link the wrong npm badge.

✅ Highlights:
- `package.json:3` correctly bumps the release version to `1.0.1`.
- `CHANGELOG.md:11` documents the release and `CHANGELOG.md:343` adds the expected compare link.
- Verified locally: `npm ci --ignore-scripts`, `npm run build`, and `npm pack --dry-run --json` all succeeded; the dry-run tarball includes `dist/` and reports `@mininglamp-oss/cc-channel-octo@1.0.1`.

PR #140, opened 2026-06-10
ci(release): tag-push npm publish, sync lockfile, fix badges

### LINKED ISSUE #139: Release plane follow-ups: npm-publish trigger, lockfile sync, badges, tag hardening
Follow-ups after the v1.0.0 / v1.0.1 release and the npm-publish wiring (#137, #138). None block usage today — `@mininglamp-oss/cc-channel-octo@1.0.1` is live on npm — but each is a rough edge worth fixing.

## 1. npm-publish trigger: switch `release: published` → tag push

**Problem.** `npm-publish.yml` currently triggers on `release: published`, but the `release-publish` workflow creates the GitHub Release with the built-in `GITHUB_TOKEN`. GitHub does **not** cascade workflow triggers from events produced by `GITHUB_TOKEN` (anti-recursion), so the publish never fired automatically — v1.0.1 had to be published via manual `workflow_dispatch`.

**Fix (pick one):**
- **(preferred)** Trigger npm-publish on `push: tags: ['v*.*.*']`. Tag pushes are not subject to the `GITHUB_TOKEN` recursion block, so `git push origin vX.Y.Z` would auto-publish. Keep `workflow_dispatch` as a manual fallback. Skip pre-release tags (`-rc`, etc.) if we don't want those on npm `latest`.
- **(alternative)** Fold the `npm publish` step directly into `release-publish.yml` (same job that creates the Release), eliminating the cross-workflow event entirely.

Acceptance: pushing a stable `vX.Y.Z` tag publishes to npm with provenance, with no manual dispatch.

## 2. Sync `package-lock.json`

**Problem.** `package-lock.json` still reads `cc-channel-octo@0.2.0` while `package.json` is `@mininglamp-oss/cc-channel-octo@1.0.1` (the name + version bumps never regenerated the lockfile). `npm ci` relies on the lock matching `package.json`; this drift can break CI installs and is just incorrect metadata.

**Fix.** Run `npm install` to regenerate `package-lock.json` (top-level `name`/`version` + `packages[""]`), commit it. Consider adding a CI check (or the org's existing lockfile-drift guard) so the two can't diverge again.

## 3. Badges: point at the scoped package

**Problem.** The README npm + Node badges still use the old unscoped name `cc-channel-octo`, which no longer exists on npm — they render "not found".

```
https://img.shields.io/npm/v/cc-channel-octo          → @mininglamp-oss/cc-channel-octo
https://img.shields.io/node/v/cc-channel-octo         → @mininglamp-oss/cc-channel-octo
https://www.npmjs.com/package/cc-channel-octo (link)  → .../@mininglamp-oss/cc-channel-octo
```

**Fix.** Update the three URLs to the scoped name (shields.io needs the scope URL-encoded, e.g. `npm/v/@mininglamp-oss/cc-channel-octo`). Optionally add an npm-publish workflow status badge.

## 4. (optional) Tag-reachability hardening

`release-publish.yml` validates the tag is strict semver and that a successful CI run exists on the tag's commit, but does not assert the tagged commit is an ancestor of (reachable from) the default branch. A tag pushed on an off-branch commit could be released. Low priority — pushing tags is already maintainer-gated — but a `git merge-base --is-ancestor <tag-sha> origin/main` guard would close it. Same guard could apply to the new tag-push npm-publish trigger so a stray tag can't publish off-branch code.

---

Context: release plane = `release-drafter.yml` + `release-publish.yml` (#130) + `npm-publish.yml` (#137). First npm publish was v1.0.1 (#138), done via manual dispatch because of issue 1 above.

lml2468:
Closes #139. Release-plane follow-ups after v1.0.0 / v1.0.1.

## 1. npm-publish trigger → tag push (+ pre-release skip)

`npm-publish.yml` triggered on `release: published`, but `release-publish` creates the Release with `GITHUB_TOKEN`, and GitHub doesn't cascade workflow triggers from `GITHUB_TOKEN`-produced events — so it never fired automatically (v1.0.1 was published by manual dispatch).

- Trigger is now `push: tags: ['v*.*.*']` (not subject to the recursion block). `workflow_dispatch` retained as manual fallback.
- Pre-release tags (`vX.Y.Z-rc.1`, …) are **skipped on automatic runs** so they never land on npm `latest`; a maintainer can still publish one deliberately via dispatch.

## 4. Tag-reachability hardening

Added a `git merge-base --is-ancestor <tag-sha> origin/main` guard (with `fetch-depth: 0`) — refuses to publish a tag whose commit isn't reachable from `main`, closing the off-branch-publish gap. (Item 4 from the issue, included here.)

## 2. Sync `package-lock.json`

Was stale at `cc-channel-octo@0.2.0` while `package.json` is `@mininglamp-oss/cc-channel-octo@1.0.1` — `npm ci` requires the lock to match. Regenerated (top-level + `packages[""]` name/version).

## 3. Badges → scoped name

README npm + node badges and the npm link used the old unscoped `cc-channel-octo`, which 404s on npm. Repointed to `@mininglamp-oss/cc-channel-octo`. Verified both shields.io endpoints resolve (`v1.0.1`, `>=22`).

## Verification

- `actionlint .github/workflows/npm-publish.yml` ok · no tabs
- `npm run type-check` ok · `npm run lint` (zero warnings) ok · `npm test` 975 passing
- `npm ci --dry-run` consistent with the regenerated lockfile
- shields.io badge JSON endpoints return correct values for the scoped name

Note: the next stable `vX.Y.Z` tag push will exercise the new auto-publish path end to end.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -2,21 +2,28 @@ name: npm Publish
 
 # Publishes the package to the public npm registry.
 #
-# Trigger: when a GitHub Release is published (the release-publish workflow flips
-# the draft to published, which fires `release: published`), or manually via
-# workflow_dispatch with an explicit tag.
+# Trigger: on a pushed stable version tag `vX.Y.Z` (NOT pre-releases like
+# `vX.Y.Z-rc.1`), or manually via workflow_dispatch with an explicit tag.
+#
+# Why tag-push instead of `release: published`: the release-publish workflow
+# creates the GitHub Release with the built-in GITHUB_TOKEN, and GitHub does not
+# cascade workflow triggers from GITHUB_TOKEN-produced events (anti-recursion),
+# so a `release: published` trigger never fires automatically. A tag push is not
+# subject to that block.
 #
 # Auth: uses the NPM_TOKEN repository secret (an npm automation/granular token
 # with publish rights on the @mininglamp-oss scope). Provenance is attached via
 # OIDC (`id-token: write` + publishConfig.provenance:true in package.json), so
 # the published version carries a verifiable build-origin attestation.
 #
-# Safety: the job refuses to publish unless the tag's `vX.Y.Z` matches the
-# version in package.json, so a mismatched tag can never push a wrong version.
+# Safety: the job refuses to publish unless (a) the tag is strict semver, (b) the
+# tagged commit is an ancestor of origin/main (no publishing off-branch code),
+# and (c) the tag's `vX.Y.Z` matches the version in package.json.
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - 'v*.*.*'
   workflow_dispatch:
     inputs:
       tag:
@@ -38,13 +45,13 @@ jobs:
       - name: Resolve tag
         id: tag
         env:
-          EVENT_TAG: ${{ github.event.release.tag_name }}
+          REF_TAG: ${{ github.ref_type == 'tag' && github.ref_name || '' }}
           INPUT_TAG: ${{ inputs.tag }}
         run: |
           set -euo pipefail
-          TAG="${EVENT_TAG:-$INPUT_TAG}"
+          TAG="${REF_TAG:-$INPUT_TAG}"
           if [ -z "$TAG" ]; then
-            echo "::error::No tag resolved from release event or dispatch input."
+            echo "::error::No tag resolved from tag push or dispatch input."
             exit 1
           fi
           # Strict semver tag (mirrors the org release-publish regex).
@@ -53,20 +60,46 @@ jobs:
             echo "::error::Tag '$TAG' is not strict semver (vMAJOR.MINOR.PATCH[-prerelease])"
             exit 1
           fi
+          # Skip pre-release tags on automatic (tag-push) runs so they never land
+          # on npm 'latest'. A maintainer can still force one via workflow_dispatch.
+          if printf '%s' "$TAG" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+-'; then
+            if [ "${{ github.event_name }}" = "push" ]; then
+              echo "::notice::Tag '$TAG' is a pre-release; skipping automatic npm publish. Use workflow_dispatch to publish a pre-release deliberately."
+              echo "skip=true" >> "$GITHUB_OUTPUT"
+            fi
+          fi
           echo "tag=$TAG" >> "$GITHUB_OUTPUT"
           echo "Resolved tag: $TAG"
 
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        if: steps.tag.outputs.skip != 'true'
         with:
           ref: ${{ steps.tag.outputs.tag }}
+          fetch-depth: 0 # need history for the ancestor check
+
+      - name: Verify tagged commit is on main
+        if: steps.tag.outputs.skip != 'true'
+        env:
+          TAG: ${{ steps.tag.outputs.tag }}
+        run: |
+          set -euo pipefail
+          git fetch --no-tags origin main
+          TAG_SHA=$(git rev-list -n 1 "$TAG")
+          if ! git merge-base --is-ancestor "$TAG_SHA" origin/main; then
+            echo "::error::Tag '$TAG' ($TAG_SHA) is not an ancestor of origin/main. Refusing to publish off-branch code."
+            exit 1
+          fi
+          echo "Tag commit $TAG_SHA is reachable from origin/main ✓"
 
       - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        if: steps.tag.outputs.skip != 'true'
         with:
           node-version: '22'
           cache: 'npm'
           registry-url: 'https://registry.npmjs.org'
 
       - name: Verify tag matches package.json version
+        if: steps.tag.outputs.skip != 'true'
         env:
           TAG: ${{ steps.tag.outputs.tag }}
         run: |
@@ -80,15 +113,19 @@ jobs:
           echo "Tag matches package.json: $EXPECTED"
 
       - name: Install dependencies
+        if: steps.tag.outputs.skip != 'true'
         run: npm ci
 
       - name: Build
+        if: steps.tag.outputs.skip != 'true'
         run: npm run build
 
       - name: Test
+        if: steps.tag.outputs.skip != 'true'
         run: npm test
 
       - name: Publish
+        if: steps.tag.outputs.skip != 'true'
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: npm publish
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@
 
 <p align="center">
   <a href="https://github.com/Mininglamp-OSS/cc-channel-octo/actions"><img src="https://github.com/Mininglamp-OSS/cc-channel-octo/actions/workflows/ci.yml/badge.svg" alt="CI"></a>
-  <a href="https://www.npmjs.com/package/cc-channel-octo"><img src="https://img.shields.io/npm/v/cc-channel-octo" alt="npm version"></a>
+  <a href="https://www.npmjs.com/package/@mininglamp-oss/cc-channel-octo"><img src="https://img.shields.io/npm/v/@mininglamp-oss/cc-channel-octo" alt="npm version"></a>
   <a href="./LICENSE"><img src="https://img.shields.io/badge/license-Apache--2.0-blue" alt="License"></a>
-  <img src="https://img.shields.io/node/v/cc-channel-octo" alt="Node.js version">
+  <img src="https://img.shields.io/node/v/@mininglamp-oss/cc-channel-octo" alt="Node.js version">
 </p>
 
 ---
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
-  "name": "cc-channel-octo",
-  "version": "0.2.0",
+  "name": "@mininglamp-oss/cc-channel-octo",
+  "version": "1.0.1",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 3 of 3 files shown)


lml2468:
## ✅ Review — PR #140 (`6eff2cb`): release-plane fixes — all three blockers resolved, verified

Cloned, built, ran the suite, and exercised the trigger logic + checked the live badge endpoints. **This correctly closes the carried-over release-plane issues from #136/#137/#138, plus two hardening extras. No blocking findings — good to merge.**

### 1. 🔴→✅ npm-publish trigger now actually fires
Switched `release: published` → `push: tags: ['v*.*.*']`. This is the right fix: the `GITHUB_TOKEN` anti-recursion limitation (which silently killed the old auto-publish — confirmed across #137/#138) does **not** apply to a human-pushed tag. `workflow_dispatch` retained as manual fallback. The comment now accurately explains *why* tag-push. ✓

I simulated the **pre-release skip** across all trigger/tag combos:
| Tag | Event | Result |
|---|---|---|
| `v1.0.1` | push | runs ✓ |
| `v1.0.1` | dispatch | runs ✓ |
| `v2.0.0-rc.1` | push | **skip** (won't pollute npm `latest`) ✓ |
| `v2.0.0-rc.1` | dispatch | runs (deliberate) ✓ |

All correct — auto-runs skip pre-releases; a maintainer can still publish one on purpose via dispatch.

### 4. ✅ Tag-reachability hardening (was yujiawei's #138 suggestion)
`git merge-base --is-ancestor <tag-sha> origin/main` with `fetch-depth: 0` — refuses to publish a tag whose commit isn't on `main`. I traced the ref resolution: checkout detaches at the tag, the step explicitly `git fetch --no-tags origin main` (populating `refs/remotes/origin/main`), and `rev-list -n 1 <tag>` dereferences both annotated and lightweight tags — so the ancestor check is sound, and an off-branch tag correctly **fails closed** (refuses to publish). ✓

### 2. 🔴→✅ Lockfile synced
`package-lock.json` regenerated to `@mininglamp-oss/cc-channel-octo@1.0.1` in all three places (top-level `name`/`version` + `packages[""]`), and the `bin` entry is now reflected. `npm ci` **passes** against it (it was the carried-over stale-`0.2.0` blocker). ✓

### 3. 🟡→✅ Badges fixed
README npm + node badges and the npm link repointed to `@mininglamp-oss/cc-channel-octo`. I curled both shields endpoints — **HTTP 200** (shields accepts the literal `@scope/pkg`, no encoding needed). ✓

### Verified dynamically
`type-check` clean · `lint` clean · `npm test` **975 passed** · `npm ci` consistent with the regenerated lock · both badge endpoints resolve.

### Non-blocking notes
- **Scope is correctly limited** to trigger/lock/badges. The other v1.0.0 doc carryovers (SemVer header line still says "major version is 0"; the 1.0.0 changelog's persistentSession Added/Removed contradiction) are **not** in this PR — fine, they're cosmetic released-history items and out of scope for a release-plane CI fix. Worth a tiny separate doc PR eventually, not here.
- **End-to-end caveat (as the PR notes)**: the new auto-publish path only gets truly exercised on the next stable `vX.Y.Z` tag push. The logic is verified by inspection/simulation; first real run is the final proof. Also still requires the `NPM_TOKEN` secret to be set (repo setup, out of PR).

### Verdict
Minimal, correct, and well-hardened. Closes the loop on the release automation that three of us flagged. LGTM. (`check-sprint` ❌ is the usual board gate, not code.)

Jerry-Xin (approved):
Release automation changes are relevant to this repository and are ready to merge; I found no blocking issues.

💬 Non-blocking

🟡 Warning: Manual prerelease dispatches still run plain `npm publish`, which defaults to the `latest` dist-tag. That appears intentional from the “deliberately via dispatch” wording, but if maintainers expect prereleases to publish under `next` or `rc`, add conditional `npm publish --tag ...` for tags containing `-`. Reference: `.github/workflows/npm-publish.yml:63` and `.github/workflows/npm-publish.yml:127`.

✅ Highlights

- The PR is in scope: it updates release publishing, package lock metadata, and README package badges for `@mininglamp-oss/cc-channel-octo`.
- The tag resolution, strict semver validation, prerelease auto-skip, main-reachability guard, and package version match checks are appropriate safeguards. Reference: `.github/workflows/npm-publish.yml:45`.
- `package-lock.json` now matches `package.json` name/version and includes the existing package bin metadata. Reference: `package-lock.json:2`.
- README npm badge and package link now point at the scoped package. Reference: `README.md:9`.
- I verified `git diff --check` and `npm ci --dry-run`; both passed.

lml2468:
## Confirming Jerry-Xin's 🟡 — it's a real (scoped) npm dist-tag gap, with a clean fix

I verified this and it's correct — worth fixing, though it's narrow.

**The gap**: `npm publish` with no `--tag` **always** publishes to the `latest` dist-tag — even for a prerelease version like `1.0.0-rc.1`. (Unlike `npm version`, `npm publish` does *not* auto-detect prereleases.) The workflow's pre-release **skip only fires on `push`** events; a maintainer who **`workflow_dispatch`es a `-rc` tag deliberately** sails past the skip → reaches the `Publish` step → `npm publish` (no `--tag`) → the RC becomes `latest`, so `npm install @mininglamp-oss/cc-channel-octo` would hand stable users a release candidate.

**Scope/severity**: only the *deliberate dispatch-a-prerelease* path. The automatic tag-push path already skips prereleases entirely, and stable releases are unaffected (they belong on `latest`). So it's not a blocker for the normal flow — but it quietly defeats the very reason dispatch-publishing a prerelease exists (you'd want it on a `next`/`rc` tag, not `latest`).

**Suggested fix** — compute the dist-tag from the version and pass it explicitly:
```yaml
- name: Publish
  if: steps.tag.outputs.skip != 'true'
  env:
    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
    TAG: ${{ steps.tag.outputs.tag }}
  run: |
    set -euo pipefail
    if printf '%s' "$TAG" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+-'; then
      npm publish --tag next      # prerelease → 'next' (or 'rc'), never 'latest'
    else
      npm publish                 # stable → 'latest'
    fi
```
This keeps stable on `latest` and routes any deliberately-dispatched prerelease to `next`, closing the gap.

Everything else in the PR is verified and correct (trigger fix, reachability guard, lockfile sync, badges — all confirmed in my review above; 975 tests green). This is the one refinement I'd fold in before merge if you want the prerelease-dispatch path to be safe; if prereleases are never dispatch-published in practice, it's a low-risk follow-up. Either way, good catch by Jerry-Xin.

yujiawei (approved):
# Code Review — PR #140 (cc-channel-octo)

**Scope:** CI/release plumbing — switch npm publish from `release: published` to tag-push, harden the publish guards, sync `package-lock.json` to the scoped `1.0.1` package, and fix the README badges to the scoped name. 3 files, +56/-16. No runtime/source behavior change.

## Verdict

**APPROVED.** The change is correct, well-reasoned, and CI is green (TypeScript type-check, ESLint zero-warnings, Vitest 460+ tests, secret-scan, dependency-review, history common-ancestor check all pass). No correctness, security, or build-breaking issues found. The notes below are minor/non-blocking.

## What's verified ✅

- **Trigger swap is sound.** `release: published` never fires when the Release is created with the built-in `GITHUB_TOKEN` (GitHub's anti-recursion rule). Switching to `on: push: tags: 'v*.*.*'` is the correct, standard fix. Rationale is documented inline (`.github/workflows/npm-publish.yml:8-12`).
- **Tag resolution handles both paths.** `REF_TAG: ${{ github.ref_type == 'tag' && github.ref_name || '' }}` resolves to the tag on a tag-push and to `''` on `workflow_dispatch` (where `ref_type` is `branch`), correctly falling back to `INPUT_TAG` (`npm-publish.yml:48-52`).
- **Pre-release safety.** The `v*.*.*` glob does match pre-release tags (e.g. `v1.0.0-rc.1`), but the in-job skip step gates them out on automatic `push` runs only, leaving `workflow_dispatch` free to publish one deliberately (`npm-publish.yml:63-70`). This keeps pre-releases off the npm `latest` tag. Good.
- **New off-branch guard is a genuine improvement.** `git merge-base --is-ancestor "$TAG_SHA" origin/main` refuses to publish a tag whose commit isn't reachable from `main`, closing a real hole (publishing code that was tagged on a stray branch). `fetch-depth: 0` is correctly added so the history is available (`npm-publish.yml:78, 80-92`).
- **Every publish step is gated** behind `if: steps.tag.outputs.skip != 'true'`, so a skipped pre-release run does nothing and exits green rather than half-running.
- **Least-privilege permissions preserved:** top-level `permissions: {}` with job-level `contents: read` + `id-token: write` (only what OIDC provenance needs). No `contents: write` since this job no longer creates Releases.
- **Lockfile sync is necessary, not cosmetic.** `npm ci` (used in the publish job) hard-fails when `package-lock.json` is out of sync with `package.json`. The root `name`/`version`/`bin` are now aligned to `@mininglamp-oss/cc-channel-octo@1.0.1`, and the new `bin` entry matches `package.json` exactly. The `bin` target `dist/index.js` is backed by a real `#!/usr/bin/env node` shebang in `src/index.ts`, which `tsc` (outDir `dist`) preserves — so the published CLI is executable.
- **Badges** correctly point to the scoped package; shields.io supports the `@scope/name` form for both the npm-version and node-version badges.

## Minor notes (non-blocking)

1. **Single-branch assumption in the ancestor check.** The guard hard-codes `origin/main`. If a future maintenance/backport release is ever cut from a `release/x.y` branch, this step will reject a legitimate tag. Fine for the current single-`main` flow; worth a one-line comment so the assumption is explicit for future maintainers.
2. **Header comment vs. trigger glob.** The top comment (`npm-publish.yml:5-6`) reads as if the trigger itself excludes pre-releases, but the trigger fires for them and the exclusion happens in the skip step. The net behavior (auto-publish stable only) is accurate; a half-sentence clarifying *where* the filtering happens would avoid confusing a future reader debugging a "why did my rc tag start a run" question.
3. **`npm test` inside the publish pipeline** re-runs the full 460+ suite on every release. Correct as a last-line safety net; just noting it adds wall-clock to each publish. No change requested.

None of these block merge.

PR #142, opened 2026-06-10
fix(router): per-message dispatch timeout to unblock hung sessions

### LINKED ISSUE #141: P0: per-message dispatch timeout — a hung agent turn permanently blocks the session queue
## Problem

`SessionRouter.withSessionLock` (`src/session-router.ts:89`) serializes message handling per `sessionKey` via a Promise chain: each message awaits the previous one's `gate` before running. The gate only resolves in the `finally` of `withSessionLock`, i.e. after `handler(result)` returns.

`handler` runs the full pipeline including `queryAgent()` + `streamRelay.deliver()` (`src/index.ts:773,785`). **If an agent turn hangs** (a stuck SDK `query()`, a wedged tool subprocess, a stalled stream), `handler` never returns, the gate never resolves, and **every subsequent message for that session is blocked forever** — silent permanent stuck. There is currently no per-message timeout anywhere in the cc pipeline (`AbortSignal.timeout` is used for individual HTTP calls in `octo/api.ts`, but nothing bounds the whole turn).

This is the one stability gap identified in the GAP analysis vs `openclaw-channel-octo`, which already guards against it (issue #75 there).

## Reference implementation (openclaw)

`openclaw-channel-octo/src/inbound.ts:2481-2630` wraps the dispatch in `Promise.race([dispatch, dispatchTimeoutPromise])`:
- `DISPATCH_TIMEOUT_DEFAULT_MS = 5 * 60 * 1000` (generous — longer than any normal LLM round-trip)
- A **per-invocation `Error` object** identifies "this is OUR timeout" by reference equality, never string comparison (avoids misclassifying a same-text upstream error).
- On timeout: warn-log, send a **bounded** apology (`AbortSignal.timeout(DISPATCH_TIMEOUT_APOLOGY_MS=10_000)` — so a sick Octo API can't hang the apology too), then let the queue advance.
- Scope note: it does NOT try to cancel the in-flight dispatch or gate "ghost" late callbacks — worst case is a delayed real reply arriving after the apology. Out of scope.

## Proposed fix

Wrap the `handler(result)` call inside `SessionRouter.routeAndHandle` (`src/session-router.ts:123`) in a timeout race so the gate always resolves and the session unblocks:
- Add config `dispatchTimeoutMs` (default `300_000`) so it's tunable per-bot; tests can set it small.
- On timeout: log a warning attributed to the `sessionKey`, send a bounded `⚠️ 处理超时，请稍后重试。` apology via the existing `replySafe` (itself bounded), and let the lock release.
- Use a per-invocation Error + reference equality, mirroring openclaw.
- Keep scope tight: no cancellation of the in-flight turn (the SDK `query()` keeps running to completion in the background; only the queue is unblocked). Document this.

## Acceptance

- A handler that never resolves no longer blocks subsequent messages on the same session (covered by a unit test with a small timeout).
- The user gets a single bounded apology on timeout.
- Normal (fast) turns are unaffected.
- `npm test` green.

lml2468:
Closes #141.

## Problem

`SessionRouter.withSessionLock` serializes per-session handling via a Promise chain — each message awaits the previous one's `gate`, which only resolves after `handler()` returns. `handler` runs the full pipeline including `queryAgent()` + `streamRelay.deliver()`. **If an agent turn hangs** (stuck SDK `query()`, wedged tool subprocess, stalled stream), `handler` never returns, the gate never resolves, and **every subsequent message on that session is blocked forever** — silent permanent stuck. cc had no per-turn timeout (only individual HTTP calls are bounded).

This is the P0 stability gap from the GAP analysis vs `openclaw-channel-octo` (which guards against it in its issue #75).

## Fix

Race `handler(result)` against a configurable timeout in `routeAndHandle` so the session lock always releases:

- **`config.dispatchTimeoutMs`** (default `300_000` = 5 min), tunable per-bot; `0`/negative disables the guard.
- On timeout: warn-log attributed to the `sessionKey`, send a **bounded** one-shot apology (`⚠️ 处理超时，请稍后重试。`) via the existing `replySafe` (itself error-swallowing + the underlying `sendMessage` is time-bounded), then let the lock release so the queue advances.
- **Per-invocation `Error`** identifies our timeout by **reference equality**, never string comparison (a same-text upstream error can't be misclassified).
- **Scope (mirrors openclaw #75):** we do NOT cancel the in-flight turn — the SDK query keeps running to completion in the background; we only unblock the queue. Worst case is a delayed real reply arriving after the apology.
- The orphaned `handler` promise gets a no-op `.catch` so a late rejection (from a future caller whose handler isn't self-contained) can't surface as an `unhandledRejection`. Today's caller (`index.ts`) is fully try/caught — defense-in-depth.

## Tests (5 new, 980 total)

- A hung handler no longer blocks the next message on the same session.
- Timeout sends exactly one apology containing `处理超时`.
- A normal fast handler doesn't trigger the timeout.
- `dispatchTimeoutMs: 0` runs the handler unguarded.
- A *rejecting* (non-timeout) handler doesn't wedge the session and does NOT send the timeout apology.

## Verification

- `npm run type-check` ok · `npm run lint` (zero warnings) ok · `npm test` — 980 passing
- `code-review` skill (high effort) run; the one PLAUSIBLE finding (orphaned-promise unhandled rejection) was addressed with the defensive `.catch`.

### DIFF
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -469,6 +469,112 @@ describe('routeAndHandle concurrency', () => {
   });
 });
 
+// ─── #141: Dispatch timeout ────────────────────────────────────────────────
+
+describe('dispatch timeout (#141)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('a hung handler does not block the next message on the same session', async () => {
+    const router = new SessionRouter(
+      makeConfig({ rateLimit: { maxPerMinute: 100 }, dispatchTimeoutMs: 50 }),
+      ROBOT_ID,
+    );
+    const completed: number[] = [];
+
+    // Message 1: handler never resolves (simulates a hung agent turn).
+    const p1 = router.routeAndHandle(
+      makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'same-user' }),
+      () => new Promise<void>(() => { /* never resolves */ }),
+    );
+
+    // Message 2: a normal fast handler on the SAME session. Without the
+    // timeout it would be blocked forever behind message 1's session lock.
+    const p2 = router.routeAndHandle(
+      makeMsg({ message_id: '2', channel_type: ChannelType.DM, from_uid: 'same-user' }),
+      async () => { completed.push(2); },
+    );
+
+    await Promise.all([p1, p2]);
+
+    // Message 2 ran — the lock was released after message 1 timed out.
+    expect(completed).toEqual([2]);
+  });
+
+  it('sends a single bounded apology on timeout', async () => {
+    const router = new SessionRouter(
+      makeConfig({ rateLimit: { maxPerMinute: 100 }, dispatchTimeoutMs: 30 }),
+      ROBOT_ID,
+    );
+
+    await router.routeAndHandle(
+      makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'u1' }),
+      () => new Promise<void>(() => { /* never resolves */ }),
+    );
+
+    expect(sendMessage).toHaveBeenCalledTimes(1);
+    expect(vi.mocked(sendMessage).mock.calls[0][0]).toMatchObject({
+      content: expect.stringContaining('处理超时'),
+    });
+  });
+
+  it('does not fire for a normal fast handler', async () => {
+    const router = new SessionRouter(
+      makeConfig({ rateLimit: { maxPerMinute: 100 }, dispatchTimeoutMs: 1000 }),
+      ROBOT_ID,
+    );
+    let ran = false;
+
+    await router.routeAndHandle(
+      makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'u1' }),
+      async () => { ran = true; },
+    );
+
+    expect(ran).toBe(true);
+    expect(sendMessage).not.toHaveBeenCalled();
+  });
+
+  it('disabled (0) runs the handler unguarded', async () => {
+    const router = new SessionRouter(
+      makeConfig({ rateLimit: { maxPerMinute: 100 }, dispatchTimeoutMs: 0 }),
+      ROBOT_ID,
+    );
+    let ran = false;
+
+    await router.routeAndHandle(
+      makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'u1' }),
+      async () => { ran = true; },
+    );
+
+    expect(ran).toBe(true);
+    expect(sendMessage).not.toHaveBeenCalled();
+  });
+
+  it('a rejecting handler does not wedge the session (next message still runs)', async () => {
+    const router = new SessionRouter(
+      makeConfig({ rateLimit: { maxPerMinute: 100 }, dispatchTimeoutMs: 1000 }),
+      ROBOT_ID,
+    );
+    const completed: number[] = [];
+
+    // Message 1: handler rejects fast (a real handler error, not a timeout).
+    await router.routeAndHandle(
+      makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'same-user' }),
+      async () => { throw new Error('boom'); },
+    );
+    // Message 2: same session — must still run (lock released, no timeout apology).
+    await router.routeAndHandle(
+      makeMsg({ message_id: '2', channel_type: ChannelType.DM, from_uid: 'same-user' }),
+      async () => { completed.push(2); },
+    );
+
+    expect(completed).toEqual([2]);
+    // A non-timeout error must NOT trigger the timeout apology.
+    expect(sendMessage).not.toHaveBeenCalled();
+  });
+});
+
 // ─── Q10: Message length limit ─────────────────────────────────────────────
 
 describe('Message length limit (Q10)', () => {
--- a/src/config.ts
+++ b/src/config.ts
@@ -174,6 +174,15 @@ export interface Config {
   };
   /** Maximum response length in chars before truncation (Q32). */
   maxResponseChars: number;
+  /**
+   * Per-message dispatch timeout in ms (#141). Bounds the full handler
+   * pipeline (agent query + stream) under the per-session lock. If a turn
+   * hangs past this, the session lock is released (a hung turn would otherwise
+   * block every subsequent message on that session forever) and the user gets
+   * a one-shot apology. Does NOT cancel the in-flight turn — only unblocks the
+   * queue. Default 5 minutes.
+   */
+  dispatchTimeoutMs: number;
   botBlocklist?: string[];
   /**
    * G14: Bots in this list are allowed to DM the bot even if their uid matches
@@ -247,6 +256,7 @@ type PartialConfig = {
   rateLimit?: Partial<Config['rateLimit']>;
   context?: Partial<Config['context']>;
   maxResponseChars?: number;
+  dispatchTimeoutMs?: number;
   botBlocklist?: string[];
   allowedBotUids?: string[];
   mentionFreeGroups?: string[];
@@ -282,6 +292,7 @@ function defaults(): Config {
       historyLimit: 40,
     },
     maxResponseChars: 524_288, // 512 KB (Q32)
+    dispatchTimeoutMs: 300_000, // 5 min (#141)
   };
 }
 
@@ -345,6 +356,7 @@ function mergeConfig(base: Config, override: PartialConfig): Config {
       ...(override.context ?? {}),
     },
     maxResponseChars: override.maxResponseChars ?? base.maxResponseChars,
+    dispatchTimeoutMs: override.dispatchTimeoutMs ?? base.dispatchTimeoutMs,
     botBlocklist: override.botBlocklist ?? base.botBlocklist,
     allowedBotUids: override.allowedBotUids ?? base.allowedBotUids,
     mentionFreeGroups: override.mentionFreeGroups ?? base.mentionFreeGroups,
--- a/src/session-router.ts
+++ b/src/session-router.ts
@@ -128,10 +128,76 @@ export class SessionRouter {
     return this.withSessionLock(key, async () => {
       const result = await this.processMessage(msg, key);
       if (result && result.shouldProcess) {
-        await handler(result);
+        await this.runHandlerWithTimeout(result, handler);
       }
       return result;
     });
+  }
+
+  /**
+   * #141: Run the handler under a dispatch timeout so a hung turn (a stuck SDK
+   * query, a wedged tool subprocess, a stalled stream) cannot block the session
+   * forever. The handler runs inside withSessionLock — if it never returns, the
+   * lock's gate never resolves and EVERY subsequent message for this session is
+   * stuck permanently (silent). Racing the handler against a timeout guarantees
+   * the lock releases.
+   *
+   * Scope (mirrors openclaw #75): we do NOT cancel the in-flight turn — the SDK
+   * query keeps running to completion in the background; we only unblock the
+   * queue. Worst case is a delayed real reply arriving after the apology.
+   *
+   * `timeoutError` is a per-invocation Error so the catch identifies OUR timeout
+   * by reference equality, never by string comparison (a same-text upstream
+   * error must not be misclassified).
+   */
+  private async runHandlerWithTimeout(
+    result: RouteResult,
+    handler: (result: RouteResult) => Promise<void>,
+  ): Promise<void> {
+    const timeoutMs = this.config.dispatchTimeoutMs;
+    if (!timeoutMs || timeoutMs <= 0) {
+      // Timeout disabled — run unguarded.
+      await handler(result);
+      return;
+    }
+
+    let timeoutHandle: ReturnType<typeof setTimeout> | undefined;
+    const timeoutError = new Error(`dispatch timed out after ${timeoutMs}ms`);
+    const timeoutPromise = new Promise<never>((_, reject) => {
+      timeoutHandle = setTimeout(() => reject(timeoutError), timeoutMs);
+    });
+
+    // The handler keeps running after a timeout (we don't cancel the in-flight
+    // turn — see scope note above). Once the race settles on timeoutError, that
+    // orphaned promise is no longer awaited; attach a no-op catch so a late
+    // rejection from a handler that doesn't self-contain its errors can't surface
+    // as an unhandledRejection. Today's caller (index.ts) is fully try/caught, so
+    // this is defense-in-depth for future callers.
+    const handlerPromise = handler(result);
+    handlerPromise.catch(() => { /* swallow late rejection after timeout */ });
+
+    try {
+      await Promise.race([handlerPromise, timeoutPromise]);
+    } catch (err) {
+      if (err === timeoutError) {
+        console.warn(
+          `session-router: dispatch hung past ${timeoutMs}ms, releasing session lock (session=${result.sessionKey})`,
+        );
+        // Bounded apology — replySafe swallows its own errors, and the
+        // underlying sendMessage in octo/api.ts is itself time-bounded, so a
+        // sick Octo API can't re-hang us here.
+        await this.replySafe(result.message, '⚠️ 处理超时，请稍后重试。');
+        return; // swallow: the lock releases, the queue advances
+      }
+      // A real handler error — index.ts's handler already catches and replies
+      // internally, so reaching here is unexpected. Swallow to keep the lock
+      // release path identical (never let an error wedge the queue).
+      console.error(
+        `session-router: handler error (session=${result.sessionKey}): ${String(err)}`,
+      );
+    } finally {
+      if (timeoutHandle) clearTimeout(timeoutHandle);
+    }
   }
 
   async route(msg: BotMessage): Promise<RouteResult | null> {


lml2468:
## ✅ Review — PR #142 (`c65a067`): dispatch timeout — verified, good to merge

Cloned, ran the suite, and adversarially tested the two trickiest concurrency paths (late resolve / late reject after timeout). **This is a solid P0 stability fix. No blocking findings.**

### The fix is correct
`runHandlerWithTimeout` races the handler against a `setTimeout` so the per-session lock in `withSessionLock` always releases — closing the "one hung turn blocks the session forever" gap. The design choices are all sound:
- **Reference-equality timeout identification** (`err === timeoutError`, per-invocation Error) — correctly avoids misclassifying a same-text upstream error as a timeout. ✓
- **Doesn't cancel the in-flight turn** (matches openclaw #75 scope) — pragmatic; worst case is a late real reply after the apology, documented. ✓
- **`finally { clearTimeout }`** — no timer leak on the fast path. ✓
- **Bounded apology via `replySafe`** (error-swallowing, and the underlying `sendMessage` is time-bounded) — a sick Octo API can't re-hang the release path. ✓

### Config plumbing complete
`dispatchTimeoutMs` added to `Config` interface, `PartialConfig`, `defaults()` (300_000), and `mergeConfig` (`override ?? base`). I verified the `??` is deliberate and correct: `0 ?? 300000 === 0`, so `dispatchTimeoutMs: 0` (disable) survives the merge — `||` would have wrongly reset it to the default. ✓

### Verified dynamically (ran it, didn't trust the diff)
- `type-check` clean · `lint` clean · `npm test` **980 passed** (5 new).
- The 5 new tests assert real invariants (not tautological): hung handler doesn't block the next message, single apology, fast handler doesn't fire, `0` disables, rejecting handler doesn't wedge + sends no apology.
- **My own adversarial tests on the two paths the suite doesn't cover**, both pass:
  - Handler that **resolves after** the timeout → exactly **one** apology, no second `sendMessage` from the late completion. ✓
  - Handler that **rejects after** the timeout → **no `unhandledRejection`** (the defensive no-op `.catch` on the orphaned promise works as intended — this validates the exact finding from the author's code-review pass). ✓

### Non-blocking notes
- **Timeout doesn't cancel the turn** (by design). Over time a pathological hang could accumulate background SDK queries that never finish — not a concern at the 5-min default + normal traffic, but if hangs become common, a future iteration might track/cap concurrent orphaned turns. Out of scope here; the doc is honest about it.
- **No config validation** on `dispatchTimeoutMs` (e.g. a tiny positive value like `1` would time out almost everything). Defaults are safe and `0` is handled; a clamp/min-warn could be a nicety but isn't needed.

### Verdict
Correct, well-tested, honestly-scoped fix for a real P0 (silent permanent session wedge). The concurrency edges — late resolve, late reject, disable path — all hold up under direct testing. LGTM. (`check-sprint` ❌ is the usual board gate.)

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#142 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo#142
**Head SHA:** `c65a067f502fe6f03c9327ef473637e66e1092d4`
**Title:** fix(router): per-message dispatch timeout to unblock hung sessions
**Files:** `src/session-router.ts` (+67/-1), `src/config.ts` (+12/-0), `src/__tests__/session-router.test.ts` (+106/-0)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Problem is real | ✅ | `withSessionLock` (session-router.ts:89-104) chains promises via gate; if `fn()` never resolves, gate never resolves → permanent session lock |
| Timeout race correct | ✅ | `runHandlerWithTimeout` (session-router.ts:148-196): `Promise.race([handlerPromise, timeoutPromise])` with per-invocation `timeoutError` for reference equality |
| Timer cleanup | ✅ | `clearTimeout(timeoutHandle)` in `finally` block (line 195) |
| Orphaned promise safety | ✅ | `handlerPromise.catch(() => {})` (line 175) prevents `unhandledRejection` after timeout |
| Apology is bounded | ✅ | `replySafe` (session-router.ts:380-393) swallows errors; underlying `sendMessage` uses `AbortSignal.timeout(30_000)` (api.ts:14,68) |
| Config merge preserves `0` | ✅ | `override.dispatchTimeoutMs ?? base.dispatchTimeoutMs` (config.ts:359) — `??` preserves `0` (not nullish) |
| Reference equality | ✅ | `err === timeoutError` (line 180) — per-invocation `new Error()`, no string matching |
| Tests cover key paths | ✅ | 5 tests: hung handler unblocks, apology sent, fast handler ok, disabled (0), rejecting handler |

## 2. Findings

### F1 — P2: Non-timeout handler errors silently swallowed (behavioral change)

**Diff-scope:** NEW — introduced by this PR.

**Pre-PR:** `await handler(result)` at session-router.ts:131 (old). If handler threw, the error propagated through `withSessionLock` → `routeAndHandle` → `handleMessage` → `onInbound`'s `.catch()` (index.ts:247), logged as "Unhandled message handler error."

**Post-PR:** `runHandlerWithTimeout` catches non-timeout errors at line 189-193, logs via `console.error`, and returns normally. The error never reaches the caller.

**Impact:** The PR justifies this ("index.ts's handler already catches and replies internally, so reaching here is unexpected"). In practice the handler callback in index.ts:321 wraps everything in try/catch, so this path is effectively dead. However, the error propagation contract of `routeAndHandle` has changed — callers can no longer observe handler failures. If a future handler callback doesn't self-contain errors (contrary to the current assumption), failures would be silently lost.

**Recommendation:** Consider re-throwing non-timeout errors after logging, so the caller's error handling chain remains intact. Alternatively, document the swallow as part of the method's contract.

### F2 — P2: Test `makeConfig` missing `dispatchTimeoutMs`

**Diff-scope:** NEW — the field was added to `Config` but the test helper wasn't updated.

The test helper `makeConfig` (session-router.test.ts:17-28) doesn't include `dispatchTimeoutMs` in its base object, yet returns `Config` (which now requires the field). At runtime, `this.config.dispatchTimeoutMs` is `undefined` for tests not passing it explicitly, and the guard `if (!timeoutMs || timeoutMs <= 0)` treats `undefined` as disabled — functionally correct for existing tests, but a type-safety gap.

**Recommendation:** Add `dispatchTimeoutMs: 300_000` (or `0`) to `makeConfig`'s base object.

### F3 — Nit: Hardcoded Chinese apology string

`'⚠️ 处理超时，请稍后重试。'` (line 186) is hardcoded. This matches the existing pattern (`'请稍后再试'` for rate limiting), so it's consistent. Noting for completeness only.

## 3. Suggestions

1. **F1 fix:** Re-throw non-timeout errors after logging, or add a JSDoc `@remarks` noting the intentional swallow.
2. **F2 fix:** Add `dispatchTimeoutMs: 0` to `makeConfig` base (disabled by default in tests, matching the pre-PR behavior).

## 4. Additional Observations

- **No in-flight cancellation:** The PR explicitly does NOT cancel the hung handler — the SDK query keeps running. This means a truly stuck turn continues consuming resources (SDK subprocess, API connections) indefinitely. This matches openclaw #75's scope and is an acceptable first step, but a follow-up adding `AbortController`-based cancellation would be valuable.
- **Single apology:** Only one apology per timeout. If the user sends multiple messages while a turn is hung, each gets its own timeout + apology (correct — each is a separate `routeAndHandle` invocation).

## 5. Data Flow Tracing

| Consumed Data | Upstream Source | Reaches Consumer? |
|---|---|---|
| `result.sessionKey` in warn/error logs | `processMessage` → `sessionKey(msg)` → `msg.from_uid`/`msg.channel_id` | ✅ validated by throws on missing fields |
| `result.message` in `replySafe` | Original `msg` parameter passed through `processMessage` | ✅ `replySafe` guards on `channel_id`/`channel_type` |
| `this.config.dispatchTimeoutMs` | `defaults()` → `300_000`; `mergeConfig` with `??` | ✅ `0` preserved correctly |
| `timeoutError` in `===` check | Per-invocation `new Error()` at line 168 | ✅ reference equality guaranteed |
| `handlerPromise` resolution | Caller-provided `handler(result)` callback | ✅ race settles correctly; orphan catch prevents unhandledRejection |

## 6. Three Blindspot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No symmetric paths touched. The timeout applies only to the handler execution path; `route()` (no handler) is unaffected by design.
- **C2 — Control-flow ordering / nesting reuse:** N/A. `runHandlerWithTimeout` is called from exactly one site. `replySafe` is reused but its behavior is unchanged.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This is a reliability mechanism, not a security change.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The implementation correctly addresses a real P0 stability gap (permanent session lock on hung handler). The two P2 findings (error propagation change, test helper type gap) are non-blocking — the error propagation change is justified by the current caller's internal try/catch, and the test helper gap is functionally benign. Recommended improvements noted but not required for landing.

Jerry-Xin (changes requested):
Summary: The PR is in scope and the timeout mechanism is mostly well placed, but the new config option is not actually per-bot in the project’s runtime config resolver.

**🔴 Blocking**
- 🔴 Critical — `dispatchTimeoutMs` is dropped by `resolveBotConfigs()`. `PartialConfig` accepts the field in [src/config.ts](/Users/mlamp/.openclaw/tmp/pr-review-ju9phdn1/repo/src/config.ts:251), but the per-bot resolver builds `resolved` from `...config` and only reapplies selected per-bot fields in [src/config.ts](/Users/mlamp/.openclaw/tmp/pr-review-ju9phdn1/repo/src/config.ts:548). It never applies `perBotFile.dispatchTimeoutMs`, and `BotOverride` also has no inline override field at [src/config.ts](/Users/mlamp/.openclaw/tmp/pr-review-ju9phdn1/repo/src/config.ts:230). In the standard two-layer deployment, a bot-specific `dispatchTimeoutMs` in `<baseDir>/<id>/config.json` will be silently ignored and the bot will use the global/default value instead. Since the PR advertises this as tunable per bot, this should be fixed and covered with a config test.

**💬 Non-blocking**
- 🔵 Suggestion — Add the new setting to `config.example.json`, `config.bot.example.json`, and the README config table so operators can discover it. Right now the behavior exists in code but is not documented in the user-facing config surface.

**✅ Highlights**
- The timeout is applied at the right level: `routeAndHandle()` still owns the per-session lock, and `runHandlerWithTimeout()` releases that lock after timeout.
- The per-invocation timeout error identity check avoids misclassifying unrelated handler errors.
- The orphaned handler promise has a defensive rejection handler, which avoids late `unhandledRejection` noise after the race has already timed out.

lml2468:
## Verified mochashanyao's three 🟡 — one is worth more than it looks, two are judgment calls

**#2 (test helper missing `dispatchTimeoutMs`) — real, and slightly more than a nit.** Confirmed: `makeConfig()` in `session-router.test.ts` returns an object literal that omits `dispatchTimeoutMs`, yet its return type is `Config` where the field is **required** (`config.ts:185`, no `?`). This doesn't error only because `tsconfig.json` **excludes `src/__tests__/**`** from type-check — so the missing required field is never caught, and vitest (esbuild) doesn't type-check at runtime. The existing tests pass because every timeout test supplies `dispatchTimeoutMs` via `overrides`; but a bare `makeConfig()` yields `config.dispatchTimeoutMs === undefined` → `runHandlerWithTimeout` takes the `!timeoutMs` disabled branch. Harmless to current tests, but the helper silently violates the `Config` contract. **Worth fixing**: add `dispatchTimeoutMs: 300_000` to `makeConfig`'s base literal so the helper matches `Config` and future tests get the real default. Low effort, prevents a latent footgun.

**#1 (non-timeout error swallowed) — a defensible design trade-off, not a defect.** The non-timeout branch does `console.error` (so it's logged, not silent) and returns. I traced the caller: `index.ts`'s handler wraps its entire body in try/catch (`index.ts:325`), so reaching this branch is genuinely "shouldn't happen" — the author's comment is accurate. Re-throwing (as suggested) would propagate into `withSessionLock`'s gate, which is exactly the lock-release path this PR exists to protect — so swallow-and-log is arguably the *more* correct choice here, not less. I'd keep it; if anything, the log line already preserves the error text for forensics. Non-blocking either way.

**#3 (hardcoded Chinese string → i18n) — out of scope.** The whole codebase uses Chinese user-facing strings (`请稍后再试`, etc.); introducing i18n is a project-wide change, not something this one-line apology should carry. Fine as-is.

None of these block. #2 is the one I'd fold into this PR (one line in the test helper); #1 and #3 are fine as designed. Core fix remains verified-correct (see my review above: race logic, late-resolve/late-reject paths, config `??` 0-disable, 980 tests). LGTM.

lml2468:
## Verified Jerry-Xin's 🔴 — directionally right, but it's 🟡-severity and a pre-existing pattern, not a #142 regression

I traced `resolveBotConfigs` (the multi-bot per-bot config builder, `config.ts:~548`). The precise behavior:

- `resolved` is built as `{ ...config, ...explicit overrides }`. So **`dispatchTimeoutMs` set in the global `config.json` DOES propagate to every bot** (carried by `...config`). It is **not** globally ignored — the common case (one timeout for all bots) works.
- What's missing: `dispatchTimeoutMs` is **not** in the per-bot-file 3-layer merge list (`perBotFile.X ?? bot.X ?? config.X`). So a **per-bot `<baseDir>/<id>/config.json` override of `dispatchTimeoutMs` is dropped** — only the global value applies.

**Crucial nuance — this is the existing design, not introduced here.** Only a small allowlist of fields gets per-bot-file override (`botToken`, `apiUrl`, `sdk.model`, `botBlocklist`, `allowedBotUids`, `mentionFreeGroups`, `groupConfigDir`). The **siblings `rateLimit`, `context`, and `maxResponseChars` are ALSO only carried via `...config`** and likewise can't be overridden per-bot-file. `dispatchTimeoutMs` simply follows the exact same pattern as `maxResponseChars`. So:

- It's **not "config ignored"** — the global value is honored per-bot.
- It's a **consistency limitation** (no per-bot-file granularity), shared with `rateLimit`/`maxResponseChars`, **pre-existing**, not a defect this PR introduced.

**Recommendation — downgrade to 🟡, optional.** If per-bot-file timeout granularity is wanted, add one line to `resolved`:
```ts
dispatchTimeoutMs: perBotFile.dispatchTimeoutMs ?? bot.dispatchTimeoutMs ?? config.dispatchTimeoutMs,
```
…but then `maxResponseChars`/`rateLimit` arguably deserve the same treatment — that's a small separate "per-bot override coverage" cleanup, not a blocker for this timeout fix. The PR's actual goal (a working dispatch timeout) functions correctly at the global level today.

Jerry-Xin's 🟡 #2 (document the new config + add to example config) is a fair, plain follow-up — agreed, worth adding `dispatchTimeoutMs` to `config.example.json` and the README config table.

Net: I would **not treat this as a merge blocker** — the timeout works; per-bot-file override is a pre-existing gap affecting a whole class of fields. Worth a tracked follow-up (either add `dispatchTimeoutMs` to the per-bot merge, or — better — document which fields are per-bot-file-overridable vs global-only). Core fix remains verified-correct.

yujiawei (approved):
# Code Review — PR #142 (cc-channel-octo)

## Summary

This PR adds a per-message dispatch timeout (`#141`) so a hung handler turn (stuck SDK query, wedged tool subprocess, stalled stream) can no longer block a session's serial queue forever. The handler is raced against a `setTimeout`; if the timeout wins, the session lock is released and a one-shot apology is sent. The in-flight turn is intentionally **not** cancelled — it continues to completion in the background. Default is 5 minutes, configurable via `dispatchTimeoutMs`, with `0`/negative disabling the guard.

**Verdict: APPROVED.** The change is correct for its stated goal and is a strict improvement over the prior behavior (a hung turn previously deadlocked the entire session silently). The remaining observations are P2/nits about the inherent trade-offs of the "do not cancel" design — none block merge.

## 1. Verification

- ✅ **Core mechanism is correct.** `runHandlerWithTimeout` (`src/session-router.ts:153-201`) races `handlerPromise` against `timeoutPromise`. On timeout it returns normally, so the enclosing `withSessionLock` `finally` (`src/session-router.ts:100-105`) runs `resolveGate()` and the next queued message proceeds. The deadlock described in `#141` is genuinely resolved.
- ✅ **Timeout identified by reference, not string.** `timeoutError` is a per-invocation `Error` and the catch compares `err === timeoutError` (`src/session-router.ts:165, 182`). A same-text upstream error cannot be misclassified as a timeout — good, this is the right call.
- ✅ **No unhandled rejection on the losing branch.** When the handler wins, `clearTimeout` in `finally` (`:198-200`) cancels the timer so `timeoutPromise` never rejects. When the timeout wins, the detached `handlerPromise.catch(() => {})` (`:177`) swallows a late rejection. Both orphan paths are covered.
- ✅ **Disabled path is clean.** `!timeoutMs || timeoutMs <= 0` (`:158`) runs the handler unguarded with no timer allocated; `NaN` from a malformed config also falls into the disabled branch (`!NaN === true`).
- ✅ **Apology send is bounded.** `replySafe` swallows its own errors (`:559-572`) and the underlying `sendMessage` uses `AbortSignal.timeout(DEFAULT_TIMEOUT_MS=30_000)` (`src/octo/api.ts:14,163`), so a sick Octo API cannot re-hang the release path. The code comment's claim is accurate.
- ✅ **Config plumbing complete.** `dispatchTimeoutMs` is declared on `Config` and `PartialConfig`, defaulted (`300_000`), and merged in `mergeConfig` (`src/config.ts:185,259,295,359`). Consistent with the existing `maxResponseChars` pattern.
- ✅ **Tests pass and cover the key paths.** 5 new tests (hung-handler-unblocks, single bounded apology, normal fast handler no-fire, disabled mode, rejecting-handler-does-not-wedge). Full suite: 980/980 passing; `tsc --noEmit` clean.

## 2. Findings

### P2-1 — Background turn races subsequent turns on the same SDK session / store

This is the headline trade-off and it is documented, but worth making explicit for reviewers. After a timeout, the handler keeps running while the next message acquires the lock and starts its own `queryAgent` resuming the **same** stored SDK session id (`src/index.ts:626, 773`). The two turns can then race on shared mutable state when the timed-out turn eventually finishes:

- `store.setSdkSessionId(...)` / `onSessionId` (`src/agent-bridge.ts:322-327`) — last-writer-wins; the timed-out turn completing *after* turn 2 could overwrite turn 2's session id, leaving the next turn to resume a stale id.
- `store.appendAssistant(...)` and `store.setLastBotReplySeq(...)` (`src/index.ts:801-808`).
- Two concurrent `streamRelay.deliver(...)` to the same channel (`src/index.ts:785`) — interleaved output.

Worst case is broader than the comment's "a delayed real reply arriving after the apology": it includes out-of-order/interleaved replies and a momentarily inconsistent session pointer. Given this only triggers on an actual hang (rare), and the alternative is a permanent silent deadlock, the trade-off is acceptable for this PR. Suggestion: a follow-up could tag the timed-out turn so its late completion skips the session-id/history writes (e.g. an `aborted` flag checked before `setSdkSessionId`/`appendAssistant`). Not required to merge.

### P2-2 — "请稍后重试" can compound concurrency

The apology `⚠️ 处理超时，请稍后重试。` (`src/session-router.ts:189`) invites the user to retry while the original turn is *still running* in the background. A retry spins up a second concurrent turn against the same session, amplifying P2-1. Consider softer wording (e.g. "正在继续处理，可能稍后回复" / "still working on it, a reply may follow") so the message matches the actual semantics — the turn was not abandoned, only unblocked. Nit-level wording, non-blocking.

### Nit — No test for late completion of a timed-out handler

The new tests confirm the lock releases and a single apology fires, but none asserts what happens when the hung handler *eventually resolves after* the timeout (the P2-1 path) — e.g. that it does not throw or double-send. Since the production `index.ts` handler is fully try/caught this is low risk, but a regression test pinning "late resolution is a no-op for the queue" would lock in the intended contract.

### Nit — `dispatchTimeoutMs` not range-validated

Like the sibling fields, the config value is accepted from JSON without validation (`src/config.ts:321`). The runtime guard handles `0`/negative/`NaN` gracefully (disabled), and a stringified number would be coerced by `setTimeout`, so this is consistent with existing code and not a defect — just flagging that none of these numeric config knobs are validated, should the project ever add a config schema.

## 3. Recommendations

- Merge as-is. The fix delivers its stated value with correct lock-release semantics and good test coverage.
- Track P2-1 as a follow-up: guard the timed-out turn's late writes (`setSdkSessionId`, `appendAssistant`, stream) so a background completion cannot clobber a newer turn's session state.
- Optionally soften the apology copy (P2-2) to reflect that the turn continues.

## 4. Additional observations

- The implementation mirrors the referenced `openclaw #75` scope ("unblock the queue, don't cancel the turn") consistently, including the documentation of the non-cancellation trade-off — good cross-repo consistency.
- `console.warn`/`console.error` include `result.sessionKey`, which for DMs embeds `spaceId:from_uid`. This matches existing logging in this file (e.g. `src/index.ts:822`), so it introduces no new exposure, but be aware these logs carry a user identifier.

lml2468:
## Confirming yujiawei's 🟡 — the orphaned-turn state race is real, and sharper than the PR's "late reply" note

yujiawei's two 🟡 point at the documented "don't cancel the in-flight turn" trade-off — but the **"session pointer confusion"** angle is more specific than the PR comment admits, and worth recording. I traced the write timeline.

**The race (after a timeout):**
1. Turn A hangs inside `queryAgent`/`streamRelay.deliver` → times out → lock releases + apology sent.
2. Turn B (next message, same session) proceeds: reads `getSdkSessionId` (`index.ts:626`), `appendUser` (`:615`), and registers its own `onSessionId` (`:726`).
3. **Turn A later revives in the background** and runs its tail, which mutates the *same* session state:
   - `:726 onSessionId` → `store.setSdkSessionId(sessionKey, …)` can **overwrite** the SDK session id B just set, with the id from A's (older) query.
   - `:801 appendAssistant` → A's late reply is appended to history **out of order** (after B's user turn).
   - `:808 setLastBotReplySeq` → may **move the reply-seq cursor backward** to A's `msg_seq`.

So the worst case isn't just "a late reply appears" (as the comment says) — it's **SDK-session-id clobber + out-of-order history + reply-seq cursor regression**, which can make the *next* resume attach to the wrong session or replay history oddly. yujiawei's "会话指针混乱" is an accurate label.

**Severity — still non-blocking, but document it honestly:**
- This is the explicitly-accepted scope (PR comment + openclaw #75 parity). The fix's core purpose — never wedge the session lock — is achieved and verified.
- Trigger is narrow: a turn must hang a full `dispatchTimeoutMs` (5 min default) **and then revive and complete**. Most real hangs (stuck subprocess, dead socket) never revive, so they never reach these writes. It's the "slow-but-not-dead" turn that hits this.
- It's state inconsistency, not corruption/crash.

**Suggested follow-up (not required for this PR):** when the timeout fires, mark the orphaned turn stale (e.g. an `aborted` flag captured in the closure) and have its `onSessionId` / `appendAssistant` / `setLastBotReplySeq` writes **no-op** if the session has since advanced. That closes the write-race without having to cancel the SDK query. Worth a tracked issue; the PR comment should at least be upgraded to name the state-write race, not just "a delayed reply."

Net: yujiawei's finding is valid and well-aimed; I'd keep it 🟡 (the PR's stated goal holds, trigger is narrow), but the known-consequence note in the code deserves to be more honest about the session-state writes. Core deadlock fix remains verified-correct.

lml2468:
## Status: #142 merged — and unlike #137/#138, the merged code has no true blocker

For the record: #142 merged with Jerry-Xin's 🔴 CHANGES_REQUESTED still **active on the current HEAD `c65a067`** (not stale). That's the 4th PR in this batch merged over an unresolved 🔴.

**But the merge outcome here is fine, and I want to say that plainly so this doesn't read as alarm:**
- Jerry-Xin's 🔴 ("`dispatchTimeoutMs` not applied per-bot") I verified down to **🟡 severity**: the global `config.json` value **does** propagate to every bot (via `...config` in `resolveBotConfigs`); only a *per-bot-file* override is dropped — and that's a **pre-existing pattern shared with `rateLimit`/`maxResponseChars`**, not a defect this PR introduced. The timeout works.
- The core fix (race logic, late-resolve/reject paths, `0`-disable, 980 tests) is **verified-correct** — I ran it.
- The two 🟡 design-tradeoff notes (orphaned-turn state race) are real but explicitly-scoped and narrow-trigger.

So nothing in the merged code needs an urgent fix. The recommended follow-ups remain **optional/tracked**, not blockers:
1. (🟡) Add `dispatchTimeoutMs` to the per-bot-file merge in `resolveBotConfigs` — or document which fields are global-only vs per-bot-overridable (applies to `rateLimit`/`maxResponseChars` too).
2. (🟡) Document `dispatchTimeoutMs` + add to `config.example.json` and the README config table (Jerry-Xin #2).
3. (🟡) Harden the orphaned-turn write-race (stale-flag the timed-out turn so its `onSessionId`/`appendAssistant`/`setLastBotReplySeq` no-op) — or at least upgrade the code comment to name the state-write race.
4. (🟡) `makeConfig` test helper should include `dispatchTimeoutMs` to match the `Config` contract (mochashanyao #2).

### Process note (4th occurrence)
#136, #137, #138, #142 all merged with an open finding unresolved at merge time (#137/#138 over a *real* 🔴; #142 over a 🔴-that-was-actually-🟡). The merge gate appears to not block on active CHANGES_REQUESTED. For #142 it happened to be harmless; for #137/#138 it shipped genuine breakage. Strongly worth a maintainer fixing the gate so the next time it's a real blocker, it's caught. I'm not re-opening this PR — the code is sound; flagging the pattern only.

PR #145, opened 2026-06-10
fix(mentions): authoritative member set for outbound @uid validation (reworks #144)

### LINKED ISSUE #143: P1: validate outbound @uid mentions against the live group member list (drop hallucinated uids)

lml2468:
Reworks #144 to fix **Jerry-Xin's blocking 🔴**. Re-closes #143.

This branch is **revert of #144** + a corrected reimplementation, so the history shows exactly what changed vs the merged-then-reverted version.

## What was wrong with #144

#144 was merged prematurely over an unresolved blocking review. Jerry-Xin correctly flagged: the membership check **wasn't based on an authoritative member set**.

`refreshMembers` only **upserted** members returned by `getGroupMembers` — it **never pruned** members who left. So `isMember()` accepted:
- a user who left the group (still cached), and
- stale uids restored from the `group_members` DB table on restart.

→ the outbound guard's core guarantee (`uid ∉ members → downgrade to plain text`) was defeated. The "stronger than openclaw, validates real membership" claim was hollow because the set only ever grew.

## Fix

**1. Authoritative roster in `refreshMembers`** — members absent from a (non-empty) server response are pruned from memory (`memberMap` / `nameToUid` / `robotFlags`) **and** the persisted `group_members` row (new `deleteMember` statement). The server roster is the source of truth (verified non-paginated).

**2. Empty-roster guard** (found by the post-fix code-review) — `getGroupMembers` returns `[]` both for a genuinely empty group **and** for a malformed 200 (`data.members` not an array → silently `[]`). A group the bot is in always has ≥1 member, so we **prune only when `present.size > 0`**. Otherwise one transient quirk would wipe the whole roster — and since `lastRefresh` is set on that "success", it wouldn't re-fetch for 1h, downgrading every mention to plain text in that window. A real emptying still prunes member-by-member across non-empty shrinking responses.

**3. A8 guard on the now-authoritative set** — `isMember` / `getNameToUidMap` getters; `resolveMentions` gains an optional `isValidUid` predicate (a v2 `@[uid:name]` whose uid isn't a member → downgraded to plain `@name`, no entity/uid); `deliver` + `index.ts` wired (groups only, DMs skip). Also activates the previously-**dead** v1 `@name` resolution (`deliver`'s `memberMap` param was never passed).

**Best-effort, documented:** only as fresh as the throttled (1h) refresh + DB seed on restart. Defense-in-depth — the server still enforces real permissions.

## Tests (8 new, 988 total)

- **Jerry-Xin's required case**: refresh `{u1,u2}` then `{u1}` → `isMember(u2)` false, reverse name entry gone.
- Departed member's **DB row** pruned and **not resurrected** on `loadMembersFromDb`.
- **Empty-roster does NOT mass-prune** (transient-quirk guard).
- A8 hit / miss (downgrade) / mixed / legacy-no-predicate; `isMember` + `getNameToUidMap` getters.

## Verification

- `npm run type-check` ok · `npm run lint` (zero warnings) ok · `npm test` — 988 passing
- `code-review` skill (medium) run — surfaced the empty-roster mass-prune, fixed in this branch before pushing.

cc @Jerry-Xin — this addresses the stale-membership 🔴 directly; please re-review.

### DIFF
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -413,4 +413,97 @@ describe('G23: robot flag', () => {
     expect(ctx.isRobot('ch1', 'shared')).toBe(true);
     expect(ctx.isRobot('ch2', 'shared')).toBe(false);
   });
+
+  // --- A8 (#143): authoritative membership + outbound validation getters ---
+
+  it('isMember / getNameToUidMap reflect a learned member', () => {
+    ctx.learnMember('ch1', 'u1', 'Alice');
+    expect(ctx.isMember('ch1', 'u1')).toBe(true);
+    expect(ctx.isMember('ch1', 'ghost')).toBe(false);
+    expect(ctx.isMember('ch2', 'u1')).toBe(false); // per-channel isolation
+    expect(ctx.getNameToUidMap('ch1').get('Alice')).toBe('u1');
+    expect(ctx.getNameToUidMap('nope').size).toBe(0);
+  });
+
+  it('refreshMembers prunes a departed member (Jerry-Xin 🔴): memory + reverse map', async () => {
+    // First authoritative refresh: u1 + u2 present.
+    (getGroupMembers as ReturnType<typeof vi.fn>)
+      .mockResolvedValueOnce([
+        { uid: 'u1', name: 'Alice', role: 0 },
+        { uid: 'u2', name: 'Bob', role: 0 },
+      ])
+      .mockResolvedValueOnce([{ uid: 'u1', name: 'Alice', role: 0 }]); // u2 left
+    await ctx.refreshMembers('ch-prune', 'https://api', 'token');
+    expect(ctx.isMember('ch-prune', 'u1')).toBe(true);
+    expect(ctx.isMember('ch-prune', 'u2')).toBe(true);
+
+    // Bypass the 1h throttle so the second refresh is effective.
+    vi.useFakeTimers();
+    vi.setSystemTime(Date.now() + 61 * 60 * 1000);
+    try {
+      await ctx.refreshMembers('ch-prune', 'https://api', 'token');
+    } finally {
+      vi.useRealTimers();
+    }
+    // Guard: the second refresh actually ran (not silently throttled).
+    expect(getGroupMembers).toHaveBeenCalledTimes(2);
+
+    // u2 left → must no longer be a member, and its reverse name entry is gone.
+    expect(ctx.isMember('ch-prune', 'u1')).toBe(true);
+    expect(ctx.isMember('ch-prune', 'u2')).toBe(false);
+    expect(ctx.getNameToUidMap('ch-prune').get('Bob')).toBeUndefined();
+    expect(ctx.getNameToUidMap('ch-prune').get('Alice')).toBe('u1');
+  });
+
+  it('refreshMembers pruning removes the persisted DB row for a departed member', async () => {
+    // Roster shrinks from {gone, keep} → {keep}. A non-empty response is trusted
+    // as authoritative, so `gone` is pruned from memory AND the DB.
+    (getGroupMembers as ReturnType<typeof vi.fn>)
+      .mockResolvedValueOnce([
+        { uid: 'gone', name: 'Ghost', role: 0 },
+        { uid: 'keep', name: 'Keeper', role: 0 },
+      ])
+      .mockResolvedValueOnce([{ uid: 'keep', name: 'Keeper', role: 0 }]);
+    await ctx.refreshMembers('ch-db', 'https://api', 'token');
+    expect(ctx.isMember('ch-db', 'gone')).toBe(true);
+
+    vi.useFakeTimers();
+    vi.setSystemTime(Date.now() + 61 * 60 * 1000);
+    try {
+      await ctx.refreshMembers('ch-db', 'https://api', 'token');
+    } finally {
+      vi.useRealTimers();
+    }
+    expect(ctx.isMember('ch-db', 'gone')).toBe(false);
+    expect(ctx.isMember('ch-db', 'keep')).toBe(true);
+
+    // A fresh GroupContext loading from the same DB must NOT resurrect the row.
+    const ctx2 = new GroupContext(adapter, 6000);
+    ctx2.loadMembersFromDb('ch-db');
+    expect(ctx2.isMember('ch-db', 'gone')).toBe(false);
+    expect(ctx2.isMember('ch-db', 'keep')).toBe(true);
+  });
+
+  it('refreshMembers does NOT mass-prune on an empty roster response (transient-quirk guard)', async () => {
+    (getGroupMembers as ReturnType<typeof vi.fn>)
+      .mockResolvedValueOnce([
+        { uid: 'u1', name: 'Alice', role: 0 },
+        { uid: 'u2', name: 'Bob', role: 0 },
+      ])
+      .mockResolvedValueOnce([]); // empty: treated as a quirk, not "everyone left"
+    await ctx.refreshMembers('ch-empty', 'https://api', 'token');
+    expect(ctx.isMember('ch-empty', 'u1')).toBe(true);
+
+    vi.useFakeTimers();
+    vi.setSystemTime(Date.now() + 61 * 60 * 1000);
+    try {
+      await ctx.refreshMembers('ch-empty', 'https://api', 'token');
+    } finally {
+      vi.useRealTimers();
+    }
+    expect(getGroupMembers).toHaveBeenCalledTimes(2);
+    // Prior roster is kept — an empty response does not wipe everyone.
+    expect(ctx.isMember('ch-empty', 'u1')).toBe(true);
+    expect(ctx.isMember('ch-empty', 'u2')).toBe(true);
+  });
 });
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -170,24 +170,6 @@ describe('GroupContext', () => {
     expect(ctx.resolveMentions('@Alice and @Alice again', 'ch1')).toEqual(['u1']);
   });
 
-  // --- A8 (#143): outbound mention validation getters ---
-
-  it('isMember reflects the live member list per channel', () => {
-    ctx.learnMember('ch1', 'u1', 'Alice');
-    expect(ctx.isMember('ch1', 'u1')).toBe(true);
-    expect(ctx.isMember('ch1', 'ghost')).toBe(false);
-    // Isolation: a member of ch1 is not a member of ch2.
-    expect(ctx.isMember('ch2', 'u1')).toBe(false);
-  });
-
-  it('getNameToUidMap exposes the displayName→uid map for outbound @name', () => {
-    ctx.learnMember('ch1', 'u1', 'Alice');
-    const map = ctx.getNameToUidMap('ch1');
-    expect(map.get('Alice')).toBe('u1');
-    // Unknown channel → empty (not undefined).
-    expect(ctx.getNameToUidMap('nope').size).toBe(0);
-  });
-
   // --- refreshMembers ---
 
   it('refreshMembers throttles within 1h', async () => {
--- a/src/__tests__/mention-utils.test.ts
+++ b/src/__tests__/mention-utils.test.ts
@@ -236,40 +236,6 @@ describe('resolveMentions', () => {
     expect(result.mentionEntities[1].uid).toBe('u1');
     expect(result.mentionUids).toEqual(['u2', 'u1']);
   });
-
-  // A8 (#143): membership validation of outbound structured @[uid:name]
-  describe('isValidUid membership guard', () => {
-    it('keeps a structured mention whose uid passes the predicate', () => {
-      const members = new Set(['u1']);
-      const result = resolveMentions('@[u1:Alice] hi', undefined, (uid) => members.has(uid));
-      expect(result.finalContent).toBe('@Alice hi');
-      expect(result.mentionUids).toEqual(['u1']);
-      expect(result.mentionEntities).toHaveLength(1);
-    });
-
-    it('downgrades a hallucinated uid to plain text (drops entity, keeps @name)', () => {
-      const members = new Set(['u1']);
-      const result = resolveMentions('@[deadbeef:Ghost] hi', undefined, (uid) => members.has(uid));
-      // @name text survives, but no entity/uid is emitted → no bogus notify.
-      expect(result.finalContent).toBe('@Ghost hi');
-      expect(result.mentionUids).toEqual([]);
-      expect(result.mentionEntities).toEqual([]);
-    });
-
-    it('keeps real members and drops hallucinated ones in the same message', () => {
-      const members = new Set(['u1']);
-      const result = resolveMentions('@[u1:Alice] and @[fake:Bob]', undefined, (uid) => members.has(uid));
-      expect(result.finalContent).toBe('@Alice and @Bob');
-      expect(result.mentionUids).toEqual(['u1']);
-      expect(result.mentionEntities).toHaveLength(1);
-      expect(result.mentionEntities[0].uid).toBe('u1');
-    });
-
-    it('without a predicate, trusts every structured uid (legacy behavior)', () => {
-      const result = resolveMentions('@[whatever:Ghost] hi');
-      expect(result.mentionUids).toEqual(['whatever']);
-    });
-  });
 });
 
 describe('patterns', () => {
--- a/src/__tests__/mention-utils.test.ts
+++ b/src/__tests__/mention-utils.test.ts
@@ -236,6 +236,39 @@ describe('resolveMentions', () => {
     expect(result.mentionEntities[1].uid).toBe('u1');
     expect(result.mentionUids).toEqual(['u2', 'u1']);
   });
+
+  // A8 (#143): membership validation of outbound structured @[uid:name]
+  describe('isValidUid membership guard', () => {
+    it('keeps a structured mention whose uid passes the predicate', () => {
+      const members = new Set(['u1']);
+      const result = resolveMentions('@[u1:Alice] hi', undefined, (uid) => members.has(uid));
+      expect(result.finalContent).toBe('@Alice hi');
+      expect(result.mentionUids).toEqual(['u1']);
+      expect(result.mentionEntities).toHaveLength(1);
+    });
+
+    it('downgrades a hallucinated uid to plain text (drops entity, keeps @name)', () => {
+      const members = new Set(['u1']);
+      const result = resolveMentions('@[deadbeef:Ghost] hi', undefined, (uid) => members.has(uid));
+      expect(result.finalContent).toBe('@Ghost hi');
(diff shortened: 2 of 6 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#145 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo#145
**Head SHA:** c68a75cb3b8467fc4c51642721233800fe387604
**Scope:** fix(mentions): authoritative member set for outbound @uid validation (reworks #144)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Root cause addressed (stale membership) | ✅ | `refreshMembers` now prunes departed members from memory + DB (`group-context.ts:248-280`) |
| Empty-roster guard | ✅ | `if (present.size > 0)` gates pruning (`group-context.ts:261`) |
| DB cleanup (deleteMember) | ✅ | New prepared statement (`group-context.ts:87-89`), invoked per pruned uid (`group-context.ts:275-278`) |
| Robot flag cleanup | ✅ | `rfMap?.delete(uid)` in pruning loop (`group-context.ts:271`) |
| Reverse name-map safety | ✅ | Conditional delete only when entry still points at pruned uid (`group-context.ts:268-270`) |
| index.ts wiring (A8 guard) | ✅ | Pre-existing on main, comments updated (`index.ts:785-793`) |
| resolveMentions isValidUid predicate | ✅ | Pre-existing on main, JSDoc updated (`mention-utils.ts:208-218`) |
| Test coverage (8 new tests) | ✅ | Prune memory+DB, DB resurrection guard, empty-roster guard, A8 hit/miss/mixed/legacy |

## 2. Findings

### No P0/P1 issues found.

### P2 Observations

**P2-1: `lastRefresh` set before pruning completes**
- **Diff-scope:** pre-existing (the `lastRefresh.set()` at `group-context.ts:205` was already before the loop body in main; this PR extends the loop body with pruning but doesn't change the `lastRefresh` placement).
- **Risk:** If a future change introduces an uncaught exception in the pruning loop, `lastRefresh` would be set but pruning incomplete. Currently safe because each item's `deleteMember.run` is individually try/caught.
- **Recommendation:** Consider moving `lastRefresh.set()` to after both loops complete, or add a brief comment noting the invariant that per-item errors are caught.

**P2-2: No operational logging of prune count**
- **Diff-scope:** new (this PR adds pruning but no info-level log of how many members were pruned).
- **Risk:** Low — operational debugging would benefit from knowing "3 members pruned from ch-xyz" when investigating stale-mention issues.
- **Recommendation:** Add a `console.log` or `console.debug` after the pruning loop with the count of pruned members.

**P2-3: Members with valid uid but empty name get pruned**
- **Diff-scope:** new behavior (the `if (!m.uid || !m.name) continue` guard at `group-context.ts:222` skips adding to `present`, so such members are pruned).
- **Risk:** Low — a member without a display name is unusable for mention resolution anyway. If the server temporarily returns an empty name, the member is pruned and re-added on the next successful refresh. Within the documented best-effort window.
- **Recommendation:** No action needed; documenting for awareness.

## 3. Suggestions

- Consider adding a prune-count log line for operational visibility (P2-2 above).
- The test suite could benefit from one additional case: `learnMember` re-adding a user after `refreshMembers` pruned them (simulating rejoin + first message). The behavior is correct (learnMember adds to map, next refresh confirms or re-prunes), but an explicit test would prevent future regressions.

## 4. Additional Findings

None beyond the P2 observations above.

## 5. Data Flow Trace

|---|---|---|
| `present` set | `getGroupMembers()` → API `/v1/bot/groups/{id}/members` → filtered by `!m.uid \|\| !m.name` | ✅ All valid-uid+name members captured |
| `memberMap` pruning | `[...memberMap.keys()]` snapshot → `present.has(uid)` check → `memberMap.delete(uid)` | ✅ Snapshot iteration avoids mutation issues |
| `nameMap` pruning | `nameMap.get(staleName) === uid` guard → only delete if still points at pruned uid | ✅ Rename-safe |
| `deleteMember` DB stmt | `adapter.prepare('DELETE FROM group_members WHERE group_id = ? AND uid = ?')` in `initStatements()` | ✅ Initialized before use |
| `isMember()` after prune | `memberMapByChannel.get(channelId)?.has(uid)` → departed uid removed from map | ✅ Returns false post-prune |
| `getNameToUidMap()` after prune | Returns live `nameToUid` map reference → stale entries removed by pruning loop | ✅ Correct |
| `isValidMentionUid` → `resolveMentions` | Closure over `groupContext.isMember(channelId, uid)` → filters `converted.entities` | ✅ Non-member uids dropped |
| `loadMembersFromDb` post-prune | `selectMembers.all(channelId)` → pruned rows deleted from DB → not resurrected | ✅ Verified by test |

## 6. R5 Blindspot Checklist

- **C1 — Dual-path parity:** `learnMember` (add via inbound message) vs `refreshMembers` pruning (remove). Both touch `memberMap`, `nameMap`, and DB. `learnMember` doesn't touch `rfMap` (correct — inbound messages don't carry robot flag info; only `refreshMembers` manages robot flags). **Clear — symmetric where it matters.**

- **C2 — Control-flow ordering / nesting reuse:** Pruning loop runs AFTER upsert loop — correct ordering (add/update first, then remove stale). No double-action risk: upsert populates `present`, pruning checks against it. No nested/compound calls on sanitized data. **Clear.**

- **C3 — Authorization boundary ≠ capability boundary:** This PR doesn't change authorization boundaries. The outbound mention guard is defense-in-depth; the server enforces real permissions. **N/A.**

---

**[Octo-Q] verdict: APPROVE**

The PR correctly fixes the stale-membership root cause identified in #144's blocking review. The authoritative roster pruning is sound, the empty-roster guard prevents the most dangerous failure mode (mass-wipe on transient API quirk), and the data flows from server response through memory/DB to the outbound mention validation predicate are all verified correct. Test coverage is thorough (8 new tests covering the key scenarios). Only P2/nit observations — no blocking issues.

Jerry-Xin (approved):
The PR is in scope for `cc-channel-octo`, but there is one blocking bug in the outbound mention guard that defeats the stated downgrade guarantee.

**🔴 Blocking**
- 🔴 Critical: Invalid structured mentions can be re-promoted by the v1 `@name` fallback. In [src/mention-utils.ts](/Users/mlamp/.openclaw/tmp/pr-review-5xmgtvea/repo/src/mention-utils.ts:238), invalid structured entities are filtered out, but then fallback resolution runs over `finalContent` at [src/mention-utils.ts](/Users/mlamp/.openclaw/tmp/pr-review-5xmgtvea/repo/src/mention-utils.ts:243) and only skips offsets already present in the filtered `entities` set. For a group call path that now passes both `memberMap` and `isValidUid`, `@[deadbeef:Alice]` becomes `@Alice`; if `Alice` exists in `memberMap`, fallback emits a valid entity for Alice anyway. That contradicts the PR’s guarantee that an invalid structured uid is downgraded to plain text with no entity/uid. Fix by treating all structured mention replacement offsets as consumed, including invalid ones, before running fallback.

**💬 Non-blocking**
- 🟡 Warning: I could not run verification locally because dependencies are not installed in this checkout: `vitest: command not found` and `tsc: command not found`.

**✅ Highlights**
- The `refreshMembers` pruning in [src/group-context.ts](/Users/mlamp/.openclaw/tmp/pr-review-5xmgtvea/repo/src/group-context.ts:261) addresses the stale member cache issue in both memory and persistence.
- The empty-roster guard is a good defensive choice for the current `getGroupMembers` behavior.

lml2468:
## Confirming Jerry-Xin's 🔴 — and correcting my own earlier call on #144

Jerry-Xin is right, and this overturns a judgment I made on #144. I reproduced it.

### The bug (verified)
`resolveMentions('@[fake-uid:Alice] hi', memberMap={Alice→real-alice}, isValidUid)` →
```
finalContent: "@Alice hi"
mentionUids : ["real-alice"]      ← a real @ notification fired
entities    : [{uid:"real-alice", offset:0, length:6}]
```
Sequence: v2 parses `@[fake-uid:Alice]`, A8 **correctly drops** the entity (fake-uid ∉ members) leaving plain `@Alice` text — but then the **v1 `@name` fallback re-resolves `@Alice`** against `memberMap` and **recovers it as a real entity for `real-alice`**. The offset check doesn't prevent it: the dropped v2 entity left no offset behind to skip.

### Why this is a real vulnerability, not a "helpful rebind"
This **violates the PR's own stated contract** — "`uid ∉ members → downgrade to plain text.`" The mention is *not* downgraded to text; it's silently **redirected to a different uid than the LLM specified**. In `@[uid:name]`, the **uid is the authoritative target**, name is just display. When the uid is hallucinated/forged, A8 makes the right call (untrusted → strip). The v1 fallback then **second-guesses that decision using the display name** and re-issues a notification — to whoever currently holds that name in the group. Display names are non-unique and attacker-influenceable (a new/renamed member can take the name), so a mention A8 just neutralized comes back as a real @ to a possibly-unintended person. That's exactly the bogus-notify class #143/A8 exists to close, slipping in via the v2-drop → v1-recover path.

### Correcting myself
On #144 I wrote an adversarial test for this exact interaction and called it "correct rebind — if there's a real Alice, @-ing the real Alice is reasonable." **That was wrong.** I only modeled the benign case (LLM meant the real Alice, fat-fingered the uid) and missed the adversarial/contract case: a security *downgrade decision must not be silently reversed by later fallback logic in the same function*. Jerry-Xin modeled it correctly. My earlier "verified correct" on that path was a miss — on the record.

### Fix
The v1 `@name` fallback must not resurrect a name whose v2 entity was just dropped by the A8 predicate. Options:
1. **Track dropped names/offsets and exclude them from v1 fallback** — if `@Alice` at this position came from a dropped `@[fake:Alice]`, don't re-resolve it.
2. **Run v1 fallback only over text that never contained a v2 mention** (compute v1 candidates from the *original* pre-conversion content, skipping spans that were `@[…]`), or
3. Simpler: when `isValidUid` is provided, also gate the **v1 fallback** entities through it — though here the recovered uid (real-alice) *is* a member, so this alone doesn't fix it; the real issue is recovering a *dropped* mention at all. Option 1/2 is the correct fix.
Add a regression test: `@[fake:Alice]` with a real member Alice → `mentionUids` empty, plain `@Alice` text only.

### Other notes
- mochashanyao's 🟡 on `lastRefresh.set()` timing: worth checking — if the prune loop throws, `lastRefresh` is already set so it won't retry for 1h. The prune `deleteMember.run` is individually try/caught, so a throw escaping the loop is unlikely, but moving `lastRefresh.set` to a `finally`-after-success or documenting why it's safe would harden it. Non-blocking.
- The **stale-membership fix itself is correct and verified**: `refreshMembers` now prunes memory + DB + robotFlags against the authoritative `present` set, the empty-roster guard correctly avoids mass-prune on a transient `[]`, and 988 tests pass. That part fully addresses #144's 🔴.

### Verdict
The membership-prune rework (the headline fix) is solid. But Jerry-Xin found a **separate, real 🔴** in the A8 downgrade path that **I missed on #144**: a dropped hallucinated v2 mention is recovered by v1 fallback, defeating the downgrade contract. Should-fix before merge. (`check-sprint` ❌ is the usual board gate.)

yujiawei (approved):
# Code Review — PR #145 (cc-channel-octo)

**Verdict: APPROVED** — this correctly fixes the blocking review on #144. No P0/P1. Two P2 items below (one external-dependency assumption to confirm, one nit).

## Summary

#144's outbound @uid guard was hollow because `refreshMembers` only **upserted** the members the server returned and never **pruned** the ones who left. `isMember()` therefore kept accepting departed users (and uids restored from the `group_members` DB table on restart), defeating the guard's core guarantee (`uid ∉ members → downgrade to plain text`). This PR makes the server roster authoritative: it tracks who was returned (`present`), then drops anyone cached/persisted who is no longer present — from `memberMap`, `nameToUid`, `robotFlags`, and the DB (new `deleteMember` statement).

## 1. Verification

- ✅ **Departed member is pruned** (`group-context.ts:261-279`) — memory, reverse name map, robot flags, and the persisted DB row are all cleared. Reverse-map deletion is correctly guarded (`nameMap.get(staleName) === uid`) so a rename that re-pointed the name to another user isn't clobbered (mirrors `learnMember`).
- ✅ **DB row not resurrected on restart** — `deleteMember` removes the `group_members` row; the test spins up a fresh `GroupContext`, calls `loadMembersFromDb`, and asserts the departed uid stays gone (`group-context.test.ts`).
- ✅ **Empty-roster mass-prune guard** (`group-context.ts:261`, `present.size > 0`) — `getGroupMembers` returns `[]` for a malformed/unexpected-shape 200 (`data.members` not an array → silently `[]`, see `octo/api.ts:324-329`) as well as a genuinely empty group. A group the bot is in always has ≥1 member, so `[]` is treated as a transient quirk, not "everyone left". The reasoning is sound: `lastRefresh` is set on the API success (`group-context.ts:206`) **before** the prune, so a naive mass-prune would both wipe the roster and suppress re-fetch for 1h — the guard avoids exactly that.
- ✅ **A8 downgrade wiring** — `index.ts:791-795` passes `getNameToUidMap` + an `isMember`-backed `isValidUid` predicate (groups only; DMs leave both `undefined`); `resolveMentions` (`mention-utils.ts:238-241`) drops entities for uids failing the predicate while keeping the `@name` text → harmless plain text, no bogus notify.
- ✅ **type-check / lint / tests** — `npm run type-check` clean, `npm run lint` clean, `npm test` → **988 passing** (reproduced locally). The 8 new tests directly cover Jerry-Xin's required case (refresh `{u1,u2}` → `{u1}` ⇒ `isMember(u2)` false, reverse entry gone), DB-row pruning + non-resurrection, and the empty-roster guard.
- ✅ **Concurrency** — `refreshMembers` runs inside the per-channel session lock in `handleMessage` (`index.ts:317`, `:383`), so the snapshot-iterate-and-mutate prune loop isn't racy.

## 2. Findings

### P2 — Pruning correctness depends on the members endpoint being non-paginated (please confirm)

The whole prune is built on the premise that `getGroupMembers` returns the **complete** current roster in one call. `GET /v1/bot/groups/{groupNo}/members` (`octo/api.ts:314-329`) sends **no** `limit`/`offset`/`page`/`cursor` params, which is consistent with the PR's "verified non-paginated" claim. But if the server ever caps or paginates this response, every refresh would prune the legitimate members beyond the cap on a large group — silently downgrading their @mentions to plain text, i.e. re-defeating the feature for exactly the big groups that need it. The empty-roster guard does **not** protect against this (a partial non-empty response still prunes the remainder).

This is the cross-file/external-dependency blind spot worth an explicit confirmation rather than a build-time check. Not blocking — the author states it's verified, and the prior no-prune behavior was the actual bug. Suggestion: if there's any doubt, add a defensive cap (e.g. only prune when the returned count is within an expected band) or a comment pinning the server contract.

### P2 (nit) — Members with a valid uid but empty `name` are treated as "not present"

The upsert loop skips entries failing `if (!m.uid || !m.name) continue;` (`group-context.ts:224`), so such a member is never added to `present` and will be **pruned** if previously cached. A real member the server momentarily returns with a blank display name would be dropped (memory + DB), and an in-flight `@[uid:name]` to them downgraded. Narrow edge (a nameless member can't be @name-resolved anyway, and it self-heals on the next clean refresh), but the behavior is slightly stronger than "prune who actually left." Optional: add the uid to `present` before the name check so only *absent* uids are pruned.

### Nit — Per-member DB deletes aren't transactional

The prune issues one `deleteMember.run` per stale uid, each in its own try/catch (`group-context.ts:273-277`). If one throws, memory is already mutated but the DB row remains, so a restart's `loadMembersFromDb` would resurrect that single member until the next successful refresh. Self-healing and low-impact; wrapping the prune in a transaction (as a batch) would tighten memory/DB consistency.

## 3. Suggestions

- Confirm and document the members-endpoint contract (single-shot full roster) inline near `getGroupMembers`, since `refreshMembers` now silently depends on it for correctness.
- Consider `present.add(m.uid)` before the `!m.name` skip so the prune only ever targets genuinely-absent uids.

## 4. Extra observations

- Good call reactivating the previously-dead v1 `@name` resolution path (`deliver`'s `memberMap` was never passed before) — surfaced and noted in the PR body.
- Comments are unusually thorough and accurately capture the failure modes (the `lastRefresh`-before-prune trap and the empty-`[]` ambiguity). This will save the next reader real time.

Net: a focused, well-tested fix that resolves the stale-membership blocker directly. Approving; the pagination assumption is the one thing worth a quick out-of-band confirmation.

PR #130, opened 2026-06-09
ci: add release-drafter + release-publish callers

lml2468:
Completes release-plane coverage for cc-channel-octo (matching sibling channel adapters). release-drafter keeps a rolling draft from merged PRs; release-publish publishes on manual dispatch with CI-run validation.

PR #146, opened 2026-06-16
chore: merge main into feat/agent-runtime

caster-Q:
## What

Merge `main` into `feat/agent-runtime` to bring the runtime branch up to date with the released gateway (v1.0.0/v1.0.1) and two functional fixes that the runtime branch was missing.

## Why

`feat/agent-runtime` had diverged: it added the supervisor CLI (`src/cli.ts`) but was missing 8 commits from `main`, including two real fixes:

- **#142** — per-message dispatch timeout to unblock hung sessions
- **#144 / #145** — outbound `@mention` validated against the live group member set (drops hallucinated mentions)

…plus the npm release machinery (v1.0.0/v1.0.1, `npm-publish.yml`, release-drafter).

## Conflict resolution

Only `package.json` conflicted (src code had **zero** overlap — main touched session-router/mention-utils/group-context/etc., the runtime branch only added `cli.ts`):

- `bin` → kept the runtime branch's `dist/cli.js` (the process-supervisor entry).
- `version` → took main's `1.0.1` (already published; must not regress).
- `files` / `repository` / `homepage` / `bugs` / `publishConfig` → kept main's publish fields.
- `package-lock.json` regenerated via `npm install` (bin → `dist/cli.js`, version → `1.0.1`; no dependency-tree drift).

## Follow-up fix included (`fix(cli)`)

The `bin` switch to the supervisor meant a bare `cc-channel-octo` / `npx @mininglamp-oss/cc-channel-octo` returned usage+exit 2 instead of starting the gateway — contradicting the README/CHANGELOG. Restored the documented behavior: an empty command now runs `cmdStart(foreground)`. Daemon-style management stays opt-in via `start`/`stop`/`restart`/`status`. CHANGELOG `[Unreleased]` updated.

## Verification

- `npm run build` (tsc) — clean
- `npm test` (vitest) — **1002/1002 passed** (51 files), incl. `cli.test.ts`
- `npm run lint` (eslint `--max-warnings 0`) — clean
- codex review — merge correctness + cli fix confirmed, no blockers

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,20 @@
+# Release Drafter config — https://github.com/release-drafter/release-drafter
+name-template: "$NEXT_PATCH_VERSION"
+tag-template: "v$NEXT_PATCH_VERSION"
+template: |
+  ## What's Changed
+
+  $CHANGES
+
+  **Full Changelog**: https://github.com/$OWNER/$REPOSITORY/compare/$PREVIOUS_TAG...v$RESOLVED_VERSION
+categories:
+  - title: "Features"
+    labels: ["type:feature"]
+  - title: "Bug Fixes"
+    labels: ["type:bug"]
+  - title: "CI and Infrastructure"
+    labels: ["type:chore", dependencies-changed, dependencies]
+  - title: "Documentation"
+    labels: ["type:docs"]
+exclude-labels: [skip-changelog]
+change-template: "- $TITLE @$AUTHOR (#$NUMBER)"
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,94 @@
+name: npm Publish
+
+# Publishes the package to the public npm registry.
+#
+# Trigger: when a GitHub Release is published (the release-publish workflow flips
+# the draft to published, which fires `release: published`), or manually via
+# workflow_dispatch with an explicit tag.
+#
+# Auth: uses the NPM_TOKEN repository secret (an npm automation/granular token
+# with publish rights on the @mininglamp-oss scope). Provenance is attached via
+# OIDC (`id-token: write` + publishConfig.provenance:true in package.json), so
+# the published version carries a verifiable build-origin attestation.
+#
+# Safety: the job refuses to publish unless the tag's `vX.Y.Z` matches the
+# version in package.json, so a mismatched tag can never push a wrong version.
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to publish (e.g. v1.0.0). Must be an existing tag."
+        required: true
+        type: string
+
+permissions: {}
+
+jobs:
+  publish:
+    name: Publish to npm
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      id-token: write # required for npm provenance (OIDC)
+    steps:
+      - name: Resolve tag
+        id: tag
+        env:
+          EVENT_TAG: ${{ github.event.release.tag_name }}
+          INPUT_TAG: ${{ inputs.tag }}
+        run: |
+          set -euo pipefail
+          TAG="${EVENT_TAG:-$INPUT_TAG}"
+          if [ -z "$TAG" ]; then
+            echo "::error::No tag resolved from release event or dispatch input."
+            exit 1
+          fi
+          # Strict semver tag (mirrors the org release-publish regex).
+          if ! printf '%s' "$TAG" | grep -Eq \
+            '^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*)(\.(0|[1-9][0-9]*|[0-9]*[A-Za-z-][0-9A-Za-z-]*))*))?([+][0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$'; then
+            echo "::error::Tag '$TAG' is not strict semver (vMAJOR.MINOR.PATCH[-prerelease])"
+            exit 1
+          fi
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+          echo "Resolved tag: $TAG"
+
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          ref: ${{ steps.tag.outputs.tag }}
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: '22'
+          cache: 'npm'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Verify tag matches package.json version
+        env:
+          TAG: ${{ steps.tag.outputs.tag }}
+        run: |
+          set -euo pipefail
+          PKG_VERSION=$(node -p "require('./package.json').version")
+          EXPECTED="v${PKG_VERSION}"
+          if [ "$TAG" != "$EXPECTED" ]; then
+            echo "::error::Tag '$TAG' does not match package.json version '$EXPECTED'. Refusing to publish."
+            exit 1
+          fi
+          echo "Tag matches package.json: $EXPECTED"
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build
+        run: npm run build
+
+      - name: Test
+        run: npm test
+
+      - name: Publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -2,21 +2,28 @@ name: npm Publish
 
 # Publishes the package to the public npm registry.
 #
-# Trigger: when a GitHub Release is published (the release-publish workflow flips
-# the draft to published, which fires `release: published`), or manually via
-# workflow_dispatch with an explicit tag.
+# Trigger: on a pushed stable version tag `vX.Y.Z` (NOT pre-releases like
+# `vX.Y.Z-rc.1`), or manually via workflow_dispatch with an explicit tag.
+#
+# Why tag-push instead of `release: published`: the release-publish workflow
+# creates the GitHub Release with the built-in GITHUB_TOKEN, and GitHub does not
+# cascade workflow triggers from GITHUB_TOKEN-produced events (anti-recursion),
+# so a `release: published` trigger never fires automatically. A tag push is not
+# subject to that block.
 #
 # Auth: uses the NPM_TOKEN repository secret (an npm automation/granular token
 # with publish rights on the @mininglamp-oss scope). Provenance is attached via
 # OIDC (`id-token: write` + publishConfig.provenance:true in package.json), so
 # the published version carries a verifiable build-origin attestation.
 #
-# Safety: the job refuses to publish unless the tag's `vX.Y.Z` matches the
-# version in package.json, so a mismatched tag can never push a wrong version.
+# Safety: the job refuses to publish unless (a) the tag is strict semver, (b) the
+# tagged commit is an ancestor of origin/main (no publishing off-branch code),
+# and (c) the tag's `vX.Y.Z` matches the version in package.json.
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - 'v*.*.*'
   workflow_dispatch:
     inputs:
       tag:
@@ -38,13 +45,13 @@ jobs:
       - name: Resolve tag
         id: tag
         env:
-          EVENT_TAG: ${{ github.event.release.tag_name }}
+          REF_TAG: ${{ github.ref_type == 'tag' && github.ref_name || '' }}
           INPUT_TAG: ${{ inputs.tag }}
         run: |
           set -euo pipefail
-          TAG="${EVENT_TAG:-$INPUT_TAG}"
+          TAG="${REF_TAG:-$INPUT_TAG}"
           if [ -z "$TAG" ]; then
-            echo "::error::No tag resolved from release event or dispatch input."
+            echo "::error::No tag resolved from tag push or dispatch input."
             exit 1
           fi
           # Strict semver tag (mirrors the org release-publish regex).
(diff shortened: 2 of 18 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#146 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo #146 — `chore: merge main into feat/agent-runtime`
**Head SHA:** `9c25b4455270123c93ff06b08056f6ae07a9cb0b`
**Base:** `feat/agent-runtime` ← `main`
**Files:** 18 changed (+830 / −116)
**Author:** caster-Q

---

## 1. Summary

This is a merge-PR bringing 8 commits from `main` into `feat/agent-runtime`, plus two PR-specific commits:

1. **Conflict resolution** in `package.json` (only conflicted file) — kept runtime branch's `bin` → `dist/cli.js`, took main's `version` 1.0.1 and publish metadata.
2. **fix(cli)** — bare `cc-channel-octo` invocation now runs the gateway in foreground (backward compat with pre-supervisor behavior documented in README/CHANGELOG).

The main-branch commits include: CI release plane (#130), npm packaging (#137/#138/#140), dispatch timeout (#142), and authoritative mention validation (#145, reworking the reverted #144).

---

## 2. Verification Conclusions

| Area | Status | Evidence |
|------|--------|----------|
| Conflict resolution (package.json) | ✅ | `bin` → `dist/cli.js` (runtime), `version` → `1.0.1` (main), publish fields from main. Lockfile regenerated. Consistent. |
| fix(cli): bare invocation | ✅ | `src/cli.ts:233` — `case ''` now calls `cmdStart(paths, true)`. Verified `cmdStart(paths: SupervisorPaths, foreground: boolean): Promise<number>` at `cli.ts:110`. `true` = foreground. Matches documented behavior. |
| Dispatch timeout (#141) | ✅ | `session-router.ts:153-199` — `runHandlerWithTimeout` races handler vs per-invocation `timeoutError` (reference equality, not string match). Lock releases via `withSessionLock` finally block (`session-router.ts:100-105`). Orphaned handler gets no-op `.catch()`. `replySafe` at line 559 swallows its own errors. Timer cleared in `finally`. |
| Mention validation (#143/#145) | ✅ | `group-context.ts:206-282` — `refreshMembers` builds `present` Set, prunes departed members from memory + DB + robotFlags. Empty-roster guard (`present.size > 0`) prevents mass-prune on transient API quirks. `isMember()` / `getNameToUidMap()` getters exposed. |
| Mention wiring (index.ts → stream-relay → mention-utils) | ✅ | `index.ts:791-795` constructs `isValidMentionUid` predicate (groups only, DMs skip). Passed through `stream-relay.ts:168` → `mention-utils.ts:235-237` filters entities. v1 `@name` entities inherently members (come from memberMap), not re-checked. |
| isMainModule guard (index.ts) | ✅ | `index.ts:1007-1016` — `realpathSync` both sides for symlink-aware comparison. Fallback to direct href if realpath fails. Shebang added at line 1. |
| CI workflows | ✅ | `npm-publish.yml` — pinned action SHAs, strict semver validation, ancestor-of-main check, tag↔package.json version match, OIDC provenance. `release-drafter.yml` / `release-publish.yml` — reusable workflows from org repo. |
| Tests | ✅ | 18 new tests: 4 group-context (prune memory, prune DB, empty-roster guard, getter basics), 4 mention-utils (hit/miss/mixed/legacy), 5 session-router (hung handler, apology, fast handler, disabled timeout, rejecting handler), plus lockfile/package consistency. |

---

## 3. Findings

### No P0 or P1 findings.

### P2 Findings (non-blocking)

**P2-1: `getNameToUidMap()` returns live internal Map reference**
- **File:** `src/group-context.ts:466-468`
- **Diff-scope:** new (introduced by this PR's #145)
- **Detail:** The getter returns the live internal `Map`, not a defensive copy. Callers could accidentally mutate the member-name→uid mapping. JSDoc says "Read-only use by callers" but this is not enforced.
- **Risk:** Low — the only production caller (`index.ts:791`) passes it through to `deliver` → `resolveMentions`, which only iterates. No mutation path exists today.
- **Suggestion:** Consider returning `new Map(this.getNameToUid(channelId))` for defensive isolation, or document the contract more strongly.

**P2-2: Orphaned handler rejection silently swallowed**
- **File:** `src/session-router.ts:175`
- **Diff-scope:** new (introduced by this PR's #142)
- **Detail:** `handlerPromise.catch(() => { /* swallow late rejection after timeout */ })` — after a timeout, if the orphaned handler eventually rejects, the error is silently discarded. This is intentional defense-in-depth (the comment explains it), but could mask real bugs in future handler implementations.
- **Risk:** Low — today's caller (`index.ts`) is fully try/caught internally, so late rejections are unlikely. The swallow is a safety net, not the primary error path.
- **Suggestion:** Consider logging the swallowed error at `debug` level for future diagnosability: `.catch((err) => { console.debug('late handler rejection swallowed:', err); })`.

**P2-3: CHANGELOG reorganization moves large blocks**
- **File:** `CHANGELOG.md` (+137/−93)
- **Diff-scope:** new (restructuring for v1.0.0/v1.0.1 release sections)
- **Detail:** The `[Unreleased]` section's Security/Changed/Removed/Fixed entries are moved under `[1.0.0]`, and new `[Unreleased]` / `[1.0.1]` sections are added. Content is accurate but the large move could cause merge conflicts with other branches editing CHANGELOG.
- **Risk:** Low — operational concern, not a code defect.

---

## 4. Data-Flow Tracing

### Dispatch timeout chain
```
config.dispatchTimeoutMs (default 300_000ms)
  → mergeConfig: override.dispatchTimeoutMs ?? base.dispatchTimeoutMs
  → SessionRouter constructor stores config
  → routeAndHandle → withSessionLock(key, fn)
    → fn: processMessage → if shouldProcess → runHandlerWithTimeout(result, handler)
      → Promise.race([handlerPromise, timeoutPromise])
      → timeout: reject(timeoutError) → catch identifies by reference equality
        → replySafe(msg, '⚠️ 处理超时') → sendMessage (time-bounded in api.ts)
        → return (lock releases via withSessionLock finally → resolveGate())
      → handler wins: normal completion, clearTimeout in finally
```
**Verdict:** Data flows correctly. Timeout releases lock. Apology is bounded. No hang path.

### Mention validation chain
```
getGroupMembers(apiUrl, botToken, channelId) → MemberRow[]
  → refreshMembers: build `present` Set from m.uid (skip !uid/!name)
  → upsert present members (existing logic)
  → prune: for uid in [...memberMap.keys()]: if !present.has(uid) → delete from memberMap, nameMap, robotFlags, DB
  → isMember(channelId, uid) → memberMapByChannel.get(channelId)?.has(uid) ?? false
  → index.ts: isValidMentionUid = (uid) => groupContext.isMember(channelId, uid) [groups only]
  → stream-relay.deliver(... memberMap, isValidUid)
    → resolveMentions(accumulated, memberMap, isValidUid)
      → v2 entities: isValidUid ? entities.filter(e => isValidUid(e.uid)) : [...entities]
      → v1 entities: from memberMap (inherently members)
```
**Verdict:** Data flows correctly. Hallucinated uids are filtered. Empty-roster guard prevents mass-prune. DB pruning uses dedicated `deleteMember` prepared statement.

### CLI bare invocation
```
process.argv → parseArgs(argv) → { cmd: '', foreground: false, ... }
  → run(): case '' → cmdStart(paths, true)
    → cmdStart(paths, foreground=true): spawns gateway in foreground
```
**Verdict:** Correct. `true` maps to `foreground` parameter. Matches pre-supervisor behavior.

### isMainModule guard
```
process.argv[1] → realpathSync → pathToFileURL → href
import.meta.url → realpathSync(fileURLToPath(...)) → pathToFileURL → href
  → compare (symlink-aware)
  → fallback: direct href comparison if realpath throws
```
**Verdict:** Correct. Handles symlink under `node_modules/.bin/`. Fallback is safe.

---

## 5. R5 Blindspot Checklist

### C1 — Dual-path parity: CLEAR
- **upsert ↔ delete members:** `upsertMember` (INSERT ON CONFLICT UPDATE) and `deleteMember` (DELETE) both use `(group_id, uid)` key. Consistent.
- **v1 ↔ v2 mention paths:** v2 uses `isValidUid` predicate; v1 entities come from `memberMap` (inherently members). No re-check needed for v1 — correct asymmetry.
- **Lock acquire ↔ release:** `withSessionLock` creates gate → fn runs → finally resolves gate. `runHandlerWithTimeout` always returns (timeout or completion), so fn always returns, so finally always fires. Symmetric.

### C2 — Control-flow ordering / nesting: CLEAR
- `resolveMentions` called once per `deliver` cycle. No nested/double invocation.
- `runHandlerWithTimeout` called once per `routeAndHandle`. No nesting.
- `refreshMembers` pruning runs once per successful API fetch. `present` Set built before iteration; pruning iterates a snapshot (`[...memberMap.keys()]`). No ordering issue.
- Safety controls tested: `isValidUid('')` → `memberMap.has('')` → false (correct). `isMainModule` with relative/absolute/symlink paths all handled.

### C3 — Authorization boundary ≠ capability boundary: CLEAR
- Mention validation is defense-in-depth (server enforces real permissions). Not an auth boundary.
- Dispatch timeout is an operational safeguard. Not an auth boundary.
- Member pruning is a local cache concern. Not an auth boundary.
- No new endpoints, tools, or credential exposure in this PR.

---

## 6. Extra Observations

- **Test quality:** The 18 new tests cover the critical paths well. The "rejecting handler does not wedge the session" test is particularly valuable — it verifies that non-timeout errors also release the lock correctly. The "empty roster does NOT mass-prune" test directly validates Jerry-Xin's original red-flag concern from #144.
- **Merge cleanliness:** Only `package.json` conflicted (src code had zero overlap between branches). The resolution is correct and well-documented in the PR body.
- **npm publish workflow on feat/agent-runtime:** The `npm-publish.yml` ancestor-of-main check won't fire until this branch merges to main and a tag is pushed — correct behavior, no premature trigger risk.

---

## 7. Verdict

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The merge is clean, conflict resolution is correct, the CLI backward-compat fix is sound, and all code brought from main (dispatch timeout, mention validation, CI release plane) has been thoroughly traced and verified. Three P2 notes are non-blocking defensive-coding suggestions.

Jerry-Xin (approved):
Merge correctness verified — APPROVE.

This is a catch-up merge of `main` into the runtime branch, so the review scope is merge fidelity: no accidental reverts, drops, or corrupted conflict resolution.

**✅ Merge fidelity (verified with `git diff main:<file> HEAD:<file>`)**
- Every substantive source file pulled from `main` is **byte-identical to `main`**: `src/mention-utils.ts`, `src/session-router.ts`, `src/group-context.ts`, `src/config.ts`, `src/index.ts`, `src/stream-relay.ts`, and all three updated test files — `diff` is empty. The merge reproduced `main` with zero drift, no reverts, no dropped commits.
- The only files differing from `main` are the runtime-branch-unique additions: `src/cli.ts`, `src/__tests__/cli.test.ts`, `package.json`, `package-lock.json`, and the `CHANGELOG.md [Unreleased]` entry. That is exactly the expected delta.

**✅ Conflict resolution (`package.json`)**
- `version` → `1.0.1` (took `main`'s; correctly avoids regressing the already-published release).
- `bin` → `dist/cli.js` (kept the runtime branch's process-supervisor entry).
- Publish metadata (`files`/`repository`/`homepage`/`bugs`/`publishConfig`) kept from `main`. Matches the documented resolution.

**✅ Net-new follow-up fix (`fix(cli)`)**
- Bare `cc-channel-octo` / `npx …` now calls `cmdStart(paths, foreground)` instead of printing usage + exit 2, restoring the README/CHANGELOG-documented foreground-start behavior after the `bin` switch to the supervisor. Daemon management stays opt-in via `start`/`stop`/`restart`/`status`. Clean and well-scoped; covered by `cli.test.ts`.

**CI**: all required checks green (sanity / actionlint / PR-title-lint / labeler / add-to-project). Mergeable, no conflicts.

**Note on out-of-scope findings**: a deeper scan surfaced two pre-existing concerns in code that this PR merely brings over unchanged from `main` (both files are byte-identical to `main`): (1) in `resolveMentions`, a downgraded hallucinated `@[fake:Name]` can be re-resolved by the `@name` fallback when `Name` matches a real member — the dropped-entity offset isn't excluded from the fallback's `existingOffsets`; (2) the per-message dispatch timeout releases the session lock without cancelling the original handler. Neither was introduced by this merge, so they are not blockers here — better tracked as separate follow-up issues against `main`.

yujiawei (approved):
# Code Review — PR #146 (cc-channel-octo)

Reviewed at head `9c25b4455270123c93ff06b08056f6ae07a9cb0b`. This is a **merge PR**: it brings the released `main` (v1.0.0/v1.0.1 + #142, #144/#145) into the divergent `feat/agent-runtime` branch, plus one net-new `fix(cli)` commit.

## 1. Verification

| Check | Result |
|---|---|
| `main` fully merged (ancestor of head) | ✅ `git merge-base --is-ancestor origin/main HEAD` passes; no commits in `main` missing from head |
| Merge conflict scope | ✅ Only `package.json` conflicted (src had zero overlap); resolved correctly — `bin → dist/cli.js`, `version → 1.0.1` (no regression), publish fields kept |
| No leftover conflict markers | ✅ none in `src/` or `package.json` |
| `npm run build` (tsc) | ✅ clean |
| `npm run lint` (eslint `--max-warnings 0`) | ✅ clean |
| `npm test` (vitest) | ✅ **1002/1002 passed** (51 files), incl. new `cli`, `session-router` timeout, `group-context` prune, `mention-utils` validation tests |
| All referenced symbols resolve cross-file | ✅ `RouteResult.sessionKey`, `replySafe`, `getNameToUidMap`, `isMember`, `deliver(... isValidUid)` all wired correctly |

The merge itself is clean and the resolution is correct. **No merge regression** — every finding below is in code that either already shipped on `main` or is the small net-new CLI change. Nothing rises to a blocking severity for this merge.

## 2. Findings

### P2 — Dispatch timeout can let two turns touch the same session state (`src/session-router.ts:176`)

```ts
const handlerPromise = handler(result);
...
await Promise.race([handlerPromise, timeoutPromise]);
...
return; // lock releases, queue advances
```

On timeout the per-session lock is released while the original handler keeps running. That handler is the full `handleMessage` pipeline (SDK resume, streaming, `store.appendAssistant`, `store.setSdkSessionId`). A next same-session message can then run concurrently with the orphaned first turn, so history / session-id writes can interleave out of order — a stronger failure mode than the "late duplicate reply" the inline comment describes.

Mitigating: this is **already-released behavior from #142**, not introduced here; default `dispatchTimeoutMs` is 5 min so it only triggers on a genuinely wedged turn; and the design intentionally mirrors the documented "don't cancel the in-flight turn, just unblock the queue" tradeoff. **Recommended follow-up (not a merge blocker):** thread an `AbortSignal` into the handler so the orphaned turn is actually cancelled, or guard the post-timeout writes against a newer turn having started.

### P2 — Invalid structured mention can be re-resolved by the v1 fallback (`src/mention-utils.ts:232-240`)

```ts
entities = isValidUid ? converted.entities.filter((e) => isValidUid(e.uid)) : [...converted.entities];
...
// later: v1 @name fallback over memberMap, skipping only offsets already in `entities`
```

When `@[badUid:Alice]` fails `isValidUid`, its entity is dropped and `finalContent` becomes `@Alice`. Because the production caller passes **both** `memberMap` and `isValidUid`, the v1 fallback can then re-add an entity at that same offset if a *current* member is named `Alice`. The documented guarantee is "downgrade to plain text, no entity"; in practice it resolves to the real current `Alice`'s uid. Effect is benign-to-mild (it notifies the correctly-named member rather than a hallucinated uid), but it deviates from the stated contract.

Mitigating: **pre-existing from #145**, requires a name collision between a hallucinated mention and a real member. **Follow-up:** record the dropped offsets and exclude them from the fallback pass to honor the documented behavior, and add a test for "invalid structured uid + populated memberMap".

### P2 — Member upsert/prune writes are not batched in a transaction (`src/group-context.ts:223-278`)

Both the upsert loop and the new prune loop call `this.upsertMember.run(...)` / `this.deleteMember.run(...)` per row. With `better-sqlite3` auto-committing each statement, a large roster means one `fsync` per member, blocking the event loop. WAL mode (`journal_mode=WAL`, `busy_timeout=5000`) softens this and typical group sizes are small, so it is not a correctness issue — but wrapping the per-channel mutations in `this.adapter.transaction(...)` (the adapter already exposes it) would remove the stall on big groups. The upsert loop is pre-existing; the prune loop is new in #145.

### P2 — Empty-roster response still suppresses retry for the refresh interval (`src/group-context.ts:201-206, 261`)

`lastRefresh` is set on any successful fetch, including an empty `[]` that the code itself treats as a likely transient/malformed response and refuses to prune on. Net effect: stale members remain accepted by `isMember()` for the full `REFRESH_INTERVAL_MS` (1h) instead of being retried on the next message. Pre-existing from #145; low impact given the guard's intent.

### P2 — Bare CLI "foreground" forks a child rather than exec-ing in-process (`src/cli.ts:111-118`) — **net-new in this PR**

```ts
const child = spawn(process.execPath, [paths.indexEntry], { stdio: 'inherit', env: process.env });
return new Promise<number>((resolve) => { child.on('exit', (code) => resolve(code ?? 0)); });
```

Bare `cc-channel-octo` now forks the gateway as a child. Two minor consequences vs. the old `dist/index.js` direct-foreground behavior:
1. **Signal forwarding:** a `SIGTERM` to the wrapper PID (e.g. from a container/supervisor) is not relayed to the child, which can be orphaned. Interactive `Ctrl+C` (SIGINT to the process group) still reaches both, so the common `npx` case is fine; the container-stop case is the gap.
2. **Exit code:** `code ?? 0` reports a signal-terminated child (where `code` is `null`) as success `0`.

**Follow-up:** forward `SIGINT`/`SIGTERM` to the child and propagate `128 + signal` on signal exit. Minor, since the documented production path is the daemon `start`/`stop` supervisor, not bare foreground.

## 3. Recommendations

- This merge is safe to land as-is. The findings above are all **P2 follow-ups** and, except the CLI fork behavior, live in code that already shipped on `main` — they are not regressions from this merge and should be tracked as separate issues against the relevant features (#142, #145).
- Highest-value follow-up: the dispatch-timeout session-state concern (P2) — an `AbortSignal` or a post-timeout write-guard would close the only finding with a path to real data corruption.

## 4. Additional notes (coverage)

- Verified by building/linting/testing the actual head checkout, not just reading the diff; all green.
- **Could not verify:** the live `getGroupMembers` server response shape (empty-`[]`-on-malformed assumption is taken from the code comment); the Claude Agent SDK's internal thread-safety under the concurrent-turn scenario in the dispatch-timeout finding; and end-to-end signal behavior of the bare CLI under a real container stop.
- Test coverage gaps worth adding: invalid structured mention with a populated `memberMap`; a timed-out handler writing session state after the next turn starts; foreground CLI signal propagation / signal exit codes.

PR #148, opened 2026-06-17
feat: add upgrade subcommand and resilient multi-bot startup

caster-Q:
## Summary
Brings the cc-channel-octo gateway up to parity with the openclaw plugin for the cc-octo runtime line: a daemon-drivable self-update path, and multi-bot startup that no longer fails the whole gateway when one bot is bad.

## Changes
- **`upgrade` subcommand** — `cc-channel-octo upgrade [<version>]` runs `npm install -g @mininglamp-oss/cc-channel-octo@<version|latest>` then restarts the gateway. Version string is whitelisted to guard against argument injection.
- **Resilient multi-bot startup** — in multi-bot mode a bot that fails to register/connect is now skipped (logged) instead of tearing down every bot; startup is only fatal when no bot comes up. Single-bot keeps fail-fast.
- **Partial-startup cleanup** — `startBot` releases its own durable resources (cwd-cleanup timer, sqlite handle, gateway lock) if it throws after acquiring them, so a skipped bot leaks nothing.

## Test plan
- `npm run build` / `npm run lint` clean
- `npm test` — full suite green (incl. new `cli-upgrade` and `multibot-resilience` unit tests)
- Local: gateway runs two bots (different servers) concurrently; both online; injection guard rejects unsafe version

Part of the cc-octo parity work; sibling PRs in octo-daemon-cli / octo-fleet / octo-web.

### DIFF
--- a/src/__tests__/cli-upgrade.test.ts
+++ b/src/__tests__/cli-upgrade.test.ts
@@ -1,0 +1,42 @@
+import { describe, it, expect } from 'vitest';
+import { buildUpgradeArgs } from '../cli.js';
+
+describe('buildUpgradeArgs', () => {
+  it('builds npm args for an explicit version', () => {
+    expect(buildUpgradeArgs('1.2.3')).toEqual([
+      'install',
+      '-g',
+      '@mininglamp-oss/cc-channel-octo@1.2.3',
+    ]);
+  });
+
+  it('falls back to @latest when no version is given', () => {
+    expect(buildUpgradeArgs()).toEqual([
+      'install',
+      '-g',
+      '@mininglamp-oss/cc-channel-octo@latest',
+    ]);
+  });
+
+  it('treats blank/whitespace version as latest', () => {
+    expect(buildUpgradeArgs('   ')).toEqual([
+      'install',
+      '-g',
+      '@mininglamp-oss/cc-channel-octo@latest',
+    ]);
+  });
+
+  it('accepts a prerelease/build-metadata semver', () => {
+    expect(buildUpgradeArgs('1.2.3-rc.1+build5')).toEqual([
+      'install',
+      '-g',
+      '@mininglamp-oss/cc-channel-octo@1.2.3-rc.1+build5',
+    ]);
+  });
+
+  it('rejects a version with shell metacharacters (injection guard)', () => {
+    expect(() => buildUpgradeArgs('1.2.3; rm -rf /')).toThrow();
+    expect(() => buildUpgradeArgs('$(whoami)')).toThrow();
+    expect(() => buildUpgradeArgs('1.2.3 && curl evil')).toThrow();
+  });
+});
--- a/src/__tests__/multibot-resilience.test.ts
+++ b/src/__tests__/multibot-resilience.test.ts
@@ -1,0 +1,59 @@
+import { describe, it, expect } from 'vitest';
+import { partitionStartResults, type BotStack } from '../index.js';
+
+// A minimal BotStack-shaped stub; partitionStartResults only ever moves the
+// fulfilled value through, it never calls into it. Cast through unknown to keep
+// the test free of `any` (the repo lints with --max-warnings 0).
+function fakeStack(id: string): BotStack {
+  return {
+    botId: id,
+    connect: async () => {},
+    shutdown: async () => {},
+  } as unknown as BotStack;
+}
+
+describe('partitionStartResults', () => {
+  const configs = [{ botId: 'a' }, { botId: 'b' }, { botId: 'c' }];
+
+  it('all fulfilled → every stack, no failures', () => {
+    const results: PromiseSettledResult<BotStack>[] = [
+      { status: 'fulfilled', value: fakeStack('a') },
+      { status: 'fulfilled', value: fakeStack('b') },
+      { status: 'fulfilled', value: fakeStack('c') },
+    ];
+    const { stacks, failures } = partitionStartResults(results, configs);
+    expect(stacks.map((s) => s.botId)).toEqual(['a', 'b', 'c']);
+    expect(failures).toEqual([]);
+  });
+
+  it('partial failure → keeps successful stacks, records failures (does not throw)', () => {
+    const err = new Error('401 bad token');
+    const results: PromiseSettledResult<BotStack>[] = [
+      { status: 'fulfilled', value: fakeStack('a') },
+      { status: 'rejected', reason: err },
+      { status: 'fulfilled', value: fakeStack('c') },
+    ];
+    const { stacks, failures } = partitionStartResults(results, configs);
+    expect(stacks.map((s) => s.botId)).toEqual(['a', 'c']);
+    expect(failures).toEqual([{ id: 'b', reason: err }]);
+  });
+
+  it('all failed → no stacks, every failure recorded (caller decides fatal)', () => {
+    const results: PromiseSettledResult<BotStack>[] = [
+      { status: 'rejected', reason: new Error('e1') },
+      { status: 'rejected', reason: new Error('e2') },
+      { status: 'rejected', reason: new Error('e3') },
+    ];
+    const { stacks, failures } = partitionStartResults(results, configs);
+    expect(stacks).toEqual([]);
+    expect(failures.map((f) => f.id)).toEqual(['a', 'b', 'c']);
+  });
+
+  it('falls back to a positional id when botId is missing', () => {
+    const results: PromiseSettledResult<BotStack>[] = [
+      { status: 'rejected', reason: new Error('e') },
+    ];
+    const { failures } = partitionStartResults(results, [{}]);
+    expect(failures[0].id).toBe('#0');
+  });
+});
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -50,6 +50,8 @@ export interface ParsedArgs {
   cmd: string;
   foreground: boolean;
   timeoutMs: number;
+  /** Positional version target for `upgrade` (e.g. `upgrade 1.2.3`); undefined → latest. */
+  version?: string;
 }
 
 /**
@@ -87,15 +89,19 @@ export function parseArgs(argv: string[]): ParsedArgs {
   const [cmd = '', ...rest] = argv;
   let foreground = false;
   let timeoutSec = 10;
+  let version: string | undefined;
   for (const a of rest) {
     if (a === '--foreground' || a === '-f') {
       foreground = true;
     } else if (a.startsWith('--timeout=')) {
       const n = Number.parseInt(a.slice('--timeout='.length), 10);
       if (Number.isFinite(n) && n > 0) timeoutSec = n;
+    } else if (!a.startsWith('-') && version === undefined) {
+      // First positional token (e.g. `upgrade 1.2.3`).
+      version = a;
     }
   }
-  return { cmd, foreground, timeoutMs: timeoutSec * 1000 };
+  return { cmd, foreground, timeoutMs: timeoutSec * 1000, version };
 }
 
 /**
@@ -228,6 +234,60 @@ function cmdStatus(paths: SupervisorPaths): number {
   return 0;
 }
 
+/** npm package name installed globally for the gateway. */
+const NPM_PKG = '@mininglamp-oss/cc-channel-octo';
+/**
+ * Semantic-version whitelist. The version reaches us from the daemon → fleet
+ * upgrade order (an untrusted boundary) and is interpolated into an npm spec,
+ * so reject anything outside `[0-9A-Za-z.-+]` to prevent argument/shell
+ * injection even though we spawn npm without a shell.
+ */
+const VERSION_RE = /^[0-9A-Za-z.\-+]+$/;
+
+/**
+ * Build the `npm install -g <pkg>@<version>` argument vector. A blank/omitted
+ * version installs `@latest`. Pure (no I/O) so the injection guard is unit
+ * testable. Throws on an unsafe version string.
+ */
+export function buildUpgradeArgs(version?: string): string[] {
+  const v = version && version.trim() ? version.trim() : 'latest';
+  if (v !== 'latest' && !VERSION_RE.test(v)) {
+    throw new Error(`unsafe version: ${v}`);
+  }
+  return ['install', '-g', `${NPM_PKG}@${v}`];
+}
+
+/**
+ * Self-update: `npm install -g @mininglamp-oss/cc-channel-octo@<version>` then
+ * restart the gateway so the new code is live. Invoked by the daemon to drive
+ * a fleet upgrade order (mirrors openclaw's daemon-driven plugin install).
+ */
+async function cmdUpgrade(paths: SupervisorPaths, timeoutMs: number, version?: string): Promise<number> {
+  let args: string[];
+  try {
+    args = buildUpgradeArgs(version);
+  } catch (err) {
+    console.error(`cc-channel-octo: ${(err as Error).message}`);
+    return 2;
+  }
+  console.log(`cc-channel-octo: upgrading via npm ${args.join(' ')}`);
+  const code = await new Promise<number>((resolve) => {
+    const child = spawn('npm', args, { stdio: 'inherit', env: process.env });
+    child.on('error', (err) => {
+      console.error(`cc-channel-octo: failed to spawn npm: ${err.message}`);
+      resolve(1);
+    });
+    child.on('exit', (c) => resolve(c ?? 1));
+  });
+  if (code !== 0) {
+    console.error(`cc-channel-octo: npm install failed (exit ${code})`);
+    return code;
+  }
+  // Restart so the freshly installed code is running.
+  await cmdStop(paths, timeoutMs);
+  return cmdStart(paths, false);
+}
+
 function usage(): string {
   return `cc-channel-octo ${readVersion()} — gateway process supervisor
 
@@ -236,6 +296,7 @@ Usage:
   cc-channel-octo stop [--timeout=<s>]   gracefully stop (SIGTERM, then SIGKILL)
   cc-channel-octo restart                stop (if running) then start
   cc-channel-octo status                 show running state
+  cc-channel-octo upgrade [<version>]    npm install -g the gateway (default latest) then restart
   cc-channel-octo version                print the version
 
 Paths (under ~/.cc-channel-octo):
@@ -246,7 +307,7 @@ POSIX only (macOS/Linux). On Windows, run under a service manager.`;
 }
 
 export async function run(argv: string[], baseDir?: string): Promise<number> {
-  const { cmd, foreground, timeoutMs } = parseArgs(argv);
+  const { cmd, foreground, timeoutMs, version } = parseArgs(argv);
   const paths = resolveSupervisorPaths(baseDir);
   switch (cmd) {
     case 'start':
@@ -258,6 +319,8 @@ export async function run(argv: string[], baseDir?: string): Promise<number> {
       return cmdStart(paths, false);
     case 'status':
       return cmdStatus(paths);
+    case 'upgrade':
+      return cmdUpgrade(paths, timeoutMs, version);
     case 'version':
     case '--version':
     case '-v':
--- a/src/gateway.ts
+++ b/src/gateway.ts
@@ -269,7 +269,12 @@ export class OctoGateway {
     }
   }
 
-  private releaseLock(): void {
+  /**
+   * Release the per-bot startup lock if we still hold it. Best-effort and
+   * idempotent (nonce-guarded), so it is safe to call from a partial-startup
+   * cleanup path even if connect()/services were never started.
+   */
+  releaseLock(): void {
     try {
       if (existsSync(this.lockFilePath)) {
         const content = readFileSync(this.lockFilePath, 'utf-8').trim();
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,17 +64,24 @@ async function main(): Promise<void> {
   // already succeeded must be torn down, or their locks/stores/intervals leak.
   // Promise.all would discard the resolved stacks on first rejection, so settle
   // all and clean up the successful ones before rethrowing.
+  // startBot() acquires the gateway.lock, opens the SQLite store, and arms the
+  // cwd-cleanup interval; on its own failure it releases those (see startBot's
+  // try/catch), so a rejected bot leaves nothing behind. Settle all, then apply
+  // the multi-bot resilience policy: skip the bots that failed, keep the ones
+  // that came up, and only treat startup as fatal when NONE start. (Single-bot
+  // mode has one entry, so "none started" == that bot failed → fatal, same as
+  // the old fail-fast behavior.) This stops one bot's bad token / held lock from
+  // taking the whole gateway — and every sibling bot — offline.
   const startResults = await Promise.allSettled(botConfigs.map((c) => startBot(c, multi)));
-  const stacks: BotStack[] = [];
-  let startError: unknown;
-  for (const r of startResults) {
-    if (r.status === 'fulfilled') stacks.push(r.value);
-    else startError = startError ?? r.reason;
+  const { stacks, failures } = partitionStartResults(startResults, botConfigs);
+  for (const f of failures) {
+    const reason = f.reason instanceof Error ? f.reason.message : String(f.reason);
+    console.error(`[cc-channel-octo] bot "${f.id}" failed to start, skipping: ${reason}`);
   }
-  if (startError) {
-    console.error('[cc-channel-octo] Startup failed; cleaning up bots that did start...');
-    await Promise.allSettled(stacks.map((s) => s.shutdown()));
-    throw startError;
+  if (stacks.length === 0) {
+    // Nothing came up — surface the first failure and exit (process supervisor
+    // / operator restarts once the underlying cause is fixed).
+    throw failures[0]?.reason ?? new Error('no bots started');
   }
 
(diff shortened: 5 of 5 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#148 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo#148
**Head SHA:** `7c4687aa05ef4a26f87a86287a7f20170b483ec2`
**Title:** feat: add upgrade subcommand and resilient multi-bot startup
**Files:** 5 changed (+397 / -157)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| `upgrade` subcommand — injection guard | ✅ | `VERSION_RE` whitelist (`cli.ts:244`), `spawn` without shell (`cli.ts:268`) |
| `upgrade` subcommand — restart flow | ✅ | `cmdStop` → `cmdStart` (`cli.ts:280-281`), PID file cleaned by stop |
| Multi-bot resilience — partition logic | ✅ | `partitionStartResults` is pure, no I/O (`index.ts:148-163`) |
| Multi-bot resilience — connect phase | ✅ | Per-bot try/catch with shutdown on failure (`index.ts:113-123`) |
| startBot resource cleanup on failure | ✅ | try/catch releases timer, scheduler, lock, store (`index.ts:349-360`) |
| `releaseLock()` visibility change | ✅ | Nonce-guarded, idempotent (`gateway.ts:272-286`) |
| Single-bot backward compatibility | ✅ | 1-entry array → "none started" = fatal, same as old fail-fast |
| Test coverage | ✅ | `cli-upgrade.test.ts` (5 cases), `multibot-resilience.test.ts` (4 cases) |

## 2. Findings

### No P0/P1 findings.

### P2 — Minor observations (non-blocking)

**P2-1: `parseArgs` captures positional `version` for all commands**
- **File:** `cli.ts:99-101`
- The positional-argument capture (`!a.startsWith('-') && version === undefined`) runs for every command, not just `upgrade`. Running `cc-channel-octo start someJunk` silently captures `someJunk` as `version` (unused by `start`). Harmless — only `cmdUpgrade` reads `version` — but slightly surprising.
- **Suggestion:** Consider gating the positional capture on `cmd === 'upgrade'` or logging a warning for unexpected positional args.

**P2-2: `VERSION_RE` allows non-semver strings**
- **File:** `cli.ts:244`
- **Diff-scope:** new
- The regex `/^[0-9A-Za-z.\-+]+$/` accepts strings like `abc`, `1.2.3.4.5`, or even `+` which aren't valid semver. This is by design (the regex guards against shell/argument injection, not semver validity — npm validates the specifier itself). Just noting that dist-tags like `beta`, `next` pass through intentionally.
- **No action needed** — documenting the design decision.

## 3. Suggestions

- **P2-1 above:** Gate positional arg capture on `cmd === 'upgrade'` for tighter CLI parsing hygiene. Low priority.
- Consider adding a `--dry-run` flag to `upgrade` for operator confidence in production deployments (future enhancement, not blocking).

## 4. Additional Findings

- **Lock-release double-call safety verified:** `register()` internally calls `releaseLock()` on its own failure (`gateway.ts:108`). The `startBot` catch also calls `releaseGatewayLock?.()`. Confirmed safe: `releaseLock()` is nonce-guarded — the second call finds the lock file already gone and no-ops (`gateway.ts:275-285`).
- **`gateway.stop()` safety on connect-failure shutdown:** When a bot's `connect()` throws, `shutdown()` calls `gateway.stop(activeHandlers)`. At that point no messages have been processed (socket never opened), so `activeHandlers` is empty and `this.socket` is null. `stop()` handles both gracefully (empty set drain skipped, null socket skipped).

## 5. Data-Flow Tracing

### Upgrade path
```
process.argv → parseArgs() → version (validated: no shell meta) 
  → cmdUpgrade() → buildUpgradeArgs(version) → VERSION_RE check
  → spawn('npm', args, {no shell}) → exit code
  → cmdStop(paths, timeoutMs) → SIGTERM → PID cleanup
  → cmdStart(paths, false) → new process with updated code
```
**Verdict:** Data flows correctly. Version string is sanitized before reaching spawn. No shell interpolation. ✅

### Multi-bot startup (success path)
```
resolveBotConfigs() → botConfigs[]
  → Promise.allSettled(botConfigs.map(c => startBot(c, multi)))
  → partitionStartResults(results, configs) → {stacks, failures}
  → cross-register sibling botIds via router.registerKnownBot()
  → per-bot connect() with individual try/catch
  → connected[] → shutdownAll wired to SIGINT/SIGTERM
```
**Verdict:** Data flows correctly. Failed bots' resources released inside startBot's catch. ✅

### Multi-bot startup (failure path)
```
startBot() throws → catch: clearInterval, cronScheduler.stop, releaseGatewayLock, store.close → re-throw
  → Promise.allSettled captures as rejected
  → partitionStartResults → failures[]
  → stacks.length === 0 → throw failures[0].reason (original error preserved)
```
**Verdict:** Original error object preserved through the chain (not stringified). Resource cleanup complete. ✅

### Connect-phase failure cleanup
```
s.connect() throws → s.shutdown().catch(() => {})
  → shutdown: clearInterval(cwdCleanupTimer), cronScheduler?.stop(),
    gateway.stop(activeHandlers), store.close()
```
**Verdict:** All durable resources released. The `.catch(() => {})` on shutdown is appropriate — a cleanup failure during a startup failure should not mask the original error. ✅

- **C1 — Dual-path parity:** N/A. The upgrade path is unidirectional (install → restart). No add/remove symmetry required.
- **C2 — Control-flow ordering / nested reuse:** Clear. `releaseLock()` is called from two sites (register's own catch + startBot's catch) but they are mutually exclusive (register-fail: internal release only; post-register-fail: startBot catch releases). Nonce guard prevents double-release. `VERSION_RE` is anchored (`^...$`) and rejects spaces — tested with `;`, `$()`, `&&` metacharacters.
- **C3 — Authorization boundary ≠ capability boundary:** N/A for this PR. The `upgrade` command runs with the process's own permissions (global npm install). The daemon/fleet layer is the authorization gate for who can trigger upgrades. No new network endpoints or tool exposure.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth/permission changes in this PR.

---

## Conclusion

This is a well-structured PR with clean implementation:
- **Upgrade subcommand:** Injection guard is properly designed (whitelist regex + no-shell spawn). Restart flow is correct.
- **Resilient multi-bot startup:** The resilience policy (skip failed, keep working, fatal only when none start) is well-implemented. Resource cleanup on failure is thorough and verified.
- **Tests:** Good coverage of both new features with pure-function unit tests.
- **Backward compatibility:** Single-bot mode preserves fail-fast behavior via the "none started" check.

No P0 or P1 findings. Two P2 observations (positional arg capture scope, VERSION_RE non-semver acceptance) are non-blocking design notes.

[Octo-Q] verdict: **APPROVE** — no blocking issues found.

Jerry-Xin (changes requested):
Summary: The PR is relevant to `cc-channel-octo` and the main design is sound, but the partial-startup cleanup still misses a realistic failure path in multi-bot mode.

**🔴 Blocking**
- 🔴 Critical — `startBot` can still leak the cwd cleanup timer if SQLite adapter creation fails. The timer is armed at `src/index.ts:191-196`, but `createAdapter(dbPath)` and `new SessionStore(adapter)` run at `src/index.ts:199-201` before the cleanup `try/catch` begins at `src/index.ts:219`. If `createAdapter` throws, for example due to an inaccessible `dataDir` or SQLite open failure, the rejected bot is skipped by the new `Promise.allSettled` flow while the interval remains alive for the lifetime of the surviving gateway process. This directly breaks the PR’s “skipped bot leaks nothing” guarantee. Move adapter/store creation inside the protected region, or add an outer cleanup guard that clears `cwdCleanupTimer` for all failures after it is acquired.

**💬 Non-blocking**
- 🔵 Suggestion — Add a focused test for the failure window above: mock `createAdapter` to throw after the cleanup timer is armed, start multiple bots, and assert the failed bot’s timer is cleared. The current `partitionStartResults` tests cover the policy split, but not resource cleanup.
- 🔵 Suggestion — I could not run the targeted tests locally because `vitest` is not installed in this checkout (`sh: vitest: command not found`). Please ensure CI covers `cli`, `cli-upgrade`, `multibot-resilience`, and gateway lifecycle tests.

**✅ Highlights**
- The PR passes the project relevance gate: it extends the gateway CLI and multi-bot runtime behavior in this repository.
- The `Promise.allSettled` startup policy is a good fit for resilient multi-bot startup.
- Exposing `releaseLock()` with nonce-guarded behavior is reasonable for best-effort partial-start cleanup.
- The `upgrade` command uses argument-vector spawning and a conservative version whitelist, avoiding shell injection in the npm invocation.

caster-Q:
Fixed the blocking Critical (f38183d).

**🔴 cwd cleanup timer leak on adapter creation failure** — moved `createAdapter(dbPath)` and `new SessionStore(adapter)` inside the cleanup `try` (they previously ran before it at index.ts:199-201). Now any SQLite open failure is caught and `clearInterval(cwdCleanupTimer)` runs, so a skipped bot leaks nothing. The store handle for cleanup is tracked via a hoisted `storeForCleanup` and closed (guarded) in the catch.

**🔵 Test suggestion** — kept as-is for now: `startBot` is integration-heavy (gateway REST register + sockets + sqlite), so a focused unit test would require mocking the whole stack. The leak fix is structural (the fallible creation now sits inside the already-tested `try/catch` region); `partitionStartResults` continues to cover the skip-vs-fatal policy. Happy to add an integration test if you'd prefer.

`npm run build` / `lint` clean; `npm test` 1020/1020 green.

yujiawei (approved):
# Code Review — PR #148 (cc-channel-octo)

Reviewed at head SHA `7c4687aa05ef4a26f87a86287a7f20170b483ec2` against base `feat/agent-runtime`. Verdict reflects a line-anchored read of the full post-change source of `src/index.ts`, `src/cli.ts`, and `src/gateway.ts`, cross-checked by an independent second reviewer pass.

## Verdict: APPROVE

No correctness, security, data-loss, or build-blocking issues. The two features are well-isolated and the resource-cleanup invariants hold. Findings below are all non-blocking (P2 / nit) — polish and test-coverage suggestions for a follow-up, not merge gates.

## What I verified (clean)

- **Injection guard is sound.** `buildUpgradeArgs` always interpolates the version as the single argv element `@mininglamp-oss/cc-channel-octo@<v>`, and `spawn('npm', args)` uses no shell. A leading-dash version (`-foo`) becomes `...@-foo` (not a standalone flag), and `@` / `/` / `:` / whitespace are all rejected by `VERSION_RE`, so npm-spec pivots (`pkg@npm:other`, `git+url`, `file:`, path) cannot slip through. (`src/cli.ts` `VERSION_RE` / `buildUpgradeArgs`)
- **Partial-startup cleanup in `startBot` is correct.** `cwdCleanupTimer` is declared before the `try` so the `catch` can always clear it; `cronScheduler?.stop()` is null-safe; `releaseGatewayLock?.()` is only set after gateway construction and is a no-op until `register()` actually takes the lock; `store.close()` is wrapped best-effort. No timer/lock/sqlite leak on the failure path. (`src/index.ts` `startBot`)
- **`releaseLock()` double-release is safe.** `register()` self-releases on its own failure, and `startBot`'s catch calls `releaseLock()` again — but it's `existsSync`-gated and nonce-guarded, so the second call is a no-op and it can never delete a lock another instance re-acquired after a PID-reuse race. (`src/gateway.ts` `releaseLock`)
- **`partitionStartResults`** preserves positional mapping with a `#<index>` id fallback and leaves cleanup to `startBot` — appropriate separation. The per-bot `connect()` loop correctly isolates a synchronous connect failure to that one bot and shuts it down. `shutdownAll` drains `connected` (the live set), which is correct.
- **Single-bot fatality is preserved**: a sole bot that fails still takes the process down (`connected.length === 0` → throw).

## Non-blocking findings (P2 / nit)

**1. (P2) The upgrade version path is permissive, not fail-closed.** Two independent angles on the same boundary:
   - `VERSION_RE` (`/^[0-9A-Za-z.\-+]+$/`) is a charset whitelist, not a semver validator. npm treats `pkg@<non-semver>` as a **dist-tag**, so `beta` / `next` / `canary` pass the guard and install whatever the registry has tagged there. Impact is bounded (still scoped to this package — no arbitrary-package install), but the "semantic-version whitelist" comment overstates the guarantee.
   - `parseArgs` drops any leading-dash positional (`upgrade --bad` / `upgrade -x`) and silently falls back to `latest` rather than erroring. For a daemon-driven fleet upgrade order from an untrusted boundary, malformed input rolling the fleet to `latest` is fail-*open*.
   
   Suggestion: if only real versions are intended, validate against a semver shape (e.g. `^\d+\.\d+\.\d+(?:[-+][0-9A-Za-z.-]+)?$`) and reject (don't drop) leading-dash input; otherwise relax the comment wording to "charset / spec-injection guard." (`src/cli.ts` `VERSION_RE`, `parseArgs`)

**2. (P2) `shutdown()` skips `store.close()` if `gateway.stop()` rejects.** The per-bot shutdown closure awaits `gateway.stop()` (which awaits `socket.disconnectAndWait()`, capable of rejecting) before `store.close()`. A rejection leaks the SQLite handle — and in the new connect-failure path the swallowed `s.shutdown().catch(() => {})` means a stop failure leaves that bot's lock/store behind while siblings keep the process alive. Suggestion: `try { await gateway.stop(...); } finally { store.close(); }`, mirroring the defensive cleanup already in `startBot`. (`src/index.ts` shutdown closure / connect loop)

**3. (P2) Single-bot connect failure surfaces a generic fatal message.** The start phase rethrows `failures[0].reason` (real cause), but the connect phase throws a generic `new Error('no bots connected')`; the actual cause (e.g. the `isAllowedWsUrl` "refusing insecure WebSocket URL" message) is only emitted via `console.error` one line earlier. The process still dies — only the final "Fatal error" summary loses the root cause. Suggestion: capture the first connect error and rethrow it when `connected.length === 0`, mirroring the start phase. (`src/index.ts` connect loop)

**4. (P2) The load-bearing partial-failure cleanup is untested.** The new tests cover only the pure `partitionStartResults` filter and `buildUpgradeArgs`. The central correctness claim — a failed bot releases its own lock/sqlite/timer so survivors don't leak — has no test, and neither does `releaseLock()`'s nonce-guard idempotency nor `parseArgs` positional-version extraction (incl. flag ordering). Suggestion: add a `gateway.releaseLock()` idempotency/foreign-nonce test and `parseArgs` cases for `['upgrade','1.2.3']`, `['upgrade','--foreground','1.2.3']`, `['upgrade']`. (`src/__tests__/*`)

**5. (nit) Injection-guard tests check only shell metacharacters** — which were never the live threat (no shell). Consider a positive-boundary case pinning that `@`/`/`-bearing strings throw and that a leading-dash stays inside the spec. (`src/__tests__/cli-upgrade.test.ts`)

## Coverage note

I could not verify at runtime: `WKSocket.disconnectAndWait()` rejection behavior (finding 2's trigger), npm's exact resolution of the installed spec, the cross-repo fleet/daemon contract that produces the version string, or sibling PRs in octo-daemon-cli / octo-fleet / octo-web. A second independent reviewer pass agreed on all findings and added no blockers; a third (long-context) advisory pass was attempted but failed to return and is therefore absent from this synthesis — its absence is not treated as agreement.

yujiawei (changes requested):
# Code Review — PR #148 (cc-channel-octo)

Reviewed at head SHA `7c4687aa05ef4a26f87a86287a7f20170b483ec2` against base `feat/agent-runtime`. **This review supersedes my earlier Approve on this PR.** After a closer pass (and reconciling with another reviewer's finding), I confirmed a real blocking gap on the resilience path that I missed the first time. The design is sound and the feature is close, but it does not yet meet its own stated guarantee.

## Verdict: REQUEST CHANGES

One P1 blocker (below). Everything else is non-blocking polish.

## 🔴 Blocking (P1)

**`startBot` leaks the cwd-cleanup timer when adapter/store construction fails.** `src/index.ts`:

```ts
const cwdCleanupTimer = setInterval(() => { cleanupExpiredCwds(cwdBase); }, CWD_CLEANUP_INTERVAL_MS);
cwdCleanupTimer.unref();
// ...
const adapter = createAdapter(dbPath);     // <-- OUTSIDE the try
const store = new SessionStore(adapter);   // <-- OUTSIDE the try
// ...
try {
  store.init();
  // ...
} catch (err) {
  clearInterval(cwdCleanupTimer);          // never reached if createAdapter threw
  // ...
}
```

`createAdapter(dbPath)` → `new BetterSqliteAdapter(dbPath)` calls `mkdirSync(dir, …)` and `new Database(dbPath)` in its constructor — both throw **synchronously** on an inaccessible/unwritable per-bot `dataDir`, a full disk, an `ENOTDIR`, or a sqlite open failure. Because this runs *before* the `try`, the already-armed `cwdCleanupTimer` is never cleared: with the new `Promise.allSettled` flow the failed bot is skipped while its 6-hourly interval keeps firing for the lifetime of the surviving gateway process. That is exactly the "a skipped bot leaks nothing" guarantee from the PR description, broken on the very per-bot failure the resilience feature is meant to tolerate.

Impact is bounded — the timer is `unref()`'d so it does not hold the process alive, no gateway lock / sqlite handle / socket is leaked, and it cannot accumulate beyond one per failed bot per process. So this is not data-loss or a crash. But it falsifies an explicit contract on a realistic path, and the fix is trivial:

- Move `createAdapter` / `new SessionStore` **inside** the protected `try` (the catch already best-effort-closes the store), **or**
- arm `cwdCleanupTimer` inside the `try`, **or**
- add an outer guard that clears the timer for any failure after it is armed.

Please also add a regression test: mock `createAdapter` to throw after the timer is armed, start multiple bots, and assert the failed bot's interval was cleared. (Credit to the parallel reviewer who first flagged this window.)

## What I verified clean

- **Injection guard is sound.** Version is always the single argv element `@mininglamp-oss/cc-channel-octo@<v>` and `spawn('npm', args)` uses no shell; `@` / `/` / `:` / whitespace are rejected, so a leading-dash version can't become a flag and no npm-spec pivot (`npm:`, `git+url`, `file:`, path) is possible.
- **`releaseLock()` double-release is safe** — `existsSync`-gated and nonce-guarded, so register()'s self-release + the catch's release can't double-unlink or delete another instance's lock.
- **`partitionStartResults`** preserves positional mapping with a `#<index>` fallback; per-bot `connect()` isolation is correct; `shutdownAll` drains the live `connected` set; single-bot fatality is preserved.

## Non-blocking findings (P2 / nit)

1. **(P2) Upgrade version path is fail-open.** `VERSION_RE` is a charset whitelist, not a semver validator — npm treats `pkg@<non-semver>` as a dist-tag, so `beta`/`next`/`canary` install whatever the registry tags (still scoped to this package). Separately, `parseArgs` silently drops a leading-dash positional (`upgrade --bad`) and falls back to `latest` instead of erroring — fail-open on a daemon-driven boundary. Suggest a semver-shaped check (e.g. `^\d+\.\d+\.\d+(?:[-+][0-9A-Za-z.-]+)?$`) and rejecting (not dropping) leading-dash input, or relax the "semantic-version whitelist" comment to "charset/spec-injection guard."
2. **(P2) `shutdown()` skips `store.close()` if `gateway.stop()` rejects.** `socket.disconnectAndWait()` can reject; wrap as `try { await gateway.stop(...) } finally { store.close() }`. Compounds with the swallowed `s.shutdown().catch(() => {})` in the connect-failure path.
3. **(P2) Single-bot connect failure throws a generic `no bots connected`** — the real cause (e.g. the `isAllowedWsUrl` refusal) is only on `console.error`. Mirror the start phase and rethrow the first connect error when `connected.length === 0`.
4. **(P2) Test coverage gaps.** No test for the partial-failure cleanup (the P1 above), `releaseLock()` nonce-guard idempotency, or `parseArgs` positional-version extraction incl. flag ordering. The new tests only cover the pure `partitionStartResults` filter and `buildUpgradeArgs`.
5. **(nit) Injection-guard tests only check shell metacharacters** (never the live threat). Add positive-boundary cases pinning that `@`/`/` strings throw and a leading-dash stays inside the spec.

## Coverage note

Could not verify at runtime: `WKSocket.disconnectAndWait()` rejection behavior, npm's resolution of the installed spec, the cross-repo fleet/daemon contract producing the version string, or sibling PRs. Findings were cross-checked by a second independent reviewer pass (which confirmed the clean areas and added no further blockers); a third long-context advisory pass was attempted but failed to return and is therefore absent — its silence is not treated as agreement.

yujiawei (approved):
# Code Review — PR #148 (cc-channel-octo)

Re-review at **new head SHA `f38183de85aef328b5d2e92054aac732de8fde2f`** (previous review was at `7c4687aa`; a new commit was pushed after my request for changes, so this is a fresh verdict — the prior one does not carry over).

## Verdict: APPROVE

The single blocking issue from my prior review has been fixed correctly. Only non-blocking P2/nit items remain.

## Blocker from prior round — now RESOLVED ✅

Commit `f38183de` ("move sqlite store creation into cleanup try so cwd timer is released") fixes the cwd-cleanup-timer leak:

```ts
let storeForCleanup: SessionStore | undefined;
// ...
try {
  const adapter = createAdapter(dbPath);   // now INSIDE the try
  const store = new SessionStore(adapter);
  storeForCleanup = store;
  store.init();
  // ...
} catch (err) {
  clearInterval(cwdCleanupTimer);
  // ...
  storeForCleanup?.close();
}
```

`createAdapter` → `new BetterSqliteAdapter` (`mkdirSync` + `new Database()`, both synchronous-throw on a bad/inaccessible dataDir or sqlite-open failure) now runs inside the protected region, so that failure path reaches the `catch` and releases the interval. The hoisted `storeForCleanup` keeps the catch's best-effort `close()` working without re-introducing an out-of-try reference. Scoping checks out: all `store`/`adapter` uses (groupContext, the handler closures, the success-path `store.close()`) are within the try, and the returned `shutdown` closure captures `store` from that scope. The PR's "a skipped bot leaks nothing" guarantee now holds on this path.

## Verified clean (unchanged from prior round)

- **Injection guard** sound — version is the single argv element `@mininglamp-oss/cc-channel-octo@<v>`, `spawn('npm', args)` uses no shell, and `@`/`/`/`:`/whitespace are rejected (no flag or npm-spec pivot).
- **`releaseLock()`** double-release safe (existsSync + nonce-guarded).
- **`partitionStartResults`** positional fallback, per-bot connect isolation, `shutdownAll` over the live `connected` set, single-bot fatality — all correct.

## Non-blocking findings (P2 / nit) — carried over, not merge gates

1. **(P2) Upgrade version path is fail-open.** `VERSION_RE` is a charset whitelist, not a semver validator — npm treats `pkg@<non-semver>` as a dist-tag, so `beta`/`next` install whatever the registry tags (still scoped to this package). And `parseArgs` silently drops a leading-dash positional (`upgrade --bad`) → falls back to `latest`. Consider a semver-shaped check and rejecting (not dropping) malformed input, or relax the "semantic-version whitelist" comment to "charset/spec-injection guard."
2. **(P2) `shutdown()` skips `store.close()` if `gateway.stop()` rejects** (success-path shutdown closure). `socket.disconnectAndWait()` can reject; wrap as `try { await gateway.stop(...) } finally { store.close() }`.
3. **(P2) Single-bot connect failure throws a generic `no bots connected`** — the real cause (e.g. the `isAllowedWsUrl` refusal) is only on `console.error`. Rethrow the first connect error when `connected.length === 0` to mirror the start phase.
4. **(P2) Test gaps.** This fix landed without a regression test. Worth adding: mock `createAdapter` to throw after the timer is armed and assert the interval is cleared; plus `releaseLock()` nonce-guard idempotency and `parseArgs` positional-version (incl. flag ordering).
5. **(nit) Injection-guard tests only check shell metacharacters** (never the live threat under no-shell spawn). Add positive-boundary cases pinning that `@`/`/` strings throw and a leading-dash stays inside the spec.

## Coverage note

Could not verify at runtime: `WKSocket.disconnectAndWait()` rejection behavior (P2-2's trigger), npm's resolution of the installed spec, or the cross-repo fleet/daemon contract producing the version string. This review re-read only the incremental diff `7c4687aa..f38183de` plus the surrounding `startBot` scope; the rest of the PR is unchanged from the prior full review.

Jerry-Xin (approved):
Summary: The PR is in scope for `cc-channel-octo` and the implementation looks merge-ready; the upgrade command and resilient multi-bot startup fit the gateway/supervisor responsibilities.

🔴 Blocking: None.

💬 Non-blocking:
- 🟡 Warning: Test coverage for multi-bot resilience is mostly on `partitionStartResults`, not the lifecycle behavior that carries the real risk. Consider adding mocked integration coverage for partial `startBot` cleanup and connect-phase failures around `src/index.ts:70`, `src/index.ts:108`, and `src/index.ts:357`.
- 🔵 Suggestion: In single-bot connect failure, the fatal error becomes `no bots connected`, while the original reason is only logged. Preserving the first connect error as `cause` at `src/index.ts:112` and `src/index.ts:118` would improve diagnostics.
- 🔵 Suggestion: `parseArgs` captures a positional argument for every command, not only `upgrade` (`src/cli.ts:88`). This is harmless currently, but command-specific validation would avoid surprising ignored/accepted arguments later.

✅ Highlights:
- The npm upgrade path uses `spawn` without a shell and validates the version string before constructing the package spec (`src/cli.ts:252`).
- Partial startup cleanup now closes the store, stops cron, clears the cwd timer, and releases the nonce-guarded lock (`src/index.ts:357`, `src/gateway.ts:277`).
- The multi-bot policy correctly keeps successful stacks and only fails the process when none start/connect (`src/index.ts:70`, `src/index.ts:118`).

Verification caveat: I attempted `npm test -- --runInBand` and `npm run lint`, but this checkout does not have dev dependencies installed (`vitest` and `eslint` were not found).

PR #150, opened 2026-06-22
feat: add configure subcommand and sdk.apiKey for one-click install

caster-Q:
## Summary

Adds a `configure` subcommand and an `sdk.apiKey` config field so the gateway's LLM credentials can be set programmatically (for daemon-driven one-click install) instead of relying on an `ANTHROPIC_API_KEY` environment prerequisite.

## Changes

- `configure --gateway-url <url> [--api-key <key>]` subcommand: raw read-merge-write of the global `~/.cc-channel-octo/config.json`, setting `sdk.anthropicBaseUrl` + `sdk.apiKey`, then `chmod 600`. Independent of `loadConfig` (which requires `apiUrl`) so it can run before any bot/apiUrl exists. Reuses `isAllowedApiUrl` (SSRF policy) for the gateway URL.
- `sdk.apiKey` config field, forwarded to the SDK subprocess as `ANTHROPIC_API_KEY` via an extracted pure `buildSdkEnv` (mirrors the existing `ANTHROPIC_BASE_URL` forwarding).
- `configure` reads the key from the `CC_OCTO_CONFIGURE_API_KEY` env var when `--api-key` is omitted, so a caller need not pass the key in argv (keeps it out of `ps`).

## Testing

- `npm run build && npm test && npm run type-check && npm run lint` — all green (1035 tests).
- `configure` exercised end-to-end against a temp HOME: fresh write, merge-preserve, mode 600, missing-arg exit 2, unsafe-URL reject.

Backward compatible: new subcommand + new optional config field + new optional env var; existing behavior unchanged.

### DIFF
--- a/src/__tests__/agent-bridge-env.test.ts
+++ b/src/__tests__/agent-bridge-env.test.ts
@@ -1,0 +1,23 @@
+import { describe, it, expect } from 'vitest'
+import { buildSdkEnv } from '../agent-bridge.js'
+
+describe('buildSdkEnv', () => {
+  it('injects ANTHROPIC_API_KEY when sdk.apiKey set', () => {
+    const env = buildSdkEnv({ apiKey: 'sk-test', anthropicBaseUrl: 'https://gw.example.com' }, { HOME: '/h' })
+    expect(env?.ANTHROPIC_API_KEY).toBe('sk-test')
+    expect(env?.ANTHROPIC_BASE_URL).toBe('https://gw.example.com')
+    expect(env?.HOME).toBe('/h')
+  })
+  it('injects ANTHROPIC_BASE_URL only when apiKey absent', () => {
+    const env = buildSdkEnv({ anthropicBaseUrl: 'https://gw.example.com' }, { HOME: '/h' })
+    expect(env?.ANTHROPIC_BASE_URL).toBe('https://gw.example.com')
+    expect(env?.ANTHROPIC_API_KEY).toBeUndefined()
+  })
+  it('returns undefined when nothing to inject', () => {
+    expect(buildSdkEnv({}, { HOME: '/h' })).toBeUndefined()
+  })
+  it('apiKey alone (no baseUrl) still injects key', () => {
+    const env = buildSdkEnv({ apiKey: 'sk-only' }, { HOME: '/h' })
+    expect(env?.ANTHROPIC_API_KEY).toBe('sk-only')
+  })
+})
--- a/src/__tests__/cli.test.ts
+++ b/src/__tests__/cli.test.ts
@@ -133,6 +133,17 @@ describe('readVersion', () => {
   });
 });
 
+describe('parseArgs configure', () => {
+  it('parses --gateway-url and --api-key (space form)', () => {
+    const a = parseArgs(['configure', '--gateway-url', 'https://gw', '--api-key', 'sk-1'])
+    expect(a.cmd).toBe('configure'); expect(a.gatewayUrl).toBe('https://gw'); expect(a.apiKey).toBe('sk-1')
+  })
+  it('parses = form', () => {
+    const a = parseArgs(['configure', '--gateway-url=https://gw', '--api-key=sk-2'])
+    expect(a.gatewayUrl).toBe('https://gw'); expect(a.apiKey).toBe('sk-2')
+  })
+})
+
 describe('run version command', () => {
   it.each(['version', '--version', '-v'])('prints bare version and exits 0 for %s', async (arg) => {
     const spy = vi.spyOn(console, 'log').mockImplementation(() => {});
--- a/src/__tests__/cli.test.ts
+++ b/src/__tests__/cli.test.ts
@@ -142,6 +142,18 @@ describe('parseArgs configure', () => {
     const a = parseArgs(['configure', '--gateway-url=https://gw', '--api-key=sk-2'])
     expect(a.gatewayUrl).toBe('https://gw'); expect(a.apiKey).toBe('sk-2')
   })
+  it('throws when --gateway-url value is missing', () => {
+    expect(() => parseArgs(['configure', '--gateway-url'])).toThrow(/requires a value/)
+  })
+  it('throws when --gateway-url value is another flag', () => {
+    expect(() => parseArgs(['configure', '--gateway-url', '--api-key', 'sk'])).toThrow(/requires a value/)
+  })
+  it('throws when --api-key value is missing', () => {
+    expect(() => parseArgs(['configure', '--api-key'])).toThrow(/requires a value/)
+  })
+  it('throws when --api-key value is another flag', () => {
+    expect(() => parseArgs(['configure', '--api-key', '--gateway-url', 'https://gw'])).toThrow(/requires a value/)
+  })
 })
 
 describe('run configure with env var fallback', () => {
--- a/src/__tests__/cli.test.ts
+++ b/src/__tests__/cli.test.ts
@@ -144,6 +144,65 @@ describe('parseArgs configure', () => {
   })
 })
 
+describe('run configure with env var fallback', () => {
+  let dir: string;
+  let cfgPath: string;
+  const originalEnv = process.env.CC_OCTO_CONFIGURE_API_KEY;
+
+  beforeEach(() => {
+    dir = mkdtempSync(join(tmpdir(), 'cc-cli-cfg-'));
+    cfgPath = join(dir, 'config.json');
+  });
+
+  afterEach(() => {
+    rmSync(dir, { recursive: true, force: true });
+    if (originalEnv === undefined) {
+      delete process.env.CC_OCTO_CONFIGURE_API_KEY;
+    } else {
+      process.env.CC_OCTO_CONFIGURE_API_KEY = originalEnv;
+    }
+  });
+
+  it('reads API key from CC_OCTO_CONFIGURE_API_KEY when --api-key is absent', async () => {
+    process.env.CC_OCTO_CONFIGURE_API_KEY = 'sk-from-env';
+    const spy = vi.spyOn(console, 'log').mockImplementation(() => {});
+    try {
+      const code = await run(['configure', '--gateway-url', 'https://gw'], dir);
+      expect(code).toBe(0);
+      const parsed = JSON.parse(readFileSync(cfgPath, 'utf-8'));
+      expect(parsed.sdk.apiKey).toBe('sk-from-env');
+      expect(parsed.sdk.anthropicBaseUrl).toBe('https://gw');
+    } finally {
+      spy.mockRestore();
+    }
+  });
+
+  it('CLI --api-key flag takes precedence over env var', async () => {
+    process.env.CC_OCTO_CONFIGURE_API_KEY = 'sk-from-env';
+    const spy = vi.spyOn(console, 'log').mockImplementation(() => {});
+    try {
+      const code = await run(['configure', '--gateway-url', 'https://gw', '--api-key', 'sk-from-cli'], dir);
+      expect(code).toBe(0);
+      const parsed = JSON.parse(readFileSync(cfgPath, 'utf-8'));
+      expect(parsed.sdk.apiKey).toBe('sk-from-cli');
+    } finally {
+      spy.mockRestore();
+    }
+  });
+
+  it('exits 2 when both --api-key and env var are missing', async () => {
+    delete process.env.CC_OCTO_CONFIGURE_API_KEY;
+    const spy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    try {
+      const code = await run(['configure', '--gateway-url', 'https://gw'], dir);
+      expect(code).toBe(2);
+      expect(spy).toHaveBeenCalledWith(expect.stringContaining('required'));
+    } finally {
+      spy.mockRestore();
+    }
+  });
+});
+
 describe('run version command', () => {
   it.each(['version', '--version', '-v'])('prints bare version and exits 0 for %s', async (arg) => {
     const spy = vi.spyOn(console, 'log').mockImplementation(() => {});
--- a/src/__tests__/configure.test.ts
+++ b/src/__tests__/configure.test.ts
@@ -41,4 +41,25 @@ describe('configure', () => {
   it('rejects an unsafe (non-http/https) gateway url', () => {
     expect(() => configure('ftp://gw', 'sk', cfgPath)).toThrow()
   })
+  it('ensures final file mode is 0600 even when existing file has broader perms', () => {
+    // Pre-create config with mode 0644
+    writeFileSync(cfgPath, JSON.stringify({ sdk: {} }), { mode: 0o644 })
+    configure('https://gw', 'sk-secret', cfgPath)
+    const mode = statSync(cfgPath).mode & 0o777
+    expect(mode).toBe(0o600)
+    const content = JSON.parse(readFileSync(cfgPath, 'utf-8'))
+    expect(content.sdk.apiKey).toBe('sk-secret')
+  })
+  it('throws a clear error when existing config root is not a plain object', () => {
+    writeFileSync(cfgPath, JSON.stringify(null))
+    expect(() => configure('https://gw', 'sk', cfgPath)).toThrow(/is not a JSON object/)
+  })
+  it('throws a clear error when existing config root is an array', () => {
+    writeFileSync(cfgPath, JSON.stringify([1, 2, 3]))
+    expect(() => configure('https://gw', 'sk', cfgPath)).toThrow(/is not a JSON object/)
+  })
+  it('throws a clear error when existing config root is a number', () => {
+    writeFileSync(cfgPath, JSON.stringify(42))
+    expect(() => configure('https://gw', 'sk', cfgPath)).toThrow(/is not a JSON object/)
+  })
 })
--- a/src/__tests__/configure.test.ts
+++ b/src/__tests__/configure.test.ts
@@ -1,0 +1,44 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import { mkdtempSync, rmSync, readFileSync, writeFileSync, statSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { configure } from '../configure.js'
+
+let dir: string
+let cfgPath: string
+beforeEach(() => { dir = mkdtempSync(join(tmpdir(), 'ccfg-')); cfgPath = join(dir, 'config.json') })
+afterEach(() => { rmSync(dir, { recursive: true, force: true }) })
+
+describe('configure', () => {
+  it('creates a fresh config with sdk gateway + apiKey', () => {
+    configure('https://gw.example.com', 'sk-test', cfgPath)
+    const parsed = JSON.parse(readFileSync(cfgPath, 'utf-8'))
+    expect(parsed.sdk.anthropicBaseUrl).toBe('https://gw.example.com')
+    expect(parsed.sdk.apiKey).toBe('sk-test')
+  })
+  it('merges into an existing config, preserving other fields', () => {
+    writeFileSync(cfgPath, JSON.stringify({ apiUrl: 'https://octo.example.com', sdk: { model: 'claude-x', anthropicBaseUrl: 'https://old' } }))
+    configure('https://new-gw', 'sk-new', cfgPath)
+    const parsed = JSON.parse(readFileSync(cfgPath, 'utf-8'))
+    expect(parsed.apiUrl).toBe('https://octo.example.com')
+    expect(parsed.sdk.model).toBe('claude-x')
+    expect(parsed.sdk.anthropicBaseUrl).toBe('https://new-gw')
+    expect(parsed.sdk.apiKey).toBe('sk-new')
+  })
+  it('writes the file mode 600', () => {
+    configure('https://gw', 'sk', cfgPath)
+    expect(statSync(cfgPath).mode & 0o777).toBe(0o600)
+  })
+  it('creates the parent dir if missing', () => {
+    const nested = join(dir, 'sub', 'config.json')
+    configure('https://gw', 'sk', nested)
+    expect(JSON.parse(readFileSync(nested, 'utf-8')).sdk.apiKey).toBe('sk')
+  })
+  it('throws on empty gatewayUrl or apiKey', () => {
+    expect(() => configure('', 'sk', cfgPath)).toThrow()
+    expect(() => configure('https://gw', '', cfgPath)).toThrow()
+  })
+  it('rejects an unsafe (non-http/https) gateway url', () => {
+    expect(() => configure('ftp://gw', 'sk', cfgPath)).toThrow()
+  })
+})
--- a/src/agent-bridge.ts
+++ b/src/agent-bridge.ts
@@ -19,6 +19,32 @@ import { trustedText, escapeSectionMarkers, CURRENT_MESSAGE_ANCHOR } from './pro
 import type { SafeText } from './prompt-safety.js';
 
 
+/**
+ * Build the SDK subprocess env overlay. The SDK's `env` option REPLACES the
+ * subprocess environment, so we spread the base env first to keep
+ * PATH/HOME/etc., then layer operator-declared vars and gateway routing.
+ * Returns undefined when there is nothing to add (subprocess just inherits).
+ * Pure (base env injected) so the injection matrix is unit-testable.
+ *
+ * Param is narrowed to the fields it reads (not the whole Config['sdk']) so
+ * tests pass plain literals — repo lint is `--max-warnings 0` with
+ * no-explicit-any, so an `as any` cast in the test would fail the build.
+ */
+export function buildSdkEnv(
+  sdk: Pick<Config['sdk'], 'apiKey' | 'anthropicBaseUrl' | 'env'>,
+  baseEnv: NodeJS.ProcessEnv,
+): NodeJS.ProcessEnv | undefined {
+  const extraEnv = sdk.env;
+  const hasExtraEnv = extraEnv !== undefined && Object.keys(extraEnv).length > 0;
+  if (!sdk.anthropicBaseUrl && !sdk.apiKey && !hasExtraEnv) return undefined;
+  return {
+    ...baseEnv,
+    ...(hasExtraEnv ? extraEnv : {}),
+    ...(sdk.anthropicBaseUrl ? { ANTHROPIC_BASE_URL: sdk.anthropicBaseUrl } : {}),
+    ...(sdk.apiKey ? { ANTHROPIC_API_KEY: sdk.apiKey } : {}),
+  };
+}
+
 const VALID_PERMISSION_MODES: Set<string> = new Set([
   'default', 'acceptEdits', 'bypassPermissions', 'plan', 'dontAsk', 'auto',
 ]);
@@ -225,25 +251,7 @@ export async function* queryAgent(
     if (sources.length > 0) linkSkillsIntoSandbox(cwd, sources);
   }
 
-  // Q1: forward scoped env to the SDK subprocess via the `env` option instead
-  // of mutating the gateway's global process.env. The SDK's `env` REPLACES the
-  // subprocess environment entirely, so spread process.env first to preserve
-  // PATH/HOME/ANTHROPIC_API_KEY. Scoping here means overrides never leak across
-  // requests. When nothing needs adding, omit `env` so the subprocess simply
-  // inherits process.env.
-  //   - sdk.env (#107): operator-declared extra vars (e.g. OCTO_BOT_ID so a
-  //     multi-bot deploy's octo-cli picks the right profile). Generic — cc does
-  //     not interpret them.
-  //   - ANTHROPIC_BASE_URL: model-gateway routing (set last so it wins).
-  const extraEnv = config.sdk.env;
-  const hasExtraEnv = extraEnv !== undefined && Object.keys(extraEnv).length > 0;
-  const env = config.sdk.anthropicBaseUrl || hasExtraEnv
-    ? {
-        ...process.env,
-        ...(hasExtraEnv ? extraEnv : {}),
-        ...(config.sdk.anthropicBaseUrl ? { ANTHROPIC_BASE_URL: config.sdk.anthropicBaseUrl } : {}),
-      }
-    : undefined;
+  const env = buildSdkEnv(config.sdk, process.env)
 
   // Build + iterate the SDK stream for a given resume id and prompt. Extracted so
   // a stale/expired `resume` (the SDK throws "No conversation found with session
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -312,7 +312,7 @@ Usage:
   cc-channel-octo restart                stop (if running) then start
   cc-channel-octo status                 show running state
   cc-channel-octo upgrade [<version>]    npm install -g the gateway (default latest) then restart
-  cc-channel-octo configure --gateway-url <url> --api-key <key>   write LLM gateway + key to config
+  cc-channel-octo configure --gateway-url <url> [--api-key <key>]   write LLM gateway + key to config (or set CC_OCTO_CONFIGURE_API_KEY)
   cc-channel-octo version                print the version
 
 Paths (under ~/.cc-channel-octo):
@@ -337,15 +337,18 @@ export async function run(argv: string[], baseDir?: string): Promise<number> {
       return cmdStatus(paths);
     case 'upgrade':
       return cmdUpgrade(paths, timeoutMs, version);
-    case 'configure':
+    case 'configure': {
+      const resolvedApiKey = apiKey ?? process.env.CC_OCTO_CONFIGURE_API_KEY ?? '';
+      const configPath = baseDir ? join(baseDir, 'config.json') : undefined;
       try {
-        configure(gatewayUrl ?? '', apiKey ?? '');
+        configure(gatewayUrl ?? '', resolvedApiKey, configPath);
         console.log('cc-channel-octo: configured gateway + api key');
         return 0;
       } catch (err) {
         console.error(`cc-channel-octo: ${(err as Error).message}`);
         return 2;
       }
+    }
     case 'version':
     case '--version':
     case '-v':
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -105,11 +105,19 @@ export function parseArgs(argv: string[]): ParsedArgs {
       const n = Number.parseInt(a.slice('--timeout='.length), 10);
       if (Number.isFinite(n) && n > 0) timeoutSec = n;
     } else if (a === '--gateway-url') {
-      gatewayUrl = rest[++i];
+      const next = rest[++i];
+      if (next === undefined || next.startsWith('--')) {
+        throw new Error('configure: --gateway-url requires a value')
+      }
+      gatewayUrl = next;
     } else if (a.startsWith('--gateway-url=')) {
       gatewayUrl = a.slice('--gateway-url='.length);
     } else if (a === '--api-key') {
-      apiKey = rest[++i];
+      const next = rest[++i];
+      if (next === undefined || next.startsWith('--')) {
+        throw new Error('configure: --api-key requires a value')
+      }
+      apiKey = next;
     } else if (a.startsWith('--api-key=')) {
       apiKey = a.slice('--api-key='.length);
     } else if (!a.startsWith('-') && version === undefined) {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -23,6 +23,7 @@ import { dirname, join } from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import { setTimeout as sleep } from 'node:timers/promises';
 import { DEFAULT_CONFIG_PATH } from './config.js';
+import { configure } from './configure.js';
 
 export interface SupervisorPaths {
   baseDir: string;
@@ -52,6 +53,10 @@ export interface ParsedArgs {
   timeoutMs: number;
   /** Positional version target for `upgrade` (e.g. `upgrade 1.2.3`); undefined → latest. */
   version?: string;
+  /** `configure --gateway-url <url>`. */
+  gatewayUrl?: string;
+  /** `configure --api-key <key>`. */
+  apiKey?: string;
 }
 
 /**
@@ -90,18 +95,28 @@ export function parseArgs(argv: string[]): ParsedArgs {
   let foreground = false;
   let timeoutSec = 10;
   let version: string | undefined;
-  for (const a of rest) {
+  let gatewayUrl: string | undefined;
+  let apiKey: string | undefined;
+  for (let i = 0; i < rest.length; i++) {
+    const a = rest[i];
     if (a === '--foreground' || a === '-f') {
       foreground = true;
     } else if (a.startsWith('--timeout=')) {
       const n = Number.parseInt(a.slice('--timeout='.length), 10);
       if (Number.isFinite(n) && n > 0) timeoutSec = n;
+    } else if (a === '--gateway-url') {
+      gatewayUrl = rest[++i];
+    } else if (a.startsWith('--gateway-url=')) {
+      gatewayUrl = a.slice('--gateway-url='.length);
+    } else if (a === '--api-key') {
+      apiKey = rest[++i];
+    } else if (a.startsWith('--api-key=')) {
+      apiKey = a.slice('--api-key='.length);
     } else if (!a.startsWith('-') && version === undefined) {
-      // First positional token (e.g. `upgrade 1.2.3`).
       version = a;
     }
   }
-  return { cmd, foreground, timeoutMs: timeoutSec * 1000, version };
+  return { cmd, foreground, timeoutMs: timeoutSec * 1000, version, gatewayUrl, apiKey };
 }
 
 /**
@@ -297,6 +312,7 @@ Usage:
   cc-channel-octo restart                stop (if running) then start
   cc-channel-octo status                 show running state
   cc-channel-octo upgrade [<version>]    npm install -g the gateway (default latest) then restart
+  cc-channel-octo configure --gateway-url <url> --api-key <key>   write LLM gateway + key to config
   cc-channel-octo version                print the version
 
 Paths (under ~/.cc-channel-octo):
@@ -307,7 +323,7 @@ POSIX only (macOS/Linux). On Windows, run under a service manager.`;
 }
 
 export async function run(argv: string[], baseDir?: string): Promise<number> {
-  const { cmd, foreground, timeoutMs, version } = parseArgs(argv);
+  const { cmd, foreground, timeoutMs, version, gatewayUrl, apiKey } = parseArgs(argv);
   const paths = resolveSupervisorPaths(baseDir);
   switch (cmd) {
     case 'start':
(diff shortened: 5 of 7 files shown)


Jerry-Xin (changes requested):
The PR is in scope for `cc-channel-octo`, but it has a blocking secret-handling issue in the new `configure` write path.

**🔴 Blocking**

🔴 Critical: [src/configure.ts](/Users/mlamp/.openclaw/tmp/pr-review-82qh_utq/repo/src/configure.ts:43) writes the API key into the existing config file before tightening permissions with `chmodSync` at line 46. If the existing `~/.cc-channel-octo/config.json` is group/world-readable, the new secret is briefly written under the old permissions. This is especially relevant because existing config files are only warned about when too broad, not rejected. Fix by setting permissions before writing when the file exists, or preferably write a `0600` temp file and atomic-rename it over the target, then `chmod 600`.

**💬 Non-blocking**

🟡 Warning: [src/configure.ts](/Users/mlamp/.openclaw/tmp/pr-review-82qh_utq/repo/src/configure.ts:26) assumes the existing JSON root is an object. A valid but malformed-for-this-app config like `null` will throw a less intentional `Cannot read properties of null` error. Consider validating `parsed && typeof parsed === 'object' && !Array.isArray(parsed)` before merging.

🔵 Suggestion: [src/cli.ts](/Users/mlamp/.openclaw/tmp/pr-review-82qh_utq/repo/src/cli.ts:107) accepts missing option values as the next token, including another flag. Explicit parser errors for `--gateway-url` / `--api-key` without a value would make CLI failures clearer.

**✅ Highlights**

The new `buildSdkEnv` extraction in [src/agent-bridge.ts](/Users/mlamp/.openclaw/tmp/pr-review-82qh_utq/repo/src/agent-bridge.ts:33) is a good refactor and preserves the existing inheritance behavior while adding `ANTHROPIC_API_KEY`.

The tests cover fresh configure, merge-preserve behavior, env-var fallback, CLI precedence, and SDK env injection. I could not execute them locally because dependencies are not installed in this checkout (`vitest`, `tsc`, and `eslint` were not found).

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#150 Review Report
**PR:** https://github.com/Mininglamp-OSS/cc-channel-octo/pull/150
**Head SHA:** `886280433548650aa527141056650f8f7fe2315f`
**Title:** feat: add configure subcommand and sdk.apiKey for one-click install
**Scope:** +250/−23 across 7 files (3 new, 4 modified)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `buildSdkEnv` refactor preserves existing behavior | ✅ | `src/agent-bridge.ts:34-48` — same spread order: baseEnv → extraEnv → ANTHROPIC_BASE_URL, plus new ANTHROPIC_API_KEY |
| `configure` validates gateway URL via SSRF policy | ✅ | `src/configure.ts:21` — calls `isAllowedApiUrl()` (same check as `loadConfig`'s `anthropicBaseUrl`) |
| File permissions set to 600 | ✅ | `src/configure.ts:43-45` — `writeFileSync` mode 0o600 + explicit `chmodSync` for pre-existing files |
| `sdk.apiKey` forwarded as `ANTHROPIC_API_KEY` to SDK subprocess | ✅ | `src/agent-bridge.ts:46` — last in spread chain, wins over baseEnv and sdk.env |
| Env var fallback `CC_OCTO_CONFIGURE_API_KEY` | ✅ | `src/cli.ts:340` — `apiKey ?? process.env.CC_OCTO_CONFIGURE_API_KEY ?? ''` |
| Test coverage (1035 tests per PR body) | ✅ | 3 new test files cover: buildSdkEnv matrix (4 cases), configure e2e (6 cases), CLI parseArgs + env-var fallback (3 cases), plus existing cli.test.ts additions |
| Backward compatibility | ✅ | New subcommand + optional config field + optional env var; no existing behavior changed |

---

## 2. Findings

### P2 — Missing semicolon in `agent-bridge.ts`

**File:** `src/agent-bridge.ts:253` (in the `queryAgent` function)
**Diff-scope:** new (introduced by this PR's refactor)

```typescript
const env = buildSdkEnv(config.sdk, process.env)  // ← missing semicolon
```

The rest of the file consistently uses semicolons. The original inline code this replaced had a semicolon. Likely a lint catch (`npm run lint` is listed as green in the PR body — if the linter didn't flag this, the eslint config may not enforce semicolons, making it a style-only issue).

**Severity:** P2/nit — cosmetic inconsistency, no runtime impact.

### P2 — `parseArgs` flag-at-end-of-argv produces generic error

**File:** `src/cli.ts:103-108`
**Diff-scope:** new (new flags introduced by this PR)

When `--gateway-url` or `--api-key` is the last token with no value following it:

```typescript
} else if (a === '--gateway-url') {
  gatewayUrl = rest[++i];  // i goes past end → undefined
```

`rest[++i]` returns `undefined` (JS array OOB). This flows to `configure(gatewayUrl ?? '', ...)` which throws `'configure: --gateway-url is required'` — technically correct but misleading (the flag *was* provided, just without a value). A more specific message like `'--gateway-url requires a value'` would improve UX.

**Severity:** P2 — minor UX, error is still caught and reported.

---

## 3. Suggestions

1. **Semicolon:** Add `;` to `src/agent-bridge.ts:253` for consistency.
2. **Flag value validation:** In `parseArgs`, when `rest[++i]` is `undefined` for a value-bearing flag, consider setting a sentinel or throwing early with a specific message. Alternatively, in the `configure` case handler, check `gatewayUrl === undefined` separately from `gatewayUrl === ''`.

---

## 4. Additional Observations

- **TOCTOU on config file:** `configure()` does read-parse-modify-write without file locking. Acceptable for a one-time install CLI (single operator, single invocation), but worth noting if this is ever called concurrently (e.g., parallel bot provisioning).
- **API key on disk:** `sdk.apiKey` persists in `config.json`. This is consistent with the existing trust model (botToken is also in config files; the file is chmod 600 + warned if too open). No regression.
- **API key in argv:** `--api-key <key>` exposes the key in `ps` output. The PR mitigates this with the `CC_OCTO_CONFIGURE_API_KEY` env var alternative — good defense-in-depth.
- **`sdk.env` cannot automated review `ANTHROPIC_API_KEY`:** In `buildSdkEnv`, `ANTHROPIC_API_KEY` is spread last, so operator-declared `sdk.env` vars cannot accidentally override it. Correct precedence.

---

## 5. Data-Flow Trace

| Consumed data | Upstream source | Reaches consumption point? |
|---|---|---|
| `sdk.apiKey` (config) | `configure()` writes to `config.json` → `loadConfig()` → `mergeConfig()` → `Config.sdk.apiKey` | ✅ Yes — `buildSdkEnv(config.sdk, process.env)` reads it, sets `ANTHROPIC_API_KEY` |
| `sdk.anthropicBaseUrl` (config) | Same path as above, pre-existing field | ✅ Yes — unchanged from base branch |
| `CC_OCTO_CONFIGURE_API_KEY` (env) | `process.env` in `cli.ts:340` | ✅ Yes — resolved to `resolvedApiKey`, passed to `configure()` which writes to config |
| `--gateway-url` (CLI arg) | `parseArgs` → `gatewayUrl` | ✅ Yes — passed to `configure()` after `isAllowedApiUrl` check |
| `--api-key` (CLI arg) | `parseArgs` → `apiKey` | ✅ Yes — takes precedence over env var via `??` chain |
| `buildSdkEnv` return value | Called at `agent-bridge.ts:253` | ✅ Yes — passed as `env` to SDK `options`, same as the old inline code |

---

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths in this PR. The `configure` command is a one-way write; there is no `unconfigure` to be symmetric with.
- **C2 — Control-flow ordering / nested reuse:** Clear. `buildSdkEnv` is called once per `queryAgent` invocation. The spread order (baseEnv → extraEnv → ANTHROPIC_BASE_URL → ANTHROPIC_API_KEY) is intentional and tested. No nested/double invocation path exists.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This PR adds a local config-writing subcommand and env-var forwarding. No new endpoints, tools, or remote capabilities are exposed.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth/authz changes.
- **C5 — Build ≠ runtime:** N/A. This is a Node.js CLI tool installed via `npm install -g`. No browser extension, no relative-URL-in-different-container scenario. The `configure` command writes to an absolute path (`~/.cc-channel-octo/config.json`) — no relative-path ambiguity.
- **C6 — Governance/policy document consistency:** N/A. No documentation or policy changes.

---

## 7. Cross-round Blocker Recheck (R6)

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 findings. Two P2 nits (missing semicolon, generic error message on flag-without-value) — neither blocks landing. The PR is well-structured: clean extraction of `buildSdkEnv` into a testable pure function, proper SSRF validation on the gateway URL, correct file permissions, sensible env-var fallback with argv precedence, and thorough test coverage across all new paths.

yujiawei (approved):
# Code Review — PR #150 (cc-channel-octo)

**Head SHA reviewed:** `886280433548650aa527141056650f8f7fe2315f`
**Base:** `feat/agent-runtime` (merge-base `ed2859f`) · 3 commits, 7 files (+250 / −23)

## Verdict: APPROVE

Verified locally against the head SHA: `tsc --noEmit` clean, `eslint --max-warnings 0` clean, full suite **1035/1035 passing** (incl. the 14 new tests for `configure`/`buildSdkEnv`/parseArgs). The core refactor (`buildSdkEnv` extraction) faithfully preserves the prior inline env-overlay behavior while adding the new `ANTHROPIC_API_KEY` injection, and it's now unit-testable as a pure function — a genuine improvement.

No correctness, build, or functional-blocker issues found. The findings below are all **hardening / advisory (P2)** on what is a local, operator-run install command. None blocks merge, but the first is cheap and worth doing.

## 1. Findings

### P2 — Secret written before permissions are tightened (re-configure case)
`src/configure.ts:44-46`
```ts
writeFileSync(path, JSON.stringify(merged, null, 2) + '\n', { mode: 0o600 })
chmodSync(path, 0o600)
```
`writeFileSync`'s `mode` option only applies when the file is **created**. I confirmed this empirically: writing over a pre-existing `0644` file keeps it `0644` for the duration of the write, and only the trailing `chmodSync` tightens it. So a **re-configure** over an existing world-readable `config.json` writes the plaintext API key under loose permissions for a brief window before the chmod lands — a local-disclosure window on a multi-user host.

Fresh install (the stated one-click use case) is safe: the file is created at `0600`. The risk is confined to overwriting a pre-existing loose-perm file.

**Recommendation:** write atomically with the tight mode up front — write to a sibling temp file created at `0600`, then `renameSync` over the target. This also fixes the crash/partial-write corruption concern below in one stroke.

### P2 — API key accepted as a command-line argument
`src/cli.ts:111-114, 315, 341`
`--api-key <key>` is captured from `argv`, advertised in the usage text, and takes precedence over the env var. A secret in `argv` is visible via process listings (`ps`) and frequently lands in shell history / installer logs.

This is partially mitigated already: the PR ships `CC_OCTO_CONFIGURE_API_KEY` as a documented fallback, which is the right channel for the daemon-driven install path. **Recommendation:** prefer the env var (or stdin) as the canonical input and consider dropping or de-advertising the `--api-key` flag, or at minimum note the exposure in the help text.

### P2 — Non-atomic in-place overwrite can corrupt config
`src/configure.ts:44`
An OOM kill, crash, or full disk mid-write leaves `config.json` truncated, bricking gateway state. The temp-file + `renameSync` fix recommended above resolves this as well.

### P2 — `sdk.apiKey` is forwarded without runtime type validation
`src/config.ts:143` / `src/agent-bridge.ts:44`
`loadConfig` validates `sdk.anthropicBaseUrl` (SSRF) but does not type-check `sdk.apiKey`. `buildSdkEnv` forwards any truthy value into `ANTHROPIC_API_KEY`, so a malformed config (`"apiKey": 123`, or an object) reaches the SDK subprocess as a coerced/garbage value. Low impact (operator-controlled file), but a non-empty-string guard at load time would fail fast with a clear message.

### Nit — trailing flag with no value is silently swallowed
`src/cli.ts:107-114`
`--gateway-url` / `--api-key` as the final token sets the value to `undefined` (`rest[++i]` is out of bounds), surfacing as `"--gateway-url is required"` rather than `"missing value for --gateway-url"`. Minor UX.

## 2. Informational (not a blocker)
**SSRF on the gateway URL:** `configure` validates `--gateway-url` with `isAllowedApiUrl`, which permits any `https:` host without DNS resolution (so `https://name-that-resolves-to-169.254.169.254` passes). This is **not introduced by this PR** — it is the established repo-wide policy already applied to `apiUrl` and `sdk.anthropicBaseUrl` in `loadConfig`, and the DNS-rebinding residual risk is explicitly documented in `url-policy.ts`. `configure` correctly mirrors the existing check rather than weakening it. If the project wants stricter gateway validation (DNS + private-range rejection), that's a repo-wide hardening item, not a change this PR should be gated on.

## 3. Coverage notes
- Ran the full vitest suite locally (1035 pass), type-check, and lint — all green at the head SHA.
- Confirmed `buildSdkEnv` is behavior-equivalent to the prior inline logic in `queryAgent` (base `ed2859f`), with `ANTHROPIC_API_KEY` as the only added injection and `undefined` returned when there's nothing to add (subprocess inherits `process.env`).
- Not exercised: whether a running daemon hot-reloads `config.json` after `configure` (assumed the install flow runs `configure` before `start`, which is correct ordering); the external lifecycle of `CC_OCTO_CONFIGURE_API_KEY` prior to invocation.

yujiawei (changes requested):
# Code Review — PR #150 (cc-channel-octo)

**Head SHA reviewed:** `886280433548650aa527141056650f8f7fe2315f`
**Base:** `feat/agent-runtime` (merge-base `ed2859f`) · 3 commits, 7 files (+250 / −23)

## Verdict: CHANGES REQUESTED

> Revising my own earlier review on this PR: I initially scored the permissions issue below as advisory because the fresh-install path creates the file at `0600`. On closer analysis the realistic, **documented** install path does hit a world-readable window, and another reviewer independently flagged it as blocking. I agree — for a command whose sole job is to store a secret, this should be fixed before merge. It's a small change.

Build health is good: at the head SHA, `tsc --noEmit` clean, `eslint --max-warnings 0` clean, full suite **1035/1035 passing** (incl. 14 new tests). The `buildSdkEnv` extraction is a clean, behavior-preserving refactor. The single blocker is the secret-write ordering.

## 1. Blocking (P1)

### Secret written under loose permissions before they are tightened
`src/configure.ts:44-46`
```ts
writeFileSync(path, JSON.stringify(merged, null, 2) + '\n', { mode: 0o600 })
chmodSync(path, 0o600)
```
`writeFileSync`'s `mode` only applies when the file is **created** — I confirmed empirically that writing over a pre-existing `0644` file leaves it `0644` through the write, and only the trailing `chmodSync` tightens it. So the plaintext API key is written under the old (loose) permissions first.

Why this is reachable, not theoretical:
- The repo **only warns** about group/world-readable config files (`config.ts:315`, `mode & 0o077`) — it does **not** reject them. So a loose-perm `config.json` is an accepted state.
- `loadConfig`'s own migration message points operators to copy `config.example.json` into place; a `cp` under a default umask lands the file at **0644**. Running `configure` on it then exposes the key.
- The window is normally brief, but if the process is killed/crashes between the write and the `chmodSync`, the secret stays world-readable **permanently**.

**Fix (small):** write atomically with the tight mode up front — write to a sibling temp file created at `0600`, then `renameSync` over the target (followed by a defensive `chmodSync`). This also closes the non-atomic-corruption concern in §3 in the same change.

## 2. Non-blocking — recommend addressing

### API key accepted as a command-line argument
`src/cli.ts:111-114, 315, 341`
`--api-key <key>` is captured from `argv`, advertised in usage, and takes precedence over the env var. A secret in `argv` is visible via `ps` and often lands in shell history / installer logs. Partially mitigated by the shipped `CC_OCTO_CONFIGURE_API_KEY` fallback (the correct channel for daemon-driven install). Prefer the env var / stdin as canonical; consider de-advertising the flag or noting the exposure in help text.

### `sdk.apiKey` forwarded without runtime type validation
`src/config.ts:143` / `src/agent-bridge.ts:44`
`loadConfig` validates `sdk.anthropicBaseUrl` (SSRF) but does not type-check `sdk.apiKey`. `buildSdkEnv` forwards any truthy value into `ANTHROPIC_API_KEY`, so a malformed config (`"apiKey": 123`, or an object) reaches the SDK subprocess coerced. Low impact (operator-controlled), but a non-empty-string guard at load time fails fast.

### Existing-config root assumed to be an object
`src/configure.ts:26-31`
A valid-JSON-but-non-object root (e.g. `null`) reaches the merge and throws a less intentional error. The `sdk` block is already narrowed defensively; doing the same for the root (`parsed && typeof parsed === 'object' && !Array.isArray(parsed)`) keeps the failure clean.

## 3. Nits
- **Non-atomic in-place overwrite** (`configure.ts:44`): a crash/OOM/full-disk mid-write truncates `config.json`. The temp-file + rename fix above resolves this too.
- **Trailing flag with no value silently swallowed** (`cli.ts:107-114`): `--gateway-url`/`--api-key` as the final token sets `undefined` (`rest[++i]` OOB), surfacing as `"... is required"` rather than `"missing value for ..."`. Minor UX.

## 4. Informational (not a blocker)
**SSRF on the gateway URL:** `configure` validates `--gateway-url` with `isAllowedApiUrl`, which permits any `https:` host without DNS resolution. This is **not introduced by this PR** — it is the established repo-wide policy already applied to `apiUrl` and `sdk.anthropicBaseUrl` in `loadConfig`, with the DNS-rebinding residual risk documented in `url-policy.ts`. `configure` correctly mirrors the existing check. Stricter gateway validation (DNS + private-range rejection) would be a repo-wide hardening item, not a gate on this PR.

## 5. Coverage notes
- Ran the full vitest suite (1035 pass), type-check, and lint locally — all green at the head SHA.
- Confirmed `buildSdkEnv` is behavior-equivalent to the prior inline logic in `queryAgent` (base `ed2859f`), with `ANTHROPIC_API_KEY` the only added injection and `undefined` returned when nothing needs adding.
- Empirically confirmed `writeFileSync({mode})` does not re-apply perms to a pre-existing file.
- Not exercised: daemon hot-reload of `config.json` after `configure` (install flow is assumed to `configure` before `start`); external lifecycle of `CC_OCTO_CONFIGURE_API_KEY` prior to invocation.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#150 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo #150
**Head SHA:** `b7ae49e35b31f189fabcd169dcb2f10d161b6258`
**Scope:** automated review / automated review

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `buildSdkEnv` pure extraction | ✅ | `src/agent-bridge.ts:22-47` — correctly narrows `Pick<Config['sdk'], ...>`, returns `undefined` when nothing to inject, spread order preserves backward compat |
| `configure()` atomic write | ✅ | `src/configure.ts:56-67` — temp file with 0600 → renameSync → chmodSync belt-and-suspenders; cleanup on error |
| `configure()` config shape validation | ✅ | `src/configure.ts:30-40` — rejects null/array/number roots with clear error |
| `configure()` SSRF guard | ✅ | `src/configure.ts:22-24` — reuses `isAllowedApiUrl` (same policy as `loadConfig`'s `anthropicBaseUrl` check) |
| CLI flag parsing guards | ✅ | `src/cli.ts:103-120` — rejects missing values and `--` prefixed values for space form |
| Env var fallback chain | ✅ | `src/cli.ts:348` — `apiKey ?? process.env.CC_OCTO_CONFIGURE_API_KEY ?? ''`; CLI flag wins over env |
| Test coverage | ✅ | 3 test files (agent-bridge-env, cli, configure) covering injection matrix, arg parsing, env fallback, atomic write, shape validation, perms |
| Backward compatibility | ✅ | When `sdk.apiKey` is unset, `buildSdkEnv` returns `undefined` (no env override) — identical to pre-PR behavior |

## 2. Findings

### P2-1: `parseArgs` processes `--gateway-url` / `--api-key` for all subcommands

**Diff-scope:** new (introduced by this PR)
**Severity:** P2 (nit/maintainability)

`parseArgs` (`src/cli.ts:95-127`) parses `--gateway-url` and `--api-key` regardless of the subcommand. A user running `cc-channel-octo start --gateway-url https://x` would silently consume the flags with no effect. Not a bug — the values are only read in the `configure` case — but could confuse users who typo a flag on the wrong subcommand and expect an error.

**Suggestion:** Consider gating these flags behind `cmd === 'configure'` or emitting a warning when they're set for non-configure commands. Low priority.

### P2-2: `--api-key=` (equals form) accepts empty string without parse-level guard

**Diff-scope:** new (introduced by this PR)
**Severity:** P2 (defense-in-depth gap, not exploitable)

The space form `--api-key <value>` guards against missing/flag values at parse time (`src/cli.ts:113-116`). The equals form `--api-key=<value>` (`src/cli.ts:118-119`) does not — `--api-key=` yields an empty string. The downstream `configure()` function catches this (`if (!apiKey) throw`), so the user still gets exit code 2 with a clear error. Defense-in-depth only.

### P2-3: Temp file path predictability

**Diff-scope:** new (introduced by this PR)
**Severity:** P2 (negligible risk for CLI tool)

`src/configure.ts:58`: `const tmpPath = \`${path}.tmp-${process.pid}-${Date.now()}\`` is predictable. In a shared-directory scenario, an attacker could pre-create a symlink at that path. Mitigated by: (a) config lives in `~/.cc-channel-octo/` (user-owned, typically 0700), (b) `writeFileSync` with `{ mode: 0o600 }`, (c) atomic `renameSync` replaces the symlink. Acceptable for a CLI tool's threat model.

## 3. Suggestions

1. **P2-1 fix direction:** Add `if (cmd !== 'configure' && (gatewayUrl || apiKey))` warning in `run()`, or move flag parsing into a configure-specific branch.
2. **P2-2 fix direction:** Add `if (!a.slice('--api-key='.length)) throw` for the equals form, mirroring the space form guard.

## 4. Additional Observations

- **Commits are well-structured:** 4 commits with clear separation (env injection → configure cmd → env var fallback → hardening). The last commit (atomic write + shape validation + flag guards) was co-authored with Claude Opus and addresses review feedback.
- **`buildSdkEnv` spread order** (`baseEnv → extraEnv → ANTHROPIC_BASE_URL → ANTHROPIC_API_KEY`) means `sdk.apiKey` wins over `sdk.env.ANTHROPIC_API_KEY`. This is the correct precedence (dedicated field > generic env) and matches the pre-PR behavior for `ANTHROPIC_BASE_URL`.
- **No `loadConfig` validation for `sdk.apiKey`:** Unlike `anthropicBaseUrl` which is SSRF-checked at load time, `apiKey` has no load-time validation. This is correct — `apiKey` is an opaque credential, not a URL. The only validation needed is non-empty (handled by `configure()`).

## 5. Data Flow Trace

| Consumed Data | Upstream Source | Reaches Consumer? |
|---|---|---|
| `sdk.apiKey` → `ANTHROPIC_API_KEY` env | `config.json` `sdk.apiKey` field → `loadConfig()` → `config.sdk.apiKey` → `buildSdkEnv()` → SDK `options.env` | ✅ Yes, when set. When unset, `buildSdkEnv` returns `undefined` and SDK inherits `process.env` (backward compat). |
| `sdk.anthropicBaseUrl` → `ANTHROPIC_BASE_URL` env | Same path as before PR. `buildSdkEnv` preserves the existing injection. | ✅ Yes, unchanged behavior. |
| `sdk.env` → extra env vars | `config.json` `sdk.env` field → `buildSdkEnv()` spread | ✅ Yes, unchanged. Spread before `ANTHROPIC_BASE_URL`/`ANTHROPIC_API_KEY` so dedicated fields win. |
| `CC_OCTO_CONFIGURE_API_KEY` env var | `process.env` → `cli.ts:348` `resolvedApiKey` → `configure()` | ✅ Yes, when `--api-key` flag is absent. CLI flag takes precedence. |
| `--gateway-url` flag value | `parseArgs()` → `gatewayUrl` → `configure()` → `isAllowedApiUrl()` check → config file | ✅ Yes. SSRF-checked before write. |
| Config file existing content | `readFileSync` → `JSON.parse` → shape validation → merge | ✅ Yes. Merge preserves all non-sdk fields. sdk block is narrowed and merged. |

- **C1 — Dual-path parity:** N/A. `configure` is a write-only path; `loadConfig` is the read path. They operate on the same file but serve different purposes (install-time vs runtime). No symmetric add/remove pair.
- **C2 — Control-flow ordering / nested reuse:** Clear. `buildSdkEnv` is called once in `queryAgent`. No nested/double invocation risk. Flag parsing guards use simple `startsWith('--')` check — tested with non-canonical inputs (flag-as-value).
- **C3 — Authorization boundary ≠ capability boundary:** N/A. `configure` is a local CLI command writing to a user-owned file. No remote endpoint or capability exposure.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth hierarchy in this change.
- **C5 — Build ≠ runtime path:** N/A. Node.js CLI tool — no browser extension, no relative URL resolution, no build artifact mismatch risk.
- **C6 — Governance/policy doc consistency:** N/A. No governance or policy documentation changes.

## 7. Cross-round Blocker Recheck (R6)

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. Three P2 nits (flag scope, equals-form empty guard, temp path predictability) — all defense-in-depth or maintainability concerns with no production impact. The PR is well-structured with solid test coverage, correct atomic write semantics, proper SSRF guard reuse, and clean backward compatibility.

Jerry-Xin (approved):
The PR is relevant to this repository and cleanly extends the existing CLI/config/SDK bridge behavior for one-click gateway credential setup.

💬 Non-blocking

🟡 Warning: `parseArgs()` throws before `run()` enters the command switch, so malformed configure invocations such as `configure --gateway-url` or `configure --api-key --gateway-url https://gw` escape the `configure` error handling and are handled only by the top-level script catch. That means direct `run()` callers get a rejection, and the CLI exits `1` instead of the command-style `2` used at [src/cli.ts:351](/Users/mlamp/.openclaw/tmp/pr-review-pxip6tms/repo/src/cli.ts:351). Consider wrapping argument parsing in `run()` or moving configure flag validation into the configure branch. See [src/cli.ts:107](/Users/mlamp/.openclaw/tmp/pr-review-pxip6tms/repo/src/cli.ts:107) and [src/cli.ts:334](/Users/mlamp/.openclaw/tmp/pr-review-pxip6tms/repo/src/cli.ts:334).

🔵 Suggestion: `configure()` silently replaces a non-object existing `sdk` value with a new object at [src/configure.ts:44](/Users/mlamp/.openclaw/tmp/pr-review-pxip6tms/repo/src/configure.ts:44). That is probably acceptable for recovery, but if the intent is strict preservation, rejecting invalid `sdk` shape would make corruption more visible.

✅ Highlights

The new `configure` backend is appropriately independent from `loadConfig()`, validates the gateway URL through the existing URL policy, preserves unrelated config fields, writes atomically, and forces `0600` permissions at [src/configure.ts:21](/Users/mlamp/.openclaw/tmp/pr-review-pxip6tms/repo/src/configure.ts:21) and [src/configure.ts:57](/Users/mlamp/.openclaw/tmp/pr-review-pxip6tms/repo/src/configure.ts:57).

The SDK env extraction is a good testable refactor, and the precedence is sensible: inherited env, then configured extra env, then gateway URL/key overrides at [src/agent-bridge.ts:40](/Users/mlamp/.openclaw/tmp/pr-review-pxip6tms/repo/src/agent-bridge.ts:40).

I attempted to run the relevant tests, but this checkout does not have dependencies installed: `vitest: command not found`.

yujiawei (approved):
# Code Review — PR #150 (cc-channel-octo)

Reviewed at head SHA `b7ae49e35b31f189fabcd169dcb2f10d161b6258` against base `feat/agent-runtime`. Repo checked out at head; build, type-check, lint, and the new tests were run locally.

## 1. Verdict summary

**APPROVE.** The change is well-scoped and backward compatible: a new `configure` subcommand + a new optional `sdk.apiKey` field + an extracted, unit-tested `buildSdkEnv`. No P0/P1 regression is introduced. The findings below are P2 hardening suggestions, not merge blockers.

Local verification (head SHA):
- `npm run build` ✅
- `npm run type-check` ✅
- `npm run lint` (`--max-warnings 0`) ✅
- New tests: `configure.test.ts` (10) + `cli.test.ts` (32) + `agent-bridge-env.test.ts` (4) = 46 passed ✅

## 2. What was verified

- **`buildSdkEnv` refactor is faithful** (`src/agent-bridge.ts:213`, used at `:254`). The gate only *adds* the `apiKey` branch: old `(anthropicBaseUrl || hasExtraEnv)` → new `(anthropicBaseUrl || apiKey || hasExtraEnv)`. For every pre-existing config (no `apiKey`) behavior is identical. Spread order is preserved and `ANTHROPIC_API_KEY` is layered last, so config wins over inherited env — consistent with how `ANTHROPIC_BASE_URL` is handled.
- **`configure` is correctly independent of `loadConfig`** (`src/configure.ts`). Raw read-merge-write touching only `sdk.anthropicBaseUrl` + `sdk.apiKey`, so install can run before `apiUrl`/a bot exists. Root + `sdk` are both narrowed to plain objects before merge; non-object roots throw a clear error.
- **Atomic write + perms**: temp file written `mode 0o600`, renamed, then `chmod 0o600` belt-and-suspenders; temp file cleaned up on failure. Final mode is asserted `0o600` even when an existing file had `0644`.
- **No secret logging**: the API key is never `console.log`'d or stringified into any log path. The success line prints a fixed string only (`src/cli.ts:353`).
- **Off-argv path exists**: `configure` reads `CC_OCTO_CONFIGURE_API_KEY` when `--api-key` is omitted (`src/cli.ts:349`), keeping the key out of `ps`/shell history for the daemon-driven install.

## 3. Findings (all P2 — non-blocking)

**P2-1 — SSRF policy admits `https://localhost` / private DNS names** (`src/url-policy.ts:257` `isAllowedApiUrl`, reused at `src/configure.ts:21`).
`isAllowedApiUrl` only IP-checks the host for `https:`; an https hostname that resolves to a private/loopback address (e.g. `https://localhost`, internal corp DNS) is accepted, so the persisted key + all model traffic could be routed to an internal endpoint. This is **pre-existing** — the same function already gates `apiUrl` and `sdk.anthropicBaseUrl` in `loadConfig` (`src/config.ts:411,418`), and this PR deliberately reuses it for consistency. Not a regression, so not a blocker here. Follow-up: consider hardening the shared policy (DNS-resolve like `assertPublicUrl`, or reject `localhost`/private hostnames) repo-wide rather than only at this call site.

**P2-2 — `--api-key <key>` in argv remains a documented path** (`src/cli.ts:323` usage string; precedence asserted in `cli.test.ts`).
The secure off-argv channel (`CC_OCTO_CONFIGURE_API_KEY`) is the right pattern and is present. The remaining `--api-key` flag still exposes the secret via process listing / shell history if a caller uses it. Suggestion: in the usage text, steer secrets to the env var (or stdin) and document `--api-key` as test/convenience only, so the daemon install path doesn't drift back to argv.

**P2-3 — whitespace-only `apiKey`/`gatewayUrl` passes validation** (`src/configure.ts:17-18`).
`if (!apiKey)` treats `"   "` as truthy, so a whitespace key is written and later overwrites any valid inherited `ANTHROPIC_API_KEY` with spaces via `buildSdkEnv`. Cheap fix: validate `value.trim().length === 0` (and trim before storing).

**P2-4 — missing-value guard only rejects `--` tokens** (`src/cli.ts:109,116`).
`--gateway-url https://gw --api-key -f` writes `-f` as the API key. Use `next.startsWith('-')` to also catch single-dash flags as a missing value.

**P2-5 — predictable temp path without exclusive create** (`src/configure.ts:55-57`).
`${path}.tmp-${pid}-${Date.now()}` is guessable and opened without `O_EXCL`; a pre-created symlink/stale file at that path could receive the secret or defeat the intended `0600`. Low risk in a user-owned `~/.cc-channel-octo`, but `openSync(tmp, 'wx', 0o600)` (random suffix) closes it cleanly.

## 4. Coverage notes

- Independent local run reproduced the PR's green claims (build/type-check/lint/tests) at the exact head SHA.
- Not verified: the actual daemon one-click-install invocation (whether it uses the env var vs `--api-key`) — that lives outside this repo, so P2-2's real-world exposure can't be confirmed from the diff. Recommend the installer use `CC_OCTO_CONFIGURE_API_KEY`.
- Cross-file: confirmed `config.ts` `apiKey` field, `agent-bridge` consumption, and `cli.ts` wiring are mutually consistent; `loadConfig` does not (and need not) validate `sdk.apiKey` beyond the URL.

None of the above blocks merge. Approving; the P2 items are good cheap follow-ups, with P2-3 the most worthwhile.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#150 Review Report
**PR:** https://github.com/Mininglamp-OSS/cc-channel-octo/pull/150
**Head SHA:** `d6636a8659a3fc583e53c83054ab61b4e8d38ae0`
**Base:** `feat/agent-runtime`
**Scope:** +314/-23 across 7 files (new `configure` subcommand + `sdk.apiKey` config field + `buildSdkEnv` extraction)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `configure` writes gateway URL + API key to config | ✅ | `src/configure.ts:18-69` — read-merge-write with atomic temp+rename |
| SSRF policy reused for gateway URL | ✅ | `src/configure.ts:23` — calls `isAllowedApiUrl()` from `url-policy.ts` |
| `sdk.apiKey` forwarded to SDK subprocess as `ANTHROPIC_API_KEY` | ✅ | `src/agent-bridge.ts:44` — `buildSdkEnv` layers `ANTHROPIC_API_KEY` last (wins over base env) |
| `buildSdkEnv` refactoring is behaviorally equivalent | ✅ | Compared base branch inline code (`agent-bridge.ts:228-243` on base) vs extracted function — same spread order, same conditional, adds only `apiKey` injection |
| Atomic write prevents key exposure under broader perms | ✅ | `src/configure.ts:57-67` — `writeFileSync(tmpPath, ..., { mode: 0o600 })` → `renameSync` → `chmodSync` |
| Config shape validation (null/array/number) | ✅ | `src/configure.ts:33-34` — `!(parsed && typeof parsed === 'object' && !Array.isArray(parsed))` |
| Flag value guards (missing / flag-like) | ✅ | `src/cli.ts:101-104, 110-113` — throws when next token is `undefined` or starts with `--` |
| Env var fallback + CLI precedence | ✅ | `src/cli.ts:348` — `apiKey ?? process.env.CC_OCTO_CONFIGURE_API_KEY ?? ''` |
| Test coverage | ✅ | 3 new test files + additions: 23 + 82 + 65 = 170 new test lines covering configure, CLI, and buildSdkEnv |

---

## 2. Findings

### No P0 or P1 issues found.

### P2 — Minor Observations

**P2-1: `--gateway-url=` (equals form with empty value) guard message is generic**
- `src/cli.ts:110`: the `--gateway-url=` form with empty value (`--gateway-url=`) is NOT caught by the explicit guard at line 101-104 (which only covers the space form). It flows through as `gatewayUrl = ''` and is caught by `configure()`'s `if (!gatewayUrl)` throw at `configure.ts:19`. The resulting error message is `"configure: --gateway-url is required"` — correct but less specific than the parse-level `"requires a value"` message. Same applies to `--api-key=`.
- **Impact:** Negligible — user still gets a clear error and exit code 2. Not a functional gap, just a minor UX inconsistency.

**P2-2: No API key format validation**
- **Diff-scope:** new
- `configure()` accepts any non-empty string as an API key. A typo or truncated key is persisted to disk and forwarded to the SDK, which will fail at auth time with a potentially confusing error.
- **Impact:** Low — the SDK will reject the key at first use. Adding format validation (e.g. `sk-ant-` prefix check) would be a UX improvement but is not a correctness issue. The PR intentionally avoids coupling to a specific key format.

**P2-3: `configure` doesn't clean up `CC_OCTO_CONFIGURE_API_KEY` from `process.env` after reading**
- **Diff-scope:** new
- `src/cli.ts:348`: after reading the env var, it remains in the process environment. For a short-lived CLI command this is harmless, but if the process were long-lived, the key would persist in memory.
- **Impact:** Negligible — the CLI exits immediately after configure completes.

---

## 3. Suggestions

1. Consider adding a test for the `--gateway-url=` (equals form, empty value) edge case to `cli.test.ts` — currently only the space form's missing-value guard is tested.
2. Optional: add a `delete process.env.CC_OCTO_CONFIGURE_API_KEY` after reading it in `cli.ts:348` as defense-in-depth (prevents accidental re-read or leakage to subprocess env).

---

## 4. Additional Findings

- **Multi-bot propagation verified:** `resolveBotConfigs()` in `config.ts` spreads `config.sdk` into each bot's resolved config (`sdk: { ...config.sdk, ...(perBotFile.sdk ?? {}), ... }`). The new `apiKey` field propagates correctly from global config to all bots. Per-bot configs can override it. No issue.
- **Backward compatibility confirmed:** `apiKey` is optional in the `Config` interface (`apiKey?: string`). `buildSdkEnv` returns `undefined` when nothing needs injection (same as the old inline code omitting `env`). Existing configs without `apiKey` work unchanged.

---

## 5. Data-Flow Tracing

|---|---|---|
| `gatewayUrl` in `configure()` | `parseArgs()` → `cli.ts:348` → `configure(gatewayUrl ?? '', ...)` | ✅ Empty string caught by `if (!gatewayUrl)` at `configure.ts:19` |
| `resolvedApiKey` in `configure()` | `apiKey ?? process.env.CC_OCTO_CONFIGURE_API_KEY ?? ''` → `configure(..., resolvedApiKey, ...)` | ✅ Empty string caught by `if (!apiKey)` at `configure.ts:20` |
| `sdk.apiKey` in `buildSdkEnv()` | `config.sdk` → `buildSdkEnv(config.sdk, process.env)` at `agent-bridge.ts:254` | ✅ `Pick<>` type narrows to `apiKey | anthropicBaseUrl | env`; layered last in spread → wins |
| `ANTHROPIC_API_KEY` in SDK subprocess | `buildSdkEnv()` return value → `sdk_query({ options: { env } })` | ✅ SDK `env` option replaces subprocess env; `buildSdkEnv` spreads `baseEnv` first to preserve PATH/HOME |
| `existing` config in `configure()` | `JSON.parse(readFileSync(path))` → validated as plain object → merged | ✅ Shape validation at line 33-34 prevents prototype pollution from array/null/number |
| `isAllowedApiUrl(gatewayUrl)` | `url-policy.ts` — checks `https:` to non-private host, or `http:` to localhost only | ✅ Same SSRF policy as `apiUrl` and `anthropicBaseUrl` in `loadConfig` |

---

## 6. Blind Spot Checklist (R5)

- **C1 — Dual-path parity:** N/A. `configure` is a standalone write command. No symmetric add/remove or create/delete paths.
- **C2 — Control-flow ordering / nested reuse:** Clear. `buildSdkEnv` has a single call-site (`agent-bridge.ts:254`). Spread order in the returned object is correct: `baseEnv → extraEnv → ANTHROPIC_BASE_URL → ANTHROPIC_API_KEY` (later wins). No double-application risk.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. Local CLI tool — the authorization boundary is OS-level file permissions (`chmod 600`). No remote endpoint exposure.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth hierarchy in this change.
- **C5 — Build ≠ runtime path:** N/A. Standard Node.js CLI tool — no browser extension, no relative URL resolution, no packaging concerns.
- **C6 — Governance/policy document consistency:** N/A. Code-only PR.

---

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 issues found. The PR is well-structured with proper security hygiene (atomic write, 0600 perms, SSRF policy reuse, env-var fallback to keep keys out of `ps`), clean refactoring (`buildSdkEnv` extraction is behaviorally equivalent), and solid test coverage (170 new test lines). The two P2 observations are minor UX nits, not correctness or security issues.

yujiawei (approved):
# Code Review — PR #150 (cc-channel-octo)

Reviewed at head `d6636a8` against base `feat/agent-runtime` (merge-base `ed2859f`). Verified locally: `tsc --noEmit` clean, eslint clean on changed files, and the new/affected suites pass (`configure.test.ts` 10, `agent-bridge-env.test.ts` 4, `cli.test.ts` 32).

## Summary

This is a clean, well-scoped change. It adds a `configure` subcommand plus an `sdk.apiKey` field so the gateway's LLM credentials can be written programmatically before any bot/apiUrl exists. The `buildSdkEnv` extraction is a genuine improvement: it makes the env-injection matrix pure and unit-testable, and preserves the SDK's `env`-replacement semantics (spread base env first, then operator vars, then gateway routing). `configure` does an atomic temp-write + rename, forces mode `0600`, validates the JSON root shape, creates the parent dir, and reuses the existing SSRF policy (`isAllowedApiUrl`) for the gateway URL. Test coverage for the new paths is thorough (fresh write, merge-preserve, mode 600, non-object root, unsafe URL, env-var fallback, CLI precedence, missing-arg exit 2).

No blocking (P0/P1) issues found. The items below are advisory hardening/UX suggestions; none should hold up merge.

## Findings (all advisory / non-blocking)

### 1. `configure` accepts `https://localhost` and DNS names resolving to private IPs (P2)
`src/configure.ts:21` gates the gateway URL with `isAllowedApiUrl`, which only rejects **private IP literals** for `https:` — it accepts `https://localhost:8443` and any public hostname that resolves to an RFC1918/metadata address (no DNS resolution at this layer). So a daemon-supplied gateway can still point at a local/internal target that then receives `ANTHROPIC_API_KEY` and prompt traffic.

Worth noting for context: this is **not a regression** — it is the exact same policy `loadConfig` already applies to `apiUrl` and `sdk.anthropicBaseUrl` (`src/config.ts:411-426`), and `url-policy.ts` documents the boot-time-vs-fetch-time split (DNS-resolving `assertPublicUrl` is the fetch-path guard, not this boot-time check). The PR faithfully mirrors the established contract, which is the right call for consistency. If tightening is desired, it should be done repo-wide in `isAllowedApiUrl`, not just here. Hence advisory, not blocking.

### 2. `--api-key` in argv exposes the secret to local `ps`/`/proc` (P2)
`src/cli.ts:115-122` supports `--api-key <key>`, which places the secret in process argv — readable by other local users on many POSIX systems until the process exits, before the `0600` config file ever protects it. The env-var path (`CC_OCTO_CONFIGURE_API_KEY`) already exists and is the safer default for daemon-driven installs. Suggest documenting the env var as the preferred path (and considering whether the argv flag should be discouraged in help text).

### 3. `renameSync` replaces a symlinked `config.json` with a regular file (P2)
`src/configure.ts` does `renameSync(tmpPath, path)`. If `config.json` is a symlink (operators sometimes link configs from a central deploy/CM directory), the rename replaces the **link** with a regular file rather than writing through to the target. Atomicity is preserved either way; this is only a concern for symlink-based config layouts. Optional: resolve the realpath before the temp-write+rename if that deployment shape is supported.

### 4. Space-form `--api-key` rejects values beginning with `--` (P2, minor)
`src/cli.ts:117` rejects a `--api-key` value that starts with `--` (the "next token is another flag" guard). Anthropic keys start with `sk-`, so this is harmless in practice, but a custom gateway issuing `--`-prefixed tokens would hit a spurious "requires a value" error. The `--api-key=<value>` form bypasses the check, so there is a workaround. Minor UX edge.

## Coverage notes

What was inspected: the full diff plus cross-file dependencies in `config.ts` (loadConfig validation, `Config['sdk']` type), `url-policy.ts` (`isAllowedApiUrl` semantics, verified behaviorally), `agent-bridge.ts` (env-replacement consumer of `buildSdkEnv`), `cli.ts` (arg parsing + dispatch + top-level error catch), and the three test files. Type-check, lint, and the affected test suites were run.

Not fully verified: downstream runtime validation of `sdk.apiKey` once forwarded to the SDK subprocess (out of this diff's scope); DNS-rebinding residual risk on the gateway host (documented in `url-policy.ts`, pre-existing); and `DEFAULT_CONFIG_PATH` absolute resolution (unchanged by this PR). DNS-resolving SSRF for the gateway is intentionally a fetch-time concern, not enforced at this write-time layer.

PR #155, opened 2026-06-23
feat(cc-octo): zero-bot idle startup and configurable gateway url/model

caster-Q:
## Summary

Lets a freshly-installed cc-channel-octo gateway come online with zero bots and accept a gateway URL + model from the one-click installer, removing two blockers in the cc adapter install flow: the first-bot chicken-and-egg deadlock and the `/v1/v1/messages` URL doubling.

## Related Issue

Part of Mininglamp-OSS/octo-daemon-cli#65

## Changes

- Zero-bot idle: `resolveBotConfigs` returns `[]` when there is no inline bot, no global token, and no token in the default per-bot file; `main()` then runs an idle keepalive (ref'd timer, clean SIGINT/SIGTERM shutdown) so the runtime stays online until the first bot is provisioned.
- `configure` accepts `--gateway-url`, `--model` (preserved when omitted), and `--api-url`; the API key is read from an env var, never argv.
- Strip a trailing `/v1` from the gateway URL (case-insensitive, trims whitespace) so the SDK's appended `/v1/messages` does not double.

## Testing

- `npm run build && npm test` — 56 files, 1069 tests passing (pre-push hook also runs lint + type-check clean).

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] PR description is in English
- [x] Added tests for my changes
- [x] All tests pass
- [x] Followed commit message conventions (Conventional Commits)

### DIFF
--- a/src/__tests__/cli.test.ts
+++ b/src/__tests__/cli.test.ts
@@ -38,6 +38,21 @@ describe('parseArgs', () => {
     expect(parseArgs(['stop', '--timeout=0']).timeoutMs).toBe(10_000);
     expect(parseArgs(['stop', '--timeout=abc']).timeoutMs).toBe(10_000);
   });
+
+  it('parses --model (space and = forms)', () => {
+    expect(parseArgs(['configure', '--model', 'm1']).model).toBe('m1');
+    expect(parseArgs(['configure', '--model=m2']).model).toBe('m2');
+  });
+
+  it('parses --api-url (space and = forms)', () => {
+    expect(parseArgs(['configure', '--api-url', 'http://127.0.0.1:8090']).apiUrl).toBe('http://127.0.0.1:8090');
+    expect(parseArgs(['configure', '--api-url=https://octo.test']).apiUrl).toBe('https://octo.test');
+  });
+
+  it('throws when --model or --api-url is missing a value', () => {
+    expect(() => parseArgs(['configure', '--model'])).toThrow(/--model requires a value/);
+    expect(() => parseArgs(['configure', '--api-url', '--api-key=k'])).toThrow(/--api-url requires a value/);
+  });
 });
 
 describe('isAlive', () => {
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -141,8 +141,11 @@ describe('required field validation', () => {
     expect(() => loadConfig(path)).toThrow(/apiUrl/);
   });
 
-  it('resolveBotConfigs throws when a bot has no token (single-bot)', () => {
-    const path = writeConfig({ apiUrl: 'https://api.test' });
+  it('resolveBotConfigs throws when a listed bot has no token', () => {
+    // A bots[] entry whose token is absent (no inline token, no per-bot file)
+    // still fails fast. Only the truly empty case (no bots[], no global token,
+    // no default per-bot token) idles — see the "zero-bot idle" suite.
+    const path = writeConfig({ apiUrl: 'https://api.test', bots: [{ id: 'a' }] });
     expect(() => resolveBotConfigs(loadConfig(path))).toThrow(/missing botToken/);
   });
 
@@ -695,5 +698,34 @@ describe('v1.1: SOUL.md personality (openclaw-style)', () => {
     });
     const bots = resolveBotConfigs(loadConfig(cfgPath));
     expect(bots[0].sdk.systemPrompt).toBe('file soul wins');
+  });
+});
+
+describe('resolveBotConfigs zero-bot idle', () => {
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it('returns [] when no bots[], no global botToken, no default per-bot token', () => {
+    // A freshly-installed gateway has only gateway/apiUrl, no bot yet.
+    const cfg = loadConfig(writeConfig({ apiUrl: 'https://api.test' }));
+    expect(resolveBotConfigs(cfg)).toEqual([]);
+  });
+
+  it('still returns one default bot when a global botToken is set', () => {
+    const cfg = loadConfig(writeConfig({ apiUrl: 'https://api.test', botToken: 'bf_legacy' }));
+    const bots = resolveBotConfigs(cfg);
+    expect(bots).toHaveLength(1);
+    expect(bots[0].botId).toBe('default');
+    expect(bots[0].botToken).toBe('bf_legacy');
+  });
+
+  it('still starts the default bot when only <baseDir>/default/config.json has a token', () => {
+    // No global botToken and no bots[], but a legacy single bot keeps its token
+    // in the default per-bot file — must NOT idle, must start that bot.
+    writeBotConfig('default', { botToken: 'bf_perbot' });
+    const cfg = loadConfig(writeConfig({ apiUrl: 'https://api.test' }));
+    const bots = resolveBotConfigs(cfg);
+    expect(bots).toHaveLength(1);
+    expect(bots[0].botToken).toBe('bf_perbot');
   });
 });
--- a/src/__tests__/configure.test.ts
+++ b/src/__tests__/configure.test.ts
@@ -2,7 +2,7 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest'
 import { mkdtempSync, rmSync, readFileSync, writeFileSync, statSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
-import { configure } from '../configure.js'
+import { configure, normalizeGatewayUrl } from '../configure.js'
 
 let dir: string
 let cfgPath: string
@@ -61,5 +61,23 @@ describe('configure', () => {
   it('throws a clear error when existing config root is a number', () => {
     writeFileSync(cfgPath, JSON.stringify(42))
     expect(() => configure('https://gw', 'sk', cfgPath)).toThrow(/is not a JSON object/)
+  })
+  it('strips a trailing /v1 from the stored gateway url', () => {
+    configure('https://gw.test/v1', 'sk-test', cfgPath)
+    expect(JSON.parse(readFileSync(cfgPath, 'utf-8')).sdk.anthropicBaseUrl).toBe('https://gw.test')
+  })
+})
+
+describe('normalizeGatewayUrl', () => {
+  it('strips a trailing /v1 or /v1/', () => {
+    expect(normalizeGatewayUrl('https://gw.test/v1')).toBe('https://gw.test')
+    expect(normalizeGatewayUrl('https://gw.test/v1/')).toBe('https://gw.test')
+  })
+  it('leaves a bare host or non-version path intact', () => {
+    expect(normalizeGatewayUrl('https://gw.test')).toBe('https://gw.test')
+    expect(normalizeGatewayUrl('https://gw.test/api')).toBe('https://gw.test/api')
+  })
+  it('does not strip a mid-path v1', () => {
+    expect(normalizeGatewayUrl('https://gw.test/v1/foo')).toBe('https://gw.test/v1/foo')
   })
 })
--- a/src/__tests__/configure.test.ts
+++ b/src/__tests__/configure.test.ts
@@ -66,6 +66,22 @@ describe('configure', () => {
     configure('https://gw.test/v1', 'sk-test', cfgPath)
     expect(JSON.parse(readFileSync(cfgPath, 'utf-8')).sdk.anthropicBaseUrl).toBe('https://gw.test')
   })
+  it('writes sdk.model when a model is provided', () => {
+    configure('https://gw.test', 'sk', cfgPath, { model: 'vertexai/claude-opus-4-8' })
+    expect(JSON.parse(readFileSync(cfgPath, 'utf-8')).sdk.model).toBe('vertexai/claude-opus-4-8')
+  })
+  it('PRESERVES an existing sdk.model when no model is provided', () => {
+    writeFileSync(cfgPath, JSON.stringify({ sdk: { model: 'old/model' } }))
+    configure('https://gw.test', 'sk', cfgPath) // no model → keep existing
+    expect(JSON.parse(readFileSync(cfgPath, 'utf-8')).sdk.model).toBe('old/model')
+  })
+  it('writes the top-level apiUrl when provided', () => {
+    configure('https://gw.test', 'sk', cfgPath, { apiUrl: 'http://127.0.0.1:8090' })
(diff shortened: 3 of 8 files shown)


OctoBoooot (commented):
## Review: feat(cc-octo): zero-bot idle startup and configurable gateway url/model (#155)

**Verdict:** Approve with comments (no blockers, no majors).

CI is red — but the failing check is `check-sprint / check-sprint` (sprint-board attachment), not a code/test failure. The TS, lint, and Vitest jobs are all green. Worth linking the PR to project #2 to clear the badge before merge.

The change is small, well-scoped, and adversarially survives the things that could have gone wrong: the idle branch keeps Node alive correctly, `normalizeGatewayUrl` is bounded to a trailing `/v1` and won't munge mid-path segments, the new `--api-url` reuses the existing `isAllowedApiUrl` SSRF policy, and re-configure preserves `sdk.model`/top-level `apiUrl` via the existing spread. Tests cover the new code paths I'd ask for.

### Praise
- `src/index.ts:55-68` — the idle keepalive uses a **ref'd** `setInterval` (not a pending Promise, not `unref()`) with an explicit comment explaining why a pending Promise alone wouldn't hold the event loop. That's exactly the trap to flag for future readers.
- `src/configure.ts:78-83` — `if (opts?.model)` with the inline comment "omitting it PRESERVES any existing sdk.model" makes the re-configure-as-key-rotation contract explicit. Without that comment a future reader would be tempted to "fix" it to always-write.
- `src/configure.ts:39-43` — funneling the new `--api-url` through `isAllowedApiUrl` rather than reinventing a check keeps the SSRF policy single-sourced.

- `parseArgs` accepts empty values for `--model=` / `--api-url=` (and the pre-existing `--api-key=`, `--gateway-url=`). Same pattern across the codebase, so not worth changing in this PR; if you ever want to tighten it, do it for all four at once.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#155 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo #155
**Head SHA:** `241bc0896d9ae62c6ef07dc05d0ddfc53beb5e5e`
**Title:** feat(cc-octo): zero-bot idle startup and configurable gateway url/model

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Zero-bot idle: `resolveBotConfigs` returns `[]` | ✅ | `src/config.ts:498-508` — checks `hasInlineBots`, `config.botToken`, then `default/config.json` per-bot token before returning `[]` |
| Idle keepalive keeps process alive | ✅ | `src/index.ts:63-76` — ref'd `setInterval` (not `unref'd`), SIGINT/SIGTERM handlers clear interval and resolve |
| `/v1` stripping (case-insensitive, trim) | ✅ | `src/configure.ts:20-22` — `raw.trim().replace(/\/v1\/?$/i, '')` |
| `--model` preserved when omitted | ✅ | `src/configure.ts:76-78` — conditional write; `existingSdk` spread preserves prior `sdk.model` |
| `--api-url` SSRF validation | ✅ | `src/configure.ts:42-44` — `isAllowedApiUrl(opts.apiUrl)` check, same policy as gateway URL |
| CLI arg parsing for new flags | ✅ | `src/cli.ts:129-146` — follows existing `--gateway-url`/`--api-key` pattern with missing-value throws |
| Test coverage | ✅ | 56 files, 1069 tests (per PR body); new suites cover zero-bot idle, normalizeGatewayUrl, model preservation, unsafe apiUrl rejection |

## 2. Findings

### No P0 or P1 findings.

### P2 — Minor observations

**P2-1: Double read of `default/config.json` at startup**
- **What:** `resolveBotConfigs` calls `readConfigFile(pathJoin(config.baseDir, 'default', 'config.json'))` at line 503 for the zero-bot check. If the zero-bot path is NOT taken (i.e., there IS a default per-bot token), the same file is read again at line 530 (`perBotFile`) inside the entries loop for the synthesized `default` entry.
- **Diff-scope:** New (introduced by this PR's zero-bot early-return guard).
- **Impact:** Negligible — startup-only, same file, synchronous read. Not blocking.
- **Suggestion:** Could restructure to reuse the read result, but not worth the complexity for a one-time startup path.

**P2-2: `normalizeGatewayUrl` does not strip `/v1` before a query string or fragment**
- **What:** `raw.trim().replace(/\/v1\/?$/i, '')` — the `$` anchor means `https://gw.test/v1?foo=bar` would NOT be stripped. After `.trim()`, the `?foo=bar` is still trailing, so the regex doesn't match.
- **Diff-scope:** New (this PR's function).
- **Impact:** Very low — gateway URLs with query strings are not a realistic install scenario. The URL still works (just with the doubled `/v1/v1/messages` path if the SDK appends). Not blocking.
- **Suggestion:** If desired, parse with `new URL()` and strip `/v1` from `pathname` only. But the current regex approach is simpler and covers all realistic inputs.

**P2-3: Idle mode has no external liveness signal**
- **What:** In idle mode (`src/index.ts:60-77`), the process stays alive with a `setInterval` keepalive but opens no WebSocket, registers no bot, and exposes no health endpoint. The supervisor sees the process as "running" (PID alive), but the Octo platform has no way to know the gateway is in idle-vs-serving state.
- **Diff-scope:** New (this PR's idle path).
- **Impact:** Low — the daemon that provisions the first bot already knows the gateway is freshly installed (it just ran `configure`). The `restart` after provisioning transitions cleanly from idle to serving. Not blocking for this PR's scope.

## 3. Suggestions

- **P2-1** is cosmetic; no action needed unless profiling shows startup latency matters.
- **P2-2** could be hardened with URL parsing, but the regex covers all realistic gateway URL formats. Consider a comment noting the query-string edge case.
- **P2-3** is a future enhancement (e.g., log line or metadata flag for "idle" state), not a landing blocker.

## 4. Extra Findings

- **Atomic write safety preserved:** The `configure()` function's atomic write pattern (temp file + rename, `wx` flag, `chmodSync 0o600`) is unchanged and correctly applied to the new fields. ✅
- **SSRF policy consistency:** Both `--gateway-url` and `--api-url` go through `isAllowedApiUrl()` in `configure()`, and `loadConfig()` re-validates `apiUrl` at boot time. Defense in depth is maintained. ✅
- **API key not on argv:** The API key is read from env var `CC_OCTO_CONFIGURE_API_KEY` (not a new flag), avoiding `/proc/*/cmdline` exposure. ✅ (Pre-existing pattern, correctly preserved.)

## 5. Data-Flow Trace

| Consumed Data | Upstream Source | Runtime Flow Verified |
|---|---|---|
| `botConfigs` (in `main()`) | `resolveBotConfigs(config)` → early-return `[]` when no bots/token/default-per-bot | ✅ `readConfigFile` returns `{}` for missing file → `.botToken` is `undefined` → falsy → returns `[]` |
| `shouldRunIdle(botConfigs)` | `botConfigs.length === 0` | ✅ Pure function, directly checks array length |
| `normalizedUrl` (in `configure()`) | `normalizeGatewayUrl(gatewayUrl)` → `raw.trim().replace(...)` | ✅ Applied after `isAllowedApiUrl` validation; normalization only strips trailing `/v1`, doesn't alter host/scheme |
| `opts.model` → `merged.sdk.model` | CLI `--model` flag → `parseArgs` → `configure(opts)` → conditional write | ✅ Only written when truthy; `existingSdk` spread preserves prior value when omitted |
| `opts.apiUrl` → `merged.apiUrl` | CLI `--api-url` flag → `parseArgs` → `configure(opts)` → `isAllowedApiUrl` gate → conditional write | ✅ SSRF check before write; `loadConfig()` re-validates at boot |
| `defaultPerBot.botToken` (zero-bot guard) | `readConfigFile(<baseDir>/default/config.json)` | ✅ Returns `{}` if file missing; `.botToken` is `undefined` → falsy → idle path taken |

## 6. Blind Spot Checklist (R5 — security_sensitive PR)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths in this PR. The zero-bot guard is a single early-return, not a paired operation. The `configure` function is write-only (no corresponding "unconfigure"). ✅ Clear.
- **C2 — Control-flow ordering / nested reuse:** Clear. `normalizeGatewayUrl` is pure and called once. The `configure()` merge order (spread existing → overwrite specific keys → conditionally add new keys) is correct and tested. No double-application risk. ✅
- **C3 — Authorization boundary ≠ capability boundary:** N/A. This PR doesn't expose new endpoints or tools. The `--api-url` flag writes to config; the SSRF check gates what values are accepted. ✅ Clear.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No auth/authorization changes. The zero-bot idle path is a startup behavior change, not an access control change. ✅ Clear.
- **C5 — Build ≠ runtime path correctness:** Clear. This is a Node.js CLI/gateway (not a browser extension or packaged app). No relative-path or origin issues. The `normalizeGatewayUrl` output is stored in config and later read by `loadConfig()` → forwarded to SDK subprocess via env var. Path is consistent from write to consumption. ✅
- **C6 — Governance/policy document self-consistency:** N/A. No governance or policy document changes in this PR. ✅ Clear.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The PR is well-structured, focused, and thoroughly tested. Security-sensitive surfaces (SSRF validation, API key handling, atomic config writes) are correctly handled and consistent with existing patterns. Three P2 observations are non-blocking (double file read at startup, regex edge case with query strings, no external idle-state signal).

yujiawei (approved):
# Code Review — PR #155 (cc-channel-octo)

**Scope:** Reviewed at head SHA `241bc08`, merge-base `8d32215`. 8 files, +204/-10. Classified security-sensitive (auth/config/SSRF surface). Built and ran the affected test suites locally (`config`, `configure`, `cli`, `index-idle` — 127 tests green); `tsc` and ESLint clean.

## 1. Summary

The PR makes three changes, all coherent and well-tested:

1. **Zero-bot idle startup** — `resolveBotConfigs()` returns `[]` when there is no `bots[]`, no global `botToken`, and no token in `<baseDir>/default/config.json`; `main()` then holds a ref'd `setInterval` keepalive with clean SIGINT/SIGTERM shutdown instead of throwing.
2. **`configure` gains `--model` and `--api-url`** — `--model` writes `sdk.model` (preserved when omitted), `--api-url` writes top-level `apiUrl`. Both go through the same `isAllowedApiUrl` SSRF gate as `--gateway-url`. API key is still read from env/argv only.
3. **`normalizeGatewayUrl()`** strips a trailing `/v1` (case-insensitive, whitespace-trimmed) so the SDK's appended `/v1/messages` doesn't double into `/v1/v1/messages`.

**Verdict: Approve.** The security-relevant invariants hold and the change is appropriately scoped. The items below are non-blocking suggestions.

## 2. Security review (no blocking findings)

- **SSRF gate is correctly applied to both new inputs.** `configure()` validates the raw `--gateway-url` and `--api-url` with `isAllowedApiUrl` *before* normalization (`configure.ts:36`, `configure.ts:42`). I confirmed validation runs on the raw URL, so a private-IP gateway with a trailing `/v1` (e.g. `https://192.168.1.1/v1`) is still rejected before the path is stripped.
- **Secrets handling is sound.** The API key is never interpolated into log/error strings; it's persisted via an atomic `0600` temp-file-then-rename with exclusive (`wx`) create. No regression here.
- **Zero-bot idle does not weaken auth.** The idle branch `return`s before `startBot()`, so an idle gateway registers no bot over REST and opens no WebSocket — it is fully inert, not an unauthenticated listener. This was my main security concern and it is clean.
- **`normalizeGatewayUrl` is correct for this SDK — and the "`/api/v1` corruption" concern is a non-issue.** Because the Anthropic SDK unconditionally appends `/v1/messages` to `ANTHROPIC_BASE_URL`, the stored base must end *before* a `/v1` segment. Stripping `https://gw/api/v1` → `https://gw/api` yields the correct final `https://gw/api/v1/messages`. A stored base ending in `/v1` is always a misconfiguration for this SDK, so the broad regex is the intended behavior, not a bug.

**For a human to spot-check:** that the one-click installer / daemon always passes `--api-url` at install time (see §3, first item) — that contract lives outside this repo.

## 3. Non-blocking suggestions (P2 / nits)

- **`apiUrl` is required at boot but `--api-url` is optional in `configure` (P2 footgun).** `loadConfig()` throws `Missing required config: apiUrl` (`config.ts:408`), which runs *before* `resolveBotConfigs()`, so a fresh install that calls `configure` without `--api-url` crashes at boot rather than idling. Keeping `--api-url` optional is the right call (re-configure for key rotation must not require re-passing it), so this is a documentation/installer-contract concern, not a code defect. The error message is clear and actionable. Consider documenting that the first `configure` on a fresh install must include `--api-url`.
- **`--model=`/`--api-url=` empty-string forms are silent no-ops (P2).** `parseArgs` accepts `--api-url=` → `apiUrl: ''`, which is falsy, so `configure()` skips both the SSRF check and the write. No unsafe value is ever persisted, but the empty `=` form silently does nothing. Consider rejecting empty values in `parseArgs` for symmetry with the space-form (`--api-url` with no value already throws).
- **`readConfigFile` on `default/config.json` can throw on a non-ENOENT FS error (P2 edge case).** The new idle check reads `<baseDir>/default/config.json` (`config.ts:505`); if that file exists but is unreadable (EACCES) it throws and aborts boot. This mirrors pre-existing `readConfigFile` behavior used elsewhere in the resolve path, so it's not a regression, but treating a read failure the same as "absent" in the idle pre-check would make a fresh install more robust.
- **Idle shutdown handler has no `try/finally` (nit).** In `main()`'s `bye()`, if anything before `clearInterval(keepalive)` threw, the timer would leak. `console.log` throwing is effectively impossible, and a second signal still exits via Node's default handler, so this is purely defensive.
- **Idle process is "up" but serves nothing (P2, by design).** A monitor seeing the process alive can't tell it's idle-awaiting-provision vs. ready. Out of scope here, but a future readiness signal (e.g. a health endpoint returning 503 while idle) would improve operability.

## 4. Test coverage notes (non-blocking)

Coverage of the three changes is good. Gaps worth a follow-up, none blocking:
- `configure.test.ts` only asserts scheme rejection (`ftp://`) for `--api-url`; the private-IP/IPv6 SSRF cases are covered at the `url-policy.test.ts` unit level but not exercised through the `configure()` call site.
- No test for the model-overwrite path (existing `sdk.model` + new `--model` provided); preservation and fresh-write are tested, overwrite is not.
- The idle keepalive/signal behavior in `main()` is only indirectly covered via the pure `shouldRunIdle()`; the `setInterval`/`process.once`/`clearInterval` lifecycle is untested (harder to unit-test, acceptable).

## 5. Conclusion

No correctness, security, or data-loss blockers. SSRF policy is consistently applied to the new URL inputs, secrets handling is unchanged and safe, the idle path is inert, and `normalizeGatewayUrl` is correct for the Anthropic SDK. Approving; the P2 items above are optional follow-ups.

PR #158, opened 2026-06-25
feat: hot-reload bots on provision instead of full gateway restart (#157)

### LINKED ISSUE #157: feat: hot-reload bots on provision instead of full gateway restart
Provisioning a new bot currently triggers a **full gateway restart**, which briefly takes every already-running bot offline. openclaw-channel-octo does this without a restart (atomic config write + a gateway file-watcher that hot-reloads). cc should reach the same.

## Current behavior

When the daemon provisions a bot for the claude runtime:

1. `octo-daemon-cli/internal/adapter/claude.go` `Provision` writes `~/.cc-channel-octo/<bot_uid>/config.json`, registers the bot in the shared `~/.cc-channel-octo/config.json` bots list, then runs **`cc-channel-octo restart`** (`claude.go:128`).
2. `restart` = `cmdStop` + `cmdStart` (`src/cli.ts`), so the whole gateway process is torn down and re-spawned.
3. On start, `src/index.ts` reloads **all** bots via `resolveBotConfigs` + `Promise.allSettled(startBot)`.

The adapter comment already flags this as interim: "a failed restart only means the running host hasn't **hot-reloaded** the new bot yet" (`claude.go:124`) — i.e. hot-reload is the intended end state, not yet implemented.

## Impact

- Every new-bot provision interrupts **all** existing bots on that gateway (sessions/streaming dropped during the restart window).
- Cost scales with bot count — a busy multi-bot gateway pays a full restart to add one bot.
- Contrast: openclaw adds a bot with no restart (`octo-daemon-cli/internal/adapter/openclaw_config.go` writes config atomically; the openclaw gateway has a **file watcher** that swaps the routing snapshot in place — see comments around `openclaw_config.go:163`).

## What we have today

cc CLI commands: `start / stop / restart / status / upgrade / configure`. There is **no** `add-bot` command and **no** config file watcher (grep for `fs.watch` / `SIGHUP` / `reload` in `src/` is empty). config.json is the durable source of truth, but picking up changes requires a restart.

## Suggested directions (pick one)

- **A — gateway file watcher (mirrors openclaw):** the running gateway watches `~/.cc-channel-octo/config.json` (and per-bot dirs); on an added bot entry it `startBot()`s just that bot; on removal it tears down just that bot. Daemon keeps writing config atomically and drops the `restart` call. Lowest operational friction, closest to openclaw.
- **B — explicit add/remove + signal:** add `cc-channel-octo add-bot <uid>` / `remove-bot <uid>` (or a SIGHUP handler) that the supervisor uses to IPC the running gateway to load/unload a single bot, instead of a full restart.

Either way the goal: **provisioning bot N does not disturb bots 1..N-1**.

## Notes / boundaries

- Watch out for the same class of pitfall openclaw documents (`openclaw_config.go:168`): a partial/atomic write must present a *complete* bot entry so a watcher never reloads a half-written config.
- Bot removal (unload) should be part of the same mechanism — relevant to the deletion-propagation work tracked in octo-fleet (orphaned/removed bots).
- Cross-repo: the daemon adapter (`claude.go`) drops/keeps `restart` depending on the chosen direction; the gateway (this repo) owns the reload mechanism.

Not urgent — captured for later.

caster-Q:
## What & why

Provisioning a new bot currently triggers a **full gateway restart**, which briefly takes every already-running bot offline (sessions/streaming dropped during the restart window). This PR makes the gateway hot-load bots: it watches `config.json` and, on a change, starts only the newly-added bot and stops only the removed one — the bots already running are never disturbed. Closes #157 (cc side; the daemon side is a companion PR in octo-daemon-cli).

## How it works

- **`src/bot-manager.ts`** — a long-lived `BotManager` owns the live bot set (`Map<configId, BotStack>`) with a serial mutation queue, so a burst of config events can never run two add/remove against the set concurrently. `addBot` follows the two-phase order: `startBot` (register + install handler, no socket) → bidirectionally cross-register the loop-guard known-bot uids → `connect` (open socket), rolling all three back if connect throws.
- **`src/config-watcher.ts`** — watches the config's *directory* (so an atomic temp+rename swap doesn't drop the subscription), debounces, and runs a serialized `applyLatestConfig` that re-reads the latest desired set inside the task and reconciles. A generation counter (bumped the moment a file event is observed) makes an in-flight reconcile abandon its remaining actions as soon as a newer event arrives. An invalid/half-written config keeps the current bots untouched.
- **`src/session-router.ts`** — adds `unregisterKnownBot` so a removed bot is symmetrically dropped from every remaining router's loop guard.
- **`src/index.ts`** — `main()` now drives everything through `BotManager` + the watcher; idle (zero bots) is a steady runtime state, not just a boot branch. A non-empty config where every bot fails to start still fails boot (preserved); an empty config idles.

The loop-guard cross-registration is bidirectional on purpose: it prevents two bots in the same mention-free group from replying to each other in a loop. It is unrelated to session isolation (each bot already has its own store/cwd/session).

## Identity discipline

`configId` (config.json `bots[].id`, the diff/Map key) and `robotUid` (Octo register id, the loop-guard key) are kept strictly separate and never mixed.

## Known limitation (follow-up)

The watcher reacts to bot **add/remove** only, not to a change of an existing bot's per-bot config. Re-provisioning the same bot with a new token/apiUrl rewrites only the per-bot file; the global `bots[]` is unchanged, so the change needs a manual `cc-channel-octo restart` to take effect. Per-bot config hot-reload is a deliberate follow-up.

## Companion PR

Requires the daemon to write config atomically and stop restarting the gateway — see the octo-daemon-cli PR (#157). This PR should land together with / before that one (the daemon assumes the gateway has the watcher).

## Tests

- New: `bot-manager.test.ts` (diff, bidirectional register, add/remove, rollback, serial queue, shutdown no-resurrection), `config-watcher.test.ts` (reconcile, generation/staleness, half-write keep-current, atomic-rename pickup), `session-router.test.ts` (unregister symmetry).
- `npm run type-check` clean, `npm run lint` clean (max-warnings 0), `npm test` — 1088 passing.
- Manually verified end-to-end: adding a 2nd bot via the web UI leaves the 1st bot's gateway PID unchanged (no restart) and both bots stay connected.


Jerry-Xin (changes requested):
REQUEST_CHANGES stands on one real lifecycle ordering bug. After independent re-verification at head e5e85a3, I am retracting my first blocking item — it does not hold.

## 🔴 Blocking (1)

### removeBot unregisters siblings before draining the removed bot's socket
`src/bot-manager.ts` removeBot:
```
this.bots.delete(configId);
const remaining = [...this.bots.values()];
crossUnregisterOnRemove(bot, remaining);   // siblings forget removed bot's robotUid
await bot.shutdown().catch(() => {});       // THEN drain + close socket
```
`shutdown()` → `gateway.stop(activeHandlers)` (`src/gateway.ts:172`) sets `_draining=true` (drops only NEW inbound) and then awaits in-flight handlers up to 10s. An in-flight handler runs the full pipeline including `streamRelay.deliver(...)` → outbound `sendMessage`, so the removed bot can still emit a reply over its still-open socket during the drain window.

Because `crossUnregisterOnRemove` runs first, sibling routers have already forgotten the removed bot's robotUid by then. Failing sequence:
1. "@A help" in group G containing bots A and B; A's handler starts (long agent call).
2. `removeBot(A)` fires from the watcher.
3. A dropped from map; `B.router.unregisterKnownBot(A.robotUid)` → B no longer knows A is a bot.
4. `await bot.shutdown()` drains; A's in-flight handler completes and emits its reply.
5. B receives it; `from_uid === A.robotUid` but it is no longer in B's known-bot set → B treats A's reply as user input → potential bot-to-bot loop.

This is the exact inversion of addBot's correct ordering (`start (no socket) → crossRegister → connect`). The symmetric inverse for remove is **drain/close first, then unregister**: swap the two calls so `await bot.shutdown()` runs before `crossUnregisterOnRemove`. Add a test that holds an in-flight handler on the removed bot and asserts siblings still recognize its robotUid until shutdown resolves.

## ⤴️ Retracted (was my Blocking #1)
I previously flagged the initial multi-bot boot loop as breaking the two-phase anti-loop invariant. **That does not hold.** Each `addBot` does start(register+handler, no socket) → `crossRegisterOnAdd` against the already-running set → `connect()`. A not-yet-connected bot cannot emit or receive, and by the time bot X's socket opens, every existing bot already knows X's robotUid and X knows all of them (cross-register precedes connect). There is no window where a connected bot receives an unrecognized sibling's message. Withdrawn — the per-bot ordering preserves the invariant across the initial set.

## 🟡 Non-blocking (2)
- `config-watcher.ts`: a stale `add` can connect before the next reconcile removes it. `isStale()` is checked at the top of each loop iteration but not after `await addBot`; the gen-bump narrows the window but does not eliminate it. Next reconcile self-heals it — worth a comment/test, not a blocker.
- `config-watcher.ts` `watchConfig`: if `fs.watch` throws at startup it is caught and logged, leaving hot-reload silently disabled while the process keeps running as if healthy. A health signal or louder warning would help operators. (raised by mochashanyao)

## ✅ Highlights
- Clean configId vs robotUid two-identity split; serial mutation queue (`enqueue`) prevents interleaving; generation guard invalidates in-flight reconciles on new events; invalid/half-written config keeps the running set; remove-then-add reconcile frees overlapping resources before the replacement claims them; addBot rolls back fully on connect failure. Solid test coverage outside the removeBot drain window.

Security: configId/botId originate from operator-side `config.json` on disk, not chat-derived — no path-traversal/command-injection surface; no auth/validation-order endpoints in scope.

src/__tests__/bot-manager.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    expect(mgr.size()).toBe(0);
+  });
+
+  it('serializes concurrent add/remove (no interleaving on the same set)', async () => {
+    const order: string[] = [];
+    const mgr = new BotManager(async (configId) => {
+      order.push(`start:${configId}`);
+      // simulate async start work
+      await new Promise((r) => setTimeout(r, 5));
+      order.push(`started:${configId}`);
```
OctoBoooot, line 282: **[minor]** test name says "concurrent add/remove" but the body fires `addBot('a')` + `addBot('b')` — no remove involved. And the `shutdownAll` no-resurrection test (above) only covers `addBot` AFTER `shutdownAll()` has resolved, while the `shuttingDown` flag's load-bearing case is `addBot` enqueued *before* `shutdownAll()` is called (the flag is set synchronously when `shutdownAll()` runs, so the queued `addBot` task sees it on dequeue). Worth adding: `const p = mgr.addBot('c'); const s = mgr.shutdownAll(); await Promise.all([p, s]); expect(mgr.size()).toBe(0);` and a real add+remove interleave.

src/__tests__/config-watcher.test.ts
```diff
@@ -0,0 +1,192 @@
...
+      await h.applyNow();
+      // current set untouched
+      expect(mgr.added).toEqual([]);
+      expect(mgr.removed).toEqual([]);
+      h.close();
+    } finally {
+      cleanup();
+    }
+  });
+
```
OctoBoooot, line 151: **[minor]** this asserts watcher delivery indirectly: after the 200ms sleep, `await h.applyNow()` is called, which would add `b` even if fs.watch never delivered the event. The test passes for the right reason and the wrong reason both. Either drop the `applyNow()` call (rely on the debounce-driven reconcile), or pass a `log` spy and assert the reconcile message fired from the watcher path.

src/__tests__/bot-manager.test.ts
```diff
@@ -0,0 +1,207 @@
...
+  };
+}
+
+function fakeBot(
+  configId: string,
+  robotUid: string,
+  opts?: { connect?: () => Promise<void>; onShutdown?: () => void },
+): ManagedBot & { router: FakeRouter } {
+  const router = fakeRouter(robotUid);
+  return {
```
OctoBoooot, line 62: **[minor]** `fakeBot` adds a `configId` field on the returned `ManagedBot`, but `ManagedBot` deliberately does not carry `configId` per the doc comment in `bot-manager.ts` ("configId is intentionally NOT carried on the stack… see plan B2 for the identity split"). The cast `as ManagedBot` papers over it. If the test wants to track configId for assertions, keep it as a sibling local; don't bake it onto the fixture in a way that contradicts the interface contract.

OctoBoooot (commented):
## Review: feat: hot-reload bots on provision instead of full gateway restart (#158)

**Verdict:** Approve with comments — solid design, only test-quality nits on substantive code.

CI is red — `check-sprint / check-sprint` is failing (project-board governance gate, not code). Vitest / ESLint / type-check are all green. Already labeled `needs-human-review` (operationally appropriate for a lifecycle change of this scope).

The diff matches the linked plan (`docs/superpowers/plans/2026-06-25-bot-hot-reload.md`): serial mutation queue on `BotManager`, generation guard in `apply()` that bumps on every fs event (not on debounce expiry), `loadDesired` throw → keep current set (so a half-written config never tears down healthy bots), removes-before-adds in `reconcile`, `unregisterKnownBot` with self-protection, two-identity discipline (`configId` for the manager map, `robotUid` for the loop guard) — all in place.

### Minor
- `src/__tests__/bot-manager.test.ts:189-205` — test named "serializes concurrent add/remove" only fires two `addBot` calls, not an actual add+remove interleave. And the "no resurrection" test (`shutdownAll tears down all bots`) only asserts that `addBot` AFTER `shutdownAll()` resolves is a no-op — but the `shuttingDown` flag's load-bearing case is `addBot` enqueued *before* `shutdownAll()` is called. Worth adding: enqueue `addBot('c')` without awaiting → call `shutdownAll()` → await both → assert nothing was added.
- `src/__tests__/config-watcher.test.ts:131-153` — "fires a reconcile on a file change (debounced)" awaits 200ms past debounce and then calls `await h.applyNow()`. If fs.watch never delivered, the explicit `applyNow()` still adds `b` and the test passes. Either drop the `applyNow()` call, or assert via a `log` spy that the reconcile reason was the watcher (not the manual call).
- `src/__tests__/bot-manager.test.ts:51-69` — `fakeBot()` includes a `configId` field on the returned object (cast through `as ManagedBot`). `ManagedBot` deliberately does NOT carry `configId` per the doc comment in `bot-manager.ts` ("configId is intentionally NOT carried on the stack… see plan B2 for the identity split"). Either drop the field from the fixture or, if the test wants to track it, keep it as a sibling local — don't paper over the interface.

### Questions
- `src/index.ts:78-89` — initial sync is now `for (const id of initialIds) await manager.addBot(id)`, sequentialized through the serial queue. Pre-PR, `Promise.allSettled(botConfigs.map(startBot))` ran the register-over-REST + handler-install phase in parallel for all bots. The connect phase was already sequential, so the new regression is roughly N × (REST register latency) extra at boot. Intentional tradeoff for queue simplicity? Worth one line in the PR body so operators know to expect it on multi-bot deployments.

### Praise
- The two-identity discipline (`configId` vs `robotUid`) is explicit in code AND documented in both module headers and the plan. `crossUnregisterOnRemove` × `unregisterKnownBot` self-protection (won't drop the router's own `robotId`) gracefully handles the pathological case where two configs share a `robotUid`.
- `apply()` captures `myGen` at invocation time and bumps `latestGen` *in* `schedule()` (not at debounce expiry) — so an in-flight reconcile is invalidated the moment a newer event lands, not 200ms later. Correct and the comment explains why.
- `addBot` rollback is complete and symmetric: `crossUnregisterOnRemove(bot, existing)` undoes the sibling-side cross-registration before `bot.shutdown().catch(() => {})`; the new bot's own router state doesn't need cleanup because the bot is gone after shutdown.
- Watching the *directory* (not the file) sidesteps the atomic-rename inode-swap trap that would silently break a file-targeted watcher after the first daemon write. Plus the `loadDesired` throw-keeps-current contract makes the half-write window safe.

- The plan calls out per-bot config hot-reload (token/apiUrl change without `bots[]` changing) as a deliberate follow-up — re-provisioning the same bot still needs a manual `cc-channel-octo restart`. Already documented in the PR body; flagging so future #157-style reports don't surprise.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# cc-channel-octo PR#158 Review Report — automated review
**PR:** https://github.com/Mininglamp-OSS/cc-channel-octo/pull/158
**Head SHA:** `e5e85a3653554b4e8ba4fc1bd694def2bb97c658`
**Routing:** complexity=security_sensitive (automated review / automated review leg)

---

## 1. 验证结论

| Area | Status | Evidence |
|------|--------|----------|
| BotManager serial queue (C/C6) | ✅ | `enqueue()` chains via `.then(task, task)`, queue stays alive on rejection (`bot-manager.ts:113-120`) |
| Two-phase start ordering (B) | ✅ | `addBot`: `start(configId)` → `crossRegisterOnAdd` → `connect()` with rollback on connect failure (`bot-manager.ts:128-142`) |
| Bidirectional loop-guard (B) | ✅ | `crossRegisterOnAdd` registers both directions; `crossUnregisterOnRemove` symmetric (`bot-manager.ts:62-78`) |
| Identity discipline (B2) | ✅ | Map keyed by `configId` (addBot arg); `SessionRouter` uses `robotUid` only. `ManagedBot` interface has no `configId` field. |
| Generation guard (C6) | ✅ | `latestGen++` in `schedule()` (event-time, not debounce-expiry); `isStale` checked per-op in `reconcile` (`config-watcher.ts:117-122`, `index.ts:155-156`) |
| Half-write protection (D) | ✅ | `loadDesired` throws on any invalid config → caught in `apply()` → current bots untouched (`config-watcher.ts:134-138`) |
| Empty→idle steady state (E) | ✅ | `keepalive` setInterval (ref'd, not unref'd) holds event loop; zero bots is a valid runtime state (`index.ts:119-120`) |
| Shutdown no-resurrection (N4) | ✅ | `shuttingDown = true` in `shutdownAll()` → `addBot` early-returns (`bot-manager.ts:108-109`, `bot-manager.ts:127`) |
| `unregisterKnownBot` symmetry | ✅ | Guards self (`uid !== this.robotId`), idempotent for unknown/empty uids. Unit tests cover register→unregister→re-process and self-protection. (`session-router.ts:83-88`) |
| Boot fail semantics preserved | ✅ | Non-empty config + all bots fail → throws (preserved). Empty config → idle. (`index.ts:97-100`) |
| `configPathForWatch()` → `DEFAULT_CONFIG_PATH` | ✅ | Fixed `~/.cc-channel-octo/config.json`, matches daemon write target. (`index.ts:52-54`, `config.ts:31`) |

## 2. 发现的问题

### P2-1 — Watcher start failure degrades hot-reload silently

**File:** `src/config-watcher.ts:155-157`

```typescript
} catch (err) {
    log(`[hot-reload] failed to start watcher on ${dir}: ${errMsg(err)}`);
}
```

If `fs.watch(dir)` throws (permissions, inotify limit, etc.), the gateway boots normally but **hot-reload is permanently disabled** for the process lifetime. The companion daemon PR (#157) removes `cc-channel-octo restart` from provision — so with a broken watcher, new provisions are silently never picked up.

**Diff-scope:** new (introduced by this PR). The old code had no watcher; the daemon always restarted.
**Severity rationale:** P2, not P1 — the log line IS emitted, the gateway still serves existing bots correctly, and the daemon PR (which removes restart) is a separate deploy. However, in the **combined deploy** scenario (gateway + daemon upgraded together, which is the stated assumption per plan F), a broken watcher means provision is silently broken. Consider escalating the log to `console.error` or adding a health-check endpoint / stderr banner so operators notice.

### P2-2 — Shutdown path drops `process.exit(0)`

**File:** `src/index.ts:121-130`

Old multi-bot shutdown:
```typescript
await Promise.allSettled(connected.map((s) => s.shutdown()));
process.exit(0);
```

New shutdown:
```typescript
await manager.shutdownAll();
resolve();  // main() returns — but no process.exit()
```

After `shutdownAll()` clears the map and `resolve()` fires, `main()` returns. The process should exit naturally once the event loop drains (keepalive was cleared). But if any unref'd resource, lingering TCP connection, or third-party module holds the loop open, the process hangs indefinitely. The old code's `process.exit(0)` was a safety net against this.

**Diff-scope:** new behavioral change (amplified by the architecture shift to long-lived process).
**Severity rationale:** P2 — the risk is theoretical and depends on event-loop state, but removing a known-safe exit for an implicit drain is a reliability regression in edge cases.

### P2-3 — Test helper adds `configId` not in `ManagedBot` interface

**File:** `src/__tests__/bot-manager.test.ts:53-64`

The `fakeBot` helper returns `{ configId, robotUid, router, connect, shutdown }`, but `ManagedBot` (`src/bot-manager.ts:28-33`) has no `configId` field. The `as ManagedBot & { router: FakeRouter }` cast hides the extra property. Tests pass because the BotManager never reads `configId` from the stack (it's the Map key, not a field — plan B2). Not broken, but the test shape diverges from production, weakening the test's value as a contract check.

**Diff-scope:** new.
**Suggestion:** Remove `configId` from `fakeBot`'s return object to match `ManagedBot` exactly.

## 3. 建议

1. **Watcher failure visibility (P2-1):** Consider `console.error` instead of routing through the `log` callback (which goes to `console.log`). Or emit a one-line stderr banner at boot: `[cc-channel-octo] hot-reload: active|degraded`.
2. **Shutdown safety (P2-2):** Add a `setTimeout(() => process.exit(0), 5000).unref()` after `resolve()` as a drain timeout, or restore `process.exit(0)` after `shutdownAll`.
3. **Test fidelity (P2-3):** Drop `configId` from `fakeBot`'s return to match `ManagedBot` interface exactly.

## 4. 额外发现

### 4a — `loadDesired` is synchronous I/O on the chain

`loadDesired` calls `loadConfig()` (synchronous `readFileSync`) inside the serialized chain. Config files are small, so this is fine today. But if the config ever grows (e.g., many bots with inline overrides), it blocks the event loop during reconcile. Not actionable now — just noting the synchronous I/O on the hot path.

### 4b — `resolveBotConfigs(config)` is called for every `addBot`

The factory (`index.ts:82-85`) calls `resolveBotConfigById(loadConfig(), configId)` which expands ALL bots via `resolveBotConfigs` and searches for one. For N bots, each add is O(N) config expansions including per-bot file reads. At runtime this is typically one add at a time, so it's fine. But a batch provision of many bots simultaneously would be O(N²) config reads. Again, not actionable for typical workloads.

### 4c — Known limitation is well-documented

Plan G (per-bot config hot-reload not supported; requires manual restart) is clearly stated in the PR body, plan doc, and code comments. No action needed — just confirming the limitation is properly surfaced.

## 5. 数据流回溯

| Consumed data | Upstream source | Traced? |
|---|---|---|
| `configId` (Map key in BotManager) | `config.json` `bots[].id` → `resolveBotConfigs` → `c.botId` → `.filter(Boolean)` | ✅ configId flows correctly from disk to Map key. Same path in initial sync (`index.ts:90`) and watcher `loadDesired` (`index.ts:112-113`). |
| `robotUid` (loop-guard key) | `OctoGateway.register()` → `reg.robot_id` → `gateway.botId` getter → `startBot` return `{ robotUid: gateway.botId }` | ✅ Traced through `gateway.ts:64` (getter returns `this.robotId`) → `gateway.ts:96` (`this.robotId = reg.robot_id`). robotUid is the Octo-issued robot ID, distinct from configId. |
| `desired` in `reconcile()` | `loadDesired()` → `resolveBotConfigs(loadConfig()).map(c => c.botId).filter(Boolean)` | ✅ Re-reads from disk inside serialized task (not precomputed). `loadConfig()` reads `DEFAULT_CONFIG_PATH`; `resolveBotConfigs` validates tokens/ids/apiUrls and throws on invalid. |
| `isStale()` predicate | Closure over `latestGen` and `myGen` in `watchConfig` | ✅ `myGen` captured synchronously in `apply()` before chain. `latestGen` bumped synchronously in `schedule()` on file event (before debounce). Stale check `myGen !== latestGen` correctly detects newer events. |
| `manager.runningKeys()` in diff | `[...this.bots.keys()]` — snapshot of Map keys | ✅ Returns current Map keys. Serial queue ensures no concurrent mutation during diff. |
| `existing` in `addBot` | `[...this.bots.values()]` — snapshot before cross-register | ✅ Taken after `start(configId)` resolves. The new bot is NOT in `existing` yet (added to Map only after `connect()` succeeds). Cross-register uses this snapshot correctly. |

## 6. 盲点 checklist (R5)

**C1 — 双路径 parity (含守卫/lint 工具自身等价写法覆盖):**
- `registerKnownBot` ↔ `unregisterKnownBot`: ✅ Symmetric. Both guard `uid` truthiness. `unregisterKnownBot` additionally guards `uid !== this.robotId` (self-protection). Verified in `session-router.ts:76-88`.
- `addBot` ↔ `removeBot`: ✅ Symmetric. `addBot`: start → crossRegister → connect (rollback on connect fail). `removeBot`: delete from map → crossUnregister → shutdown. Both on serial queue.
- `crossRegisterOnAdd` ↔ `crossUnregisterOnRemove`: ✅ Both iterate over the other bots and call the corresponding router method.
- **Create-gate ↔ execute-path:** N/A — no authorization gates in this PR.

**C2 — control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿:**
- Serial queue prevents interleaving: ✅ `enqueue` chains promises; same queue for add/remove/shutdownAll.
- Generation guard ordering: ✅ `latestGen++` in `schedule()` (synchronous, event-time) → captured as `myGen` in `apply()` → checked in `reconcile`. No way for a stale reconcile to act on an outdated desired set without detection.
- No nested/reused safety controls (regex/escape/sanitize) in this PR. N/A for non-canonical probing.

**C3 — 授权边界 ≠ 能力边界:**
- No new endpoints, tools, or capabilities exposed. The watcher reads a local file the gateway already trusted. `addBot`/`removeBot` are internal-only. N/A.

**C4 — 授权生命周期 / 容器-成员状态级联:**
- No authorization changes. Bot lifecycle management only. N/A.

**C5 — build/note 通过 ≠ 运行期路径正确:**
- Not a build-artifact / extension / CLI-packaging PR. The watcher watches a directory (not file) specifically to handle atomic rename at runtime — this IS the runtime path concern, and it's correctly addressed. N/A for build≠runtime.

**C6 — 治理/策略/安全文档自洽性:**
- `docs/superpowers/plans/2026-06-25-bot-hot-reload.md` is a design plan, not a governance/security document. No conflict with existing SECURITY.md or disclosure policies. N/A.

## 7. 跨轮 blocker 复检 (R6)

First review — N/A.

---

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. Three P2 observations (watcher failure visibility, missing process.exit safety net, test fidelity) — none block landing. The core architecture (serial queue, generation guard, two-phase start, bidirectional loop-guard, rollback on connect failure, shutdown no-resurrection) is sound and well-tested. Data-flow tracing confirms all consumed data reaches its consumption point correctly.

src/bot-manager.ts
```diff
@@ -0,0 +1,175 @@
...
+        throw err;
+      }
+      this.bots.set(configId, bot);
+    });
+  }
+
+  /** Remove one bot by configId: shutdown + symmetric unregister + drop. */
+  removeBot(configId: string): Promise<void> {
+    return this.enqueue(async () => {
+      const bot = this.bots.get(configId);
```
OctoBoooot, line 152: **[major]** ordering inversion vs `addBot`. `addBot` correctly does `start (no socket) → crossRegister → connect` so siblings learn a new bot's uid before its socket can emit. The symmetric inverse should be `shutdown (close socket) → crossUnregister`, but here `crossUnregisterOnRemove` runs synchronously BEFORE `await bot.shutdown()`. During the awaited shutdown, the removed bot's WS is still open; any in-flight handler completing during that window emits a reply that sibling routers no longer recognize as a known-bot uid → they route it as user input.

Failing input: user sends `@A help` in a group with A and B. A's handler starts (slow LLM call). `removeBot(A)` fires from the watcher. (1) A dropped from map. (2) `B.router.unregisterKnownBot(A.robotUid)` runs — B stops knowing A is a bot. (3) `await bot.shutdown(A)` begins. (4) A's in-flight handler completes and sends a reply via A's still-open socket. (5) B receives A's reply, loop-guard misses (uid not in knownBotUids), B processes as user input.

Fix: shut down first, then unregister.

OctoBoooot (changes requested):
## Review (follow-up): #158

**Updated verdict:** Request changes — credit to Jerry-Xin for catching the `removeBot` ordering inversion; I missed it in my first pass.

### Major
- `src/bot-manager.ts` `removeBot` — ordering is the inverse of `addBot`'s and creates a sibling-loop-guard window during shutdown. addBot is `start (no socket) → crossRegister → connect`, so a sibling never sees a new bot's first message before knowing it's a bot. The symmetric inverse for removal should be `shutdown (close socket) → crossUnregister`, but the current code does:

  ```
  this.bots.delete(configId);
  crossUnregisterOnRemove(bot, remaining);   // B forgets A's robotUid
  await bot.shutdown().catch(() => {});       // THEN close A's socket
  ```

  Failing input: user sends `@A help` in a group containing bots A and B. A's WS receives and A's async handler starts processing. `removeBot(A)` is then triggered by the config watcher. Steps:
  1. A is dropped from the map and `B.router.unregisterKnownBot(A.robotUid)` runs → B stops recognizing A as a bot.
  2. `await bot.shutdown()` begins; A's in-flight handler completes and emits its reply via A's still-open socket.
  3. Octo broadcasts A's reply to the group; B's WS receives it.
  4. B's loop guard checks `from_uid === A.robotUid` — but A's uid is no longer in `knownBotUids`, so B treats A's farewell as a user message and may reply.

  Fix: shut the bot down (await its socket close) BEFORE unregistering its uid from siblings, so any in-flight A→Octo→B messages still land while B's router knows to drop them. Drop from the map can happen at either end — the load-bearing pair is `shutdown` precedes `crossUnregister`.

### Notes on Jerry-Xin's other findings
- 🔴 "初始多机器人启动未保持两阶段注册连接顺序" — I read this as not a defect: per-bot `addBot` already does `start (no socket) → crossRegister(bot, existing) → connect`, so at the moment any bot's socket opens, every already-running bot's router knows its `robotUid`. The pre-PR batched cross-register and the new per-bot one are equivalent for the invariant "no bot's socket is live before siblings know about it." Happy to be shown a concrete scenario where the per-bot version fails.
- 🟡 stale-config check timing — agreed it's a minor: `isStale` is checked between add/remove iterations, not inside them, so an in-flight add/remove during a newer config event finishes before the next iteration bails. Eventual consistency reconciles it on the next reconcile pass. Acceptable as-is unless a single add/remove can take seconds.

yujiawei (dismissed):
# Code Review — PR #158 (cc-channel-octo)

**Verdict: APPROVED.** No blocking (P0/P1) issues. Three non-blocking follow-ups (P2) are listed below; none should hold the merge, but they're worth a quick look before/after landing.

This is a clean, well-reasoned refactor. The two-identity discipline (`configId` for the Map / diff key vs `robotUid` for the loop-guard) is held consistently, the serial mutation queue + generation guard are sound, and the half-written-config protection (keep current bots on any `loadDesired` throw) is exactly right. Verified locally on `e5e85a3`: `type-check` clean, `lint` clean (max-warnings 0), `npm test` → 1088 passing.

## What I verified

- **Resource lifecycle on remove/re-add is safe.** `removeBot` → `bot.shutdown()` → `gateway.stop()` → `releaseLock()`, so the `gateway.lock` and SQLite store are freed; a later re-add of the same id won't hit a held lock. The `removes-before-adds` ordering in `reconcile` further protects an in-place bot swap that shares a dir/lock.
- **connect() rollback is complete.** On a `connect` throw, `addBot` undoes the cross-registration, shuts the bot down, and never inserts it into the Map — no half-added bot lingers in sibling routers' known-bot sets. Covered by the rollback test.
- **Concurrency.** `enqueue` serializes all mutations and keeps the chain alive across a rejected task; `shutdownAll` sets `shuttingDown` so a late apply can't resurrect a bot. The generation counter is bumped on the *event* (in `schedule`), not at debounce expiry, so an in-flight reconcile is invalidated immediately by a newer write.
- **Directory watch + atomic rename.** Watching `dirname(configPath)` (not the file) survives a temp+rename inode swap; the `changed === null` permissive branch is reasonable.
- **No scope creep.** Dead code (`shouldRunIdle`, `partitionStartResults`, the duplicate idle branch) is removed; the removed tests correspond to removed exports. Idle-as-runtime-state is correctly generalized.

## Non-blocking follow-ups (P2)

### 1. Boot-window config-change blind spot — `src/index.ts:88-124`
The initial reconcile (`for (const id of initialIds) await manager.addBot(id)`) runs *before* `watchConfig(...)` installs the watcher. A config write that lands during initial bot registration produces no watcher event and no post-subscription reconcile, so a bot deprovisioned in that window could stay up (or a freshly provisioned one be missed) until the *next* config edit. Narrow (boot only) and self-healing on the next write, hence P2. Consider installing the watcher first, then calling `watcher.applyNow()` as the initial sync so any write during boot is reconciled.

### 2. `unregisterKnownBot` happens before the removed bot finishes draining — `src/bot-manager.ts:155-158`
`removeBot` deletes the entry and calls `crossUnregisterOnRemove` *before* `await bot.shutdown()`. While the departing bot drains an in-flight turn, the surviving routers have already forgotten its `robotUid`; if it emits a message during that drain window, a sibling in a mention-free group may treat it as user input and reply — the exact bot-to-bot interaction the loop-guard prevents. Bounded (a draining bot, one-shot, not an infinite loop), so P2. Safer order: `await bot.shutdown()` first, then `crossUnregisterOnRemove`.

### 3. Watcher `chain` lacks a terminal `.catch` (defense-in-depth) — `src/config-watcher.ts:127-137`
`BotManager.enqueue` hardens its queue against a rejected task, but the watcher's own `chain = chain.then(async () => { ... })` does not. `reconcile` swallows per-bot errors and `loadDesired` is wrapped, so today nothing should escape — but if any unexpected synchronous throw ever did, `chain` becomes permanently rejected and every future `apply()` short-circuits, silently killing hot-reload until a process restart. Terminating with `.catch(err => log(...))` matches the resilience already applied to `enqueue`.

## Notes (not action items)

- The "same-bot re-provision (token/apiUrl change) needs a manual restart" limitation is correctly documented in the PR body and the watcher's header comment — good. Worth ensuring the daemon-side companion change keeps that contract visible to operators.
- Companion dependency: the daemon must write config atomically (temp+rename) and stop restarting the gateway. This PR assumes that; if the daemon side lands later, an old daemon doing a non-atomic write could momentarily expose a half-written file — mitigated here by the keep-current-on-invalid behavior, but worth confirming the land order.
- `gemini`-style concern about legacy unnamed configs being dropped by the `.filter(id => !!id)` is a non-issue: `resolveBotConfigs` always assigns an id (`bot.id ?? bot${i}`, and the single-bot path synthesizes `'default'`), so the filter is type-narrowing only, never a behavioral drop.

yujiawei:
Follow-up to the approval: an additional automated pass surfaced three more non-blocking (P2) observations that I've confirmed against the code. None change the verdict (still no P0/P1), but they're worth a look before/after landing.

**1. A failed watcher start silently disables hot-reload — `src/config-watcher.ts:154-164`**
If `fs.watch(dir)` throws (inotify limit, permissions), the `catch` only logs through the `log` callback (→ `console.log`); the gateway boots and serves existing bots normally, but hot-reload is dead for the process lifetime. Because the companion daemon change removes the `restart`-on-provision path, in the combined-deploy scenario a broken watcher means new provisions are silently never picked up. Consider escalating this to `console.error` / a stderr boot banner (`hot-reload: active|degraded`) so operators notice.

**2. Clean shutdown no longer forces `process.exit(0)` — `src/index.ts:129-140`**
The new shutdown awaits `manager.shutdownAll()` then `resolve()`s and lets `main()` return; the entrypoint `.catch` only calls `process.exit(1)` on error, so a clean SIGTERM now relies on the event loop draining naturally. The pre-refactor multi-bot path had an explicit `process.exit(0)` safety net. The hang risk is theoretical (depends on some unref'd handle / lingering socket), but a `setTimeout(() => process.exit(0), 5000).unref()` after `resolve()` would restore the guarantee cheaply.

**3. Test fidelity nit — `src/__tests__/bot-manager.test.ts` `fakeBot`**
`fakeBot` returns a `configId` field that `ManagedBot` (`src/bot-manager.ts:24-29`) doesn't declare; the `as ManagedBot & {...}` cast hides it. Tests pass because `BotManager` never reads `configId` off the stack (it's the Map key, per the two-identity split), but dropping `configId` from the fake would keep the test shape aligned with the production interface.

All three are P2 / non-blocking.

OctoBoooot (approved):
## Re-review (delta): #158

**Verdict:** Approve.

Delta from `e5e85a3` → `c5ef173` resolves my prior major + all three minors:

- ✓ `src/bot-manager.ts` `removeBot` — order is now `await bot.shutdown()` then `crossUnregisterOnRemove(bot, remaining)`, the correct symmetric inverse of addBot's `start → crossRegister → connect`. The new comment explains the drain-window rationale. Test `removeBot closes the socket (shutdown) BEFORE unregistering from siblings` asserts the recorded order is `['shutdown:a', 'unregister:a']` via a spied `unregisterKnownBot`.
- ✓ `addBotsBatch` added and wired through `main()` for initial sync — start-all → cross-register-the-whole-batch → connect-all, with a rollback path that excludes the failing bot from the unregister target. Tests `addBotsBatch cross-registers the WHOLE batch before any connect` and `addBotsBatch rolls back a bot whose connect fails, keeps the rest` cover both paths.
- ✓ New test `shutdownAll cancels an addBot enqueued BEFORE it (shuttingDown flag)` exercises the load-bearing race: addBot enqueued (`const addP = mgr.addBot('a')`), then `shutdownAll()` enqueued — start() never runs (`expect(started).toBe(0)`). Closes my "no-resurrection test doesn't cover the actual race" minor.
- ✓ `src/__tests__/config-watcher.test.ts` "fires a reconcile on a file change" — `applyNow()` removed, replaced with a poll-loop on `mgr.added.includes('b')` plus a `log` spy asserting `'added bot b'`. Now proves fs.watch delivery rather than passing for the wrong reason.
- ✓ `fakeBot` — `configId` field dropped, param renamed `_configId`, comment cites plan B2.

CI is green on this SHA (only `code-review` pending — that's this review).

### Notes (carrying forward, not blocking)
- yujiawei's #1 (install watcher BEFORE initial sync to catch a config write that lands during the sync) — still open. Narrow race, minor.
- mochashanyao's "watcher startup failure silent" (`fs.watch` throw is caught and logged, hot-reload is then permanently disabled while the process keeps running) — still open. Minor; consider escalating that one log line to a louder warning so operators notice.

### Praise
- Clean, surgical response to the round-1 findings — every blocker/major/minor I raised has a code change AND a targeted test asserting the fix's premise (the `shutdown:a`-then-`unregister:a` event ordering test in particular is exactly how I'd want to see it pinned).
- `addBotsBatch`'s connect-fail rollback target `[...existing, ...batch.filter((b) => b !== s.bot)]` correctly excludes the failing bot from its own unregister set — easy to get wrong.

### Retraction
- My round-1 "boot is sequential vs prior parallel" question is moot now; `addBotsBatch` is the right shape for the initial set and the per-bot serial concern doesn't apply.

Jerry-Xin (approved):
🔄 Re-review at `c5ef1739f632` (delta since my prior review at `e5e85a3`).

**Summary:** My one surviving 🔴 from the last review — the `removeBot` ordering inversion — is fixed correctly at this head, with a dedicated drain-window regression test. No new blocking defect is introduced by the delta. Approving.

**✅ Resolved 🔴 (was blocking)**
- `src/bot-manager.ts:160` `removeBot` now closes the socket FIRST, then unregisters from siblings: `await bot.shutdown()` runs **before** `crossUnregisterOnRemove(bot, remaining)`. This is the exact swap required. The previous order let an in-flight handler (draining for up to the gateway drain timeout) emit a reply over the still-open socket while siblings had already dropped its `robotUid`, so they'd misclassify it as user input → bot-to-bot loop. Order is now the symmetric inverse of `addBot`. ✅
- Backed by a real regression test in `src/__tests__/bot-manager.test.ts`: *"removeBot closes the socket (shutdown) BEFORE unregistering from siblings"* asserts `events == ['shutdown:a', 'unregister:a']` by spying on the sibling's `unregisterKnownBot`. Exactly the drain-window ordering guarantee. ✅

**✅ Initial multi-bot boot (the item I retracted last time) is now properly handled**
- `src/index.ts:88` no longer adds bots one-by-one (which would let the first bot connect before later bots exist, so it never learns them). It now uses the new `manager.addBotsBatch(initialIds)` two-phase batch in `src/bot-manager.ts`: start all (register + handler, no socket) → cross-register the whole set → connect all. The "no bot opens its socket before every sibling knows its robotUid" invariant now holds across the initial set, with rollback on per-bot connect failure. Covered by new tests for batch cross-registration and connect-failure rollback. Nice.

**🟡 Non-blocking (pre-existing, not introduced by this delta — keep as hardening)**
- `src/config-watcher.ts:154` `fs.watch` startup failure is logged and swallowed. With the companion daemon no longer restarting the gateway, a failed watcher silently disables hot-reload until manual intervention. Consider surfacing a hard health/error state when hot-reload is unavailable. (Unchanged by this delta.)
- `src/config-watcher.ts:41` stale-add window: `isStale()` is checked before each `addBot`/`removeBot`, not mid-flight, so an in-flight add completes even if a newer generation supersedes it. This is bounded and self-healing — the next reconcile (triggered by the newer generation) removes the now-undesired bot. A hardening note, not a blocker; this file is **unchanged** vs my prior review. If you want to tighten it, propagate cancellation/staleness into the add path or re-check the desired set after `start` and before `connect`/Map insertion.
- `src/bot-manager.ts:161` `removeBot` swallows shutdown errors (`.catch(() => {})`) and reconcile still logs a successful removal. If teardown ever fails before releasing a lock/socket/store, the manager forgets the bot while the resource lingers. At minimum log the shutdown failure.

**✅ Highlights**
- `configId` vs `robotUid` separation is clean and removes a real class of multi-bot loop-guard mistakes; the test fixture deliberately drops `configId` to match.
- New unit tests cover diffing, batch cross-registration, rollback, directory watch, unregister ordering, and the shutdown-before-shutdownAll cancellation path well.

The remaining items are all 🟡 hardening notes on code that is unchanged by this delta. The blocking concurrency bug is fixed and tested. Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# PR#158 Review Report — feat: hot-reload bots on provision instead of full gateway restart
Head SHA: `c5ef1739f632c60a2f95ec3b4df9a9463c153ff4`
Repo: `Mininglamp-OSS/cc-channel-octo`
Complexity: `security_sensitive` (automated review)

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Two-phase add ordering (start → cross-register → connect) | ✅ | `bot-manager.ts:126-148` — `addBot` calls `start()`, then `crossRegisterOnAdd`, then `bot.connect()` with rollback on failure |
| Two-phase batch boot (addBotsBatch) | ✅ | `bot-manager.ts:168-209` — Phase 1: start all, Phase 2: cross-register whole batch pairwise, Phase 3: connect all |
| Remove ordering (shutdown before unregister) | ✅ | `bot-manager.ts:152-164` — `await bot.shutdown()` before `crossUnregisterOnRemove` |
| Serial mutation queue | ✅ | `bot-manager.ts:105-115` — promise-chain `enqueue` serializes all add/remove |
| Generation guard (staleness) | ✅ | `config-watcher.ts:42-48,115-125` — `latestGen` bumped in `schedule()` (event arrival), captured in `apply()` for `isStale` check |
| Symmetric unregisterKnownBot | ✅ | `session-router.ts:83-88` — never drops `this.robotId` (self-guard) |
| Invalid config resilience | ✅ | `config-watcher.ts:130-135` — `loadDesired()` throw → keep current bots |
| Shutdown race (shuttingDown flag) | ✅ | `bot-manager.ts:92` — set synchronously in `shutdownAll()` body before enqueue; `addBot` checks it at task start |
| Directory watch (atomic rename) | ✅ | `config-watcher.ts:148-157` — `watch(dir, …)` filters by filename, `changed === null` permissive |
| Boot fail semantics preserved | ✅ | `index.ts:96-99` — non-empty config with all bots failed → throws first error |
| Idle keepalive | ✅ | `index.ts:120` — `setInterval(() => {}, 60_000)` ref'd timer, cleared on signal |
| Tests coverage | ✅ | 293 (bot-manager) + 197 (config-watcher) + 61 (session-router) lines of new tests; deleted obsolete test files |

## 2. Findings

### No P0/P1 findings.

---

### P2-1: `process.exit(0)` removed — process exit relies on event loop drain

**Diff-scope**: new (introduced by this PR)
**Severity**: P2 (maintainability / defensive posture)

The old multi-bot shutdown called `process.exit(0)` after draining bots:
```
await Promise.allSettled(connected.map((s) => s.shutdown()));
process.exit(0);
```

The new code resolves the keepalive promise → `main()` returns → Node exits when the event loop empties. After shutdown: keepalive cleared, watcher closed, `gateway.stop()` (drains handlers, stops heartbeat, disconnects socket, releases lock), `store.close()`, cron stopped, cwd-cleanup timer cleared.

This is correct IF no SDK-internal handle (agent connection, unresolved promise, lingering timer) survives bot shutdown. In practice this is likely fine — `gateway.stop()` is thorough. But the explicit `process.exit(0)` was a stronger guarantee against leaked handles in third-party SDK code.

**Suggestion**: consider adding `process.exit(0)` after `manager.shutdownAll()` as a defensive measure, or at least a timeout fallback (e.g., `setTimeout(() => process.exit(1), 15_000).unref()` after shutdown). Not blocking.

---

### P2-2: Per-bot config hot-reload acknowledged as follow-up

**Diff-scope**: pre-existing limitation, documented by PR
**Severity**: P2 (documented known limitation)

The watcher reacts to `config.json` changes only (the `bots[]` array). A per-bot config change (token, apiUrl) rewrites only `<baseDir>/<id>/config.json` without touching the global config, so the watcher doesn't fire. The PR explicitly documents this as a deliberate follow-up.

The data flow is sound for the covered case: the daemon writes per-bot config first, then atomically writes `config.json` → watcher fires → `loadConfig()` inside `resolveBotConfigById` picks up the already-written per-bot file.

Not blocking — the documented limitation doesn't make any currently-working path worse.

---

### P2-3: `configPathForWatch()` wrapper is trivial

**Diff-scope**: new
**Severity**: nit

`configPathForWatch()` at `index.ts:52-54` is a single-line wrapper returning `DEFAULT_CONFIG_PATH`. It adds no abstraction value. Could be inlined. Not blocking.

---

### P2-4: Error message changed from "no bots connected" to "no bots started"

**Diff-scope**: new (minor behavioral change)
**Severity**: nit

The old code threw `'no bots connected'` when all connects failed. The new code throws `failed[0]?.error` (the first start/connect error), which is strictly more informative. The `?? new Error('no bots started')` fallback is also clearer. No issue — just noting the change.

## 3. Suggestions

1. **Defensive process exit** (P2-1): add a post-shutdown fallback timer that forces `process.exit(1)` if the event loop doesn't drain within ~15s. Protects against SDK handle leaks.
2. **Per-bot hot-reload** (P2-2): the documented follow-up. Consider watching per-bot config dirs as a future enhancement.

## 4. Extra Observations

- **`multi` parameter is now always `true`**: `startBot` is always called with `multi=true` (via the manager's start function at `index.ts:82`). The `handleSignals: !multi` flag is always `false`. This is correct — main() always owns signals. The `multi` parameter could be removed in a follow-up cleanup, but it's harmless as-is.
- **`BotStack` is now a type alias for `ManagedBot`**: the `export type BotStack = ManagedBot` at `index.ts:141` preserves backward compatibility for any external consumer. Good defensive aliasing.
- **Test quality is high**: tests cover rollback ordering, batch two-phase cross-registration, generation-guard staleness, atomic rename pickup, shutdown cancellation, and self-unregister guard. The `removeBot` ordering test (`shutdown:a` before `unregister:a`) directly validates the most critical invariant.

## 5. Data-Flow Tracing

|---|---|---|
| `configId` (Map key) | `resolveBotConfigs(config).map(c => c.botId)` → watcher `loadDesired()` → `addBot(configId)` | ✅ — `.filter(id => !!id)` drops undefineds at both call sites (`index.ts:91`, `index.ts:117`) |
| `robotUid` (loop-guard key) | `startBot()` returns `{ robotUid: gateway.botId, … }` (`index.ts:329`) | ✅ — `gateway.botId` is the registered bot id from REST |
| `resolveBotConfigById` config | `loadConfig()` called inside manager's start function (`index.ts:81`) — re-reads from disk on each add | ✅ — picks up latest config.json for newly added bots |
| `latestGen` (generation guard) | Bumped in `schedule()` (`config-watcher.ts:118`) on file event, captured in `apply()` (`config-watcher.ts:101`) | ✅ — bumps at event time, not debounce expiry, so in-flight reconciles are immediately invalidated |
| `existing` list in addBot | `[...this.bots.values()]` captured BEFORE `crossRegisterOnAdd` (`bot-manager.ts:131`) | ✅ — rollback uses same list |
| `remaining` list in removeBot | `[...this.bots.values()]` captured AFTER `this.bots.delete(configId)` (`bot-manager.ts:156`) | ✅ — removed bot not in its own remaining list |
| Batch cross-registration | `[...existing, ...batch]` includes self in batch (`bot-manager.ts:192`) | ✅ — `crossRegisterOnAdd` self-guard (`e.robotUid === newBot.robotUid`) skips self |
| `shuttingDown` flag | Set synchronously in `shutdownAll()` body (`bot-manager.ts:92`) BEFORE enqueue | ✅ — queued addBot tasks see `true` when they execute |

## 6. Blind-spot Checklist (R5 — security_sensitive, all items mandatory)

- **C1 Dual-path parity (add ↔ remove, register ↔ unregister)**: ✅ Verified. `addBot` = start → cross-register → connect (with rollback). `removeBot` = delete from map → shutdown → cross-unregister. Order is correct: shutdown before unregister prevents draining handlers from being misidentified as user input by siblings. `registerKnownBot` ↔ `unregisterKnownBot` are symmetric; `unregisterKnownBot` correctly guards against self-removal. `addBotsBatch` ↔ `shutdownAll` — batch brings up as one unit, shutdownAll tears down concurrently (order doesn't matter at full teardown).
- **C1b Guard/lint tool self-coverage**: N/A — no lint/guard tools are modified by this PR.
- **C2 Control-flow ordering / nested reuse**: ✅ Verified. The serial queue (`enqueue`) correctly serializes all mutations. Nested enqueue (reconcile → addBot/removeBot → enqueue) works because the watcher's `chain` serializes reconcile runs and the manager's `queue` serializes mutations independently. The generation guard correctly prevents stale reconciles from acting.
- **C3 Authorization boundary ≠ capability boundary**: N/A — no new endpoints or authorization changes. Config file access is delegated to existing `loadConfig`/`resolveBotConfigs` which enforce `isAllowedApiUrl` and other validation.
- **C4 Authorization lifecycle / container-member state cascade**: N/A — no auth changes.
- **C5 Build ≠ runtime path correctness**: N/A — no build artifacts, extensions, or relative URL changes. This is a Node.js process with no packaging concerns.
- **C6 Governance/policy/security document self-consistency**: N/A — no governance or policy documents modified.

## 7. Cross-round Blocker Re-check (R6)

N/A — this is the automated review review. No prior blockers to re-check.

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 findings. The PR demonstrates careful attention to concurrency correctness (serial queue, two-phase batch boot, generation guard, shutdown race prevention), loop-guard invariant preservation (shutdown before unregister, bidirectional cross-registration), and resilience (invalid config keeps current bots, connect-failure rollback). Test coverage is thorough and directly validates the critical ordering invariants.

The two P2 items (process.exit removal, per-bot hot-reload follow-up) are non-blocking: the shutdown code is thorough enough that event-loop drain is a reasonable exit strategy, and the per-bot limitation is explicitly documented with no regression in currently-working paths.

yujiawei (changes requested):
# Code Review — PR #158 (cc-channel-octo)

Reviewed at head `c5ef173`. Built and ran the full suite against this SHA: `type-check` clean, `lint` clean (`--max-warnings 0`), `vitest` **1092 passing**. The design is solid — the serial mutation queue, two-phase batch boot, generation-counter staleness guard, and shutdown `shuttingDown` flag are all well-reasoned and well-tested. The new unit tests directly target the tricky races. One correctness gap blocks merge.

## Verdict: Changes Requested

---

## 1. Spec compliance

The diff implements exactly what #157 calls for (hot add/remove of bots from a config-dir watcher, no full gateway restart) and nothing out of scope. The "known limitation" (no in-place per-bot config reload) is explicitly declared in the PR body, so it is a documented scope boundary, not a miss. Backward-compatible idle/empty-config and "non-empty but all failed → fatal boot" semantics are preserved.

No missing, extra, or divergent functionality.

---

## 2. Code quality

### 🔴 P1 — Initial sync runs *before* the watcher is armed: a config change during boot is lost permanently

`src/index.ts:79-117`

```
const initialIds = resolveBotConfigs(config).map(...).filter(...);
const { failed } = await manager.addBotsBatch(initialIds);   // line 79 — SLOW (REST register + media probe per bot)
...
const watcher = watchConfig({ ... });                         // line 113 — fs.watch armed only here
```

The initial reconcile happens, and only *afterwards* is `fs.watch` installed — with no `applyNow()` to re-read disk after arming. The window between `loadConfig()` (line 68) and `watch()` (inside `watchConfig`, line 113) spans the entire `addBotsBatch`, which performs a REST `registerBot` and a media-CDN probe per bot (each bounded by a 30s timeout). Any write to `config.json` in that window produces **no** `fs.watch` event the watcher will see, and nothing schedules a later catch-up reconcile.

Impact: a provision (or removal) that lands during gateway startup is silently dropped. The bot stays offline indefinitely — until some *unrelated* later write to `config.json` happens to trigger a reconcile. Critically, this PR's whole purpose is to remove the full-gateway-restart on provision, so the previous self-healing path ("provision → restart → re-read config") is gone. A lost provision now has no recovery.

This is the standard watcher TOCTOU: arm the watcher *first*, then do the initial reconcile, so any event during setup is captured. Two equivalent fixes:
- Move `watchConfig({...})` above the `addBotsBatch` call and then `await watcher.applyNow()` for the initial sync; or
- Keep current order but `await watcher.applyNow()` immediately after `watchConfig(...)` to re-read the latest on-disk set.

`applyNow` already serializes through the same chain and is idempotent against the running set, so the catch-up reconcile is a no-op in the common case and only adds/removes the delta when a write was missed.

### 🟡 P2 — In-flight `addBot` ignores a newer generation, so a just-removed bot can still connect

`src/config-watcher.ts:55-70`, `src/bot-manager.ts:118-140`

`reconcile` checks `isStale()` *before* each `addBot`/`removeBot`, but not after the `await` returns. If a newer config event bumps `latestGen` while `manager.addBot(id)` is mid-flight (register → cross-register → connect), the add completes and the bot connects and is logged as added, even though the newer desired set no longer contains it. The next debounced reconcile then removes it. The documented guarantee ("an earlier, slower reconcile never starts a bot a later edit removed") is therefore not fully met — the bot can briefly be live and process messages. Self-healing and short-lived, hence P2, but worth either a post-await staleness re-check or a note that the guarantee is "eventually consistent."

### 🟡 P2 — Removed bot's loop-guard uid is dropped as soon as `shutdown()` returns

`src/bot-manager.ts:148-165`

On hot remove, the order is `shutdown()` → `crossUnregisterOnRemove()`, which is the right ordering. But `gateway.stop()` resolves after the **drain timeout** (default 10s) even if a long-running handler is still executing and may still emit a REST reply afterwards (or a delivery already queued in Octo arrives late). Once the uid is unregistered, remaining routers in a mention-free group no longer recognize those late messages as bot-authored and can treat them as human input — re-opening the very bot↔bot loop the guard prevents. Narrow (mention-free group; uid not ending in `_bot`, which `looksLikeBot` still catches), so P2. Consider retaining the removed uid in the loop guard for a grace period beyond shutdown, or document the residual window.

### 🟡 P2 — SIGTERM path no longer force-exits after a bounded drain

`src/index.ts:125-135`

The unified shutdown resolves `main()`'s promise instead of calling `process.exit(0)` (which the previous multi-bot shutdown did). If `gateway.stop()` hits its drain timeout with handler promises, SDK subprocesses, or in-flight fetches still holding event-loop handles, `main()` returns but the process can linger. A supervised `stop` will SIGKILL after its own timeout, so this is not data-loss, but a foreground/service `SIGTERM` may not exit promptly. Consider `process.exit(0)` once the bounded drain completes, matching prior behavior.

### Nit — `addBotsBatch` intra-batch duplicate handling is unreachable but unguarded

`src/bot-manager.ts:168-205`

Phase 1's `if (this.bots.has(configId)) continue;` does not catch duplicates *within* the same batch (the map is written only in Phase 3), so two entries with the same key would both start/connect and the first would be orphaned by the Phase-3 `set` overwrite. In practice this is unreachable: `resolveBotConfigs` throws on duplicate ids (`config.ts:528`) and duplicate tokens (`config.ts:553`) before `initialIds` is built, so the batch never contains dupes. Optional defense-in-depth: dedupe `configIds` at the top of `addBotsBatch`.

---

## 3. Items a human should verify (security-sensitive)

- **Companion-daemon atomicity (cross-repo):** correctness depends on the daemon writing `config.json` atomically (temp + rename). The watcher correctly watches the *directory* and keeps current bots on a parse failure, but a daemon that does a non-atomic in-place truncating write could still expose a transient empty/half file. Confirm the companion PR's write path is atomic, and that #158 lands together with / before it as the PR notes.
- **`fs.watch` reliability:** `fs.watch` on the config directory can miss events or behave inconsistently on some platforms and on network/overlay filesystems (NFS, certain container mounts). Worth confirming the deployment target's filesystem fires directory rename events reliably; the P1 fix above also reduces exposure to a single missed event.

---

## Summary

Strong, well-tested change. The blocker is the startup ordering: arm the watcher before (or re-read via `applyNow()` after) the initial batch so a provision during boot is never silently lost — important precisely because this PR removes the restart that used to recover that case. The P2s are smaller hardening items.

PR #161, opened 2026-06-26
fix(read-receipt): omit empty message_ids and silence best-effort failures

caster-Q:
## What

Harden the bot read-receipt path so an empty / unresolvable message id can never produce a 400 or log noise.

- `sendReadReceipt` filters out empty/blank ids and omits `message_ids` entirely when none are valid (the request still clears the conversation unread badge).
- Both call sites' `.catch` are now true no-ops — a read receipt is fire-and-forget and must never surface failures.
- Tests: empty / blank / mixed / non-empty message id cases assert the `message_ids` field presence and contents.

## Why

Read receipts are best-effort and decoupled from message delivery (no `await`, failures already caught). But the gateway sends `message_ids` unconditionally; when an id can't be resolved to a message_seq the server returns 400 and the gateway logged an error each time. Chatting was never affected, but the error log was noise.

Verified against the server handler: with no `message_ids`, the readReceipt path only clears the unread badge and returns 2xx — it never reaches the message-level lookup that can fail. Omitting the field is therefore the correct way to avoid the failure at the source.

This is a defensive guard ahead of going to production. The server-side error-downgrade (returning 400 instead of skipping an unresolvable id) is tracked separately in its own repository.

## Test

- `npm run type-check` / `npm run build` / `npm test` (1096 passing) / `npm run lint` (0 warnings)

### DIFF
--- a/src/__tests__/api.test.ts
+++ b/src/__tests__/api.test.ts
@@ -3,7 +3,8 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterAll } from 'vitest';
-import { getUploadCredentials } from '../octo/api.js';
+import { getUploadCredentials, sendReadReceipt } from '../octo/api.js';
+import { ChannelType } from '../octo/types.js';
 
 // Mock global fetch
 const fetchMock = vi.fn();
@@ -135,5 +136,47 @@ describe('getUploadCredentials', () => {
     // Error prefix + 200 chars max + nothing more
     expect(captured!.message.length).toBeLessThan(300);
     expect(captured!.message).not.toMatch(/x{300,}/);
+  });
+});
+
+describe('sendReadReceipt — message_ids guard', () => {
+  const RR_BASE = {
+    apiUrl: 'https://api.example.com',
+    botToken: 'test-token',
+    channelId: 'chan-1',
+    channelType: ChannelType.DM,
+  };
+
+  function bodyOf(callIndex = 0): Record<string, unknown> {
+    const init = fetchMock.mock.calls[callIndex][1] as RequestInit;
+    return JSON.parse(init.body as string);
+  }
+
+  it('includes message_ids when non-empty', async () => {
+    fetchMock.mockResolvedValueOnce(mockJsonResponse({}));
+    await sendReadReceipt({ ...RR_BASE, messageIds: ['123', '456'] });
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    expect(bodyOf()).toMatchObject({ message_ids: ['123', '456'] });
+  });
+
+  it('omits message_ids when the array is empty', async () => {
+    fetchMock.mockResolvedValueOnce(mockJsonResponse({}));
+    await sendReadReceipt({ ...RR_BASE, messageIds: [] });
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    expect('message_ids' in bodyOf()).toBe(false);
+    // still clears the unread badge — channel_id is present
+    expect(bodyOf()).toMatchObject({ channel_id: 'chan-1' });
+  });
+
+  it('omits message_ids when every id is empty or blank', async () => {
+    fetchMock.mockResolvedValueOnce(mockJsonResponse({}));
+    await sendReadReceipt({ ...RR_BASE, messageIds: ['', '   '] });
+    expect('message_ids' in bodyOf()).toBe(false);
+  });
+
+  it('filters out blank ids and keeps the real ones', async () => {
+    fetchMock.mockResolvedValueOnce(mockJsonResponse({}));
+    await sendReadReceipt({ ...RR_BASE, messageIds: ['x', '', '  '] });
+    expect(bodyOf()).toMatchObject({ message_ids: ['x'] });
   });
 });
--- a/src/index.ts
+++ b/src/index.ts
@@ -416,7 +416,7 @@ export async function handleMessage(
               channelId: msg.channel_id,
               channelType: msg.channel_type,
               messageIds: [msg.message_id],
-            }).catch((err) => console.error(`[cc-channel-octo] readReceipt failed: ${String(err)}`));
+            }).catch(() => { /* read receipt is best-effort; never surface failures */ });
           }
           return; // skip context, history, and the agent query entirely
         }
@@ -852,7 +852,7 @@ export async function handleMessage(
           channelId: msg.channel_id,
           channelType: msg.channel_type,
           messageIds: [msg.message_id],
-        }).catch((err) => console.error(`[cc-channel-octo] readReceipt failed: ${String(err)}`));
+        }).catch(() => { /* read receipt is best-effort; never surface failures */ });
       }
 
       // --- Store assistant response in history ---
--- a/src/octo/api.ts
+++ b/src/octo/api.ts
@@ -293,10 +293,15 @@ export async function sendReadReceipt(params: {
   messageIds: string[];
   signal?: AbortSignal;
 }): Promise<void> {
+  // Only send message-level ids that are actually present. An empty / blank id
+  // makes the server resolve a message_seq it cannot find, which the IM backend
+  // rejects — so omit message_ids entirely when there is nothing valid to ack
+  // (the request then just clears the conversation unread badge).
+  const ids = (params.messageIds ?? []).filter((id) => id && id.trim() !== '');
   await postJson(params.apiUrl, params.botToken, '/v1/bot/readReceipt', {
     channel_id: params.channelId,
     channel_type: params.channelType,
-    message_ids: params.messageIds,
+    ...(ids.length > 0 ? { message_ids: ids } : {}),
   }, params.signal);
 }
 


OctoBoooot (commented):
## Review: fix(read-receipt): omit empty message_ids and silence best-effort failures (#161)

**Verdict:** Approve with comments — clean defensive fix, only the project-board CI gate stands in the way.

CI is red — `check-sprint / check-sprint` is failing (project-board gate, not code). Vitest / ESLint / type-check all green.

Risk tier: normal. 3 files, +52/-4.

### Praise
- The filter `(params.messageIds ?? []).filter((id) => id && id.trim() !== '')` is precise: `id &&` rejects `null`/`undefined`/empty string in one short-circuit, `id.trim() !== ''` then rejects whitespace-only. No false-positive on string `'0'` (truthy, `.trim() !== ''`).
- Conditional-spread omission `...(ids.length > 0 ? { message_ids: ids } : {})` correctly drops the field instead of sending `"message_ids": []`. PR body confirms server treats missing-field as "clear unread badge only," so this is the right shape — not an "empty list" semantic.
- Tests map 1:1 to the filter branches: non-empty / empty-array / all-blank / mixed. Each asserts both the field's presence/absence AND that the request still carries `channel_id` so the badge-clear path keeps working. That's exactly the test matrix I'd want.
- PR body cites the server-side verification ("with no `message_ids`, the readReceipt path only clears the unread badge and returns 2xx") and notes the corresponding server-side fix is tracked separately — so the gateway-side defensive guard is honestly framed as belt-and-suspenders, not a workaround hiding a real server bug.
- The silent `.catch(() => {})` change is paired with the api.ts filter — the previously-noisy "unresolvable id → 400 → log" case is now PREVENTED at the source, so the silent swallow only covers genuine network/auth failures (operational rather than per-message). That's the right way to drop log noise without losing the signal.

- `check-sprint` red is the same project-board gate as #156/#158/#159/#160 — coordinate with whoever owns project #2 if it's required for merge.
- Coverage-sniff flagged `src/octo/api.ts` and `src/index.ts` as untested by naming convention, but the new `src/__tests__/api.test.ts` block covers `sendReadReceipt` directly; the index.ts changes are `.catch` swaps inside existing well-tested `handleMessage` paths. False positive on both.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#161 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo#161
**Head SHA:** `ab116395b6de1840748a5ac84645a881e3babb6e`
**Title:** fix(read-receipt): omit empty message_ids and silence best-effort failures
**Files changed:** 3 (+52 / -4)

---

## 1. 验证结论

| Item | Status | Evidence |
|------|--------|----------|
| Empty/blank id filtering in `sendReadReceipt` | ✅ | `src/octo/api.ts` L298: `(params.messageIds ?? []).filter((id) => id && id.trim() !== '')` — correctly rejects `''`, whitespace-only, `null`, `undefined`, `0` |
| Conditional `message_ids` field omission | ✅ | `src/octo/api.ts` L299–302: `...(ids.length > 0 ? { message_ids: ids } : {})` — spread-merges the field only when valid ids exist; absent otherwise |
| `.catch` silenced at both call sites | ✅ | `src/index.ts` L419 (command path) + L855 (normal path): `.catch(() => { /* … */ })` — both symmetric, both fire-and-forget |
| Test coverage for new guard | ✅ | `src/__tests__/api.test.ts` L141–181: 4 test cases covering non-empty, empty array, all-blank, and mixed ids |
| `channel_id` still sent when `message_ids` omitted | ✅ | Test at L162: `expect(bodyOf()).toMatchObject({ channel_id: 'chan-1' })` confirms unread-badge clearing still works |

## 2. 发现的问题

**No P0 or P1 findings.**

### P2 — `.catch` silencing reduces operational observability (note, non-blocking)

- **Diff-scope:** new (previously logged errors via `console.error`; now silent)
- The change from `console.error` to a no-op `.catch` is consistent with fire-and-forget semantics and the PR's stated goal of eliminating log noise. However, if a systemic issue arises (e.g. auth token expired, server down), all read-receipt failures become invisible — no log, no metric, no alert signal. This is a design trade-off, not a bug. Consider adding a `debug`-level log or a counter metric in the future if observability becomes a concern.

## 3. 建议

- None required. The change is minimal, focused, and well-tested.

## 4. 额外发现

- None.

## 5. 数据流回溯

### `params.messageIds` at `sendReadReceipt` entry

- **Source (call site 1, command path):** `src/index.ts` L413–419. Guard `if (msg.message_id && …)` ensures `msg.message_id` is truthy before wrapping in `[msg.message_id]`. Empty string `''` would be excluded by this guard.
- **Source (call site 2, normal path):** `src/index.ts` L849–855. Identical guard pattern.
- **`BotMessage.message_id` type:** `string` (non-optional in `src/octo/types.ts` L18). The WS message parser populates this field; a missing field would be `undefined` at runtime despite the type, but the `msg.message_id &&` guard catches both cases.
- **Runtime flow:** At both call sites, when the guard passes, `messageIds` is always a single-element array with a truthy string. The new filter in `sendReadReceipt` is therefore a no-op for current callers — its value is as a defense-in-depth guard for future callers or edge cases where `message_id` might be a non-empty-but-blank string.

### `ids` consumed by spread operator

- **Construction:** `const ids = (params.messageIds ?? []).filter(...)` — always produces a clean string array.
- **Consumption:** `...(ids.length > 0 ? { message_ids: ids } : {})` — correctly omits the key when empty. `postJson` then `JSON.stringify`s the payload, so the field is genuinely absent from the wire request (not `null` or `[]`).

### `.catch` at both call sites

- **Before:** `.catch((err) => console.error(...))` — logged the error but didn't rethrow. Failure was already swallowed.
- **After:** `.catch(() => { /* … */ })` — same swallow behavior, just removes the log line.
- **No behavioral change** to the main message-handling pipeline.

## 6. 盲点 checklist（R5）

- **C1 双路径 parity（含守卫/lint 工具自身等价写法覆盖）:** The two `sendReadReceipt` call sites (command path L413, normal path L849) are structurally identical — same guard, same arguments, same `.catch`. Parity confirmed. ✅
- **C2 control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿:** `sendReadReceipt` is a leaf function with no nested reuse or ordering concerns. The `.filter((id) => id && id.trim() !== '')` guard handles the non-canonical blank-string variants (whitespace, tabs). N/A for ordering. ✅
- **C3 授权边界 ≠ 能力边界:** No authorization or capability changes. The `botToken` is the same as before. N/A. ✅
- **C4 授权生命周期 / 容器-成员状态级联:** No auth changes. N/A. ✅
- **C5 build/note 通过 ≠ 运行期路径正确:** No build artifacts, extensions, or relative-path changes. The runtime HTTP call to `/v1/bot/readReceipt` is unchanged; only the payload shape varies. N/A. ✅
- **C6 治理/策略/安全文档自洽性:** No documentation or policy changes. N/A. ✅

## 7. 跨轮 blocker 复检（R6）

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 findings. The change is a clean, well-scoped defensive guard with good test coverage. The only note (P2) is a minor observability trade-off from silencing `.catch`, which is a reasonable design choice for fire-and-forget read receipts.

yujiawei (approved):
# Code Review — PR #161 (cc-channel-octo)

Reviewed at head `ab116395b6de1840748a5ac84645a881e3babb6e`. Independent review with a cross-model adversarial pass.

## 1. Spec compliance

**Spec: ✅**

The diff does exactly what the PR description states, no more, no less:
- `sendReadReceipt` filters empty/blank ids and omits `message_ids` entirely when none are valid (`src/octo/api.ts:300-306`).
- Both call-site `.catch` handlers are now true no-ops (`src/index.ts:419`, `src/index.ts:855`).
- Tests cover non-empty / empty-array / all-blank / mixed cases asserting `message_ids` presence and contents (`src/__tests__/api.test.ts:142-182`).

No missing items, no out-of-scope additions, no deviation from the agreed approach.

## 2. Code quality

**Quality: Approved**

The change is small, correct, and the conditional spread `...(ids.length > 0 ? { message_ids: ids } : {})` is structurally sound (no TypeScript issue). The filter is reachable only defensively: both production call sites guard `if (msg.message_id && ...)` and pass `[msg.message_id]`, so an empty/blank id never originates there — the guard hardens the exported function for any future caller, which is reasonable.

No P0/P1 findings. The following are **P2 / advisory** (not merge blockers):

- **P2 — Silent error swallowing (`src/index.ts:419`, `src/index.ts:855`).** `.catch(() => {})` now discards every failure class: auth/token expiry, network outages, malformed payloads, request-construction bugs. Best-effort and fire-and-forget is the right contract, but this removes the only local signal that read receipts have stopped working entirely. Consider a `debug`-level log or a metrics counter instead of total silence, so a systemic regression is still observable. *(Both independent review passes flagged this — highest-confidence point.)*

- **P2 — Test assertion overclaims (`src/__tests__/api.test.ts:168`).** The "still clears the unread badge" comment is stronger than what the test proves: it only asserts `channel_id` is serialized into the request body, not that the server accepts a body without `message_ids` or actually clears unread state. The omit-`message_ids`-clears-badge behavior is the load-bearing assumption of this PR and lives in the server repo; it is not — and arguably cannot be — verified by this unit test. Worth softening the comment or noting the behavior is verified server-side, not here.

- **P2 — Blank filter does not trim (`src/octo/api.ts:302`, `:306`).** `(id) => id && id.trim() !== ''` rejects blank ids but passes whitespace-padded non-blank ids unchanged, so `' 123 '` would be sent verbatim. Low impact in practice (call sites pass server-provided ids), but if the intent is "only valid ids," `map(id => id.trim()).filter(Boolean)` normalizes, or add a test documenting that padding is intentionally preserved.

### Coverage notes
- The core claim — "omitting `message_ids` still clears the unread badge and returns 2xx" — depends on the IM server handler, which is in a different repo not present in this diff. Taken on the author's stated server-side verification; not independently confirmable here.
- `postJson` error/timeout behavior, read-receipt endpoint rate limits, and whether any external telemetry already covers these now-silent failures were out of scope for this diff.

## 3. Verdict

**APPROVED.** Spec ✅ and no blocking quality issues. The three P2 items are observability/test-clarity improvements that can land in a follow-up; none should hold up this defensive hardening.

Jerry-Xin (approved):
The PR is relevant to `cc-channel-octo` and cleanly hardens the Octo read-receipt path without affecting message delivery.

🔴 Blocking: None.

💬 Non-blocking:
- 🔵 Suggestion: `src/octo/api.ts:300` defensively accepts `params.messageIds ?? []`, but the public type still requires `messageIds: string[]`. That is fine for current call sites, though making it optional would better match the runtime guard if callers may omit it later.

✅ Highlights:
- The `message_ids` omission logic in `src/octo/api.ts` correctly filters blank/whitespace ids and uses a conditional spread so an empty list omits the field entirely (rather than sending `[]`), matching the stated server behavior — the request still clears the conversation unread badge via `channel_id`.
- The fire-and-forget `.catch` handlers in `src/index.ts:419` and `src/index.ts:855` are scoped to best-effort read-receipt calls only, consistent with the PR's "read receipts must not surface failures" contract; message delivery error handling is unaffected.
- The added tests in `src/__tests__/api.test.ts` cover empty, blank-only, mixed, and normal id payloads (1:1 with the filter branches).

PR #167, opened 2026-06-27
fix(agent-bridge): match Bedrock ValidationException in isResumeError (#154)

### LINKED ISSUE #154: fix(agent-bridge): interrupted turns leave orphan tool_use blocks causing ValidationException on next resume

lml2468:
## Summary

When an agent turn is interrupted mid-execution (process kill, timeout, or restart) while the Claude Agent SDK has already emitted `tool_use` blocks, the SDK session is left with orphaned `tool_use` blocks (no corresponding `tool_result`). On the next resume, the API returns a `ValidationException` that `isResumeError()` did not match, so `onResumeFailed()` was never called, the corrupt `sdk_session_id` was never cleared, and every subsequent turn failed with the same error.

## Root Cause

`isResumeError()` in `agent-bridge.ts` only matched:
```
/No conversation found with session ID|--resume requires a valid session/i
```

It did **not** match the Bedrock `ValidationException`:
```
ValidationException: messages.805: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_...
```

## Fix

Extended the `isResumeError` regex to also match:
- `tool_use.*ids were found without.*tool_result` — Bedrock error shape
- `tool_result.*blocks.*immediately after` — Anthropic-direct variant

The existing recovery path (clear stale `sdk_session_id` → retry once without resume using `fallbackRetryPrompt`) now fires for this error class, matching the behavior for stale/expired sessions.

## Tests

Added 3 regression tests to `agent-bridge-query.test.ts`:

1. **`recovers from a Bedrock ValidationException for orphaned tool_use blocks (no prior output)`** — full recovery: `onResumeFailed` called, retry WITHOUT resume, fresh session id captured
2. **`clears the stale id when a Bedrock ValidationException arrives AFTER output (no retry)`** — no retry (partial output already streamed), but stale id still cleared so next turn recovers
3. **`matches Anthropic-direct error shape for orphaned tool_use (variant wording)`** — covers alternative phrasing from Anthropic-direct API

All 1099 tests pass.

## Verification

- Pre-commit hook (eslint + type-check): ✅
- Pre-push hook (full vitest run, 1099 tests): ✅
- TypeScript build (`tsc`): ✅

Fixes #154

### DIFF
--- a/src/agent-bridge.ts
+++ b/src/agent-bridge.ts
@@ -307,10 +307,14 @@ export async function* queryAgent(
     });
 
   // Detect the SDK's stale/invalid-resume signal (verified via spike): it throws
-  // an Error whose message names the missing/invalid session id.
+  // an Error whose message names the missing/invalid session id. Also matches
+  // the Bedrock/Anthropic-direct ValidationException for orphaned tool_use
+  // blocks — when a turn is interrupted mid-execution, the SDK session is left
+  // with tool_use blocks that have no matching tool_result, and the next resume
+  // attempt must be treated as stale (#154).
   const isResumeError = (err: unknown): boolean => {
     const m = err instanceof Error ? err.message : String(err);
-    return /No conversation found with session ID|--resume requires a valid session/i.test(m);
+    return /No conversation found with session ID|--resume requires a valid session|tool_use.*ids were found without.*tool_result|tool_result.*blocks.*immediately after/i.test(m);
   };
 
   const stream = runStream(opts?.resume, userMessage);
--- a/src/__tests__/agent-bridge-query.test.ts
+++ b/src/__tests__/agent-bridge-query.test.ts
@@ -639,6 +639,94 @@ describe('queryAgent', () => {
     expect(mockQuery).toHaveBeenCalledTimes(1); // no retry after a tool_use side effect
     expect(resumeFailed).toBe(true); // but the stale id was cleared
   });
+
+  // --- orphaned tool_use ValidationException recovery (#154) ---
+  // Bedrock's ValidationException when a session carries tool_use blocks without
+  // corresponding tool_result blocks was NOT matched by isResumeError, so
+  // onResumeFailed was never called and the corrupt session id persisted.
+
+  it('recovers from a Bedrock ValidationException for orphaned tool_use blocks (no prior output)', async () => {
+    // The exact error shape Bedrock returns when a session carries tool_use
+    // blocks without matching tool_result blocks (observed in production, #154).
+    const bedrockErr = new Error(
+      'ValidationException: messages.805: `tool_use` ids were found without ' +
+      '`tool_result` blocks immediately after: toolu_757190a2ce1b41e48ebf8ebd, ' +
+      'toolu_c6e1a12e9f234cf493debbb5. Each `tool_use` block must have a ' +
+      'corresponding `tool_result` block.',
+    );
+    const throwing = createMockStream([]);
+    throwing[Symbol.asyncIterator] = async function* () { throw bedrockErr; };
+    const ok = createMockStream([
+      { type: 'assistant', session_id: 'fresh-sid', message: { content: [{ type: 'text', text: 'recovered' }] } },
+    ]);
+    mockQuery.mockReturnValueOnce(throwing).mockReturnValueOnce(ok);
+
+    let resumeFailed = false;
+    const ids: string[] = [];
+    const chunks: string[] = [];
+    for await (const chunk of queryAgent('hi', makeConfig(), undefined, undefined, {
+      resume: 'stale-sid',
+      onSessionId: (id) => ids.push(id),
+      onResumeFailed: () => { resumeFailed = true; },
+      fallbackRetryPrompt: 'fallback',
+    })) {
+      chunks.push(chunk);
+    }
+
+    expect(resumeFailed).toBe(true);
+    expect(chunks).toEqual(['recovered']);
+    expect(mockQuery).toHaveBeenCalledTimes(2);
+    expect(mockQuery.mock.calls[0][0].options.resume).toBe('stale-sid');
+    expect(mockQuery.mock.calls[1][0].options).not.toHaveProperty('resume');
+    expect(ids).toEqual(['fresh-sid']);
+  });
+
+  it('clears the stale id when a Bedrock ValidationException arrives AFTER output (no retry)', async () => {
+    const partial = createMockStream([]);
+    partial[Symbol.asyncIterator] = async function* () {
+      yield { type: 'assistant', session_id: 's', message: { content: [{ type: 'text', text: 'partial' }] } };
+      throw new Error(
+        'ValidationException: messages.12: `tool_use` ids were found without ' +
+        '`tool_result` blocks immediately after: toolu_abc. Each `tool_use` block ' +
+        'must have a corresponding `tool_result` block.',
+      );
+    };
+    mockQuery.mockReturnValue(partial);
+    const chunks: string[] = [];
+    let resumeFailed = false;
+    await expect(async () => {
+      for await (const chunk of queryAgent('hi', makeConfig(), undefined, undefined, {
+        resume: 'sid', onResumeFailed: () => { resumeFailed = true; },
+      })) { chunks.push(chunk); }
+    }).rejects.toThrow('ValidationException');
+    expect(chunks).toEqual(['partial']);
+    expect(mockQuery).toHaveBeenCalledTimes(1); // no retry after partial output
+    expect(resumeFailed).toBe(true); // stale id still cleared
+  });
+
+  it('matches Anthropic-direct error shape for orphaned tool_use (variant wording)', async () => {
+    // The Anthropic-direct API may phrase it differently. The regex must cover:
+    // "tool_result blocks immediately after" variant too.
+    const throwing = createMockStream([]);
+    throwing[Symbol.asyncIterator] = async function* () {
+      throw new Error('tool_result blocks immediately after: toolu_xyz');
+    };
+    const ok = createMockStream([
+      { type: 'assistant', session_id: 'new', message: { content: [{ type: 'text', text: 'ok' }] } },
+    ]);
+    mockQuery.mockReturnValueOnce(throwing).mockReturnValueOnce(ok);
+
+    let resumeFailed = false;
+    const chunks: string[] = [];
+    for await (const chunk of queryAgent('hi', makeConfig(), undefined, undefined, {
+      resume: 'stale',
+      onResumeFailed: () => { resumeFailed = true; },
+    })) {
+      chunks.push(chunk);
+    }
+    expect(resumeFailed).toBe(true);
+    expect(chunks).toEqual(['ok']);
+  });
 });
 
 describe('queryAgent — sdk.env injection (#107)', () => {


Jerry-Xin (approved):
This PR is in scope and correctly extends the existing resume-recovery path for corrupted SDK sessions with orphaned `tool_use` state.

💬 Non-blocking
- 🔵 Suggestion: [src/agent-bridge.ts](src/agent-bridge.ts:317) could be split into a named multi-line regex or helper list for readability. The current pattern is functionally fine, but it is getting dense as more resume-error shapes are added.

✅ Highlights
- The new matching remains gated by `opts.resume`, so unrelated non-resume calls are not affected. See [src/agent-bridge.ts](src/agent-bridge.ts:399) and [src/agent-bridge.ts](src/agent-bridge.ts:429).
- The no-output retry path and after-output clear-only path are both covered, which is the key behavioral distinction for avoiding duplicate side effects. See [src/__tests__/agent-bridge-query.test.ts](src/__tests__/agent-bridge-query.test.ts:648) and [src/__tests__/agent-bridge-query.test.ts](src/__tests__/agent-bridge-query.test.ts:684).
- The alternative provider wording is covered by a regression test at [src/__tests__/agent-bridge-query.test.ts](src/__tests__/agent-bridge-query.test.ts:707).

Local verification note: I attempted `npm test -- src/__tests__/agent-bridge-query.test.ts`, but this checkout does not have dependencies installed, so `vitest` was not available.

src/agent-bridge.ts
```diff
@@ -307,10 +307,14 @@ export async function* queryAgent(
...
-  // an Error whose message names the missing/invalid session id.
+  // an Error whose message names the missing/invalid session id. Also matches
+  // the Bedrock/Anthropic-direct ValidationException for orphaned tool_use
+  // blocks — when a turn is interrupted mid-execution, the SDK session is left
+  // with tool_use blocks that have no matching tool_result, and the next resume
+  // attempt must be treated as stale (#154).
   const isResumeError = (err: unknown): boolean => {
     const m = err instanceof Error ? err.message : String(err);
-    return /No conversation found with session ID|--resume requires a valid session/i.test(m);
+    return /No conversation found with session ID|--resume requires a valid session|tool_use.*ids were found without.*tool_result|tool_result.*blocks.*immediately after/i.test(m);
```
OctoBoooot, line 317: **[major]** this is the same `isResumeError` regex change as the still-open #159 (also Fixes #154, also 3 APPROVEs). Two open PRs with the same diff for one issue will conflict / produce a redundant change. Deconflict before merge: #167 has the stronger test (test 3 isolates the `tool_result.*blocks.*immediately after` branch, which was effectively dead in #159), so porting that test into #159 and closing this — or closing #159 and landing this — both work. Just don't merge both.

OctoBoooot (changes requested):
## Review: fix(agent-bridge): match Bedrock ValidationException in isResumeError (#154) (#167)

**Verdict:** Request changes — the fix is correct and better-tested than its twin, but it duplicates an already-open PR (#159) for the same issue; the two must be deconflicted before either merges.

Note: this PR is a **draft**. CI is green.

### Major
- **Duplicate of the still-open #159.** PR #159 ("extend isResumeError to catch Bedrock ValidationException for orphan tool_use blocks") makes the identical `agent-bridge.ts` regex change for the same issue #154 and is OPEN with 3 APPROVEs. Landing both would produce a redundant/conflicting change. Decide which one merges: #159 has reviewer sign-off already; #167 has the stronger test (see praise). Cleanest path is probably to port #167's test 3 into #159 and close this, OR close #159 and land this. Either way, two open PRs fixing one issue with the same diff needs deconfliction before merge — not a code defect, but a real "should not merge as-is alongside #159."

### Praise
- **Test 3 (`matches Anthropic-direct error shape`) closes the exact gap reviewers found in #159.** Its input `'tool_result blocks immediately after: toolu_xyz'` contains no `tool_use`, so the 3rd regex branch (`tool_use.*ids were found without.*tool_result`) cannot match — only the 4th branch (`tool_result.*blocks.*immediately after`) can. In #159 that 4th branch was effectively dead (every test string also matched branch 3, and the real multi-line Bedrock message doesn't cross `\n` without the `s` flag). Here the single-line, tool_use-free input genuinely exercises branch 4 in isolation. That's the right regression test for a multi-alternation matcher.
- Test 2 covers the no-double-side-effect boundary (ValidationException after partial output → no retry, but stale id still cleared) — same rigor I praised on #159.

### Questions
- Both branches still rely on default (non-`s`) regex flags. Test 3 passes because its input is single-line, but the real Bedrock message in test 1 puts "blocks" and "immediately after" on the same line via string concatenation (no `\n`), so it matches branch 3 anyway. Worth confirming the production error is single-line per logical match — if Bedrock ever wraps `tool_result blocks\nimmediately after`, branch 4 would silently stop matching. Adding the `s` flag or `[\s\S]*?` would make branch 4 robust to that. Non-blocking.

- Issue #154 will be closed by whichever of #159/#167 lands — make sure the other is closed manually so the issue isn't reopened by a dangling "Fixes #154".

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#167 Review Report
**PR:** https://github.com/Mininglamp-OSS/cc-channel-octo/pull/167
**Head SHA:** adf63b54b68353326414c02c0e1b417a18bda855
**Scope:** fix(agent-bridge): match Bedrock ValidationException in isResumeError (#154)

---

- ✅ **Root cause correctly identified**: `isResumeError()` regex was too narrow — only matched two SDK-origin error patterns, missed Bedrock/Anthropic-direct `ValidationException` for orphaned `tool_use` blocks.
- ✅ **Fix correctly extends regex**: Two new alternatives added, both matching real production error shapes.
- ✅ **Recovery path reused (not duplicated)**: The existing stale-session recovery (clear `sdk_session_id` → retry once without resume using `fallbackRetryPrompt`) now fires for this error class. No new control flow.
- ✅ **Tests are behavioral, not hand-fed**: All 3 new tests exercise the real `queryAgent` code path through mock SDK streams, verifying `onResumeFailed` invocation, retry/no-retry decisions, and fresh session id capture.
- ✅ **Comments updated**: The `isResumeError` comment block now documents the new matching behavior and the scenario it addresses.

## 2. Findings

### No P0 or P1 issues found.

### P2 / Nit Observations

**P2-1: Regex alternative 4 breadth**
- The fourth regex alternative `tool_result.*blocks.*immediately after` would match any error containing that phrase sequence, even without `tool_use` context. In practice, "tool_result blocks immediately after" is a highly specific API-origin phrase, so false-positive risk is negligible.
- **diff-scope**: new (this PR adds the alternative)
- **Severity**: P2/nit — not blocking.

## 3. Suggestions

- None blocking. The PR is focused, minimal, and well-tested.

## 4. Extra Findings

- None. The change is surgically scoped to the regex extension and its test coverage.

## 5. Data-Flow Tracing

### `isResumeError(err)` — consumer of SDK error messages

| Consumer | Upstream Source | Runtime Data Flow |
|---|---|---|
| `isResumeError` (line ~317) | `err` caught in `catch` block (line ~392) of `drainStream` try/catch | SDK async iterator throws → `err` propagates to catch → `err.message` extracted → regex tested. **Verified: data flows correctly.** |
| `onResumeFailed()` (line ~401) | Caller (`index.ts` line ~779): `() => store.clearSdkSessionId(sessionKey)` | When `isResumeError` returns true AND `opts?.resume` is set → callback fires → stored session id cleared. **Verified: callback chain intact.** |
| `fallbackRetryPrompt` (line ~414) | Caller (`index.ts` line ~728-730): `assembleUserMessage(historyBlock, userBody, MAX_USER_LLM_BYTES)` | Pre-assembled by caller from separate history + user body. Used only on recovery retry (no resume, no double-anchoring). **Verified: data flows correctly, no double-anchor risk.** |
| `emitted.any` gate (line ~393) | Set in `drainStream` (line ~359) on any assistant content block (text OR tool_use) | If error arrives before any output → retry safe. If after → no retry (side effect risk), but stale id still cleared. **Verified: gate logic correct for both paths.** |

### Test data-flow verification

- **Test 1 (Bedrock, no prior output)**: Mock throws immediately → `emitted.any=false` → recovery fires → retry stream yields `'recovered'` → assertions confirm: `resumeFailed=true`, retry without resume, fresh session id captured. ✅
- **Test 2 (Bedrock, after output)**: Mock yields `'partial'` then throws → `emitted.any=true` → no retry → error rethrown → but `onResumeFailed` still called. ✅
- **Test 3 (Anthropic-direct variant)**: Tests alternative regex pattern `tool_result.*blocks.*immediately after` → full recovery path. ✅

- **C1 Dual-path parity**: N/A — single `isResumeError` definition, single call-site pattern. No symmetric paths.
- **C2 Control-flow ordering / nesting**: N/A — regex extension is a pure additive match; no reordering, no nested reuse.
- **C3 Authorization boundary**: N/A — no auth/permission/capability changes.
- **C4 Authorization lifecycle / container-member cascade**: N/A — no auth changes.
- **C5 Build ≠ runtime**: N/A — no build artifacts, packaging, or relative-path changes.
- **C6 Governance / policy docs**: N/A — no documentation/policy changes.

---

[Octo-Q] verdict: APPROVE

**Rationale:** Zero P0/P1 findings. The regex extension is correct, minimal, and well-covered by 3 behavioral regression tests. Data-flow tracing confirms the existing recovery path (clear stale session id → retry without resume) fires correctly for the newly-matched error class. No false-positive risk in practice given the specificity of the API error phrases. P2 nit on regex alternative 4 breadth is non-blocking.

yujiawei (approved):
# Code Review — PR #167 (cc-channel-octo)

**Verdict: APPROVE** — the fix is correct, minimal, well-tested, and matches the production symptom in #154. The notes below are non-blocking P2 robustness suggestions only.

## Summary

Extends `isResumeError()` in `agent-bridge.ts` so that the Bedrock/Anthropic `ValidationException` for orphaned `tool_use` blocks (a turn interrupted after `tool_use` was emitted but before `tool_result`) is now treated as a stale-resume signal. This wires the failure into the existing recovery path: clear the corrupt `sdk_session_id` via `onResumeFailed()`, then retry once without resume using `fallbackRetryPrompt`. Before this change the error fell through and every subsequent turn on that channel failed until the session expired.

## Spec compliance

The change matches the proposed fix in #154 exactly, with nothing added beyond it:
- **Missing**: none — the regex is extended for both the Bedrock shape (`tool_use ... ids were found without ... tool_result`) and the Anthropic-direct variant (`tool_result ... blocks ... immediately after`), and tests cover both, as the issue requested.
- **Extra**: none — no new flags, fields, or behavior; only the regex and 3 regression tests.
- **Deviation**: none.

## Correctness

- The recovery (retry-without-resume) path is double-guarded by `opts?.resume && !emitted.any` (`agent-bridge.ts:399`), so it only fires on a resume turn before any output — preserving the existing "never retry after a side effect" guarantee.
- The after-output branch (`agent-bridge.ts:429`) correctly clears the stale id without retrying, so the *next* turn recovers cleanly. Tests assert exactly this.
- Verified locally on the head SHA: `tsc --noEmit` clean, `eslint` clean, and `agent-bridge-query.test.ts` passes 44/44 (including the 3 new cases).

1. **Matcher breadth** (`agent-bridge.ts:317`). The `tool_result.*blocks.*immediately after` alternative matches on the ordering phrase alone, without requiring a `ValidationException` / `tool_use ... without ... tool_result` context. In principle a genuine live-turn tool ordering error on a resumed request could be classified as stale-session corruption. The blast radius is small (the path is gated by `opts?.resume && !emitted.any`, so the worst case is a single bounded fresh retry), but tightening the orphaned-session alternative to require the `tool_use ... without ... tool_result` signature would make the classifier more precise.

2. **Newline spanning** (`agent-bridge.ts:317`). `.*` does not cross `\n`/`\r` without the `s` (dotAll) flag. Observed Bedrock errors are single-line, so this is defensive, but if an upstream ever line-wraps the message the match would silently fail (fail-closed → unrecoverable session). Consider `[\s\S]*` or adding the `s` flag.

3. **Nested error messages** (`agent-bridge.ts:316`). The classifier inspects `err.message` / `String(err)` only. If the SDK ever wraps the `ValidationException` in `cause.message` or a structured response body, the match is missed and the stale id persists. This mirrors the pre-existing behavior (the two original patterns matched the same way) and the issue observed the string directly on `.message`, so it is not a regression — just a hardening opportunity.

## Coverage note

Could not exercise the real Bedrock/Anthropic SDK error object shapes at runtime, nor confirm whether tool side effects can occur before any assistant content is emitted; both are covered by the existing `!emitted.any` guard and the tests' mocked error shapes rather than live traffic.

Issue #88, opened 2026-06-08
feat(threads+md): server-driven GROUP.md/THREAD.md + Thread (子区) support


lml2468:
**Priority: P3 (needed if your Octo deployment uses 子区/threads).** Parity gap vs openclaw-channel-octo.

## Problem
Two coupled gaps:
1. **Thread (CommunityTopic / 子区) support** — openclaw parses the `groupNo____shortId` channel-id format and has a full thread lifecycle API (create/list/get/delete/join/leave/members). cc has **zero thread awareness** — it treats thread channel ids as opaque.
2. **Server-driven GROUP.md / THREAD.md** — openclaw fetches per-group/thread instruction markdown from the backend (versioned, disk-cached, refreshed per session, `group_md_updated/deleted` events hot-reload, agent can write back). cc's `group-config.ts` only reads a **local operator file** `<dir>/<groupId>.md` — no API, no versioning, no events, no write-back, no THREAD.md.

## Scope
- Parse `groupNo____shortId` (THREAD_ID_SEPARATOR) so thread channels route correctly and get their own session/cwd/memory partition.
- Add the thread lifecycle REST calls to `octo/api.ts` (create/list/get/delete/join/leave/members).
- Server GROUP.md/THREAD.md: fetch via API with version+disk cache + per-session refresh; handle `group_md_updated/deleted` + `thread_md_updated/deleted` inbound events (never forward to LLM); optional agent write-back (ties into #87 Phase 2).
- Keep the existing local-file GROUP.md as a fallback / offline mode.

## Acceptance
- A thread message routes to a distinct session; thread instructions inject correctly.
- Server GROUP.md fetch + event-driven refresh works; local-file fallback preserved.
- Tests for the `____` parser, md fetch/cache, and event handling.

## Notes
Larger effort; can land in phases (thread routing first, then server-md, then write-back).

lml2468:
## 分诊结果

- **分类**:feature
- **优先级**:P3 — Parity gap only needed when a deployment uses 子区/threads;local-file `GROUP.md` fallback already works,所以存在 workaround 且受众有限。
- **module**:server
- **track**:🚀 Feature
- **size**:L

> 说明:本仓库为 default-only 配置,priority/module/track/size 不打标签,以本评论正文为准。

**相关 issue**:#89 #91 #90 #87

lml2468:
## Triage (leader / loop pipeline)

- **type**: feature
[...]

PR #169, opened 2026-06-29
feat(thread): route community-topic threads + restore lifecycle endpoints (P1 of #88)

boris-clark:
## Phase 1 of #88 — thread routing + composite channel_id parsing + thread lifecycle endpoints

Parity with `openclaw-channel-octo`. This is the first of three serial phases
(`P1 → P2 → P3`); it deliberately does **not** touch server-driven GROUP.md /
THREAD.md injection (P2/P3). Scope is routing + endpoints + types only.

### What changed

- **`src/octo/channel-id.ts`** (new) — `THREAD_ID_SEPARATOR = "____"` plus pure
  helpers `extractParentGroupNo` / `extractThreadShortId` / `isThreadChannelId`.
  Splitting always uses the first `____` occurrence, so a short id that itself
  contains the separator stays intact.
- **`src/octo/api.ts`** — restores the seven bot thread lifecycle endpoints that
  were removed at fork time (`create` / `list` / `get` / `delete` / `members` /
  `join` / `leave`). Path / method / auth are a verbatim restore of upstream;
  each call is routed through the existing `postJson` / `getJson` helpers (and a
  small no-body helper) so it shares the same timeout, `Bearer` auth, int64-safe
  JSON parse, and error-message format as the rest of the client. Adds
  `Thread` / `ThreadMember` types.
- **`src/group-context.ts`** — `refreshMembers` now resolves the **parent group
  number** before calling `/v1/bot/groups/{groupNo}/members`. A thread roster
  lookup with the composite id would otherwise 404. All caches (memory + DB +
  robot flags) stay keyed by the full `channel_id`, so a thread keeps its own
  isolated roster view.
- **`src/group-config.ts`** — `loadGroupConfig` now splits a composite thread
  `channel_id` (see path-semantics note below).

Threads get per-channel session / cwd / memory isolation for free: the composite
`channel_id` is already the session key, so a thread never shares its parent
group's partition.

### ⚠️ Path-semantics change (redline 5)

Previously a thread `channel_id` was used verbatim as the group-config filename
(`<groupNo>____<shortId>.md`). After this change:

- A plain group is unchanged: `<groupNo>.md`.
- A thread reads its **own** `<shortId>.md` and does **not** inherit the parent
  group's `<groupNo>.md` (decision A from the approved plan: a thread injects its
  own file only).
- **Backward compatibility:** if `<shortId>.md` is absent, `loadGroupConfig`
  falls back to the legacy whole-composite `<groupNo>____<shortId>.md` filename,
  so existing deployments that dropped a file under the old name keep working.

Operators who relied on the old whole-composite filename get a transparent
fallback; new thread instruction files should be named by short id.

### Testing

- `____` parsing boundary cases (no separator, multiple separators → first wins,
  empty segments).
- All seven thread endpoints assert correct path, method, and
  `Authorization: Bearer <botToken>`.
- Thread composite id routes to a session distinct from its parent group.
- Legacy whole-composite filename still loads (redline 5 compat).
- Thread roster resolves to the parent group number (redline 6).
- Full suite green: **1132 tests / 58 files** (baseline 1096 / 56, no regression).

Part of #88 (Phase 1; P2/P3 tracked separately).

> Draft: opened for line-by-line review. Promotion to a ready PR happens only
> after review + deploy + green tests are confirmed.

### DIFF
--- a/src/octo/api.ts
+++ b/src/octo/api.ts
@@ -1,13 +1,17 @@
 // Forked from openclaw-channel-octo v1.0.13 (2026-06-04)
 // Source: https://github.com/Mininglamp-OSS/openclaw-channel-octo
-// Removed: COS upload, GROUP.md API, OBO, rich text, media, thread/group management,
+// Removed: COS upload, GROUP.md API, OBO, rich text, media, group management,
 //          read receipts, bot groups list, group info, mention prefs, space members.
+// Restored: thread lifecycle (create/list/get/delete/members/join/leave) — see
+//          "Thread Lifecycle" section below; group/server-md APIs stay removed.
 
 import {
   ChannelType,
   MessageType,
   type MentionEntity,
   type SendMessageResult,
+  type Thread,
+  type ThreadMember,
 } from "./types.js";
 import { randomUUID } from "node:crypto";
 
@@ -494,4 +498,171 @@ export async function getChannelMessages(params: {
     console.error(`octo: getChannelMessages error: ${String(err)}`);
     return [];
   }
+}
+
+// ─── Thread Lifecycle ────────────────────────────────────────────────────────
+//
+// Restores the bot thread (CommunityTopic) lifecycle endpoints that were removed
+// at fork time (see file header). Path / method / auth are a verbatim restore of
+// openclaw-channel-octo api-fetch.ts (createThread … leaveThread); each call is
+// routed through this client's postJson/getJson/requestNoBody helpers so it
+// shares the same timeout, Bearer auth, int64-safe JSON parse, and error-message
+// format as the rest of the API surface.
+
+/**
+ * Issue a request that carries neither a request body nor a meaningful response
+ * body (thread delete / join / leave). Mirrors postJson's timeout, Bearer auth,
+ * and error-message format. A 2xx with an empty body resolves to void.
+ */
+async function requestNoBody(
+  apiUrl: string,
+  botToken: string,
+  method: "POST" | "DELETE",
+  path: string,
+  signal?: AbortSignal,
+): Promise<void> {
+  const url = `${apiUrl.replace(/\/+$/, "")}${path}`;
+  const effectiveSignal = signal ?? AbortSignal.timeout(DEFAULT_TIMEOUT_MS);
+  const resp = await fetch(url, {
+    method,
+    headers: { Authorization: `Bearer ${botToken}` },
+    signal: effectiveSignal,
+  });
+  if (!resp.ok) {
+    const text = await resp.text().catch(() => "");
+    throw new Error(`Octo API ${path} failed (${resp.status}): ${text || resp.statusText}`);
+  }
+}
+
+/** Create a thread under a parent group. POST /v1/bot/groups/{groupNo}/threads */
+export async function createThread(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;
+  name: string;
+  /** Optional: anchor the thread to the message it was started from. */
+  sourceMessageId?: number;
+  signal?: AbortSignal;
+}): Promise<Thread | undefined> {
+  const body: Record<string, unknown> = { name: params.name };
+  if (params.sourceMessageId != null) body.source_message_id = params.sourceMessageId;
+  return await postJson<Thread>(
+    params.apiUrl,
+    params.botToken,
+    `/v1/bot/groups/${encodeURIComponent(params.groupNo)}/threads`,
+    body,
+    params.signal,
+  );
+}
+
+/** List threads under a parent group. GET /v1/bot/groups/{groupNo}/threads */
+export async function listThreads(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;
+  signal?: AbortSignal;
+}): Promise<Thread[]> {
+  const data = await getJson<Record<string, unknown>>(
+    params.apiUrl,
+    params.botToken,
+    `/v1/bot/groups/${encodeURIComponent(params.groupNo)}/threads`,
+    params.signal,
+  );
+  // Tolerate both a bare array and a `{ threads: [...] }` envelope, mirroring
+  // getGroupMembers' defensive shape handling.
+  const threads = Array.isArray(data?.threads)
+    ? data.threads
+    : Array.isArray(data)
+      ? data
+      : [];
+  return threads as Thread[];
+}
+
+/** Get a single thread. GET /v1/bot/groups/{groupNo}/threads/{shortId} */
+export async function getThread(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;
+  shortId: string;
+  signal?: AbortSignal;
+}): Promise<Thread> {
+  return await getJson<Thread>(
+    params.apiUrl,
+    params.botToken,
+    `/v1/bot/groups/${encodeURIComponent(params.groupNo)}/threads/${encodeURIComponent(params.shortId)}`,
+    params.signal,
+  );
+}
+
+/** Delete a thread. DELETE /v1/bot/groups/{groupNo}/threads/{shortId} */
+export async function deleteThread(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;
+  shortId: string;
+  signal?: AbortSignal;
+}): Promise<void> {
+  await requestNoBody(
+    params.apiUrl,
+    params.botToken,
+    "DELETE",
+    `/v1/bot/groups/${encodeURIComponent(params.groupNo)}/threads/${encodeURIComponent(params.shortId)}`,
+    params.signal,
+  );
+}
+
+/** List a thread's members. GET /v1/bot/groups/{groupNo}/threads/{shortId}/members */
+export async function listThreadMembers(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;
+  shortId: string;
+  signal?: AbortSignal;
+}): Promise<ThreadMember[]> {
+  const data = await getJson<Record<string, unknown>>(
+    params.apiUrl,
+    params.botToken,
+    `/v1/bot/groups/${encodeURIComponent(params.groupNo)}/threads/${encodeURIComponent(params.shortId)}/members`,
+    params.signal,
+  );
+  const members = Array.isArray(data?.members)
+    ? data.members
+    : Array.isArray(data)
+      ? data
+      : [];
+  return members as ThreadMember[];
+}
+
+/** Join a thread. POST /v1/bot/groups/{groupNo}/threads/{shortId}/join */
+export async function joinThread(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;
+  shortId: string;
+  signal?: AbortSignal;
+}): Promise<void> {
+  await requestNoBody(
+    params.apiUrl,
+    params.botToken,
+    "POST",
+    `/v1/bot/groups/${encodeURIComponent(params.groupNo)}/threads/${encodeURIComponent(params.shortId)}/join`,
+    params.signal,
+  );
+}
+
+/** Leave a thread. POST /v1/bot/groups/{groupNo}/threads/{shortId}/leave */
+export async function leaveThread(params: {
+  apiUrl: string;
+  botToken: string;
+  groupNo: string;
+  shortId: string;
+  signal?: AbortSignal;
+}): Promise<void> {
+  await requestNoBody(
+    params.apiUrl,
+    params.botToken,
+    "POST",
+    `/v1/bot/groups/${encodeURIComponent(params.groupNo)}/threads/${encodeURIComponent(params.shortId)}/leave`,
+    params.signal,
+  );
 }
--- a/src/__tests__/channel-id.test.ts
+++ b/src/__tests__/channel-id.test.ts
@@ -1,0 +1,79 @@
+/**
+ * Tests for octo/channel-id.ts — composite (thread) channel-id parsing.
+ *
+ * Thread channelId format: `<groupNo>____<shortId>` (THREAD_ID_SEPARATOR, 4
+ * underscores). Group channelId has no separator.
+ */
+
+import { describe, it, expect } from 'vitest';
+import {
+  THREAD_ID_SEPARATOR,
+  extractParentGroupNo,
+  extractThreadShortId,
+  isThreadChannelId,
+} from '../octo/channel-id.js';
+
+describe('THREAD_ID_SEPARATOR', () => {
+  it('is exactly four underscores', () => {
+    expect(THREAD_ID_SEPARATOR).toBe('____');
+    expect(THREAD_ID_SEPARATOR.length).toBe(4);
+  });
+});
+
+describe('isThreadChannelId', () => {
+  it('is true for a composite id', () => {
+    expect(isThreadChannelId('12345____67890')).toBe(true);
+  });
+
+  it('is false for a plain group id (no separator)', () => {
+    expect(isThreadChannelId('12345')).toBe(false);
+    expect(isThreadChannelId('s1_group99')).toBe(false);
+  });
+
+  it('does not treat fewer than four underscores as a separator', () => {
+    expect(isThreadChannelId('a_b')).toBe(false);
+    expect(isThreadChannelId('a__b')).toBe(false);
+    expect(isThreadChannelId('a___b')).toBe(false);
+  });
+});
+
+describe('extractParentGroupNo', () => {
+  it('returns the left segment of a composite id', () => {
+    expect(extractParentGroupNo('12345____67890')).toBe('12345');
+  });
+
+  it('returns the id unchanged when there is no separator', () => {
+    expect(extractParentGroupNo('12345')).toBe('12345');
+  });
+
+  it('splits on the FIRST separator when several are present', () => {
+    // Right-hand side keeps its own separator; only the first split matters.
+    expect(extractParentGroupNo('12345____67890____extra')).toBe('12345');
+  });
+
+  it('returns an empty parent when the id starts with the separator', () => {
+    expect(extractParentGroupNo('____67890')).toBe('');
+  });
+});
+
+describe('extractThreadShortId', () => {
+  it('returns the right segment of a composite id', () => {
+    expect(extractThreadShortId('12345____67890')).toBe('67890');
+  });
+
+  it('returns null for a plain group id (no separator)', () => {
+    expect(extractThreadShortId('12345')).toBeNull();
+  });
+
+  it('keeps embedded separators in the short id (first split only)', () => {
+    expect(extractThreadShortId('12345____67890____extra')).toBe('67890____extra');
+  });
+
+  it('returns null when the short-id segment is empty (trailing separator)', () => {
+    expect(extractThreadShortId('12345____')).toBeNull();
+  });
+
+  it('returns the short id even when the parent segment is empty', () => {
+    expect(extractThreadShortId('____67890')).toBe('67890');
+  });
+});
--- a/src/__tests__/group-config.test.ts
+++ b/src/__tests__/group-config.test.ts
@@ -92,3 +92,43 @@ describe('loadGroupConfig', () => {
     expect(loadGroupConfig(dir, 'g1')).toBe('instructions');
   });
 });
+
+describe('loadGroupConfig — thread (composite channel_id) routing [#88 redline 5]', () => {
+  const GROUP = 'g123';
+  const SHORT = 't789';
+  const COMPOSITE = `${GROUP}____${SHORT}`;
+
+  it('a plain group id is unaffected (byte-identical to prior behavior)', () => {
+    writeFileSync(join(dir, `${GROUP}.md`), 'group rules');
+    expect(loadGroupConfig(dir, GROUP)).toBe('group rules');
+  });
+
+  it('NEW semantics: a thread reads its own <shortId>.md', () => {
+    writeFileSync(join(dir, `${SHORT}.md`), 'thread rules');
+    expect(loadGroupConfig(dir, COMPOSITE)).toBe('thread rules');
+  });
+
+  it('a thread does NOT inherit the parent group <groupNo>.md (decision A)', () => {
+    writeFileSync(join(dir, `${GROUP}.md`), 'group rules');
+    // No <shortId>.md and no legacy file → undefined, not the parent's file.
+    expect(loadGroupConfig(dir, COMPOSITE)).toBeUndefined();
+  });
+
+  it('COMPAT fallback: legacy whole-composite <groupNo____shortId>.md still loads', () => {
+    writeFileSync(join(dir, `${COMPOSITE}.md`), 'legacy thread file');
+    expect(loadGroupConfig(dir, COMPOSITE)).toBe('legacy thread file');
+  });
+
+  it('NEW semantics wins over the legacy whole-composite file', () => {
+    writeFileSync(join(dir, `${SHORT}.md`), 'new');
+    writeFileSync(join(dir, `${COMPOSITE}.md`), 'legacy');
+    expect(loadGroupConfig(dir, COMPOSITE)).toBe('new');
+  });
+
+  it('thread short id and parent group resolve to different files', () => {
+    writeFileSync(join(dir, `${GROUP}.md`), 'group rules');
+    writeFileSync(join(dir, `${SHORT}.md`), 'thread rules');
+    expect(loadGroupConfig(dir, GROUP)).toBe('group rules');
+    expect(loadGroupConfig(dir, COMPOSITE)).toBe('thread rules');
+  });
+});
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -507,3 +507,34 @@ describe('G23: robot flag', () => {
     expect(ctx.isMember('ch-empty', 'u2')).toBe(true);
   });
 });
+
+describe('GroupContext.refreshMembers — thread roster uses parent group [#88 redline 6]', () => {
+  let adapter: DbAdapter;
+  let ctx: GroupContext;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    adapter = createTestAdapter();
+    ctx = new GroupContext(adapter, 6000);
+  });
+
+  it('queries the PARENT group number, not the composite thread channel_id', async () => {
+    const GROUP = '99dc18164a29435f9791dc37023f98e1';
+    const COMPOSITE = `${GROUP}____2071488441815666688`;
+    (getGroupMembers as ReturnType<typeof vi.fn>).mockResolvedValueOnce([]);
+
+    await ctx.refreshMembers(COMPOSITE, 'https://test.example.com', 'bf_test');
+
+    expect(getGroupMembers).toHaveBeenCalledTimes(1);
+    const arg = (getGroupMembers as ReturnType<typeof vi.fn>).mock.calls[0][0];
+    // Hitting /groups/<composite>/members would 404 — must use the parent group.
+    expect(arg.groupNo).toBe(GROUP);
+    expect(arg.groupNo).not.toContain('____');
+  });
+
+  it('passes a plain group channel_id through unchanged', async () => {
+    (getGroupMembers as ReturnType<typeof vi.fn>).mockResolvedValueOnce([]);
+    await ctx.refreshMembers('plain-group', 'https://test.example.com', 'bf_test');
+    expect((getGroupMembers as ReturnType<typeof vi.fn>).mock.calls[0][0].groupNo).toBe('plain-group');
+  });
+});
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -943,3 +943,32 @@ describe('unsupported channel types (system messages)', () => {
     expect(result?.shouldProcess).toBe(true);
   });
 });
+
+describe('SessionRouter — thread (CommunityTopic) session isolation [#88]', () => {
+  let router: SessionRouter;
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    router = new SessionRouter(makeConfig(), ROBOT_ID);
+  });
+
+  it('a thread composite channel_id keys a session distinct from its parent group', () => {
+    const GROUP = '99dc18164a29435f9791dc37023f98e1';
+    const COMPOSITE = `${GROUP}____2071488441815666688`;
+
+    const parent = makeMsg({ channel_type: ChannelType.Group, channel_id: GROUP, from_uid: 'u1' });
+    const thread = makeMsg({ channel_type: ChannelType.CommunityTopic, channel_id: COMPOSITE, from_uid: 'u1' });
+
+    // The composite id IS the session key, so the thread never shares the
+    // parent group's session/history/cwd/memory partition.
+    expect(router.sessionKey(thread)).toBe(COMPOSITE);
+    expect(router.sessionKey(thread)).not.toBe(router.sessionKey(parent));
+  });
+
+  it('two threads under the same parent get independent sessions', () => {
+    const GROUP = 'g1';
+    const a = makeMsg({ channel_type: ChannelType.CommunityTopic, channel_id: `${GROUP}____aaa`, from_uid: 'u1' });
+    const b = makeMsg({ channel_type: ChannelType.CommunityTopic, channel_id: `${GROUP}____bbb`, from_uid: 'u1' });
+    expect(router.sessionKey(a)).not.toBe(router.sessionKey(b));
+  });
+});
--- a/src/__tests__/thread-api.test.ts
+++ b/src/__tests__/thread-api.test.ts
@@ -1,0 +1,206 @@
+/**
+ * Tests for octo/api.ts — thread lifecycle endpoints.
+ *
+ * Each test asserts the restored endpoint hits the correct path + method and
+ * carries `Authorization: Bearer <botToken>` (parity restore of openclaw
+ * api-fetch.ts; see #88 acceptance criteria).
+ */
+
+import { describe, it, expect, vi, beforeEach, afterAll } from 'vitest';
+import {
+  createThread,
+  listThreads,
+  getThread,
+  deleteThread,
+  listThreadMembers,
+  joinThread,
+  leaveThread,
+} from '../octo/api.js';
+
+const fetchMock = vi.fn();
+const originalFetch = globalThis.fetch;
+
+beforeEach(() => {
+  fetchMock.mockReset();
+  globalThis.fetch = fetchMock as unknown as typeof fetch;
+});
+
+afterAll(() => {
+  globalThis.fetch = originalFetch;
+});
+
+function mockJsonResponse(body: unknown): Response {
+  return new Response(JSON.stringify(body), {
+    status: 200,
+    headers: { 'Content-Type': 'application/json' },
+  });
+}
+
+function mockEmptyResponse(status = 200): Response {
+  // null body — a 204 (and other null-body statuses) rejects a non-null body.
+  return new Response(null, { status, statusText: 'OK' });
+}
+
+const BASE = { apiUrl: 'https://test.example.com', botToken: 'bf_test' };
+const GROUP = '99dc18164a29435f9791dc37023f98e1';
+const SHORT = '2071488441815666688';
+
+/** Pull (url, init) of the Nth fetch call. */
+function call(n = 0): { url: string; init: RequestInit } {
+  const [url, init] = fetchMock.mock.calls[n] as [string, RequestInit];
+  return { url, init };
+}
+
+function authHeader(init: RequestInit): string | undefined {
+  const h = init.headers as Record<string, string> | undefined;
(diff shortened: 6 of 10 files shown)


src/octo/api.ts
```diff
@@ -495,3 +499,170 @@ export async function getChannelMessages(params: {
...
+    signal: effectiveSignal,
+  });
+  if (!resp.ok) {
+    const text = await resp.text().catch(() => "");
+    throw new Error(`Octo API ${path} failed (${resp.status}): ${text || resp.statusText}`);
+  }
+}
+
+/** Create a thread under a parent group. POST /v1/bot/groups/{groupNo}/threads */
+export async function createThread(params: {
```
OctoBoooot, line 538: **[minor]** `sourceMessageId` is typed `number` and sent as `source_message_id`. This client's header calls out int64-safe JSON parsing for ids; a `number` source id would lose precision above 2^53 before serialization. Other id fields here are strings. If thread/message ids can exceed the JS safe-integer range, consider `string`. Non-blocking for P1 (field is optional and not on the routing path), but worth settling before P2 consumes it.

OctoBoooot (commented):
## Review: feat(thread): route community-topic threads + restore lifecycle endpoints (P1 of #88) (#169)

**Verdict:** Approve with comments — well-scoped P1, correct core logic, thorough tests. No blockers or majors; one minor (int64 id typing) worth settling before P2.

Note: PR is a **draft** (author opened for line-by-line review; promotion gated on review + deploy + green tests per the body). CI is green (1132 tests / 58 files). Repo already applied `needs-human-review` + `size/XL` — appropriate for a 697-line feature touching routing.

This is a large, well-prepared PR — the kind where I expected to find something and didn't after checking the spots that usually hide bugs. Walking the risk surface:

- **Composite-id parsing (`channel-id.ts`)** — `indexOf`-based first-separator-wins split is correct and matches the documented contract; a shortId containing `____` stays intact on the RHS. The three helpers are pure and the boundary cases are pinned by tests (no separator, multiple separators, empty parent, trailing separator → null, fewer-than-four underscores not treated as separator).
- **Path injection** — every restored endpoint wraps `groupNo`/`shortId` in `encodeURIComponent`, and there's a dedicated test asserting URL-encoding. The composite id never reaches a path unescaped.
- **`loadGroupConfig` rename `groupId`→`channelId`** — checked the sole caller (`index.ts` already passes `channelId`); parameter rename only, same type, no caller breakage.
- **`isSafeId` vs the legacy fallback** — the regex allows `_`, so the legacy whole-composite `<groupNo>____<shortId>.md` fallback filename passes the safety gate while still rejecting `/`, `.`, `..` (path escape). Redline-5 compat works without opening a traversal hole.
- **Roster resolution (redline 6)** — `refreshMembers` resolves `extractParentGroupNo(channelId)` for the `/groups/{groupNo}/members` call while keeping all caches keyed by the full composite id. For a plain group, `extractParentGroupNo` is identity, so existing behavior is byte-identical.

### Praise
- **Decision-A isolation is tested both ways**: a thread reads its own `<shortId>.md` AND explicitly does NOT inherit the parent `<groupNo>.md` (`toBeUndefined()`), plus new-semantics-wins-over-legacy precedence. That's exactly the redline-5 behavior matrix, pinned so a future refactor can't silently regress it.
- **Envelope tolerance** on `listThreads`/`listThreadMembers` (`{threads:[...]}` vs bare array vs unexpected → `[]`) mirrors the existing `getGroupMembers` defensive shape handling — consistent with the established client style rather than inventing a new one.
- The fork-provenance comments (`Forked from openclaw-channel-octo v1.0.13`, "verbatim restore of upstream path/method/auth") make the restored endpoints auditable against their source — the right call for a fork resurrecting removed code.

### Questions
- `createThread` types `sourceMessageId` as `number` and sends it as `source_message_id`. The rest of this client is careful about int64-safe JSON parsing (noted in the header) — if thread/message ids can exceed 2^53, a `number` source id would lose precision before it's even serialized. Plain groups use string ids elsewhere; is `source_message_id` guaranteed to fit in a JS safe integer, or should it be `string` like the other id fields? Non-blocking for P1 since the field is optional and unused by the routing path, but worth confirming before P2 wires it up.

- Coverage-sniff flagged `api.ts` / `channel-id.ts` / `types.ts` as untested by naming convention — false positives; `channel-id.test.ts` and `thread-api.test.ts` cover them directly with 200+ lines of assertions.
- P2/P3 (server-driven GROUP.md/THREAD.md injection) are correctly out of scope here; the body is explicit about the serial-phase boundary.

Jerry-Xin (approved):
**APPROVE** — P1 scope (thread routing + composite channel_id parsing + restored lifecycle endpoints + types) is clean and well-contained. Routing logic traced both directions, path-safety preserved, no reachable PR-owned bug. One non-blocking int64 type concern to address before a real caller is wired in P2/P3.

### 🔴 Blocking
None.

### 💬 Non-blocking

**🟡 `createThread.sourceMessageId?: number` breaks the codebase's int64-string invariant for message IDs** — `src/octo/api.ts:544`, written into the request body at `:548` (`body.source_message_id = params.sourceMessageId`). Everywhere else Octo `message_id` is deliberately typed `string` (`types.ts:16,86,157`; `api.ts:383,474`) and there is explicit inbound int64 protection (`api.ts:53` converts 16+ digit `message_id` to string pre-`JSON.parse` to avoid `Number.MAX_SAFE_INTEGER` loss). The inbound regex does **not** cover this outbound path, so a future caller anchoring a thread to a real 16+ digit message ID would silently lose precision (or hit a TS mismatch if they hold the ID as a string, which they will). Not merge-blocking for P1 because `createThread` currently has **zero callers** (restored API surface for P2/P3) — no reachable bug today. Recommend `sourceMessageId?: string` (or `string | number`) + a 16+ digit test before P2 wires a caller.

**🟡 Seven restored lifecycle endpoints have no callers yet** — `createThread/listThreads/getThread/deleteThread/listThreadMembers/joinThread/leaveThread` are exported but unused outside `api.ts`/tests. This is intentional per the PR description (parity restore; consumption lands in P2/P3), so it's a note, not a defect — flagging only so the dangling surface is tracked and exercised when P2 lands.

### ✅ Highlights
- **`channel-id.ts` is the right abstraction.** Single source of truth for the `____` composite format; first-occurrence `indexOf` split keeps a short id containing the separator intact on the RHS. Boundary tests cover no-separator, multiple-separators-first-wins, and empty segments. Verified adversarial inputs: `""`, `"____"`, `"g____"`, `"g____a____b"` all behave as documented.
- **`refreshMembers` redline-6 fix is correct.** `extractParentGroupNo(channelId)` resolves the parent group_no for the `/groups/{groupNo}/members` call (which would 404 on a composite id) while every cache (memory + DB + robot flags) stays keyed by the full `channelId` → thread keeps an isolated roster. `extractParentGroupNo` is identity for plain groups, so zero regression on the existing path (sole caller: `index.ts:433`).
- **`loadGroupConfig` path-safety preserved.** The early `isSafeId(channelId)` guard moved into `loadConfigFile(slug)`, which runs `isSafeId` on the final filename slug in *every* branch (short-id, legacy composite fallback, plain group). `____` passes `isSafeId` (underscore allowed) but `/`/`..` are still rejected → no path-traversal regression. Sole real caller is `index.ts:800` (groups only). Legacy whole-composite fallback (redline 5 compat) is correct.
- **`requestNoBody` error format is byte-identical** to `postJson`/`getJson` (`Octo API ${path} failed (${status}): ${text || statusText}`), so the restored endpoints share the same timeout/Bearer/error contract as the rest of the client — no new unsanitized sink relative to the established pattern.
- Type-check passed; focused suites (channel-id, thread-api, group-config, group-context, session-router) green.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#169 Review Report
**PR:** https://github.com/Mininglamp-OSS/cc-channel-octo/pull/169
**Head SHA:** `202724ffd5f8611fec5a96fc83dda7c49753d416`
**Title:** feat(thread): route community-topic threads + restore lifecycle endpoints (P1 of #88)
**Files changed:** 10 (+697 / −13)
**Complexity:** security_sensitive

---

## 1. 验证结论

| Item | Status | Evidence |
|------|--------|----------|
| Composite channel_id parsing (`channel-id.ts`) | ✅ | `indexOf` on first `____`; pure functions with correct edge-case handling (empty parent, empty shortId, embedded separators). Tests at `src/__tests__/channel-id.test.ts` cover all boundary cases. |
| Thread lifecycle API endpoints (`api.ts`) | ✅ | All 7 endpoints (create/list/get/delete/members/join/leave) use correct HTTP method + path. `encodeURIComponent` on all path params. Bearer auth consistent with existing helpers. Tests at `src/__tests__/thread-api.test.ts` assert path, method, and auth header. |
| `group-context.ts` parent groupNo resolution | ✅ | `extractParentGroupNo(channelId)` correctly extracts parent groupNo before calling `/groups/{groupNo}/members`. Caches remain keyed by full composite channelId (per-channel isolation preserved). Verified call chain: `index.ts:433` → `refreshMembers(compositeChannelId)` → `extractParentGroupNo` → `getGroupMembers({ groupNo: parentOnly })`. |
| `group-config.ts` thread routing (primary path) | ✅ | `isThreadChannelId` → `extractThreadShortId` → `loadConfigFile(dir, shortId)`. Each segment passes `isSafeId` independently. Short-id file takes priority over legacy composite. |
| `group-config.ts` thread routing (legacy compat fallback) | ❌ | **See Finding #1 below.** `isSafeId` rejects underscores; the whole-composite legacy filename can never pass validation. |
| Thread session isolation | ✅ | Composite channelId IS the session key (SessionRouter uses `channel_id` directly). Thread naturally gets its own session/cwd/memory partition. Verified at `index.ts:377`. |
| `isGroup` includes CommunityTopic | ✅ | `index.ts:365`: `isGroup = channelType === ChannelType.Group \|\| channelType === ChannelType.CommunityTopic`. Thread messages correctly flow through group handling paths. |
| Type definitions (`types.ts`) | ✅ | `Thread` and `ThreadMember` interfaces match the API response shapes used in tests. |

---

## 2. 发现的问题

### Finding #1 — P2: Legacy composite filename backward compat is broken (dead code + test would fail)

**Severity: P2** | **Diff-scope: NEW** (introduced by this PR)

**Root cause:** `loadGroupConfig` adds a backward-compat fallback that calls `loadConfigFile(groupConfigDir, channelId)` with the whole-composite channelId (e.g. `g123____t789`). However, `loadConfigFile` gates on `isSafeId(slug)`, and the regex `/^[a-zA-Z0-9._-]+$/` does NOT include underscore (`_`). Since the `____` separator contains four underscores, ANY composite channelId fails `isSafeId` and the fallback returns `undefined` unconditionally.

**Data-flow trace:**
1. `loadGroupConfig(dir, 'g123____t789')` → `isThreadChannelId` = true
2. `extractThreadShortId` → `'t789'` → `loadConfigFile(dir, 't789')` → `isSafeId('t789')` = true ✅ (primary path works)
3. If `t789.md` absent → fallback: `loadConfigFile(dir, 'g123____t789')` → `isSafeId('g123____t789')` = **false** ❌ → returns `undefined`

**Impact:**
- The PR description states: *"if `<shortId>.md` is absent, `loadGroupConfig` falls back to the legacy whole-composite `<groupNo>____<shortId>.md` filename, so existing deployments that dropped a file under the old name keep working."* — **This claim is false.** The fallback is dead code.
- The test "COMPAT fallback: legacy whole-composite `<groupNo____shortId>.md` still loads" (`group-config.test.ts`) would **fail** if executed: `beforeEach` creates a fresh `mkdtempSync` dir per test, so no stale `shortId.md` exists. `loadGroupConfig` returns `undefined` but the test expects `'legacy thread file'`.
- The PR claims "Full suite green: 1132 tests / 58 files" — this appears inconsistent with the analysis above.

**Fix:** Add underscore to the `isSafeId` regex: `/^[a-zA-Z0-9.__-]+$/`. Underscore is safe as a filename character (no filesystem special meaning). One-line change.

### Finding #2 — P2 (nit): `requestNoBody` lacks Bearer token sanitization in error messages

**Severity: P2 (pre-existing pattern)** | **Diff-scope: NEW code, pre-existing gap**

`requestNoBody` (`api.ts:519-534`) echoes raw `resp.text()` in error messages. The existing `getUploadCredentials` (line ~175) already strips Bearer tokens from error text defensively. The new `requestNoBody` and the pre-existing `postJson` both lack this sanitization.

Not blocking — the risk is low since thread endpoints don't typically echo auth headers — but for defense-in-depth parity with `getUploadCredentials`, consider applying the same sanitization pattern.

### Finding #3 — P2 (nit): `getJson<Record<string, unknown>>` type assertion for array responses

**Severity: P2** | **Diff-scope: NEW**

`listThreads` and `listThreadMembers` call `getJson<Record<string, unknown>>` but the server may return a bare JSON array. TypeScript allows the assignment (array ⊂ object), but `Record<string, unknown>` is technically incorrect for an array. The runtime `Array.isArray(data)` check correctly handles both shapes. Not a runtime bug, but a type-safety smell.

---

## 3. 建议

1. **Fix `isSafeId` before merge** (Finding #1): Add underscore to the regex. Verify the COMPAT fallback test passes after the fix. This is the only should-fix blocker.
2. **Re-run the full test suite** and confirm green — the current analysis suggests the COMPAT fallback test would fail.
3. (Optional) Add Bearer-token sanitization to `requestNoBody` for parity with `getUploadCredentials`.
4. (Optional) Use a union type `T | T[]` instead of `Record<string, unknown>` for list endpoints.

---

## 4. 额外发现

- **No changes to `session-router.ts` source** — only tests added. The session isolation works because the composite channelId is already the session key. Confirmed correct; no source change needed.
- **Thread shortId with embedded `____`**: The design correctly handles this via first-occurrence split (`indexOf`). However, there is no documented server-side constraint preventing shortIds from containing `____`. If the server ever generates such shortIds, round-trip `extractParentGroupNo(groupNo + '____' + shortId)` would still be correct, but `extractThreadShortId` would return the full right-hand side including embedded separators. This is by design per tests, but worth documenting as a protocol assumption.

---

## 5. 数据流回溯

|---------------|----------------|-----------------|
| `channelId` at `loadGroupConfig` call site | `index.ts:800` ← `msg.channel_id` from inbound webhook | ✅ Composite for threads, plain for groups |
| `groupNo` for `getGroupMembers` | `extractParentGroupNo(channelId)` at `group-context.ts:210` | ✅ Strips `____shortId` suffix; identity for plain groups |
| `shortId` for `loadConfigFile` | `extractThreadShortId(channelId)` → `channelId.slice(indexOf('____') + 4)` | ✅ Correct first-split semantics |
| `isSafeId` gate on slug | `loadConfigFile` line ~89 | ✅ for shortId segments; ❌ for whole-composite (underscores rejected) |
| `members` cache key | Full `channelId` (composite), NOT the extracted `groupNo` | ✅ Per-channel isolation preserved |
| Thread API URL construction | `encodeURIComponent(groupNo)` + `encodeURIComponent(shortId)` | ✅ Path traversal prevented |
| Bearer token in thread endpoints | `botToken` from caller params, same as existing endpoints | ✅ Consistent auth pattern |

---

## 6. 盲点 checklist（R5 — security_sensitive: C1–C6 all mandatory）

**C1 — 双路径 parity（对称路径自检）:** CLEAR
- Thread lifecycle endpoints: create ↔ delete are independent REST calls with no client-side state to keep symmetric. Join ↔ leave are both POST with no side effects. No create-gate / execute-path asymmetry.
- `loadGroupConfig` routing: thread path vs plain-group path both funnel through `loadConfigFile`, which applies `isSafeId` uniformly. The difference is only the slug selection (shortId vs channelId), which is correct.

**C2 — control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿:** ⚠️ HIT (Finding #1)
- `isSafeId` is a security control (path-traversal guard). The PR introduces a new code path where composite IDs are split, and the WHOLE composite is also tried as a fallback. The `isSafeId` regex was not updated to account for the `____` separator, so the fallback path is dead code.
- Non-canonical试穿: a composite ID like `g123____t789` contains underscores which are rejected. The split segments (`g123`, `t789`) pass individually. The control (isSafeId) is correct for the primary path but incorrectly rejects the legacy fallback.
- No double-application or ordering issues in the two-step lookup (shortId first, then composite).

**C3 — 授权边界 ≠ 能力边界:** CLEAR
- Thread endpoints use the same bot token as all other endpoints. Authorization is enforced server-side (bot must be group member). Client side only constructs URLs correctly (verified: `encodeURIComponent` on all path params).
- No new endpoints exposed to unauthenticated callers.

**C4 — 授权生命周期 / 容器-成员状态级联:** CLEAR
- No auth changes in this PR. Thread lifecycle endpoints use the existing Bearer token pattern. No container/member status checks are performed client-side (server responsibility).

**C5 — build/note 通过 ≠ 运行期路径正确:** CLEAR
- No build/deploy changes. No relative paths or extension packaging. Standard TypeScript module that runs in Node.js.

**C6 — 治理/策略/安全文档自洽性:** N/A
- Not a governance/policy PR.

---

## 7. 跨轮 blocker 复检

---

## [Octo-Q] verdict: APPROVE

**Rationale:** Only P2 findings per R4 rubric. The primary thread routing, session isolation, API surface, and data flows are all correct and well-tested.

**Should-fix before merge (not blocking):**
- **Finding #1 (P2):** Fix `isSafeId` regex to accept underscores (`/^[a-zA-Z0-9.__-]+$/`). Re-run test suite to confirm the COMPAT fallback test passes. Without this fix, the backward-compat claim in the PR description is false, and the COMPAT fallback test should fail.

**No P0/P1 findings.** The PR is structurally sound and the core functionality works correctly.

yujiawei (approved):
# Code Review — PR #169 (cc-channel-octo)

Reviewed at head `202724ffd5f8611fec5a96fc83dda7c49753d416` against `main` (merge-base `203dc5f`). Build, type-check, lint, and the full suite (1132 tests / 58 files) are green locally. This is a clean, well-scoped, well-tested P1: thread routing + composite-id parsing + restored lifecycle API surface, with the server-driven md work correctly deferred to P2/P3.

## 1. Scope compliance

**✅ Complete and in-scope.**

- **Missing:** none. The three P1 acceptance items are all present — `____` composite parsing (`src/octo/channel-id.ts`), the seven thread lifecycle REST calls (`src/octo/api.ts`), and thread-aware instruction injection (`src/group-config.ts`). Session isolation, parent-group roster resolution, parser edge cases, and all seven endpoints are covered by tests.
- **Over-build:** none. Server-driven GROUP.md/THREAD.md, hot-reload events, and write-back (P2/P3) are deliberately untouched, as stated.
- **Deviation:** none. The "thread injects its own `<shortId>.md` only, not the parent's `<groupNo>.md`" decision, plus the legacy whole-composite fallback, match the stated design.

## 2. Code quality

**Verdict: Approved.** No blocking (correctness / security / data-loss / build-break) defects in the wired diff. The findings below are all advisory (P2) and several are latent in the **restored-but-not-yet-called** endpoint surface — none affect a live code path in this phase, but they should be addressed before P2/P3 wires those endpoints in.

The live, wired changes are solid:
- `loadGroupConfig` routing is keyed correctly (short-id first, legacy composite fallback), and the `isSafeId` slug guard (`/^[a-zA-Z0-9._-]+$/`, rejecting `.`/`..`) plus the group/world-writable refusal hold against a malicious server-controlled `channel_id` — no traversal or unsanitized-injection path was found.
- `refreshMembers` correctly resolves the parent group number for the roster API while keeping every cache keyed by the full `channel_id`, preserving per-thread isolation.
- Session routing returns the composite id as the session key, so threads get genuine per-channel session/cwd/memory isolation.

### P2 — advisory

1. **`short_id` int64 precision is unprotected** (`src/octo/api.ts` `parseOctoJson`, ~L51). The int64 guard only stringifies `message_id`. `Thread.short_id` is typed `string`, but if the Octo server emits it as a bare JSON number (the test fixtures use a 19-digit value, `2071488441815666688`, which exceeds `MAX_SAFE_INTEGER`), `JSON.parse` rounds it (verified: → `2071488441815666700`) before the cast. This is **dormant in P1** because nothing consumes these endpoint responses yet, but once P2/P3 builds a composite `channel_id` from a server-returned `short_id`, a rounded value would mis-route the thread. Recommend extending the parse guard (or normalizing `short_id`/`source_message_id` to strings) before those endpoints are wired. **For a security-sensitive merge, a human should confirm the actual Octo wire format of `short_id`** — number vs. string — since that determines whether this is latent or immediate.

2. **`createThread` types `sourceMessageId` as `number`** (`src/octo/api.ts`, ~L544). Message IDs are int64 and handled as strings elsewhere; a JS `number` anchor would silently lose precision for large ids. Same dormant-until-wired status as (1). Consider `string` (or `string | number`).

3. **Bearer token not redacted in thrown error text** (`requestNoBody`, ~L533; same pattern in `postJson`/`getJson`). The error interpolates the raw response body. `getUploadCredentials` already demonstrates the redaction pattern (`.replace(/Bearer\s+\S+/gi, "Bearer ***")`). The realistic exposure is low (the server is the token's intended recipient, and it would have to echo it back into a body that then reaches a log sink), and the `postJson`/`getJson` pattern is pre-existing — but for consistency and defense-in-depth on a security-sensitive surface, redacting in the new `requestNoBody` (and ideally the shared helpers) is worth doing.

4. **Empty-parent edge** (`extractParentGroupNo`). A malformed `"____x"` yields `""`, producing `/v1/bot/groups//members`. It degrades gracefully (404 → caught → logged), so this is hardening only, not a bug. A guard that treats an empty parent as unroutable would be cleaner.

5. **Per-thread roster fan-out** (`refreshMembers`). Caching the parent roster under each thread's composite id means N active threads in one group issue N identical roster fetches and hold N copies. This is an explicitly documented isolation-vs-efficiency tradeoff and is bounded by the 1h refresh throttle, so it is acceptable for now — but for very large groups with many active threads, a shared parent-roster cache key would scale better. Worth a note for a later phase.

6. **`requestNoBody` does not drain the response body on success.** Under undici, an unconsumed body can delay socket reuse. Minor resource-hygiene nit; a `resp.body?.cancel()` (or reading to completion) on the success path would be tidy.

## 3. Overall verdict

**APPROVED.** Spec ✅ and no P0/P1 in the wired diff. The P2 items are advisory; items (1) and (2) in particular should be resolved before the deferred phases consume the restored endpoints.

## 4. For human attention (security-sensitive)

- Confirm the Octo server's wire encoding of thread `short_id` (string vs. int64 number) — this decides whether finding (1) is latent or live.
- Note a **pre-existing** (not introduced by this PR) item adjacent to the changed call site: `getGroupMembers` interpolates `groupNo` into the path **without** `encodeURIComponent` (the new lifecycle endpoints all encode correctly). Since `refreshMembers` feeds it a substring of the server-controlled `channel_id`, encoding it there would close a latent path-shaping vector. Out of scope for this diff, but flagging it while this area is under security review.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#169 Review Report
**Head SHA:** `202724ffd5f8611fec5a96fc83dda7c49753d416`
**Scope:** automated review / one-leg review (automated review — final verdict by Opus)
**Files:** 10 changed (+697 / −13), 5 source + 5 test files

---

| Item | Status | Evidence |
|------|--------|----------|
| Thread composite ID parsing (channel-id.ts) | ✅ | `extractParentGroupNo` / `extractThreadShortId` / `isThreadChannelId` are byte-identical to upstream `group-md.ts:36-60` except they import `THREAD_ID_SEPARATOR` constant instead of hardcoding `"____"`. 79 tests cover boundary cases (no sep, multi-sep first-wins, empty segments). |
| Thread lifecycle endpoints (api.ts) | ✅ | All 7 endpoints (create/list/get/delete/members/join/leave) match upstream `api-fetch.ts:1477-1605` on path, method, `encodeURIComponent`, and `Authorization: Bearer`. 206 tests assert URL + method + auth header. |
| loadGroupConfig thread routing (group-config.ts) | ✅ | Thread branch: shortId.md → legacy composite.md → undefined. Plain group: unchanged. `isSafeId` blocks path traversal for both branches (regex `[a-zA-Z0-9._-]` — underscores allowed). 40 new test lines cover new-semantics, legacy-fallback, and priority ordering. |
| refreshMembers parent group extraction (group-context.ts) | ✅ | `extractParentGroupNo(channelId)` called before `getGroupMembers` (line 213). Caches (memory/DB/robot flags) still keyed by full composite channelId — thread roster stays isolated. 31 new test lines verify parent extraction and plain-group passthrough. |
| Session isolation (session-router) | ✅ | Composite channelId is already the session key; no code change needed in SessionRouter. 29 new test lines confirm thread gets distinct session from parent group and sibling threads. |
| Thread / ThreadMember types (types.ts) | ✅ | Formal interfaces added (`short_id`, `name`, `creator_uid`, `status?`, `member_count?`; `uid`, `role`). Upstream uses inline types; this is a deliberate improvement, not a divergence. |
| Test suite integrity | ✅ | 1132 tests / 58 files (baseline 1096 / 56). No hand-fed data bypassing real paths detected. Mocks at `vi.mock('../octo/api')` boundary — appropriate for HTTP client tests. |

---

## 2. Findings

### F1 — getJson/postJson apply parseOctoJson to thread responses (upstream does not)
**Severity: P2 (design note, not a defect)**
**Diff-scope:** new — introduced by this PR's deliberate choice to route through fork helpers.

The PR routes `listThreads`, `getThread`, `listThreadMembers` through `getJson` and `createThread` through `postJson`, both of which apply `parseOctoJson` (int64-safe regex replacement on `message_id` fields). Upstream `api-fetch.ts:1477-1605` uses raw `fetch` + `resp.json()` for all thread endpoints — no int64 protection.

This is strictly **more defensive** than upstream: the regex is a no-op when `message_id` is absent (which it is in all thread responses). The PR body explicitly discloses this ("shares the same… int64-safe JSON parse"). No correctness risk.

### F2 — requestNoBody is a new abstraction not in upstream
**Severity: P2 (design note, not a defect)**
**Diff-scope:** new — introduced by this PR.

Upstream handles body-less POST (join/leave) by simply omitting `body` from the `fetch` init. The PR introduces `requestNoBody` as a shared helper. Behavior is equivalent: no `Content-Type` header, no body, same timeout + Bearer auth + error format. This reduces code duplication without changing semantics.

### F3 — Thread / ThreadMember are formal interfaces (upstream uses inline types)
**Severity: P2 (design note, not a defect)**
**Diff-scope:** new — introduced by this PR.

Upstream `api-fetch.ts` declares thread response shapes as inline return types. The PR defines formal `Thread` and `ThreadMember` interfaces in `types.ts`. Fields match upstream shapes exactly. Improves type safety and documentation.

### F4 — createThread returns `Thread | undefined` (upstream returns `Thread`)
**Severity: P2 (edge-case observation)**
**Diff-scope:** new — consequence of routing through `postJson`.

`postJson<T>` returns `T | undefined` (undefined when response body is empty). Upstream `createThread` returns `Thread` directly via raw `resp.json()`. In practice, the server always returns the created thread in the response body, so this is theoretical. Callers that don't check for undefined could get a runtime error, but TypeScript type checking would flag this.

---

## 3. Suggestions

1. **Consider exporting `requestNoBody`** if future endpoint additions also need body-less POST/DELETE (e.g., group management endpoints when restored). Currently private — fine for now but likely to be duplicated.
2. **Add a `source_message_id` type guard** — currently `sourceMessageId` is `number` (no validation for negative or non-integer). Low risk since it's an internal API, but a `Number.isInteger` check would be defense-in-depth.

---

## 4. Additional Findings

### Pre-existing: getGroupMembers does not encodeURIComponent(groupNo)
**Not introduced by this PR.** `api.ts:~470` constructs `/v1/bot/groups/${params.groupNo}/members` without encoding. Upstream `api-fetch.ts:497` has the same gap. The new thread endpoints correctly use `encodeURIComponent` on both `groupNo` and `shortId`. This PR's `extractParentGroupNo` output flows into `getGroupMembers` unencoded — but since upstream has the same behavior and group numbers are typically hex/alphanumeric strings, this is a pre-existing P2/note, not a regression.

---

## 5. Data-Flow Tracing

| Consumer | Data | Upstream Source | Verified? |
|----------|------|----------------|-----------|
| `refreshMembers` (group-context.ts:213) | `groupNo` for roster API | `extractParentGroupNo(channelId)` — splits on first `____`, returns parent group number for composite IDs, identity for plain groups | ✅ Data flows correctly. Cache keys remain full `channelId` (line 214-249). |
| `loadGroupConfig` (group-config.ts:62-68) | `shortId` for new-semantics file lookup | `extractThreadShortId(channelId)` — returns right segment after first `____`, or null | ✅ Short-circuits to legacy fallback when null. Both paths gated by `isSafeId`. |
| `loadGroupConfig` (group-config.ts:71) | `channelId` for legacy fallback | Full composite passed as slug → `isSafeId` accepts `[a-zA-Z0-9._-]` including underscores | ✅ Legacy `<groupNo>____<shortId>.md` filenames pass validation. |
| Thread API URL construction (api.ts:517-668) | `groupNo`, `shortId` path segments | All endpoints use `encodeURIComponent(params.groupNo)` and `encodeURIComponent(params.shortId)` | ✅ URL-safe. Test asserts `g/1` → `g%2F1`, `s 2` → `s%202`. |
| `listThreads` / `listThreadMembers` response | Parsed array | Defensive shape handling: bare array → `{ threads/members: [] }` envelope → `[]` fallback | ✅ Mirrors `getGroupMembers` pattern. Tolerates both server response shapes. |
| `createThread` request body | `name`, `sourceMessageId` | `body.name` always set; `body.source_message_id` only when `params.sourceMessageId != null` | ✅ Correctly omits field when not provided (not `undefined` in JSON). |

---

This PR is tagged `security_sensitive`. All C-items evaluated:

- **C1 — Dual-path parity:** N/A. This PR adds new thread endpoints; no symmetric add↔remove / subscribe↔unsubscribe pairs are modified. The thread lifecycle has create↔delete and join↔leave — both pairs are restored with matching path/method/auth patterns. ✅ Parity confirmed.
- **C2 — Control-flow ordering / nested reuse:** The shared `loadConfigFile` helper is called from two branches in `loadGroupConfig` (thread shortId, thread legacy, plain group). Each call is independent — no nesting, no double-application. `extractParentGroupNo` / `extractThreadShortId` are each called once at their consumption point. ✅ Clear.
- **C3 — Authorization boundary ≠ capability boundary:** Thread endpoints forward the bot's `botToken` for auth — same pattern as all other endpoints (`getGroupMembers`, `sendMessage`, etc.). Authorization (which threads the bot can access) is enforced server-side. The PR doesn't change any auth boundary. ✅ Clear.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. This PR adds client-side API wrappers; no server-side auth or container-state logic is modified.
- **C5 — Build ≠ runtime:** N/A. TypeScript library consumed via Node.js import. No browser extension, CLI binary, or relative-path-at-runtime concerns. Tests run via vitest.
- **C6 — Governance / policy docs:** N/A. No documentation or policy files modified.

---

## 7. Cross-Round Blocker Re-check (R6)

N/A — this is the first review pass for this PR.

---

## Summary

**10 files changed, +697 / −13 lines.** Clean, well-structured PR that restores thread lifecycle parity with `openclaw-channel-octo`. All 7 endpoints match upstream on path, method, auth, and URL encoding. Composite channel-id parsing is correct and well-tested (79 boundary-case tests). The `loadGroupConfig` path-semantics change (redline 5) correctly implements new-semantics-first with legacy fallback. The `refreshMembers` parent-group extraction (redline 6) correctly resolves the roster API key while keeping caches isolated by full composite channelId.

All findings are P2 design notes (deliberate improvements over upstream: int64-safe JSON parsing, formal interfaces, shared helper). No P0 or P1 defects identified.

[Octo-Q] verdict: **APPROVE** — no P0/P1 findings; only P2 design notes on deliberate improvements over upstream.

yujiawei (approved):
# Code Review — PR #169 (cc-channel-octo)

Reviewed at head SHA `202724ffd5f8611fec5a96fc83dda7c49753d416` against merge-base `203dc5ff`. Independently verified: `npm run type-check` clean, full suite green at **1132 tests / 58 files** (no regression vs. the stated 1096 baseline). This review also incorporates two independent advisory passes (a line-level adversarial pass and a long-context cross-file pass); both ran to completion.

## 1. Scope / Spec compliance — ✅

The diff matches the stated Phase-1 scope exactly: thread routing + composite `channel_id` parsing + thread lifecycle endpoints + types. No scope creep:

- **Not over-built:** no server-driven GROUP.md / THREAD.md injection (correctly deferred to P2/P3); the previously-removed GROUP.md / server-md APIs stay removed. The 7 restored endpoints are pure parity surface with **no live callers** yet — appropriate for a parity-restore phase.
- **Not under-built:** the path-semantics change (thread reads its own `<shortId>.md`, with a legacy whole-composite fallback) and the parent-group roster resolution are both implemented as described, with tests covering the boundary cases.
- The plain-group code path is byte-identical to prior behavior (the non-thread branch routes straight through `loadConfigFile(dir, channelId)`).

## 2. Code quality — Approved

No correctness, security, or data-loss defect is reachable by this diff. The findings below are all **P2 / latent** and do not block merge, but the first two should be addressed before any caller is wired up in P2/P3.

### P2 (latent — fix before wiring callers in P2/P3)

**1. `createThread` `sourceMessageId` is not int64-safe** — `src/octo/api.ts:544`
The param is typed `number`, but message IDs in this codebase exceed `Number.MAX_SAFE_INTEGER` (the tests use `2071488441815666688`). Everywhere else `message_id` is deliberately a `string` (see `parseOctoJson` int64 protection and `types.ts:86`). A future caller passing a numeric snowflake here would silently corrupt it before `JSON.stringify`. Not triggerable today (no caller), but the type should be `string` (or int64-safe) before P2/P3 calls `createThread`.

**2. `requestNoBody` never drains the response body** — `src/octo/api.ts` (`requestNoBody`)
On a 2xx it returns without consuming the body. Under `undici` (Node ≥18 global fetch), an undrained body can hold the socket out of the connection pool. `delete`/`join`/`leave` are the affected calls — all currently unused, so this is latent. A `await resp.arrayBuffer().catch(() => {})` (or `resp.body?.cancel()`) on the success path keeps it consistent with `postJson`/`getJson`, which both read the body.

### P2 (design / maintainability — within the approved plan, noted for awareness)

**3. Thread roster is the parent group's roster (by design), and bypasses the restored `listThreadMembers`** — `src/group-context.ts:212`
This is the documented redline-6 decision (a composite id would 404 against `/groups/{id}/members`). Worth stating the consequence explicitly: if topic membership is ever narrower than the parent group's, `@name→uid` resolution and the outbound-mention guard will see parent-group members inside a thread. This is acceptable because the mention guard is explicitly defense-in-depth and the server enforces real permissions — but once `listThreadMembers` is wired in a later phase, prefer the thread's own roster (or an intersection) for thread channels.

**4. Roster cache keyed by composite `channel_id` rather than parent group number** — `src/group-context.ts:200-213`
Also a deliberate redline-6 choice (per-thread isolated roster view). The tradeoff: N active threads under one parent each carry their own 1h refresh throttle and their own full copy of the (identical) parent roster — `O(active_threads × group_size)` memory and N redundant fetches/hour of the same data. Fine at current scale; revisit if thread fan-out grows (a parent-group-keyed roster cache with per-channel views would remove the duplication).

**5. `isThreadChannelId` is a pure string heuristic (`includes("____")`) with no channel-type guard** — `src/octo/channel-id.ts:53`
`channel_type` is available at every call site (`index.ts`/`group-context.ts` already gate on `isGroup`), and the live group-id grammar (`s{spaceId}_{groupNo}`, single underscores) makes a 4-underscore false positive unlikely — so this is safe in practice. But routing trusted-config and roster behavior off a substring match, rather than `channel_type === CommunityTopic`, is more fragile than it needs to be. Consider threading `channelType` in as the authoritative discriminator. Note `extractThreadShortId` already handles the empty-shortId edge (`"g____"` → null), so the config path degrades safely there.

## 3. Security review (PR flagged security_sensitive)

The trust model of `group-config.ts` is unchanged and intact: the instruction file is injected **unsanitized as trusted**, and the protection remains (a) operator-owned config dir, (b) `isSafeId` filename pinning that blocks traversal (`..`, `/`, `\`), and (c) the group/world-writable refusal. The thread change only alters which *slug* is looked up within that already-trusted dir — it does not introduce a user-controlled path or a new injection sink. The `<shortId>.md` naming does mean two threads sharing a short id (if short ids are not globally unique) would read the same operator file; since the operator names these files, that is a config-authoring concern, not an injection vector — but **please confirm short_id uniqueness scope** (global vs. per-parent-group); if per-group, document the namespacing expectation for operators.

All seven endpoints carry `Authorization: Bearer <botToken>`, url-encode path segments, and share the client's timeout/error format — verified by tests.

## 4. Coverage / blind spots

- **short_id uniqueness scope** — global vs. per-parent-group — could not be verified from the diff; it determines whether finding #1 (security section) and the `<shortId>.md` naming are fully safe.
- **Thread vs. parent-group membership semantics** — whether topics have narrower rosters/roles — not verifiable from the diff; bears on findings 3 and the eventual use of `listThreadMembers`.
- The 7 restored endpoints have no callers in this diff, so their runtime behavior against a live server is unexercised beyond the unit-level path/method/auth assertions.

## Verdict

Scope is clean and the implementation is careful and well-tested. All findings are P2/latent — none blocks this parity-restore phase. Approving, with a request to address the int64 type (#1) and body-drain (#2) before P2/P3 wires real callers.

PR #171, opened 2026-06-30
fix(session-store): dedupe cold-start double-write via partial unique index

boris-clark:
Part of #88

## Problem

On a group cold start, the G4 history backfill (`seedHistoryFromApi`) and the
live inbound handler both persist the *same* triggering message into the
`messages` table. With a plain `INSERT` and no uniqueness, that left two
identical `(session_id, role, message_seq)` rows — the "seq236" double-write
(root cause tracked in the linked issue). `append()` was not idempotent and
`messages` had no `(session_id, message_seq)` constraint.

## Fix

1. **Partial unique index** on `messages(session_id, role, message_seq)
   WHERE message_seq IS NOT NULL`.
   - Keyed on `role`: a user turn and the bot's reply legitimately share the
     inbound `message_seq` (the assistant turn is stored with `msg.message_seq`),
     so a `session_id + message_seq`-only index would wrongly forbid the reply —
     and the existing-row migration would delete every bot reply that shares a
     seq with its user message. Including `role` collapses only the true
     duplicate (the inbound counted twice, always same role).
   - **Partial** (`WHERE message_seq IS NOT NULL`): assistant / legacy / no-seq
     turns carry `NULL` seq and must remain unconstrained.
2. **`append()` → `INSERT OR IGNORE`**: a re-write of an already stored
   `(session_id, role, message_seq)` is silently dropped, so the cold-start
   double-write is a no-op. `NULL`-seq rows are exempt from the index, so
   `OR IGNORE` never suppresses them.
3. **Existing-row dedup migration** (`dedupeMessagesBySeq`, run in `init()`
   before the index is created — a unique index cannot be built over a table
   that already violates it):
   - Keeps the earliest row (`MIN(id)`) per `(session_id, role, message_seq)`,
     deletes the rest; `NULL`-seq rows are never touched.
   - Wrapped in a single transaction with explicit before/after row accounting
     that throws and rolls back on a delta mismatch (no partial migration, no
     lost data).
   - Idempotent: a second run over clean data deletes nothing.

Scope is limited to the session store + migration; no changes to
server-md / events / write-back.

## Tests

New `session-store` tests:
- seq236 cold-start double-write repro → collapses to one row.
- INSERT OR IGNORE keeps the first write (earliest wins).
- user turn + bot reply sharing a seq are preserved (role keeps them distinct).
- the index is genuinely partial — multiple `NULL`-seq rows all insert.
- dedup migration: before/after row accounting, preserves the assistant
  same-seq row and `NULL`-seq rows, and is idempotent across two runs.
- `init()` on a dirty DB dedupes then enforces uniqueness on subsequent writes.

The multi-turn DM e2e fixture now uses realistic monotonic seqs (WuKongIM
assigns `message_seq` monotonically per channel; two distinct inbound messages
never share a seq).

Full suite green: 1138 passed. `type-check` and `lint` clean.

### DIFF
--- a/src/session-store.ts
+++ b/src/session-store.ts
@@ -149,13 +149,15 @@ export class SessionStore {
     // idempotent. Keyed on role because a user turn and the bot's reply
     // legitimately share the inbound seq (appendAssistant is called with the
     // inbound message_seq), so a session_id+seq-only index would wrongly forbid
-    // the reply. WHERE message_seq IS NOT NULL keeps it partial: assistant /
-    // legacy / no-seq turns carry NULL seq and stay unconstrained (NULLs are
-    // never equal to one another, but the explicit predicate documents intent and
-    // keeps the index small).
+    // the reply. The predicate is `message_seq > 0` (not merely IS NOT NULL):
+    // a synthetic cron fire carries the seq=0 SENTINEL (no real wire seq; see
+    // index.ts), and multiple cron rounds legitimately share seq=0 — constraining
+    // them would silently drop real rounds via INSERT OR IGNORE. So only real
+    // positive wire seqs are deduped; NULL- and 0/negative-seq turns stay
+    // unconstrained. `> 0` matches the existing "real seq" predicate in index.ts.
     this.adapter.exec(
       'CREATE UNIQUE INDEX IF NOT EXISTS idx_messages_session_role_seq ' +
-        'ON messages(session_id, role, message_seq) WHERE message_seq IS NOT NULL',
+        'ON messages(session_id, role, message_seq) WHERE message_seq > 0',
     );
 
     this.selectSession = this.adapter.prepare('SELECT * FROM sessions WHERE id = ?');
@@ -168,8 +170,9 @@ export class SessionStore {
     // row at its message_seq (the "seq236" bug). The cure is a PARTIAL unique
     // index (created above) plus INSERT OR IGNORE here: a re-write of an already
     // stored (session_id, role, message_seq) is silently dropped, making append()
-    // idempotent. NULL-seq rows (assistant / legacy / no-seq turns) are exempt
-    // from the index, so OR IGNORE never suppresses them.
+    // idempotent. The index is partial on `message_seq > 0`, so NULL-seq turns
+    // AND the seq=0 cron sentinel (multiple legit rounds share it) are exempt —
+    // OR IGNORE never suppresses them.
     this.insertMessage = this.adapter.prepare(
       'INSERT OR IGNORE INTO messages (session_id, role, content, timestamp, message_seq, from_name) VALUES (?, ?, ?, ?, ?, ?)',
     );
@@ -205,11 +208,17 @@ export class SessionStore {
   /**
    * XIN-154 existing-row migration: collapse duplicate
    * (session_id, role, message_seq) rows left by the pre-fix seq236 double-write,
-   * keeping the lowest id (the earliest insert) in each group. Rows with NULL
-   * message_seq (assistant / legacy / no-seq turns) are never touched — they are
-   * not part of the uniqueness contract — and rows that merely share a seq across
-   * roles (a user turn and the bot's reply) are preserved because role is part of
-   * the grouping key.
+   * keeping the lowest id (the earliest insert) in each group.
+   *
+   * Only REAL wire seqs (message_seq > 0) participate. Rows are exempt when:
+   *   - message_seq IS NULL (assistant / legacy / no-seq turns), or
+   *   - message_seq <= 0 — notably the seq=0 SENTINEL a synthetic cron fire
+   *     carries (no real wire seq; see index.ts where the cron round stores
+   *     seq=0). Multiple cron rounds legitimately share seq=0, so folding on it
+   *     would delete real data — hence `> 0`, matching the codebase's existing
+   *     "real seq" predicate (`msg.message_seq > 0`).
+   * Rows that merely share a seq across roles (a user turn and the bot's reply)
+   * are preserved because role is part of the grouping key.
    *
    * Wrapped in a single transaction with explicit before/after row accounting: a
    * mismatch between the delete count and the row delta means the table was
@@ -226,8 +235,8 @@ export class SessionStore {
       ).n;
       const result = this.adapter
         .prepare(
-          'DELETE FROM messages WHERE message_seq IS NOT NULL AND id NOT IN (' +
-            'SELECT MIN(id) FROM messages WHERE message_seq IS NOT NULL ' +
+          'DELETE FROM messages WHERE message_seq > 0 AND id NOT IN (' +
+            'SELECT MIN(id) FROM messages WHERE message_seq > 0 ' +
             'GROUP BY session_id, role, message_seq)',
         )
         .run();
--- a/src/session-store.ts
+++ b/src/session-store.ts
@@ -138,13 +138,40 @@ export class SessionStore {
       );
     }
 
+    // XIN-154: collapse any pre-existing seq236-style duplicate rows BEFORE
+    // creating the partial unique index — a unique index cannot be built over a
+    // table that already violates it. Idempotent + transactional (see the method),
+    // so it is safe to run on every startup; on an already-clean DB it deletes
+    // nothing. After this, INSERT OR IGNORE (insertMessage) keeps the table clean.
+    this.dedupeMessagesBySeq();
+
+    // Partial unique index that makes the (session_id, role, message_seq) write
+    // idempotent. Keyed on role because a user turn and the bot's reply
+    // legitimately share the inbound seq (appendAssistant is called with the
+    // inbound message_seq), so a session_id+seq-only index would wrongly forbid
+    // the reply. WHERE message_seq IS NOT NULL keeps it partial: assistant /
+    // legacy / no-seq turns carry NULL seq and stay unconstrained (NULLs are
+    // never equal to one another, but the explicit predicate documents intent and
+    // keeps the index small).
+    this.adapter.exec(
+      'CREATE UNIQUE INDEX IF NOT EXISTS idx_messages_session_role_seq ' +
+        'ON messages(session_id, role, message_seq) WHERE message_seq IS NOT NULL',
+    );
+
     this.selectSession = this.adapter.prepare('SELECT * FROM sessions WHERE id = ?');
     this.insertSession = this.adapter.prepare(
       'INSERT INTO sessions (id, channel_id, channel_type, created_at, updated_at) VALUES (?, ?, ?, ?, ?)',
     );
     this.touchSession = this.adapter.prepare('UPDATE sessions SET updated_at = ? WHERE id = ?');
+    // XIN-145 root cause: the G4 cold-start backfill (seedHistoryFromApi) and the
+    // live inbound path both persisted the SAME triggering message, doubling the
+    // row at its message_seq (the "seq236" bug). The cure is a PARTIAL unique
+    // index (created above) plus INSERT OR IGNORE here: a re-write of an already
+    // stored (session_id, role, message_seq) is silently dropped, making append()
+    // idempotent. NULL-seq rows (assistant / legacy / no-seq turns) are exempt
+    // from the index, so OR IGNORE never suppresses them.
     this.insertMessage = this.adapter.prepare(
-      'INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name) VALUES (?, ?, ?, ?, ?, ?)',
+      'INSERT OR IGNORE INTO messages (session_id, role, content, timestamp, message_seq, from_name) VALUES (?, ?, ?, ?, ?, ?)',
     );
     this.selectRecentMessages = this.adapter.prepare(
       'SELECT role, content, message_seq, from_name FROM messages WHERE session_id = ? ORDER BY id DESC LIMIT ?',
@@ -173,6 +200,50 @@ export class SessionStore {
     this.deleteExpiredSdkSessions = this.adapter.prepare(
       'DELETE FROM sdk_sessions WHERE updated_at < ?',
     );
+  }
+
+  /**
+   * XIN-154 existing-row migration: collapse duplicate
+   * (session_id, role, message_seq) rows left by the pre-fix seq236 double-write,
+   * keeping the lowest id (the earliest insert) in each group. Rows with NULL
+   * message_seq (assistant / legacy / no-seq turns) are never touched — they are
+   * not part of the uniqueness contract — and rows that merely share a seq across
+   * roles (a user turn and the bot's reply) are preserved because role is part of
+   * the grouping key.
+   *
+   * Wrapped in a single transaction with explicit before/after row accounting: a
+   * mismatch between the delete count and the row delta means the table was
+   * mutated concurrently, so we throw and roll back rather than report a clean
+   * migration over an inconsistent result (防丢数据). Idempotent: a second run over
+   * already-clean data deletes nothing and leaves counts unchanged.
+   */
+  dedupeMessagesBySeq(): { before: number; removed: number; after: number } {
+    const tx = this.adapter.transaction(() => {
+      const before = (
+        this.adapter.prepare('SELECT COUNT(*) AS n FROM messages').get() as {
+          n: number;
+        }
+      ).n;
+      const result = this.adapter
+        .prepare(
+          'DELETE FROM messages WHERE message_seq IS NOT NULL AND id NOT IN (' +
+            'SELECT MIN(id) FROM messages WHERE message_seq IS NOT NULL ' +
+            'GROUP BY session_id, role, message_seq)',
+        )
+        .run();
+      const after = (
+        this.adapter.prepare('SELECT COUNT(*) AS n FROM messages').get() as {
+          n: number;
+        }
+      ).n;
+      if (before - after !== result.changes) {
+        throw new Error(
+          `session-store: dedup row accounting mismatch (before=${before}, after=${after}, deleted=${result.changes}) — rolling back`,
+        );
+      }
+      return { before, removed: result.changes, after };
+    });
+    return tx();
   }
 
   getOrCreate(id: string, channelId: string, channelType: number): Session {
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -595,11 +595,16 @@ describe('E2E smoke tests', () => {
   // --- 6. Multi-turn history accumulates ---
 
   it('Multi-turn DM: history accumulates across messages', async () => {
+    // message_seq is monotonic per channel in WuKongIM, so two distinct inbound
+    // messages always carry distinct seqs. Use realistic seqs here — the
+    // (session_id, role, message_seq) uniqueness contract treats a repeated
+    // (user, seq) as the same message (seq236 double-write guard), which would
+    // otherwise drop the second turn if both reused the makeDmMsg default seq.
     mockQueryYield('First reply');
-    await simulateMessage(makeDmMsg('Hello'), config, store, router, groupContext, streamRelay);
+    await simulateMessage(makeDmMsg('Hello', { message_seq: 1 }), config, store, router, groupContext, streamRelay);
 
     mockQueryYield('Second reply');
-    await simulateMessage(makeDmMsg('Follow up'), config, store, router, groupContext, streamRelay);
+    await simulateMessage(makeDmMsg('Follow up', { message_seq: 2 }), config, store, router, groupContext, streamRelay);
 
     expect(queryAgent).toHaveBeenCalledTimes(2);
 
--- a/src/__tests__/session-store.test.ts
+++ b/src/__tests__/session-store.test.ts
@@ -333,6 +333,215 @@ describe('SessionStore G10 message_seq migration (Q1-2)', () => {
   });
 });
 
+// XIN-145 root cause / XIN-154 fix: the G4 cold-start backfill (seedHistoryFromApi)
+// and the live inbound path both persist the SAME triggering message. With a plain
+// INSERT and no uniqueness, that left two identical (session_id, role, message_seq)
+// rows (the seq236 double-write). The fix is a PARTIAL unique index on
+// (session_id, role, message_seq) WHERE message_seq IS NOT NULL + INSERT OR IGNORE.
+//
+// The index is keyed on role (not just session_id+seq): a user turn and the bot's
+// reply legitimately share the inbound seq (agent-bridge stores the assistant turn
+// with msg.message_seq — see appendAssistant at index.ts), so collapsing on seq
+// alone would drop every bot reply. The index is partial (WHERE message_seq IS NOT
+// NULL) because assistant / legacy / no-seq turns carry NULL seq and must remain
+// unconstrained.
+describe('SessionStore — seq236 double-write root-cause fix (partial unique index)', () => {
+  let adapter: DbAdapter;
+  let store: SessionStore;
+
+  beforeEach(() => {
+    adapter = createAdapter(':memory:');
+    store = new SessionStore(adapter);
+    store.init();
+  });
+
+  afterEach(() => {
+    store.close();
+  });
+
+  function countSeq(sessionId: string, role: string, seq: number): number {
+    return (
+      adapter
+        .prepare(
+          'SELECT COUNT(*) AS n FROM messages WHERE session_id = ? AND role = ? AND message_seq = ?',
+        )
+        .get(sessionId, role, seq) as { n: number }
+    ).n;
+  }
+
+  it('cold-start double-write of the same inbound (seq 236) collapses to ONE row', () => {
+    store.getOrCreate('g1', 'ch1', 2);
+    // Path 1 — G4 cold-start backfill seeds the just-arrived inbound from the API.
+    store.appendUser('g1', 'message at seq 236', 236, 'Alice');
+    // Path 2 — the live inbound handler appends the very same message moments later.
+    store.appendUser('g1', 'message at seq 236', 236, 'Alice');
+
+    // Pre-fix this was 2 rows (the bug). The partial unique index + INSERT OR
+    // IGNORE make the second write a no-op, so seq 236 is no longer doubled.
+    expect(countSeq('g1', 'user', 236)).toBe(1);
+  });
+
+  it('keeps the FIRST write when a duplicate seq arrives (INSERT OR IGNORE, earliest wins)', () => {
+    store.getOrCreate('g1', 'ch1', 2);
+    store.appendUser('g1', 'first copy', 236, 'Alice');
+    store.appendUser('g1', 'second copy (ignored)', 236, 'Alice');
+
+    const history = store.buildHistoryPrefix('g1', 10);
+    expect(history).toContain('first copy');
+    expect(history).not.toContain('second copy (ignored)');
+  });
+
+  it('a user turn and the bot reply may SHARE a seq — role keeps them distinct', () => {
+    store.getOrCreate('g1', 'ch1', 2);
+    store.appendUser('g1', 'the question', 236, 'Alice');
+    // The assistant reply is stored with the inbound seq (segmentation relies on
+    // this). It must NOT be swallowed by the user row at the same seq.
+    store.appendAssistant('g1', 'the answer', 236, 'OctoBot');
+
+    expect(countSeq('g1', 'user', 236)).toBe(1);
+    expect(countSeq('g1', 'assistant', 236)).toBe(1);
+  });
+
+  it('the index is genuinely PARTIAL — NULL message_seq rows are never constrained', () => {
+    store.getOrCreate('g1', 'ch1', 2);
+    // assistant / legacy / no-seq turns all carry NULL seq; any number of them
+    // must insert fine (WHERE message_seq IS NOT NULL exempts them).
+    store.appendAssistant('g1', 'reply A', undefined, 'OctoBot');
+    store.appendAssistant('g1', 'reply B', undefined, 'OctoBot');
+    store.appendUser('g1', 'no-seq one', undefined, 'Alice');
+    store.appendUser('g1', 'no-seq two', undefined, 'Alice');
+
+    const nullCount = (
+      adapter
+        .prepare('SELECT COUNT(*) AS n FROM messages WHERE message_seq IS NULL')
+        .get() as { n: number }
+    ).n;
+    expect(nullCount).toBe(4);
+  });
+});
+
+describe('SessionStore — existing-row dedup migration (seq236 backfill)', () => {
+  // Build a pre-fix messages table (current columns, NO unique index) and seed it
+  // with the duplicate rows the bug produced, so we can exercise the migration.
+  function seedDirtyDb(): DbAdapter {
+    const adapter = createAdapter(':memory:');
+    adapter.exec(`
+      CREATE TABLE sessions (
+        id TEXT PRIMARY KEY,
+        channel_id TEXT NOT NULL,
+        channel_type INTEGER NOT NULL,
+        created_at INTEGER NOT NULL,
+        updated_at INTEGER NOT NULL
+      );
+      CREATE TABLE messages (
+        id INTEGER PRIMARY KEY AUTOINCREMENT,
+        session_id TEXT NOT NULL,
+        role TEXT NOT NULL,
+        content TEXT NOT NULL,
+        timestamp INTEGER NOT NULL,
+        message_seq INTEGER,
+        from_name TEXT
+      );
+      INSERT INTO sessions VALUES ('g1', 'ch1', 2, 0, 0);
+      -- Doubled user inbound at seq 236 (the bug). Earliest id must survive.
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'user', 'seq236 first', 10, 236, 'Alice');
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'user', 'seq236 second', 11, 236, 'Alice');
+      -- Assistant reply sharing seq 236 — NOT a duplicate, must be preserved.
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'assistant', 'seq236 answer', 12, 236, 'OctoBot');
+      -- A distinct user seq — untouched.
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'user', 'seq237', 13, 237, 'Alice');
+      -- NULL-seq rows (assistant / legacy) — never part of the uniqueness contract.
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'assistant', 'no-seq A', 14, NULL, 'OctoBot');
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'assistant', 'no-seq B', 15, NULL, 'OctoBot');
+    `);
+    return adapter;
+  }
+
+  it('collapses duplicate (session_id, role, seq) rows, preserves everything else, and is idempotent', () => {
+    const adapter = seedDirtyDb();
+    const store = new SessionStore(adapter);
+
+    // First pass: one duplicate user row removed; row accounting reported.
+    const first = store.dedupeMessagesBySeq();
+    expect(first.before).toBe(6);
+    expect(first.removed).toBe(1);
+    expect(first.after).toBe(5);
+    expect(first.after).toBe(first.before - first.removed);
+
+    // Idempotent re-entrancy: a second pass over clean data removes nothing.
+    const second = store.dedupeMessagesBySeq();
+    expect(second.before).toBe(5);
+    expect(second.removed).toBe(0);
+    expect(second.after).toBe(5);
+
+    // The earliest (MIN id) duplicate survives.
+    const userSeq236 = adapter
+      .prepare(
+        "SELECT content FROM messages WHERE session_id = 'g1' AND role = 'user' AND message_seq = 236",
+      )
+      .all() as Array<{ content: string }>;
+    expect(userSeq236).toHaveLength(1);
+    expect(userSeq236[0].content).toBe('seq236 first');
+
+    // The assistant reply sharing seq 236 is preserved (role keeps it distinct).
+    const asstSeq236 = (
+      adapter
+        .prepare(
+          "SELECT COUNT(*) AS n FROM messages WHERE session_id = 'g1' AND role = 'assistant' AND message_seq = 236",
+        )
+        .get() as { n: number }
+    ).n;
+    expect(asstSeq236).toBe(1);
+
+    // NULL-seq rows are untouched by the migration.
+    const nullCount = (
+      adapter
+        .prepare('SELECT COUNT(*) AS n FROM messages WHERE message_seq IS NULL')
+        .get() as { n: number }
+    ).n;
+    expect(nullCount).toBe(2);
+
+    store.close();
+  });
+
+  it('init() dedupes a dirty DB then enforces uniqueness on subsequent writes', () => {
+    const adapter = seedDirtyDb();
+    const store = new SessionStore(adapter);
+
+    // init() must dedupe BEFORE creating the unique index (else index creation
+    // would fail on the pre-existing duplicates).
+    expect(() => store.init()).not.toThrow();
+
+    const userSeq236 = (
+      adapter
+        .prepare(
+          "SELECT COUNT(*) AS n FROM messages WHERE session_id = 'g1' AND role = 'user' AND message_seq = 236",
+        )
+        .get() as { n: number }
+    ).n;
+    expect(userSeq236).toBe(1);
+
+    // Going forward, a re-arriving duplicate inbound is ignored, not doubled.
+    store.appendUser('g1', 'seq236 re-arrival', 236, 'Alice');
+    const afterAppend = (
+      adapter
+        .prepare(
+          "SELECT COUNT(*) AS n FROM messages WHERE session_id = 'g1' AND role = 'user' AND message_seq = 236",
+        )
+        .get() as { n: number }
+    ).n;
+    expect(afterAppend).toBe(1);
+
+    store.close();
+  });
+});
+
 describe('SessionStore — v0.3 SDK session ids (persistent sessions)', () => {
   let adapter: DbAdapter;
   let store: SessionStore;
--- a/src/__tests__/session-store.test.ts
+++ b/src/__tests__/session-store.test.ts
@@ -418,6 +418,25 @@ describe('SessionStore — seq236 double-write root-cause fix (partial unique in
     ).n;
     expect(nullCount).toBe(4);
   });
+
+  it('does NOT constrain the seq=0 cron sentinel — multiple cron rounds all persist', () => {
+    // A synthetic cron fire carries message_seq=0 (no real wire seq; see
+    // index.ts). Multiple rounds share seq=0 and are legitimately distinct. The
+    // partial index is `WHERE message_seq > 0`, so the sentinel is exempt and
+    // INSERT OR IGNORE must NOT collapse these (reviewer's data-loss blocker).
+    store.getOrCreate('g1', 'ch1', 2);
+    store.appendUser('g1', 'cron round 1', 0, 'Scheduler');
+    store.appendAssistant('g1', 'cron reply 1', 0, 'OctoBot');
+    store.appendUser('g1', 'cron round 2', 0, 'Scheduler');
+    store.appendAssistant('g1', 'cron reply 2', 0, 'OctoBot');
+
+    const count = (
+      adapter
+        .prepare("SELECT COUNT(*) AS n FROM messages WHERE session_id = 'g1' AND message_seq = 0")
+        .get() as { n: number }
+    ).n;
+    expect(count).toBe(4);
+  });
 });
 
 describe('SessionStore — existing-row dedup migration (seq236 backfill)', () => {
@@ -459,6 +478,17 @@ describe('SessionStore — existing-row dedup migration (seq236 backfill)', () =
         VALUES ('g1', 'assistant', 'no-seq A', 14, NULL, 'OctoBot');
       INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
         VALUES ('g1', 'assistant', 'no-seq B', 15, NULL, 'OctoBot');
+      -- seq=0 SENTINEL: synthetic cron fires carry message_seq=0 (no real wire
+      -- seq; see index.ts). Multiple cron rounds legitimately share seq=0 and are
+      -- DISTINCT rows. Folding them on (session, role, seq) would delete real
+      -- data — the reviewer's data-loss blocker. The migration must leave all of
+      -- these intact (predicate is message_seq > 0, not IS NOT NULL).
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'user', 'cron round 1', 16, 0, 'Scheduler');
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'user', 'cron round 2', 17, 0, 'Scheduler');
+      INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name)
+        VALUES ('g1', 'assistant', 'cron reply 1', 18, 0, 'OctoBot');
     `);
     return adapter;
   }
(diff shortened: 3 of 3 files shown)


Jerry-Xin (dismissed):
Summary: The PR is in scope and the partial unique index plus pre-index dedupe migration addresses the session-store duplicate-write issue without breaking the user/assistant same-sequence model.

💬 Non-blocking
- 🟡 Warning: `src/session-store.ts:173-175` uses `INSERT OR IGNORE`, which ignores more than the intended duplicate `(session_id, role, message_seq)` conflict. It can also suppress other SQLite constraint failures such as `NOT NULL` or `CHECK` violations. Consider targeting only this uniqueness case with an explicit conflict target, for example `ON CONFLICT(session_id, role, message_seq) WHERE message_seq IS NOT NULL DO NOTHING`, if supported by the bundled SQLite version.

✅ Highlights
- 🔵 The index key correctly includes `role`, preserving the legitimate user/assistant pair that shares an inbound `message_seq` (`src/session-store.ts:156-159`).
- 🔵 The migration runs before index creation and is idempotent, which is the right ordering for dirty existing databases (`src/session-store.ts:141-159`, `src/session-store.ts:220-247`).
- 🔵 The new tests cover the important behavioral edges: duplicate collapse, earliest-row retention, role separation, partial `NULL` behavior, migration idempotency, and dirty-DB init.

Verification note: I attempted the focused tests and type-check, but this checkout does not have dependencies installed, so `vitest` and `tsc` were not available.

src/session-store.ts
```diff
@@ -138,13 +138,40 @@ export class SessionStore {
...
+    // XIN-145 root cause: the G4 cold-start backfill (seedHistoryFromApi) and the
+    // live inbound path both persisted the SAME triggering message, doubling the
+    // row at its message_seq (the "seq236" bug). The cure is a PARTIAL unique
+    // index (created above) plus INSERT OR IGNORE here: a re-write of an already
+    // stored (session_id, role, message_seq) is silently dropped, making append()
+    // idempotent. NULL-seq rows (assistant / legacy / no-seq turns) are exempt
+    // from the index, so OR IGNORE never suppresses them.
     this.insertMessage = this.adapter.prepare(
-      'INSERT INTO messages (session_id, role, content, timestamp, message_seq, from_name) VALUES (?, ?, ?, ?, ?, ?)',
+      'INSERT OR IGNORE INTO messages (session_id, role, content, timestamp, message_seq, from_name) VALUES (?, ?, ?, ?, ?, ?)',
```
OctoBoooot, line 177: **[minor]** `INSERT OR IGNORE` suppresses every constraint violation on the row, not just the partial-unique-index conflict. Today `messages` has only the `role` CHECK plus this index, so the blast radius is small — but a future `NOT NULL`/`CHECK` added to `messages` would turn a violating write into a silent no-op instead of a surfaced error. Scoping it to `ON CONFLICT(session_id, role, message_seq) WHERE message_seq IS NOT NULL DO NOTHING` would limit the ignore to exactly the intended conflict. Non-blocking; the broad form matches existing style.

OctoBoooot (commented):
## Review: fix(session-store): dedupe cold-start double-write via partial unique index (#171)

**Verdict:** Approve with comments — correct, well-tested fix for the seq236 double-write; one question on assistant-turn collision worth settling, two minors.

CI is green. PR is a draft. `needs-human-review` + `size/L` already applied.

**Risk note:** the file-path classifier scored this `normal`, but this PR ships a **schema migration that deletes rows** (`dedupeMessagesBySeq`), so I reviewed it at migration-tier rigor — reversibility, data-loss safety, and index correctness. It holds up.

I verified the load-bearing design claim rather than taking the PR body's word: `appendAssistant` (index.ts:861) is called with the **inbound** `msg.message_seq` — the same seq as the user turn at index.ts:665. So a `(session_id, message_seq)`-only index would genuinely collide a user message with its own reply, and the backfill migration would delete bot replies sharing a user seq. Role-keying the index is therefore necessary, not incidental. The migration's transaction + before/after row-accounting rollback, the NULL-seq exemption, and `INSERT OR IGNORE` only suppressing true same-role dupes are all correct.

### Question
- `src/session-store.ts` `appendAssistant` + index.ts:861 — can a single inbound `message_seq` ever produce **two legitimate assistant rows**? If the gateway ever stores more than one assistant turn for one inbound (e.g., a split long reply, or a tool-use turn followed by a final turn, or a regenerate/retry), both would carry `(session_id, 'assistant', <inbound seq>)` and the partial unique index + `INSERT OR IGNORE` would silently drop the second. Today's streaming path stores one `fullResponse` per turn so this is safe — but the constraint now makes "one assistant row per inbound seq" a hard invariant. If that's intended, worth a one-line comment on `appendAssistant`; if multi-part assistant turns are on the #88 roadmap, this index will silently eat them. Not blocking — I can't construct the second-row call path in the current code.

### Minor
- `src/session-store.ts` `INSERT OR IGNORE` — this suppresses ALL constraint violations on the row, not just the partial-unique-index one. Today `messages` has only the `role` CHECK and this index, so the blast radius is limited, but if a future `NOT NULL`/`CHECK` is added to `messages`, a violating write would become a silent no-op instead of an error. A targeted `ON CONFLICT(session_id, role, message_seq) WHERE message_seq IS NOT NULL DO NOTHING` would scope the ignore to exactly the intended conflict. Non-blocking; the broader `OR IGNORE` matches the existing style.
- `src/session-store.ts:229-231` `dedupeMessagesBySeq` — the `DELETE ... WHERE id NOT IN (SELECT MIN(id) ... GROUP BY ...)` subquery is correct but O(n) over the full `messages` table on every `init()`. On a large historical DB this runs at every startup even when clean (the body says idempotent, which it is — but it still scans). Consider gating it behind a one-time migration marker so steady-state startups skip the full scan. Non-blocking for correctness.

### Praise
- The role-keyed partial index is the subtle-but-correct call, and the PR body explains exactly why a `session_id+seq`-only index would be wrong (delete every bot reply sharing a user seq) — that's the reasoning I had to reconstruct from the call sites, stated upfront.
- `dedupeMessagesBySeq` wraps the delete in a transaction with explicit before/after row accounting that throws + rolls back on a delta mismatch. A migration that deletes rows and refuses to report success over an inconsistent result is exactly the data-loss discipline this class of change needs.
- The e2e fixture fix (monotonic seqs 1/2 instead of a shared default) is honest: the new invariant would have made the old fixture's repeated-default-seq turns collapse, and the author updated the fixture to reflect real WuKongIM monotonic-seq behavior rather than weakening the constraint to fit the test.

- The `from_name` injection guard (`sanitizeDisplayName`) on the append path is pre-existing and correct — noting only that this PR's `append()` change preserves it.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# cc-channel-octo PR#171 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo#171
**Head SHA:** `332c7c45ab61f5c408f7cf7b0661ae29378bb826`
**Title:** fix(session-store): dedupe cold-start double-write via partial unique index

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Partial unique index correctness | ✅ | `session-store.ts:155-158` — `CREATE UNIQUE INDEX ... ON messages(session_id, role, message_seq) WHERE message_seq IS NOT NULL` |
| INSERT OR IGNORE idempotency | ✅ | `session-store.ts:171` — single prepared statement, all writes flow through `append()` → `insertMessage.run()` |
| Dedup migration correctness | ✅ | `session-store.ts:223-248` — transactional, MIN(id) keeps earliest, row accounting throws on mismatch |
| Init ordering (dedupe before index) | ✅ | `session-store.ts:141` calls `dedupeMessagesBySeq()` before `CREATE UNIQUE INDEX` at line 155 |
| NULL-seq exemption | ✅ | Both the DELETE predicate (`WHERE message_seq IS NOT NULL`) and partial index predicate (`WHERE message_seq IS NOT NULL`) correctly exempt NULL rows |
| Role-keyed index (user+assistant share seq) | ✅ | Index includes `role` column — test at `session-store.test.ts:395-404` validates user turn + bot reply at same seq are both preserved |
| E2E test fix (monotonic seqs) | ✅ | `e2e.test.ts:598-603` — assigns distinct seqs (1, 2) to multi-turn messages, matching WuKongIM's monotonic per-channel seq assignment |
| Test coverage | ✅ | 6 new test cases covering: cold-start double-write, first-write-wins, role distinction, NULL-seq exemption, dedup migration (idempotency + row accounting), init() integration |

## 2. Findings

### No P0 or P1 issues found.

### P2 Notes (non-blocking)

**P2-1: `touchSession` runs even when INSERT OR IGNORE suppresses the row**
- **Location:** `session-store.ts:283-284` — `this.touchSession.run(now, sessionId)` runs unconditionally after `this.insertMessage.run()`
- **Diff-scope:** pre-existing (the `touchSession` call was already there before this PR; INSERT OR IGNORE is new but doesn't gate the touch)
- **Impact:** When a cold-start double-write is silently suppressed, the session's `updated_at` is still bumped. This extends the 7-day expiry window by a few milliseconds. Not a correctness issue — the session should stay alive anyway since it's actively being used.
- **Suggestion:** No change needed. Documenting for completeness.

**P2-2: `dedupeMessagesBySeq()` is a public method**
- **Location:** `session-store.ts:223` — exposed as public API
- **Diff-scope:** new (added by this PR)
- **Impact:** Tests exercise it directly (needed for unit testing). It's a migration detail that ideally wouldn't be part of the public surface, but TypeScript has no `@VisibleForTesting` annotation, and the method is well-documented.
- **Suggestion:** No change needed. The JSDoc clearly documents it as a migration method.

## 3. Data Flow Trace

### Double-write root cause (confirmed)
1. **Path 1 — G4 backfill:** `index.ts:657` → `seedHistoryFromApi()` → `index.ts:1040` → `store.appendUser(sessionKey, content, m.message_seq, ...)` — persists the triggering message as part of the API history fetch
2. **Path 2 — live inbound:** `index.ts:665` → `store.appendUser(sessionKey, userContent, msg.message_seq, ...)` — persists the same triggering message again
3. Both paths share the same `(session_id, role='user', message_seq)` → pre-fix: 2 rows; post-fix: second write is a no-op via INSERT OR IGNORE

### Data flow verification for each consumed datum:

| Consumed datum | Upstream source | Reaches consumption point? |
|---|---|---|
| `msg.message_seq` (live inbound) | `BotMessage.message_seq` from WuKongIM socket | ✅ flows to `appendUser()` → `insertMessage.run()` |
| `m.message_seq` (API backfill) | `HistoricalMessage.message_seq` from `getChannelMessages()` REST API | ✅ flows to `appendUser/appendAssistant()` → `insertMessage.run()` |
| `role` column in index | Set by `append()` at `session-store.ts:274` as literal `'user'` or `'assistant'` | ✅ always one of two valid values (CHECK constraint enforces) |
| `message_seq` NULL vs non-NULL | `messageSeq ?? null` at `session-store.ts:282` — `undefined` becomes `null` | ✅ partial index correctly exempts NULL rows |

### INSERT OR IGNORE safety analysis:
The `messages` table has these constraints:
- `id INTEGER PRIMARY KEY AUTOINCREMENT` — auto-generated, never conflicts
- `role TEXT NOT NULL CHECK(role IN ('user', 'assistant'))` — programming error if violated, not suppressed
- `content TEXT NOT NULL` — always provided by callers
- `FOREIGN KEY(session_id)` — session is created before messages
- **New:** partial unique index on `(session_id, role, message_seq) WHERE message_seq IS NOT NULL`

The only constraint that fires during normal operation is the new unique index → INSERT OR IGNORE suppresses only the intended duplicates. ✅

## 4. Extra Observations

- **Performance:** The `DELETE ... WHERE id NOT IN (SELECT MIN(id) ... GROUP BY ...)` in `dedupeMessagesBySeq()` runs at every `init()`. For typical session store sizes this is negligible. On a clean DB it's a no-op (0 deletes). The `GROUP BY` subquery is well-indexed by the new unique index created immediately after.
- **Idempotency:** Running `init()` twice on a clean DB is safe — dedup deletes 0 rows, `CREATE UNIQUE INDEX IF NOT EXISTS` is a no-op. Confirmed by test at `session-store.test.ts:474-497`.

## 5. Blind-spot Checklist (R5)

- **C1 — Dual-path parity:** ✅ `appendUser()` and `appendAssistant()` both delegate to `append()` which uses the same `insertMessage` prepared statement with OR IGNORE. Symmetric. No parity gap.
- **C2 — Control-flow ordering / nesting:** ✅ `dedupeMessagesBySeq()` runs in a single transaction, called once in `init()` before the index is created. No nested/double invocation risk. The method is not called from any other site.
- **C3 — Authorization boundary ≠ capability boundary:** N/A — no authorization change.
- **C4 — Authorization lifecycle / container-member cascade:** N/A — no auth change.
- **C5 — Build ≠ runtime:** N/A — server-side SQLite, no build artifact / browser extension / CLI packaging concern.
- **C6 — Governance / policy doc self-consistency:** N/A — no documentation changes.

## 6. Cross-round Blocker Recheck (R6)

---

## [Octo-Q] verdict: APPROVE

Well-targeted data-integrity fix with correct use of SQLite partial unique index + INSERT OR IGNORE. Init ordering is correct (dedupe before index creation). Tests comprehensively cover the double-write scenario, role distinction, NULL-seq exemption, migration idempotency, and integration with init(). No P0/P1 issues found.

ploy-elison (dismissed):
## Review — seq236 dedup via partial unique index

Read the diff and then the head branch (`seq236-dedup-partial-index`) directly, tracing every `appendUser`/`appendAssistant` call site. The migration and the role-in-key design are sound, but the uniqueness contract has a blocking interaction with synthetic cron fires.

### Blocking — `message_seq = 0` (synthetic cron fires) collides under the new index

`src/cron-scheduler.ts:43` synthesizes every cron fire with `message_seq: 0`, and the doc comment is explicit that it then "runs through the entire normal pipeline" (= `onInbound`). So each fire reaches:

- `src/index.ts:665` → `store.appendUser(sessionKey, userContent, msg.message_seq /* 0 */, ...)`
- `src/index.ts:861` → `store.appendAssistant(sessionKey, fullResponse, msg.message_seq /* 0 */, botId)`

In `append()` the seq is stored verbatim: `messageSeq ?? null` keeps `0` as `0` (`session-store.ts:292`), and the partial index at `session-store.ts:157-158` only exempts `WHERE message_seq IS NOT NULL` — `0` is **not** NULL, so it is fully constrained.

Consequences:

1. **Going forward (regression):** the first cron fire in a session inserts `(sid, 'user', 0)` and `(sid, 'assistant', 0)`. Every subsequent cron fire in that same session — a different prompt at a different time — hits the constraint and is silently dropped by `INSERT OR IGNORE`. Cron-driven turns past the first vanish from history. Pre-PR (plain `INSERT`) they all accumulated as distinct rows.
2. **Migration (data loss on existing deployments):** `dedupeMessagesBySeq()` (`session-store.ts:229-230`) groups on `(session_id, role, message_seq)` for all non-NULL seqs, including `0`. On any deployment that used cron, the N accumulated seq-0 user rows and N seq-0 assistant rows collapse to the earliest one each — N−1 legitimate, distinct turns deleted per role per session. This is real data loss happening under the "seq236 dedup" banner, unannounced.

This is internally inconsistent with the codebase's own treatment of the sentinel: `src/index.ts:864-867` already documents that `message_seq = 0` is "no real wire seq" and explicitly refuses to advance the reply cursor for it. The new index does not honor that sentinel.

**Fix direction (pick one):** exclude the sentinel from the contract — e.g. `WHERE message_seq IS NOT NULL AND message_seq > 0` on both the index and the migration DELETE, or coalesce `seq <= 0` to NULL in `append()` / in `synthesizeCronMessage`. Add a regression test: two cron fires in one session must produce two user rows, and the migration must not collapse seq-0 rows.

### Non-blocking

- **Repro is store-level, not the real race.** `session-store.test.ts` proves the collapse by calling `appendUser` twice directly. It does not drive `seedHistoryFromApi` (`index.ts:1014`) + the live `appendUser` at `index.ts:665` through `handleMessage`, which is the actual XIN-145 cold-start race. The contract is well covered; the integration path that produces the duplicate is not. Worth one e2e that exercises both writers.
- **Earliest-wins can keep a backfilled placeholder over real content.** `seedHistoryFromApi` runs (`index.ts:657`) before the live `appendUser` (`index.ts:665`). If the API row for the just-arrived inbound carries empty content, the seeded row stores `resolveHistoricalMessagePlaceholder(...)` and the later real `userContent` is then dropped by `OR IGNORE`. Narrow edge (the API normally returns real content), but a `content`-divergence note in the comment would be honest.

### Confirmed good

- Migration is transactional, NULL-safe, keeps `MIN(id)`, and is genuinely idempotent; the before/after row-accounting guard is harmless (better-sqlite3 is synchronous, so it effectively never fires, but it does no harm). Order in `init()` is correct: dedupe → create index → prepare statements.
- Role-in-key correctly preserves a user turn and the bot reply that share the inbound seq (`index.ts:861` stores the assistant turn with the inbound seq), and the partial predicate correctly leaves NULL-seq turns unconstrained.
- Diff deletions are limited to the e2e seqs and `INSERT` → `INSERT OR IGNORE`; no existing safety logic or prior migration removed.
- `Part of #88` (not `Closes`), English throughout.

### Verdict: CHANGES_REQUESTED

The seq236 mechanism itself is correct, but the index/migration must stop treating `message_seq = 0` as a real key, or cron-fired history is silently dropped going forward and deleted retroactively by the migration.

src/session-store.ts
```diff
@@ -138,13 +138,40 @@ export class SessionStore {
...
+    // idempotent. Keyed on role because a user turn and the bot's reply
+    // legitimately share the inbound seq (appendAssistant is called with the
+    // inbound message_seq), so a session_id+seq-only index would wrongly forbid
+    // the reply. WHERE message_seq IS NOT NULL keeps it partial: assistant /
+    // legacy / no-seq turns carry NULL seq and stay unconstrained (NULLs are
+    // never equal to one another, but the explicit predicate documents intent and
+    // keeps the index small).
+    this.adapter.exec(
+      'CREATE UNIQUE INDEX IF NOT EXISTS idx_messages_session_role_seq ' +
+        'ON messages(session_id, role, message_seq) WHERE message_seq IS NOT NULL',
```
OctoBoooot: **[major]** the partial predicate `WHERE message_seq IS NOT NULL` covers `message_seq = 0` (`0 IS NOT NULL` is true). Cron fires hardcode `message_seq: 0` (`cron-scheduler.ts:43 synthesizeCronMessage`), and `append()`'s `messageSeq ?? null` keeps `0` (nullish coalescing doesn't null it). So two cron fires in one session both produce `(session_id, 'user', 0)` / `(session_id, 'assistant', 0)` → `INSERT OR IGNORE` silently drops the second, and `dedupeMessagesBySeq` collapses pre-existing seq=0 rows on first init(), deleting historical cron messages. Fix: give synthetic cron messages NULL seq (they have no real WuKongIM seq), or widen the predicate to `... AND message_seq > 0` and have the dedup skip seq=0 too. Credit ploy-elison.

OctoBoooot (changes requested):
## Review (follow-up): #171 — seq=0 cron collision (credit ploy-elison)

**Updated verdict:** Request changes — ploy-elison found a real data-loss path my first review missed. Upgrading.

### Major (was missed in my prior COMMENT)
- `src/session-store.ts` index + `dedupeMessagesBySeq`, interacting with `src/cron-scheduler.ts:43` — the partial index `WHERE message_seq IS NOT NULL` covers `message_seq = 0`, because `0 IS NOT NULL` is true. Every cron fire synthesizes its message with a hardcoded `message_seq: 0` (`synthesizeCronMessage`), and `append()`'s `messageSeq ?? null` preserves `0` (nullish coalescing doesn't convert it to NULL). So cron-originated rows are persisted as `(session_id, role, 0)` and ARE constrained.

  Failing sequence:
  1. A group/session fires cron task A → user+assistant rows stored at seq 0.
  2. Cron task B fires later in the same session → its rows are also `(session_id, 'user', 0)` / `(session_id, 'assistant', 0)`.
  3. `INSERT OR IGNORE` silently drops task B's rows — the second cron message and its reply never persist (history loss going forward).
  4. On a dirty DB, `dedupeMessagesBySeq` collapses ALL pre-existing seq=0 rows per (session, role) to `MIN(id)` and deletes the rest — wiping historical cron messages on the first `init()` after this ships.

  This is the inverse of the role-keying case I did verify: role keeps user/assistant distinct, but nothing keeps two *different* seq-0 messages of the same role distinct. The sentinel `0` collides with itself.

  Direction (author's call): either exclude the cron sentinel from the constraint (give synthetic cron messages a NULL `message_seq` instead of `0` — they have no real WuKongIM seq anyway), or widen the partial predicate to `WHERE message_seq IS NOT NULL AND message_seq > 0`, and make `dedupeMessagesBySeq` skip `message_seq = 0` to match. A regression test with two cron fires in one session (asserting both turns persist) would pin it.

My first pass (review 4597937129) verified the role-keyed reasoning but did not check the sentinel-`0` case against the cron path — that was the gap. ploy-elison's catch is correct. The earlier `INSERT OR IGNORE` scoping minor still stands.

Jerry-Xin (dismissed):
Flipping my earlier APPROVE to REQUEST_CHANGES. On re-review I confirmed a real data-loss bug around the `message_seq = 0` sentinel that my first pass missed. Credit to the reviewer who first raised the seq=0 concern.

## 🔴 Blocking — cron-triggered messages (message_seq = 0) collide and are lost

The partial unique index predicate is `WHERE message_seq IS NOT NULL`. In SQL, `0 IS NOT NULL` is TRUE, so **seq = 0 rows are constrained by the unique index** — they are NOT exempted the way the comment intends ("assistant / legacy / no-seq turns ... stay unconstrained").

But `synthesizeCronMessage` hard-codes a sentinel `message_seq: 0` for every synthetic cron fire (`src/cron-scheduler.ts:43`). The codebase already treats 0 as "no real wire seq" elsewhere — `src/index.ts:864-868` guards `setLastBotReplySeq` with `typeof msg.message_seq === 'number' && msg.message_seq > 0`, with the comment *"#115: a synthetic cron fire carries message_seq=0 (no real wire seq)"*.

These two paths now contradict each other:

1. **Live write loss** — every distinct cron fire in a session calls `appendUser(sessionKey, prompt, 0, ...)` → all land as `(session_id, 'user', 0)`. With `INSERT OR IGNORE` + this index, the **second and every subsequent cron prompt in the same session is silently dropped**. Two different cron tasks (or the same task firing twice) are treated as the same message and lost. This is exactly the failure mode the PR set out to prevent, but applied to legitimately-distinct messages.

2. **Migration loss** — `dedupeMessagesBySeq()` does `... id NOT IN (SELECT MIN(id) FROM messages WHERE message_seq IS NOT NULL GROUP BY session_id, role, message_seq)`. Since seq=0 passes `IS NOT NULL`, all historical seq=0 cron rows in a session collapse to a single row (earliest id wins) → **historical cron-message data loss on the one-time backfill**.

The seq=0 sentinel means "no real seq" — the same semantic class as NULL — so it must be excluded from the uniqueness contract just like NULL is. Suggested fix (both sites must agree):

- Index: `... WHERE message_seq IS NOT NULL AND message_seq > 0` (or `WHERE message_seq > 0`, since seq is non-negative).
- Dedupe: `DELETE FROM messages WHERE message_seq > 0 AND id NOT IN (SELECT MIN(id) FROM messages WHERE message_seq > 0 GROUP BY session_id, role, message_seq)`.

Either normalize 0 → NULL at the `appendUser`/`insertMessage` boundary, or scope both the index predicate and the dedupe to `message_seq > 0`. Whichever is chosen, the index predicate and the dedupe predicate MUST use the identical condition, and a regression test should cover two distinct cron fires (seq=0) in one session both surviving.

## 💬 Non-blocking

- `INSERT OR IGNORE` is broader than a targeted `ON CONFLICT(...) DO NOTHING`: it would also swallow other constraint violations on the `messages` table. There are none meaningful today, so non-blocking — but a targeted conflict clause documents intent better.

## ✅ Highlights (these parts are correct and well-done)

- Role-keyed composite index `(session_id, role, message_seq)` is the subtle-correct call: a user turn and the bot reply legitimately share the inbound seq (`appendAssistant` at index.ts:861 stores with the inbound `msg.message_seq`), so a seq-only index would wrongly collide them and the backfill would delete bot replies. Verified against the call sites.
- The migration runs `dedupeMessagesBySeq()` BEFORE creating the unique index (a unique index cannot be built over a table that already violates it), is transactional with before/after row-accounting + rollback on mismatch, and is idempotent. Good discipline.
- This is SQLite (better-sqlite3), so there is no hot-table-lock / CONCURRENTLY concern; `IF NOT EXISTS` makes index creation safe and repeatable.
- NULL-seq exemption (genuine no-seq turns) is correct; tests cover the repro, role-distinct, partial-index, and idempotency cases. The only gap is the seq=0 cron case above.

yujiawei (changes requested):
# Code Review — PR #171 (cc-channel-octo)

**Head SHA reviewed:** `332c7c45ab61f5c408f7cf7b0661ae29378bb826`

Independent review of the session-store dedupe fix. Reproduced the key behaviors against the checked-out head with standalone SQLite repros and the existing test suite (29/29 session-store tests pass, `type-check` clean). The core idea — a partial unique index plus `INSERT OR IGNORE` to make `append()` idempotent — is sound and the dedup migration SQL itself is correct. However, the uniqueness key as chosen interacts badly with one existing producer that emits a fixed sentinel `message_seq`, and that introduces both a destructive one-time migration and a silent forward regression. Requesting changes.

---

## 1. Specification compliance

The diff faithfully implements its stated goal (collapse the cold-start double-write via a partial unique index + `INSERT OR IGNORE` + an existing-row dedup migration). Nothing missing, nothing obviously extra relative to the description. The problem is not *what* was implemented but an unstated assumption baked into the uniqueness key — covered under Code Quality below.

---

## 2. Code quality

### P1 — `message_seq = 0` (cron / synthetic fires) is constrained by the new index → destructive migration + silent forward data loss

The new index is `messages(session_id, role, message_seq) WHERE message_seq IS NOT NULL`. The predicate excludes only `NULL`, **not `0`** — and the cron scheduler synthesizes every fired task with a hardcoded `message_seq: 0`:

- `src/cron-scheduler.ts:43` — `synthesizeCronMessage` sets `message_seq: 0` for every fire.
- These flow through the normal pipeline and are persisted unconditionally: `src/index.ts:665` (`appendUser(..., msg.message_seq /* =0 */, ...)`) and `src/index.ts:861` (`appendAssistant(..., msg.message_seq /* =0 */, ...)`). The `msg.message_seq > 0` guard at `src/index.ts:867` only gates the segmentation cursor — it does **not** stop the insert.
- `src/session-store.ts:292` — `insertMessage.run(..., messageSeq ?? null, ...)`; `??` does not coalesce `0`, so the row is stored with `message_seq = 0` and *is* indexed.

Two consequences, both reproduced against the head:

1. **One-time migration data loss.** `dedupeMessagesBySeq()` groups by `(session_id, role, message_seq)` and keeps only `MIN(id)` per group, ignoring `content`/`timestamp`. On any production DB that already holds cron history, every session accumulates multiple *genuinely different* rows at `(session, 'user', 0)` and `(session, 'assistant', 0)`. The startup migration deletes all but the earliest — i.e. it permanently drops distinct historical cron turns. (Repro: 6 distinct seq-0 rows → 2 survive, 4 deleted.)
2. **Silent forward regression.** Going forward, `INSERT OR IGNORE` drops the 2nd+ cron user prompt and reply per session. Two *different* cron tasks in the same channel also clobber each other at seq 0. Pre-PR, each fire wrote distinct rows.

Mitigating factor (why this is P1 and not P0): the SDK session is the source of truth for history on non-first turns (`src/index.ts:668-675`), so live model continuity is not broken on the happy path — the durable SQLite store only feeds the model on first-turn injection, stale-resume rebuild, and G4 cold-start backfill. So the damage is masked in normal operation. But it is still (a) real deletion of historical data in a migration and (b) a silent durable-store regression that surfaces exactly in the rebuild/cold-start paths this store exists to serve. "Silent" + "destructive migration" on a security-sensitive change is why this blocks.

**Suggested fix (any one):** have cron fires carry `NULL` seq instead of `0` (then they stay unconstrained, matching the design intent for "no real wire seq"); or scope the index/migration predicate to `message_seq > 0`; or use a monotonic synthetic seq for cron. The `NULL` route is the smallest and most consistent with how assistant/legacy turns are already treated.

### P2 — dedupe is content-blind: cold-start media/edited turns lose the richer copy

The key assumption "same `(session, role, seq)` ⇒ identical message" does not hold for media cold-starts. `seedHistoryFromApi` runs **before** the live append (`src/index.ts:657` then `:665`) and stores the backfill placeholder for a media trigger (`resolveHistoricalMessagePlaceholder` → bare `[图片]` / `[语音消息]`, no URL — `src/inbound.ts:530`+). The later live append for the same seq carries the richer render that includes the media URL (`src/inbound.ts:319`, `:434`, `:452`). `INSERT OR IGNORE` keeps the earlier (placeholder) row and silently drops the URL-bearing one. Same logical message, but the inferior copy wins and the media URL is lost from durable history. Worth a guard or an `ON CONFLICT … DO UPDATE` for the richer content, or ordering the live write first.

### P2 — `dedupeMessagesBySeq()` runs unconditionally on every startup

`src/session-store.ts:146` calls the migration on every `init()` with no "already clean / index present" gate. It does two `COUNT(*)` full scans plus a `DELETE … WHERE id NOT IN (SELECT MIN(id) … GROUP BY …)` correlated subquery inside a write transaction, even when nothing needs cleaning. It's idempotent (deletes 0 on a clean DB) and the table is bounded by the 7-day session TTL, so impact is limited — but gating it behind an index-presence check would remove a recurring full-table scan from the hot startup path.

### P2 — row-accounting rollback guard cannot fire under the current adapter

`src/session-store.ts:239-243` throws/rolls back if `before - after !== result.changes`, with a comment citing concurrent mutation. `better-sqlite3` is synchronous and single-connection (`src/db-adapter.ts:85-87`), so within a process there is no concurrent writer for this to catch; the assertion is effectively unreachable defensive code. Harmless, but note that *if* it ever could fire (multi-process sharing the file under WAL), it converts a benign concurrent write into a hard `init()` throw that crashes startup. Either document it as a pure invariant or reconsider crashing boot on it.

### P2 — tests under-cover the new contract (false-green risk)

- `src/__tests__/e2e.test.ts:315,322` sends two DM turns in one session at the `makeDmMsg` default `message_seq: 1`. After this PR the second `appendUser` is silently dropped by `INSERT OR IGNORE`, yet the test still passes because it only asserts on resume/mock args, never on stored history. Only the multi-turn-history test (line 604) was patched to use distinct seqs; the same latent collision exists in any multi-turn-same-session fixture. A real behavior change is masked by a green suite.
- The migration fixtures (`seedDirtyDb`) only use seqs 236/237 with *identical-content* duplicates and a hand-rolled table without the production `CHECK(role …)` / FK. They never exercise seq=0, content-divergent collisions, or the real schema — exactly the cases that break above. Adding a seq-0 / content-divergence case would have caught the P1.

---

## 3. Migration safety (security-sensitive checklist)

- Init ordering is correct: `SCHEMA → ALTER add column → dedup → CREATE UNIQUE INDEX → prepared statements` (`src/session-store.ts:117-160`). The dedup correctly precedes index creation so the unique build can't fail on pre-existing dups.
- The dedup `DELETE` is SQL-correct: `id` is `INTEGER PRIMARY KEY` so the `NOT IN (SELECT MIN(id) …)` NULL trap cannot fire; `WHERE message_seq IS NOT NULL` protects NULL-seq rows; `GROUP BY` includes `session_id` and `role`, so a user turn + bot reply sharing a seq, and the same seq across sessions, are preserved. Verified empirically.
- No new injection surface — all migration SQL is static with bound parameters.
- The one real safety gap is the **content-blind deletion of seq-0 cron rows** (P1 above): the migration is safe only under the assumption that a shared `(session, role, seq)` always means an identical message, which the cron producer violates.

---

## 4. Recommendations

1. Block on the cron `seq=0` issue: make cron/synthetic fires use `NULL` seq (preferred), or scope the index + migration to `message_seq > 0`. Re-run the migration mentally against a DB containing several distinct seq-0 rows per session before merging.
2. Make the cold-start dedupe content-aware for media (or order the live, URL-bearing write before the backfill placeholder).
3. Add tests at seq=0 and with content-divergent collisions; fix the e2e fixtures that now silently drop a turn so green means green.
4. Optional: gate `dedupeMessagesBySeq()` behind an index-presence check; downgrade or document the row-accounting guard.

---

## 5. Additional observations (out of scope — not blocking)

The `group_messages` table (`src/group-context.ts:96`, `pushMessage`) is written with a plain `INSERT` and has no seq column or idempotency key. The seq236 root cause was at-least-once / cold-start reprocessing of the same inbound; if WuKongIM can redeliver an inbound (ack loss, restart mid-handle), the same double-write class plausibly exists in this parallel store, which this PR does not touch. Worth a follow-up check, but explicitly outside the stated scope here.

**Coverage note:** I verified the SQLite/index/migration behavior directly and traced the cron, backfill, and media render paths in-repo. I did not run the full e2e suite end-to-end or benchmark startup on a realistically-sized file-backed WAL DB — the performance and WAL-checkpoint edges of the every-startup dedup are reasoned, not measured.

OctoBoooot (approved):
## Re-review (delta): #171 — seq=0 fix verified

**Verdict:** Approve. The seq=0 cron data-loss major is resolved correctly and thoroughly tested.

Delta `332c7c45ab61` → `b843c58d8c05`:

- ✓ Index predicate changed `WHERE message_seq IS NOT NULL` → `WHERE message_seq > 0`, so the seq=0 cron sentinel is no longer constrained — multiple cron rounds can share seq=0 without `INSERT OR IGNORE` dropping them.
- ✓ `dedupeMessagesBySeq` changed **consistently** — both the `DELETE ... WHERE message_seq > 0` and the inner `SELECT MIN(id) ... WHERE message_seq > 0` now use `> 0`, so the migration no longer collapses historical seq=0 cron rows. (This is the part that mattered most — index and dedup predicates must agree, and they do.)
- ✓ `> 0` matches the codebase's existing `msg.message_seq > 0` "real seq" convention (index.ts), so the sentinel handling is now uniform rather than a local special-case.
- ✓ Tests cover all three paths: live-write (4 cron rounds at seq=0 all persist), migration (seq=0 rows untouched, with row-accounting), and a post-migration fresh seq=0 round — `cronCount === 4`.

Credit ploy-elison for the original catch; the author's fix addresses both the forward-write and the migration data-loss vectors, not just one.

### Out of scope (still open, non-blocking)
- The `INSERT OR IGNORE` scoping minor from the first round (broad ignore vs targeted `ON CONFLICT ... DO NOTHING`) is unaddressed — still a minor, still non-blocking; `messages` has no other constraint `append()` would violate today.
- The assistant-same-inbound-seq question (could one inbound legitimately yield two assistant rows) remains a question, not a defect, in the current single-`fullResponse`-per-turn path.

Jerry-Xin (approved):
Re-review at head b843c58d8c05. The seq=0 cron data-loss blocker is fully fixed exactly as suggested. APPROVE.

## Fix verified (the seq=0 cron blocker)

Both predicates changed from `WHERE message_seq IS NOT NULL` to `WHERE message_seq > 0`, with the index predicate and the dedupe predicate now using the identical condition:

- Partial unique index: `CREATE UNIQUE INDEX ... ON messages(session_id, role, message_seq) WHERE message_seq > 0`.
- `dedupeMessagesBySeq()`: `DELETE FROM messages WHERE message_seq > 0 AND id NOT IN (SELECT MIN(id) FROM messages WHERE message_seq > 0 GROUP BY session_id, role, message_seq)` — both the outer DELETE and the inner MIN(id) subquery use `> 0`.

This exempts both NULL-seq turns and the seq=0 cron sentinel from the uniqueness contract, matching the codebase's existing "real seq" predicate (`msg.message_seq > 0` at index.ts). Multiple cron rounds — which all legitimately carry `message_seq: 0` from `synthesizeCronMessage` — are no longer collapsed by INSERT OR IGNORE (live) or by the migration (historical).

## Tests cover the regression

New head adds explicit seq=0 coverage:
- Live: 2 user + 2 assistant rows all at seq=0 persist (count == 4), not deduped.
- Migration: seq=0 cron rounds are NOT touched by `dedupeMessagesBySeq`, with before/after row accounting and idempotency (second pass removes nothing).
- init()-then-fresh-round: seq=0 rounds survive the migration AND a subsequent fresh seq=0 append.

Delta is scoped to `session-store.ts` + `session-store.test.ts` only — no collateral changes.

## Still-correct from the prior review (unchanged, re-confirmed)

- Role-keyed composite index is the subtle-correct call (a user turn and the bot reply legitimately share the inbound seq via `appendAssistant`), so a seq-only index would wrongly collide them.
- Migration runs dedupe BEFORE creating the unique index, is transactional with row-accounting rollback, and is idempotent.
- SQLite (better-sqlite3) → no hot-table-lock / CONCURRENTLY concern; `IF NOT EXISTS` keeps index creation safe.

## Verdict: APPROVE

The data-loss blocker is resolved correctly, both predicates agree, and there is direct regression coverage for the cron-sentinel case. Non-blocking nit carried from before: `INSERT OR IGNORE` is broader than a targeted `ON CONFLICT(...) DO NOTHING`, but there are no other meaningful constraints on `messages` today, so it is fine.

Jerry-Xin:
@clawsweeper re-review — the blocking changes have been addressed in the latest commit. PR#171: seq=0 sentinel data-loss fixed via `message_seq > 0` guard in both the partial unique index and the dedup migration (commit b843c58). PR#172: trust-poisoning fixed by switching the server-md cache to in-memory-only (no disk artifact), plus a TTL refresh. Please re-review the current head and clear the prior CHANGES_REQUESTED.

Jerry-Xin:
@clawsweeper re-review at head `b843c58d8c05`. The P1 `message_seq = 0` sentinel concern from the prior review (against `332c7c45`) is fixed in this newer commit: both the partial unique index and the dedup migration now use `WHERE message_seq > 0`, so cron/synthetic sentinel rows (seq=0) are excluded from de-duplication entirely — no destructive migration and no forward data loss. Standalone repro + suite green (1139 tests). The outstanding CHANGES_REQUESTED reviewed an older SHA; please re-review the current head and clear it.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#171 Review Report
**Head SHA:** `b843c58d8c05f114c3f55c1944d368c998b9eae1`
**Repo:** Mininglamp-OSS/cc-channel-octo

---

| Item | Status | Evidence |
|------|--------|----------|
| Root cause correctly identified (G4 backfill + live inbound double-write) | ✅ | `src/index.ts:657` (seedHistoryFromApi → appendUser) + `src/index.ts:665` (live path → appendUser) — same `(session_id, role, message_seq)` persisted twice |
| Partial unique index correctly scoped on `(session_id, role, message_seq) WHERE message_seq > 0` | ✅ | `src/session-store.ts:160-162` — keyed on role (user/assistant share seq legitimately); partial on `> 0` exempts NULL-seq and seq=0 cron sentinel |
| `INSERT OR IGNORE` makes `append()` idempotent | ✅ | `src/session-store.ts:177` — second write of same `(session_id, role, seq>0)` is a no-op |
| Migration ordering in `init()`: dedup BEFORE unique index | ✅ | `src/session-store.ts:142` calls `dedupeMessagesBySeq()` first, then `:160` creates the unique index — prevents index-creation failure on pre-existing duplicates |
| `dedupeMessagesBySeq` SQL correctness | ✅ | `src/session-store.ts:233-235` — `DELETE WHERE message_seq > 0 AND id NOT IN (SELECT MIN(id) … GROUP BY session_id, role, message_seq)` — keeps earliest per group; exempts NULL and seq=0 |
| Row accounting (防丢数据 guard) | ✅ | `src/session-store.ts:243-248` — throws + rolls back on delta mismatch |
| seq=0 cron sentinel exemption verified at source | ✅ | `src/cron-scheduler.ts:43` — `synthesizeCronMessage` sets `message_seq: 0`; `src/index.ts:867-868` — `setLastBotReplySeq` gated on `msg.message_seq > 0` (same predicate) |
| E2E test uses realistic monotonic seqs | ✅ | `src/__tests__/e2e.test.ts:604,607` — `{ message_seq: 1 }` and `{ message_seq: 2 }` — prevents false dedup in multi-turn DM test |
| Test coverage: cold-start double-write, earliest-wins, role distinction, NULL exemption, seq=0 exemption, migration idempotency, init() integration | ✅ | `src/__tests__/session-store.test.ts` — 7 new test cases covering all paths |

## 2. Findings

### F1 — Stale comment in test describe block (P2 / nit)

**Diff-scope:** pre-existing (carried over from commit 1, not updated by commit 2).

**Location:** `src/__tests__/session-store.test.ts:338-342` (the describe-block comment above the new test suite).

The comment states:
> "The fix is a PARTIAL unique index on (session_id, role, message_seq) WHERE message_seq IS NOT NULL + INSERT OR IGNORE."

But the actual predicate (updated in commit `b843c58`) is `message_seq > 0`, not `IS NOT NULL`. The individual test bodies correctly test `> 0` behavior (e.g. the seq=0 cron sentinel test), so this is a comment-only mismatch — no functional impact.

**Suggestion:** Update the describe-block comment to say `WHERE message_seq > 0` to match the implementation.

### No P0 / P1 findings.

## 3. Data Flow Tracing

### 3.1 Cold-start double-write path (root cause)

1. **Path A (G4 backfill):** `index.ts:657` → `seedHistoryFromApi()` → iterates `apiMessages` → `store.appendUser(sessionKey, content, m.message_seq, …)` (line 1040) or `store.appendAssistant(…)` (line 1038). `m.message_seq` comes from the WuKongIM sync API response (positive integer for real messages).
2. **Path B (live inbound):** `index.ts:665` → `store.appendUser(sessionKey, userContent, msg.message_seq, …)`. `msg.message_seq` comes from the WebSocket inbound `BotMessage`.
3. **Convergence:** Both call `append()` → `insertMessage.run(sessionId, role, content, now, messageSeq ?? null, safeName)` (`session-store.ts:312`). With the new `INSERT OR IGNORE`, the second write of the same `(session_id, role, message_seq>0)` is silently dropped. ✅

### 3.2 Cron fire path

1. `cron-scheduler.ts:43` — `synthesizeCronMessage()` constructs a `BotMessage` with `message_seq: 0`.
2. This flows through the same `handleMessage` → `appendUser(…, 0, …)` → `append()` → `insertMessage.run(…, 0, …)`.
3. The partial unique index predicate is `message_seq > 0`, so seq=0 is **not indexed** — `INSERT OR IGNORE` never suppresses it. Multiple cron rounds all persist. ✅

### 3.3 Assistant reply sharing inbound seq

1. `index.ts:861` — `store.appendAssistant(sessionKey, fullResponse, msg.message_seq, botId)` — stores the bot reply with the **same** `message_seq` as the inbound user message.
2. The unique index includes `role` in the key, so `(session_id, 'user', 236)` and `(session_id, 'assistant', 236)` are distinct keys. Both persist. ✅

### 3.4 Migration on dirty DB (init path)

1. `init()` → `dedupeMessagesBySeq()` — transaction: COUNT before → DELETE duplicates → COUNT after → verify delta → commit or rollback.
2. After dedup: `CREATE UNIQUE INDEX IF NOT EXISTS …` — succeeds because no violations remain.
3. Subsequent `INSERT OR IGNORE` keeps the table clean. ✅
4. Idempotent: second `init()` → dedup finds nothing → index already exists (IF NOT EXISTS) → no-op. ✅

### 3.5 Legacy DB (no message_seq column)

1. `init()` → column migration adds `message_seq INTEGER` (existing rows get NULL).
2. `dedupeMessagesBySeq()` — `WHERE message_seq > 0` matches nothing → no deletes.
3. Unique index created over empty set → succeeds.
4. Future inserts with seq>0 are deduped; NULL-seq legacy rows remain unconstrained. ✅

## 4. Blind-point Checklist (R5 — security_sensitive PR)

- **C1 — Dual-path parity:** N/A. This PR modifies a single write path (messages table append). There are no symmetric add/remove or create/delete pairs. The two callers (seedHistoryFromApi + live inbound) converge on the same `append()` — now correctly idempotent.
- **C2 — Control-flow ordering / nested reuse + safety-control non-canonical probe:** CLEAR. `INSERT OR IGNORE` is called from exactly two paths (`appendUser` / `appendAssistant`). Both pass `messageSeq ?? null`. The unique index is the safety control — tested with: seq=0 (sentinel), NULL (legacy), positive seq (normal), same seq across roles (user+assistant at seq 236). All behave correctly. The `dedupeMessagesBySeq` DELETE uses `WHERE message_seq > 0 AND id NOT IN (SELECT MIN(id) … WHERE message_seq > 0 GROUP BY …)` — the inner WHERE correctly excludes NULL/0 rows from the MIN(id) computation, preventing accidental deletion of non-duplicate rows.
- **C3 — Authorization boundary ≠ capability boundary:** N/A. No authorization changes.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No authorization changes.
- **C5 — Build/note ≠ runtime path:** N/A. No build artifacts, extensions, or CLI packaging changes. SQLite DDL is exercised at runtime via `init()`, and the test suite verifies the runtime behavior directly.
- **C6 — Governance/policy/security-doc self-consistency:** N/A. No documentation or policy changes.

## 5. Cross-round Blocker Recheck (R6)

## 6. Additional Observations

- The `INSERT OR IGNORE` approach is well-suited for SQLite's conflict resolution — it's atomic and requires no SELECT-before-INSERT round-trip.
- The `dedupeMessagesBySeq` row-accounting guard (throw on delta mismatch) is a strong safety net against concurrent mutation during migration. With better-sqlite3's serialized model this is unlikely to fire, but it's the right defense.
- The `message_seq > 0` predicate alignment with `index.ts:867-868` (`setLastBotReplySeq` guard) is intentional and well-documented.

---

[Octo-Q] verdict: **APPROVE**

Rationale: No P0/P1 findings. The fix correctly addresses the seq236 double-write root cause with defense-in-depth (partial unique index + INSERT OR IGNORE + transactional migration). The seq=0 cron sentinel data-loss risk from commit 1 is properly addressed in commit 2. Test coverage is comprehensive. One P2 stale-comment nit (F1) is non-blocking.

yujiawei (approved):
# Code Review — PR #171 (cc-channel-octo)

**Head SHA reviewed:** `b843c58d8c05f114c3f55c1944d368c998b9eae1`

Independent re-review at the current head. Reproduced the behavior against the checked-out branch: full suite green (1173 passed), `type-check` clean. This was reviewed at migration-tier rigor because it ships a schema migration that **deletes rows** (`dedupeMessagesBySeq`), and I re-verified the load-bearing design claims at the call sites rather than trusting the PR body.

## 1. Specification compliance

**Spec: ✅**

The diff faithfully implements its stated goal and nothing more:

- **Missing:** none — partial unique index, `INSERT OR IGNORE`, and the existing-row dedup migration are all present.
- **Extra:** none — scope is confined to `session-store.ts` + its two test files; no changes to server-md / events / write-back.
- **Deviation:** none — the e2e fixture change (`e2e.test.ts`, monotonic seqs `1`/`2` instead of the shared default `message_seq: 1`) is honest: the new uniqueness contract would have collapsed the old fixture's repeated-default-seq turns, and the author corrected the fixture to reflect real WuKongIM monotonic-seq behavior rather than weakening the constraint to fit the test.

## 2. Code quality

**Quality: Approved**

### The prior blocker is resolved

An earlier round flagged a P1: the index/migration predicate `WHERE message_seq IS NOT NULL` also constrained the **seq=0 cron sentinel** (`cron-scheduler.ts:43` synthesizes every fire with `message_seq: 0`), which would have (a) let `INSERT OR IGNORE` silently drop the 2nd+ cron turn per session going forward, and (b) made the one-time migration fold distinct seq=0 cron rounds and delete real history.

This head fixes it completely:

- `session-store.ts:160` — index predicate is now `WHERE message_seq > 0`.
- `session-store.ts:238-239` — the migration `DELETE` also scopes to `message_seq > 0`, so seq=0 (and NULL) rows are exempt from folding.
- `> 0` matches the codebase's existing "real wire seq" predicate (`index.ts:867`).
- Regression tests added: `session-store.test.ts:422` (multiple seq=0 rounds all persist) and `:481-594` (migration preserves all seq=0 cron rows + a fresh seq=0 round appends fine post-`init()`).

### Design verified at call sites

- **Role-in-key is necessary, not incidental.** `appendAssistant` (`index.ts:861`) is called with the **inbound** `msg.message_seq` — the same seq as the user turn at `index.ts:665`. A `(session_id, message_seq)`-only index would collide a user message with its own reply and the migration would delete bot replies sharing a user seq. Keying on `role` is the correct, subtle call.
- **Uniqueness invariant holds today.** `session_id` is per-channel (`session-router.ts:222` — group key is `channel_id`), WuKongIM assigns `message_seq` monotonically per channel, and the per-session lock + single `queryAgent` generator guarantee exactly one assistant row per inbound. So `(session_id, role, message_seq>0)` is genuinely unique for distinct legitimate messages.
- **Migration is data-safe.** Transactional, keeps `MIN(id)` (earliest), exempts NULL/seq≤0, and the before/after row-accounting throws + rolls back on a delta mismatch. Ordering in `init()` is correct: dedupe → create unique index → prepare statements.

### Non-blocking findings (P2)

- **P2 — dedupe is content-blind on a media cold-start.** `seedHistoryFromApi` runs (`index.ts:657`) *before* the live `appendUser` (`index.ts:665`). If the backfill API row for the just-arrived inbound has empty content, the seeded row stores a placeholder (`resolveHistoricalMessagePlaceholder`, e.g. `[图片]`, no URL); the later live append carries the richer render, and `INSERT OR IGNORE` / the migration's earliest-wins keeps the placeholder and drops the richer copy. Narrow (only the first turn of a group cold start, only when the API returns empty content for the trigger), and the turn is degraded rather than lost — but a `content`-divergence note, or ordering the live write first / `ON CONFLICT … DO UPDATE` for richer content, would close it. Not a merge blocker.
- **P2 — `INSERT OR IGNORE` suppresses all row constraint violations, not just this index.** Today `messages` has only the `role` CHECK plus this index, so the blast radius is limited, but a future `NOT NULL`/`CHECK` on `messages` would become a silent no-op. A targeted `ON CONFLICT(session_id, role, message_seq) WHERE message_seq > 0 DO NOTHING` would scope the ignore precisely. Matches existing style; non-blocking.
- **P2 — `dedupeMessagesBySeq()` scans the full table on every `init()`.** Idempotent and a no-op on clean data, but it's an O(n) `DELETE … WHERE id NOT IN (SELECT MIN(id) … GROUP BY …)` at every startup. On a large historical DB, gating it behind a one-time migration marker would let steady-state startups skip the scan. Non-blocking for correctness.

### Security note (security-sensitive PR)

The `append()` path preserves the pre-existing `sanitizeDisplayName` / `escapeRoleLabels` guards against turn-label forgery via user-controlled `from_name` and content — this PR does not weaken them. The one thing worth a human's eye: the correctness of the whole scheme rests on the *unstated* WuKongIM guarantee that `message_seq` is per-channel monotonic and never replayed. That holds for today's producers, but it's an implicit contract worth keeping in mind if a future producer emits reused positive seqs.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The blocking data-loss issue (seq=0 sentinel) is fully fixed at this head; remaining findings are P2 quality suggestions that do not block merge.

PR #172, opened 2026-06-30
feat(group-md): server-first GROUP.md fetch with cache and local fallback

boris-clark:
Part of #88

## P2-A: server GROUP.md fetch + cache + feature flag

Restores the server GROUP.md API removed at fork time and wires it into the per-group instruction injection, server-first with a never-lose local-file fallback, behind a feature flag. PUT write-back triggering and event-driven refresh are out of scope here (separate work items).

### Changes
- **`src/octo/api.ts`** — restore `getGroupMd` (`GET /v1/bot/groups/{groupNo}/md` → `{content, version, updated_at, updated_by}`) routed through the shared `getJson` client so it reuses the same timeout, `Bearer` auth, int64-safe JSON parse and error format as the rest of the surface. `updateGroupMd` (`PUT`, body `{content}` → `{version}`) is restored as a client function **only** — it is not connected to any write-back path.
- **`src/group-md-cache.ts`** (new) — two-tier cache keyed by `groupNo`: in-memory `Map` + durable on-disk copy (`<groupNo>.md` for content, `<groupNo>.meta.json` for version/updated_at/updated_by). Independent of the group-context caches. Path-safe `groupNo` guard; `invalidate()` hook for the future event-driven refresh.
- **`src/group-md.ts`** (new) — `resolveGroupInstructions`: server-first with a never-lose fallback to the local `groupConfigDir` file on any failure (404 / network / timeout / empty content). Threads resolve their **parent** `groupNo` (`extractParentGroupNo`) for the server call + cache key, while the local fallback keeps the full `channelId` routing intact.
- **`src/config.ts`** — `serverMd` feature flag (default **off** → pure local-file behavior; flip off to roll back).
- **`src/index.ts`** — `await resolveGroupInstructions(...)` in the group path; one cache instance per bot under `<dataDir>/group-md-cache`.
- **`config.example.json`** — document the `serverMd` flag.

### Testing
- Baseline before: 58 files / 1132 tests green. After: **60 files / 1153 tests green** (+21 new), no regressions.
- New tests: `group-md-api` (GET/PUT path, method, Bearer auth, body, parse, non-2xx errors) and `group-md` (cache memory+disk persistence/invalidate/path-safety; resolver flag gating, server-first preference, 404/empty/network/no-cache fallback, cache reuse, thread parent-group routing).
- Verified the existing local-file injection, thread routing, and group-instruction e2e paths stay green (no degradation).
- `type-check`, `lint` (`--max-warnings 0`) and `build` all clean.

### DIFF
--- a/src/group-md-cache.ts
+++ b/src/group-md-cache.ts
@@ -1,0 +1,159 @@
+/**
+ * GROUP.md server cache (P2-A).
+ *
+ * A two-tier cache for server-fetched GROUP.md, keyed by groupNo:
+ *   - in-memory Map (hot path: read on every group turn),
+ *   - on-disk durable copy (`<groupNo>.md` for content + `<groupNo>.meta.json`
+ *     for version / updated_at / updated_by) so a fetched GROUP.md survives a
+ *     restart and can serve as a fallback when the server is unreachable.
+ *
+ * This module is deliberately INDEPENDENT of the P1 GroupContext caches
+ * (group_messages / group_members in SQLite): those hold per-channel chat
+ * history + roster; this holds operator-authored instruction text. Keeping them
+ * decoupled means a GROUP.md refresh never touches the message DB and vice
+ * versa.
+ *
+ * Freshness: this class only stores and serves; it does not expire entries on a
+ * timer. A populated entry is returned until `invalidate()` is called — the
+ * event-driven refresh that decides WHEN to invalidate is a separate work item.
+ * Holding a stable value between invalidations is also what keeps the agent's
+ * cached system prompt byte-identical turn-to-turn (see CLAUDE.md "Frozen system
+ * prompt").
+ *
+ * Never throws — disk I/O failures degrade to "memory only" (or "no cache"),
+ * never failing the turn.
+ */
+
+import {
+  existsSync,
+  mkdirSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { join } from 'node:path';
+
+/** A cached GROUP.md plus the server metadata that came with it. */
+export interface GroupMdEntry {
+  content: string;
+  version: number;
+  updated_at: string | null;
+  updated_by?: string;
+}
+
+/**
+ * Only allow groupNos that are safe as a single path segment, so a crafted id
+ * can never escape the cache dir. Same policy as group-config.ts isSafeId.
+ */
+function isSafeGroupNo(groupNo: string): boolean {
+  return /^[a-zA-Z0-9._-]+$/.test(groupNo) && groupNo !== '.' && groupNo !== '..';
+}
+
+export class GroupMdCache {
+  private readonly mem = new Map<string, GroupMdEntry>();
+  private readonly cacheDir?: string;
+  /** Set once we've successfully created cacheDir (lazy: only when first written). */
+  private dirReady = false;
+
+  /**
+   * @param cacheDir directory for the durable copy. When undefined the cache is
+   *   memory-only (still correct, just not restart-durable).
+   */
+  constructor(cacheDir?: string) {
+    this.cacheDir = cacheDir;
+  }
+
+  /**
+   * Read a cached entry. Checks memory first, then the on-disk copy (hydrating
+   * memory on a disk hit so the next read is hot). Returns undefined on a miss
+   * or for an unsafe groupNo.
+   */
+  get(groupNo: string): GroupMdEntry | undefined {
+    if (!isSafeGroupNo(groupNo)) return undefined;
+    const hot = this.mem.get(groupNo);
+    if (hot) return hot;
+    const fromDisk = this.readDisk(groupNo);
+    if (fromDisk) this.mem.set(groupNo, fromDisk);
+    return fromDisk;
+  }
+
+  /** Store an entry in memory and (best-effort) on disk. */
+  set(groupNo: string, entry: GroupMdEntry): void {
+    if (!isSafeGroupNo(groupNo)) return;
+    this.mem.set(groupNo, entry);
+    this.writeDisk(groupNo, entry);
+  }
+
+  /**
+   * Drop a cached entry from memory and disk. The hook the event-driven refresh
+   * (separate work item) calls when the server reports a GROUP.md changed, so
+   * the next turn re-fetches.
+   */
+  invalidate(groupNo: string): void {
+    if (!isSafeGroupNo(groupNo)) return;
+    this.mem.delete(groupNo);
+    if (!this.cacheDir) return;
+    for (const name of [`${groupNo}.md`, `${groupNo}.meta.json`]) {
+      try {
+        rmSync(join(this.cacheDir, name), { force: true });
+      } catch (err) {
+        console.error(`[cc-channel-octo] group-md-cache: invalidate ${name} failed: ${String(err)}`);
+      }
+    }
+  }
+
+  private ensureDir(): boolean {
+    if (!this.cacheDir) return false;
+    if (this.dirReady) return true;
+    try {
+      mkdirSync(this.cacheDir, { recursive: true });
+      this.dirReady = true;
+      return true;
+    } catch (err) {
+      console.error(`[cc-channel-octo] group-md-cache: mkdir ${this.cacheDir} failed: ${String(err)}`);
+      return false;
+    }
+  }
+
+  private readDisk(groupNo: string): GroupMdEntry | undefined {
+    if (!this.cacheDir) return undefined;
+    const contentPath = join(this.cacheDir, `${groupNo}.md`);
+    const metaPath = join(this.cacheDir, `${groupNo}.meta.json`);
+    try {
+      if (!existsSync(contentPath)) return undefined;
+      const content = readFileSync(contentPath, 'utf-8');
+      let version = 0;
+      let updatedAt: string | null = null;
+      let updatedBy: string | undefined;
+      if (existsSync(metaPath)) {
+        try {
+          const meta = JSON.parse(readFileSync(metaPath, 'utf-8')) as Partial<GroupMdEntry>;
+          if (typeof meta.version === 'number') version = meta.version;
+          if (typeof meta.updated_at === 'string' || meta.updated_at === null) updatedAt = meta.updated_at ?? null;
+          if (typeof meta.updated_by === 'string') updatedBy = meta.updated_by;
+        } catch {
+          // Corrupt meta — keep the content, default the metadata.
+        }
+      }
+      return { content, version, updated_at: updatedAt, updated_by: updatedBy };
+    } catch (err) {
+      console.error(`[cc-channel-octo] group-md-cache: read ${groupNo} failed: ${String(err)}`);
+      return undefined;
+    }
+  }
+
+  private writeDisk(groupNo: string, entry: GroupMdEntry): void {
+    if (!this.ensureDir() || !this.cacheDir) return;
+    try {
+      writeFileSync(join(this.cacheDir, `${groupNo}.md`), entry.content, 'utf-8');
+      const meta = {
+        version: entry.version,
+        updated_at: entry.updated_at,
+        updated_by: entry.updated_by,
+      };
+      writeFileSync(join(this.cacheDir, `${groupNo}.meta.json`), JSON.stringify(meta), 'utf-8');
+    } catch (err) {
+      console.error(`[cc-channel-octo] group-md-cache: write ${groupNo} failed: ${String(err)}`);
+    }
+  }
+}
--- a/src/group-md-cache.ts
+++ b/src/group-md-cache.ts
@@ -1,37 +1,39 @@
 /**
- * GROUP.md server cache (P2-A).
+ * GROUP.md server cache (P2-A) — IN-MEMORY ONLY, with a TTL.
  *
- * A two-tier cache for server-fetched GROUP.md, keyed by groupNo:
- *   - in-memory Map (hot path: read on every group turn),
- *   - on-disk durable copy (`<groupNo>.md` for content + `<groupNo>.meta.json`
- *     for version / updated_at / updated_by) so a fetched GROUP.md survives a
- *     restart and can serve as a fallback when the server is unreachable.
+ * Holds server-fetched GROUP.md keyed by groupNo, in a process-local Map. There
+ * is deliberately **no on-disk persistence** (see SECURITY below).
  *
- * This module is deliberately INDEPENDENT of the P1 GroupContext caches
- * (group_messages / group_members in SQLite): those hold per-channel chat
- * history + roster; this holds operator-authored instruction text. Keeping them
- * decoupled means a GROUP.md refresh never touches the message DB and vice
- * versa.
+ * Independent of the P1 GroupContext caches (group_messages / group_members in
+ * SQLite): those hold per-channel chat history + roster; this holds operator-
+ * authored instruction text. Keeping them decoupled means a GROUP.md refresh
+ * never touches the message DB and vice versa.
  *
- * Freshness: this class only stores and serves; it does not expire entries on a
- * timer. A populated entry is returned until `invalidate()` is called — the
- * event-driven refresh that decides WHEN to invalidate is a separate work item.
- * Holding a stable value between invalidations is also what keeps the agent's
- * cached system prompt byte-identical turn-to-turn (see CLAUDE.md "Frozen system
- * prompt").
+ * SECURITY — why memory-only (review #172, 🔴 durable-cache poisoning):
+ *   The resolved GROUP.md is injected into the agent's system prompt as a TRUSTED
+ *   instruction block (same channel as the operator's local `groupConfigDir`
+ *   file). A local file earns that trust from OS permissions (operator-owned,
+ *   non-writable by others; `group-config.ts` additionally refuses a group/world-
+ *   writable file). A DISK cache cannot earn the same trust here: the gateway
+ *   runs the agent under `bypassPermissions` with `Bash`/`Write`, so the agent
+ *   process — drivable by untrusted chat input — can write any file the gateway
+ *   user owns, including a cache file. A persisted cache entry would then be read
+ *   back and injected as trusted across restarts (a chat-injection → trusted-
+ *   prompt poisoning that survives reboot), and the `0o022` perm check is useless
+ *   because agent == gateway user. So GROUP.md from the server is cached ONLY in
+ *   memory: the sole path content can enter the trusted channel is a live,
+ *   authenticated `getGroupMd` over the bot token against the SSRF-validated
+ *   `apiUrl` (server-side bot_admin-gated) — never a forgeable on-disk artifact.
+ *   The tradeoff (no cross-restart durability) is acceptable: a cold start simply
+ *   re-fetches, and the local file remains the offline fallback.
  *
- * Never throws — disk I/O failures degrade to "memory only" (or "no cache"),
- * never failing the turn.
+ * Freshness: an entry expires `ttlMs` after it was stored; an expired read is a
+ * miss (the resolver re-fetches). This is a staleness BACKSTOP so a server-side
+ * GROUP.md edit eventually takes effect (review #172, 🟡) — Stage 2 (item B) adds
+ * event-driven invalidation on top; the two do not conflict.
+ *
+ * Never throws.
  */
-
-import {
-  existsSync,
-  mkdirSync,
-  readFileSync,
-  rmSync,
-  writeFileSync,
-} from 'node:fs';
-import { join } from 'node:path';
 
 /** A cached GROUP.md plus the server metadata that came with it. */
 export interface GroupMdEntry {
@@ -41,119 +43,66 @@ export interface GroupMdEntry {
   updated_by?: string;
 }
 
+/** Default staleness backstop: re-fetch a cached GROUP.md at most this old. */
+export const DEFAULT_GROUP_MD_TTL_MS = 5 * 60 * 1000; // 5 minutes
+
 /**
- * Only allow groupNos that are safe as a single path segment, so a crafted id
- * can never escape the cache dir. Same policy as group-config.ts isSafeId.
+ * Only allow groupNos that are safe as a single Map key / log token. Mirrors
+ * group-config.ts isSafeId — cheap defense-in-depth against a crafted id even
+ * though nothing here touches the filesystem anymore.
  */
 function isSafeGroupNo(groupNo: string): boolean {
   return /^[a-zA-Z0-9._-]+$/.test(groupNo) && groupNo !== '.' && groupNo !== '..';
 }
 
+interface StoredEntry {
+  entry: GroupMdEntry;
+  storedAt: number;
+}
+
 export class GroupMdCache {
-  private readonly mem = new Map<string, GroupMdEntry>();
-  private readonly cacheDir?: string;
-  /** Set once we've successfully created cacheDir (lazy: only when first written). */
-  private dirReady = false;
+  private readonly mem = new Map<string, StoredEntry>();
+  private readonly ttlMs: number;
+  private readonly now: () => number;
 
   /**
-   * @param cacheDir directory for the durable copy. When undefined the cache is
-   *   memory-only (still correct, just not restart-durable).
+   * @param ttlMs staleness backstop in ms (entry expires this long after it was
+   *   stored). Defaults to {@link DEFAULT_GROUP_MD_TTL_MS}. A non-positive value
+   *   disables expiry (entries live until invalidate()).
+   * @param now injectable clock (testing); defaults to Date.now.
    */
-  constructor(cacheDir?: string) {
-    this.cacheDir = cacheDir;
+  constructor(ttlMs: number = DEFAULT_GROUP_MD_TTL_MS, now: () => number = () => Date.now()) {
+    this.ttlMs = ttlMs;
+    this.now = now;
   }
 
   /**
-   * Read a cached entry. Checks memory first, then the on-disk copy (hydrating
-   * memory on a disk hit so the next read is hot). Returns undefined on a miss
-   * or for an unsafe groupNo.
+   * Read a cached entry from memory. Returns undefined on a miss, an expired
+   * entry (which is also evicted), or an unsafe groupNo.
    */
   get(groupNo: string): GroupMdEntry | undefined {
     if (!isSafeGroupNo(groupNo)) return undefined;
-    const hot = this.mem.get(groupNo);
-    if (hot) return hot;
-    const fromDisk = this.readDisk(groupNo);
-    if (fromDisk) this.mem.set(groupNo, fromDisk);
-    return fromDisk;
+    const stored = this.mem.get(groupNo);
+    if (!stored) return undefined;
+    if (this.ttlMs > 0 && this.now() - stored.storedAt >= this.ttlMs) {
+      this.mem.delete(groupNo);
+      return undefined;
+    }
+    return stored.entry;
   }
 
-  /** Store an entry in memory and (best-effort) on disk. */
+  /** Store an entry in memory, stamping it for TTL expiry. */
   set(groupNo: string, entry: GroupMdEntry): void {
     if (!isSafeGroupNo(groupNo)) return;
-    this.mem.set(groupNo, entry);
-    this.writeDisk(groupNo, entry);
+    this.mem.set(groupNo, { entry, storedAt: this.now() });
   }
 
   /**
-   * Drop a cached entry from memory and disk. The hook the event-driven refresh
-   * (separate work item) calls when the server reports a GROUP.md changed, so
-   * the next turn re-fetches.
+   * Drop a cached entry. The hook the event-driven refresh (item B) calls when
+   * the server reports a GROUP.md changed, so the next turn re-fetches.
    */
   invalidate(groupNo: string): void {
     if (!isSafeGroupNo(groupNo)) return;
     this.mem.delete(groupNo);
-    if (!this.cacheDir) return;
-    for (const name of [`${groupNo}.md`, `${groupNo}.meta.json`]) {
-      try {
-        rmSync(join(this.cacheDir, name), { force: true });
-      } catch (err) {
-        console.error(`[cc-channel-octo] group-md-cache: invalidate ${name} failed: ${String(err)}`);
-      }
-    }
-  }
-
-  private ensureDir(): boolean {
-    if (!this.cacheDir) return false;
-    if (this.dirReady) return true;
-    try {
-      mkdirSync(this.cacheDir, { recursive: true });
-      this.dirReady = true;
-      return true;
-    } catch (err) {
-      console.error(`[cc-channel-octo] group-md-cache: mkdir ${this.cacheDir} failed: ${String(err)}`);
-      return false;
-    }
-  }
-
-  private readDisk(groupNo: string): GroupMdEntry | undefined {
-    if (!this.cacheDir) return undefined;
-    const contentPath = join(this.cacheDir, `${groupNo}.md`);
-    const metaPath = join(this.cacheDir, `${groupNo}.meta.json`);
-    try {
-      if (!existsSync(contentPath)) return undefined;
-      const content = readFileSync(contentPath, 'utf-8');
-      let version = 0;
-      let updatedAt: string | null = null;
-      let updatedBy: string | undefined;
-      if (existsSync(metaPath)) {
-        try {
-          const meta = JSON.parse(readFileSync(metaPath, 'utf-8')) as Partial<GroupMdEntry>;
-          if (typeof meta.version === 'number') version = meta.version;
-          if (typeof meta.updated_at === 'string' || meta.updated_at === null) updatedAt = meta.updated_at ?? null;
-          if (typeof meta.updated_by === 'string') updatedBy = meta.updated_by;
-        } catch {
-          // Corrupt meta — keep the content, default the metadata.
-        }
-      }
-      return { content, version, updated_at: updatedAt, updated_by: updatedBy };
-    } catch (err) {
-      console.error(`[cc-channel-octo] group-md-cache: read ${groupNo} failed: ${String(err)}`);
-      return undefined;
-    }
-  }
-
-  private writeDisk(groupNo: string, entry: GroupMdEntry): void {
-    if (!this.ensureDir() || !this.cacheDir) return;
-    try {
-      writeFileSync(join(this.cacheDir, `${groupNo}.md`), entry.content, 'utf-8');
-      const meta = {
-        version: entry.version,
-        updated_at: entry.updated_at,
-        updated_by: entry.updated_by,
-      };
-      writeFileSync(join(this.cacheDir, `${groupNo}.meta.json`), JSON.stringify(meta), 'utf-8');
-    } catch (err) {
-      console.error(`[cc-channel-octo] group-md-cache: write ${groupNo} failed: ${String(err)}`);
-    }
   }
 }
--- a/src/group-md.ts
+++ b/src/group-md.ts
@@ -1,0 +1,110 @@
+/**
+ * GROUP.md resolution (P2-A): server-first fetch with a never-lose local-file
+ * fallback, behind a feature flag.
+ *
+ * `resolveGroupInstructions` is the single entry point the message pipeline
+ * calls to obtain the trusted per-group instruction block injected into the
+ * agent's system prompt. Resolution order:
+ *
+ *   1. Feature flag OFF (default) or no cache wired → pure local file
+ *      (`loadGroupConfig`). Byte-identical to the pre-P2 behavior, so existing
+ *      local-file deployments are unaffected.
+ *   2. Feature flag ON:
+ *        a. serve a cached server GROUP.md if present (keyed by the PARENT group
+ *           number — a thread shares its parent group's GROUP.md);
+ *        b. otherwise fetch from the server (server-first). On success with
+ *           non-empty content, cache it and use it;
+ *        c. on ANY failure (404 "no GROUP.md", network, timeout, empty content)
+ *           fall back to the local file. The local-file path is never lost.
+ *
+ * Thread routing (P1): a thread channelId is the composite `<groupNo>____<shortId>`.
+ * The server GROUP.md endpoint is keyed by the parent group number, so we resolve
+ * it with `extractParentGroupNo` for the API call + cache key (identity for a
+ * plain group). The local-file fallback still receives the FULL channelId so its
+ * own thread/short-id routing in `loadGroupConfig` is preserved unchanged.
+ *
+ * Never throws — a server error degrades to local, a local miss degrades to
+ * "no custom instructions".
+ */
+
+import { getGroupMd } from './octo/api.js';
+import { extractParentGroupNo } from './octo/channel-id.js';
+import { loadGroupConfig, MAX_GROUP_CONFIG_BYTES } from './group-config.js';
+import type { GroupMdCache, GroupMdEntry } from './group-md-cache.js';
+
+/**
+ * Trim and byte-bound server-provided GROUP.md the same way loadGroupConfig
+ * bounds a local file, so an oversized server payload can't blow the prompt
+ * budget. Returns undefined for empty/whitespace-only content.
+ */
+function boundInstructions(content: string): string | undefined {
+  let text = content;
+  if (Buffer.byteLength(text, 'utf-8') > MAX_GROUP_CONFIG_BYTES) {
+    const buf = Buffer.from(text, 'utf-8').subarray(0, MAX_GROUP_CONFIG_BYTES);
+    // The slice may end mid-codepoint; trim a trailing replacement char.
+    text = buf.toString('utf-8').replace(/�+$/, '') + '\n[… group config truncated]';
+  }
+  const trimmed = text.trim();
+  return trimmed.length > 0 ? trimmed : undefined;
+}
+
+export interface ResolveGroupInstructionsParams {
+  /** Operator local-instruction directory (the existing fallback source). */
+  groupConfigDir?: string;
+  /** Feature flag: when false/undefined, server fetch is skipped entirely. */
+  serverMd?: boolean;
+  apiUrl: string;
+  botToken: string;
+  /** Full channelId (may be a `<groupNo>____<shortId>` thread composite). */
+  channelId: string;
+  /** Server GROUP.md cache. Omitted (or flag off) → pure local file. */
+  cache?: GroupMdCache;
+  signal?: AbortSignal;
+}
+
+export async function resolveGroupInstructions(
+  params: ResolveGroupInstructionsParams,
+): Promise<string | undefined> {
+  const { groupConfigDir, serverMd, apiUrl, botToken, channelId, cache, signal } = params;
+
+  const local = (): string | undefined => loadGroupConfig(groupConfigDir, channelId);
+
+  // Flag off (or no cache to dedupe fetches) → pure local file, unchanged behavior.
+  if (!serverMd || !cache) return local();
+
+  const groupNo = extractParentGroupNo(channelId);
+
+  // a) cached server GROUP.md wins (kept stable until invalidated by the
+  //    event-driven refresh — a separate work item).
+  const cached = cache.get(groupNo);
+  if (cached) {
+    const bounded = boundInstructions(cached.content);
+    if (bounded) return bounded;
+    // Cached-but-empty (shouldn't normally happen) → fall through to local.
+    return local();
+  }
+
+  // b) server-first fetch.
+  try {
+    const md = await getGroupMd({ apiUrl, botToken, groupNo, signal });
+    const bounded = boundInstructions(md?.content ?? '');
+    if (bounded) {
+      const entry: GroupMdEntry = {
+        content: md.content,
+        version: typeof md.version === 'number' ? md.version : 0,
+        updated_at: md.updated_at ?? null,
+        updated_by: md.updated_by,
+      };
+      cache.set(groupNo, entry);
+      return bounded;
+    }
+    // Server reachable but no/empty GROUP.md → local fallback.
+    return local();
+  } catch (err) {
+    // c) 404 / network / timeout — never-lose local fallback.
+    console.error(
+      `[cc-channel-octo] group-md: server fetch for ${groupNo} failed, falling back to local: ${String(err)}`,
+    );
+    return local();
+  }
+}
--- a/src/group-md.ts
+++ b/src/group-md.ts
@@ -10,12 +10,22 @@
  *      (`loadGroupConfig`). Byte-identical to the pre-P2 behavior, so existing
  *      local-file deployments are unaffected.
  *   2. Feature flag ON:
- *        a. serve a cached server GROUP.md if present (keyed by the PARENT group
- *           number — a thread shares its parent group's GROUP.md);
+ *        a. serve a cached server GROUP.md if present and not past its TTL (keyed
+ *           by the PARENT group number — a thread shares its parent group's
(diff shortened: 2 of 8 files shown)


src/group-md.ts
```diff
@@ -0,0 +1,110 @@
...
+  const local = (): string | undefined => loadGroupConfig(groupConfigDir, channelId);
+
+  // Flag off (or no cache to dedupe fetches) → pure local file, unchanged behavior.
+  if (!serverMd || !cache) return local();
+
+  const groupNo = extractParentGroupNo(channelId);
+
+  // a) cached server GROUP.md wins (kept stable until invalidated by the
+  //    event-driven refresh — a separate work item).
+  const cached = cache.get(groupNo);
```
OctoBoooot, line 90: **[major]** cache hit short-circuits the fetch and the cache never expires. Step (a) returns the cached entry whenever present, `getGroupMd` only runs on a miss, `GroupMdCache` has no TTL, and `invalidate()` has no caller in this PR (event-driven refresh is a separate work item per the body). The on-disk copy also survives restart. Net: with `serverMd` on, once a group's GROUP.md is fetched, an operator editing it server-side never sees the change take effect — and a restart re-hydrates the stale disk copy rather than re-fetching. Failing sequence: enable flag → fetch+cache v1 → operator PUTs v2 → every later turn and every restart still serves v1. Direction: add a short TTL to `GroupMdCache.get` (re-fetch when older than N minutes), or document in the `serverMd` config comment that edits don't propagate until the refresh phase ships. Non-blocking only because the flag is off by default.

src/group-md-cache.ts
```diff
@@ -0,0 +1,159 @@
...
+  constructor(cacheDir?: string) {
+    this.cacheDir = cacheDir;
+  }
+
+  /**
+   * Read a cached entry. Checks memory first, then the on-disk copy (hydrating
+   * memory on a disk hit so the next read is hot). Returns undefined on a miss
+   * or for an unsafe groupNo.
+   */
+  get(groupNo: string): GroupMdEntry | undefined {
```
OctoBoooot, line 83: **[minor]** `get()` returns a memory/disk hit with no freshness check. Even a documented-as-future TTL would slot in here cleanly (compare an `updated_at`/fetch timestamp against now before returning the hot entry). Flagging the spot since the resolver's cache-hit branch depends entirely on this method to decide staleness — today it never does.

OctoBoooot (changes requested):
## Review: feat(group-md): server-first GROUP.md fetch with cache and local fallback (#172)

**Verdict:** Request changes — clean, well-tested P2-A, but the cache never refreshes a fetched GROUP.md, so with `serverMd` on an operator's server-side edits never take effect. Flag-off-by-default contains the blast radius, but this should be documented (or a TTL added) before the flag is recommended.

CI is green. PR is a draft. `needs-human-review` + `size/XL` already applied.

I reviewed this at elevated rigor (network fetch → disk → system-prompt injection = SSRF + path + injection surface). The security posture is solid:
- `getGroupMd` routes through the shared `getJson` client, so it inherits the same timeout / Bearer auth / int64-safe parse / SSRF validation as the rest of the surface — no new unvalidated fetch path.
- `groupNo` is `encodeURIComponent`'d on the wire and `isSafeGroupNo`-guarded (same policy as `group-config.isSafeId`) before any disk path is built — `..`/slash can't escape the cache dir, tested.
- Server content is byte-bounded by `boundInstructions` (mirrors `MAX_GROUP_CONFIG_BYTES`), so an oversized payload can't blow the prompt budget; mid-codepoint slice is handled.
- Server content flows into `trustedText(...)` — correct, since GROUP.md is operator-authored server-side (PUT requires bot_admin), same authority class as the local file. Not a trust-boundary regression.

### Major
- `src/group-md-cache.ts` + `src/group-md.ts` — a fetched GROUP.md is cached with **no TTL and no wired invalidation**, and the resolver only calls `getGroupMd` on a cache *miss* (`group-md.ts` step a returns the cached entry whenever present). `invalidate()` exists but the PR body confirms its caller (event-driven refresh) is a separate work item — so nothing invalidates. Worse, the cache is durable on disk, so a restart re-hydrates the stale copy rather than re-fetching. Net: once a group's GROUP.md is fetched, an operator editing it server-side will **never** see the change take effect while `serverMd` is on — the exact opposite of what "server-driven instructions" implies. Failing sequence: enable `serverMd` → group turn fetches v1, caches it → operator PUTs v2 server-side → every subsequent turn (and every restart) still serves v1 indefinitely. Direction: either add a short TTL to `GroupMdCache.get` (re-fetch when an entry is older than N minutes) or, at minimum, document in the `serverMd` config comment that edits don't take effect until the event-driven refresh ships / the cache dir is cleared. Not a blocker only because the flag is off by default and the local-file path is unaffected.

### Minor
- `config.ts` / `config.example.json` `serverMd` docs — the flag is described as server-first-with-fallback but doesn't mention that fetched content is cached indefinitely (no refresh in P2-A). An operator enabling it will reasonably expect server edits to propagate. One sentence — "fetched GROUP.md is cached until the cache dir is cleared; live refresh is a later phase" — would set the right expectation. Pairs with the major.

### Praise
- The never-lose fallback is genuinely never-lose: cached-but-empty, server-reachable-but-empty, 404, network, and timeout all route to `local()`, and each branch has a dedicated test. The resolver "never throws" contract is upheld at every exit.
- Thread routing is correct and tested both ways: server fetch + cache key use `extractParentGroupNo` (a thread shares its parent's GROUP.md), while the local fallback still receives the full composite `channelId` so P1's short-id routing is preserved. That's the subtle interaction with #169 and it's handled.
- `getGroupMd` reusing `getJson` rather than hand-rolling a fetch is the right call — it's why this 80-line API restore inherits SSRF validation for free instead of reintroducing the unvalidated-fetch hole the fork originally removed.

- `updateGroupMd` (PUT) is restored as a client function with no caller — intentional per the body (write-back is a separate work item). Worth confirming it stays uncalled until that work lands; an exported-but-unused mutating endpoint is a loaded gun if a future caller wires it without the bot_admin permission check the comment assumes.
- Coverage-sniff flagged `group-md-cache.ts` / `api.ts` as untested — false positives; `group-md.test.ts` and `group-md-api.test.ts` cover both directly (cache persistence, path-safety, all fallback branches, GET/PUT path+auth).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# cc-channel-octo PR#172 Review Report
**Head SHA:** `302497b21a7ce2ddfb6b6584018ea043bc9c4ecf`
**Scope:** feat(group-md): server-first GROUP.md fetch with cache and local fallback

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Path traversal guard | ✅ | `isSafeGroupNo` (`group-md-cache.ts:47`) uses same regex as `isSafeId` in `group-config.ts:42` — rejects `..`, `.`, anything with `/` or special chars |
| Feature flag default | ✅ | `serverMd?: boolean` is optional, resolver checks `!serverMd` → pure local when unset (`group-md.ts:75`) |
| Never-lose fallback | ✅ | All three failure paths (404/network/empty) fall to `local()` (`group-md.ts:89-101`); `loadGroupConfig` itself never throws (`group-config.ts:60-105`) |
| Bearer auth on GET | ✅ | Routes through `getJson` (`api.ts:267-289`) which sets `Authorization: Bearer ${botToken}` |
| Bearer auth on PUT | ✅ | Explicitly set in `updateGroupMd` headers (`api.ts:735-737`) |
| Byte-cap parity | ✅ | `boundInstructions` (`group-md.ts:42-51`) mirrors `loadConfigFile` (`group-config.ts:112-128`): same MAX check, same UTF-8 mid-codepoint trim, same `[… truncated]` marker |
| Thread parent routing | ✅ | `extractParentGroupNo` (`channel-id.ts:34-37`) splits on first `____`; local fallback gets full channelId preserving thread short-id routing |
| Config propagation | ✅ | `serverMd` flows through `mergeConfig` (`config.ts:363`), `resolveBotConfigs` (`config.ts:596`), and `PartialConfig` (`config.ts:271`) |
| Cache disk persistence | ✅ | `writeDisk`/`readDisk` with content + meta files; type-checked on read (`group-md-cache.ts:119-148`) |
| Test coverage | ✅ | 21 new tests: API client (path, method, auth, parse, error), cache (memory, disk, invalidate, path-safety, memory-only), resolver (flag, server-first, 404/empty/network fallback, cache reuse, thread routing) |
| CI status | ✅ | TypeScript type-check, ESLint (0 warnings), Vitest (460+ tests), coverage, secret scan, dependency review — all green |

## 2. Findings

### F1 — Concurrent fetch stampede on cold cache (P2 — performance)

**What:** When the cache is cold (e.g., post-restart) and N group messages arrive near-simultaneously, `resolveGroupInstructions` fires N parallel `getGroupMd` calls for the same `groupNo` — there's no in-flight dedup / coalesce.

**Diff-scope:** **new** — introduced by this PR's server-first path (`group-md.ts:80-95`).

**Severity:** P2 (performance, not correctness). All N fetches produce the same result; the last `cache.set()` wins. No data corruption. In the typical flow, messages arrive through `activeHandlers` sequential promise-chaining per bot, so concurrent same-group arrivals are bounded.

**Suggestion:** Add an in-flight promise map (`Map<string, Promise<GroupMd>>`) keyed by `groupNo` to coalesce concurrent fetches for the same group. Low urgency.

### F2 — Server timeout (30s default) on the message hot path (P2 — latency)

**What:** `resolveGroupInstructions` is `await`-ed per group message (`index.ts:810`). `getGroupMd` routes through `getJson` which defaults to `AbortSignal.timeout(30_000)` (`api.ts:275`). If the server is reachable but slow-responding, each group message handler blocks for up to 30 seconds.

**Diff-scope:** **new** — `loadGroupConfig` was synchronous (local disk read); `resolveGroupInstructions` introduces a network round-trip on the hot path.

**Severity:** P2 (latency, not correctness). Only applies when `serverMd=true` (opt-in). Operators accept the trade-off when enabling. The 30s timeout is the same as other API calls, so it's consistent with the existing surface.

**Suggestion:** Consider a shorter timeout for the GROUP.md fetch specifically (e.g., 5–10s), or pass a dedicated `groupMdTimeoutMs` config option. Low urgency.

### F3 — `boundInstructions` truncation not directly unit-tested (P2 — test gap)

**What:** `boundInstructions` (`group-md.ts:42-51`) handles byte-bounded truncation of server-provided content, mirroring `loadConfigFile`'s truncation. No test asserts the truncation path specifically (content > `MAX_GROUP_CONFIG_BYTES`).

**Diff-scope:** **new** — the function is new in this PR.

**Severity:** P2 (test gap). The logic is a near-verbatim copy of the battle-tested `loadConfigFile` truncation, so correctness is high-confidence. But the server path can realistically receive oversized payloads (no server-side size cap visible in the client).

**Suggestion:** Add a test that feeds a `>16KiB` server response and asserts the truncated output ends with `[… group config truncated]` and is ≤ `MAX_GROUP_CONFIG_BYTES` + marker length.

## 3. Suggestions

1. **In-flight coalesce** (F1): A `Map<groupNo, Promise<GroupMd>>` guard before `cache.set()` would eliminate the stampede. Simple, no-dependency change.
2. **Dedicated timeout** (F2): Accept a `groupMdTimeoutMs` in config (or hardcode 5s) for the GROUP.md fetch, since it's a best-effort enrichment, not a critical-path API call.
3. **Truncation test** (F3): One more test case in `group-md.test.ts` for the oversized server payload.

## 4. Additional Observations

- **`updateGroupMd` is intentionally unwired** — exported from `api.ts` but not called by any resolver or handler. PR description explicitly scopes this out. No risk from dead-export; it will be wired by a follow-up work item.
- **Stale cache across restarts** is by design ("kept stable until invalidated"). The event-driven refresh is a separate work item. Operators who toggle `serverMd` should be aware the disk-cached copy persists until explicitly invalidated. The `config.example.json` comment documents this.
- **Sync disk I/O on hot path** (`readFileSync` in `GroupMdCache.get` for cold memory) is negligible for a single small file per group, but worth monitoring if a bot serves many groups on first turn after restart.

## 5. Data-Flow Trace

| Consumer | Data | Upstream Source | Traced Path | Status |
|----------|------|-----------------|-------------|--------|
| `resolveGroupInstructions` → `bounded` | `md.content` | `getGroupMd` → `getJson` → `parseOctoJson` → `GroupMd.content` | `api.ts:703-708` → `api.ts:287-289` (JSON parse) → content string | ✅ Flows correctly; null-coalesced via `md?.content ?? ''` at `group-md.ts:88` |
| `resolveGroupInstructions` → `cache.set()` | `md.version`, `md.updated_at`, `md.updated_by` | Same `getJson` response | Type-guarded: `typeof md.version === 'number'` at `group-md.ts:91` | ✅ Flows correctly |
| `cache.get(groupNo)` → `bounded` | `entry.content` | `readDisk` → `readFileSync(groupNo.md)` or `mem.get()` | `group-md-cache.ts:78-83` (mem) → `group-md-cache.ts:127-148` (disk) | ✅ Flows correctly; `isSafeGroupNo` gates both paths |
| `local()` fallback | `loadGroupConfig(groupConfigDir, channelId)` | Local filesystem `<groupConfigDir>/<slug>.md` | `group-config.ts:60-105` (synchronous, never throws) | ✅ Flows correctly; full channelId preserves thread routing |
| `index.ts:810` → `groupInstructions` | Return of `resolveGroupInstructions` | Either server (cached/fresh) or local file | `index.ts:808-818` → `sessionOpts.groupInstructions` → SDK system prompt | ✅ Flows correctly |

## 6. Blind-Spot Checklist (R5 — security_sensitive PR)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths. `getGroupMd` (GET) is the only active path; `updateGroupMd` (PUT) is exported but unwired.
- **C2 — Control-flow ordering / nested reuse:** ✅ Clear. `boundInstructions` is called at exactly two points (`group-md.ts:79` for cached, `group-md.ts:88` for fresh fetch), both after content is obtained. No double-bounding risk. The function is pure and not called from any other module.
- **C3 — Authorization boundary ≠ capability boundary:** ✅ Clear. `getGroupMd` uses the bot's own `botToken` with `Bearer` auth. The bot can only fetch GROUP.md for groups it's a member of (server-side authorization). No new endpoint is exposed — this is a client-only addition.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. This PR adds a client-side fetch of a read-only resource. No authorization logic is changed; the server handles group membership checks.
- **C5 — Build/note ≠ runtime path:** N/A. This is a Node.js server-side TypeScript change, not a browser extension / CLI / packaged artifact. `tsc` + `vitest` + `eslint` all green is sufficient for this runtime.
- **C6 — Governance / policy / security doc self-consistency:** N/A. No governance or policy changes. Only a `config.example.json` comment addition documenting the new flag.

N/A — first review round for this PR.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. Three P2 observations (concurrent stampede, server timeout on hot path, truncation test gap) are improvement suggestions, not blockers. The implementation is clean, well-tested, correctly implements the never-lose fallback pattern, and the feature flag defaults off for safe rollout.

Jerry-Xin (dismissed):
This PR is in scope, but it introduces a writable local cache as a trusted instruction source.

**🔴 Blocking**
🔴 Critical — `GROUP.md` disk cache can be poisoned and then injected as trusted instructions. `src/index.ts:219` stores the durable cache under the bot’s writable `dataDir`; `src/group-md-cache.ts:118` reads `<groupNo>.md` directly from disk; and `src/group-md.ts:79` returns cached content before contacting the server. This bypasses the existing local `GROUP.md` trust boundary documented in `src/group-config.ts:13` and enforced partly at `src/group-config.ts:114`.

A group member can potentially drive the agent/tool process to write or replace `../../data/group-md-cache/<groupNo>.md` from its session workspace. On the next turn, that file is loaded as trusted system instructions even if the server is healthy, and it persists until invalidated. This turns user-influenced local state into trusted prompt content.

Fix options include making durable cache non-authoritative unless a server fetch fails, moving it behind an operator-controlled/non-writable trust boundary, adding robust integrity validation, or avoiding disk persistence for trusted instruction content under the current runtime permissions.

**✅ Highlights**
The feature flag default is conservative, the parent-group routing for thread server fetches is handled cleanly, and the fallback behavior is well covered by focused tests.

**Verification**
I attempted to run the new tests with `npm test -- --run src/__tests__/group-md.test.ts src/__tests__/group-md-api.test.ts`, but this checkout has no installed dev dependencies: `vitest: command not found`.

Jerry-Xin (commented):
Correction to my prior review. I am retracting my three security 🔴 (disk-cache tampering / untrusted-instruction injection / trust-boundary bypass). On verification against the actual path derivation they rest on a false premise, and I should not have rated them blocking. Credit to Octo-PR for the precise rebuttal. My re-issued verdict converges with the cache-refresh finding.

## Retracted — the three security 🔴 do not hold (false reachability premise)

My 🔴 assumed the durable cache dir is agent-writable, letting the agent poison its own trusted instructions. The path derivation refutes that:

- Cache dir = `join(config.dataDir, 'group-md-cache')` (index.ts:219), and `dataDir = <baseDir>/<id>/data` (config.ts:545).
- The agent sandbox cwd = `cwdBase = <baseDir>/<id>/workspace` (config.ts:546) — a **sibling** of `data/`, not a parent.
- The SQLite session store (`data/cc-octo.db`) and memory (`data/memory`) live in the **same `data/` tree** and rely on the identical isolation. The repo threat model already treats `dataDir` as gateway-owned, not agent-writable. Only skill dirs are symlinked into the session sandbox (skill-linker); `data/` is never linked in.

So an agent able to write into `data/group-md-cache` could already poison the session DB and memory — that is a pre-existing `dataDir`-trust property, not something this PR introduces. The `trustedText` framing is correct: server content authored via bot_admin PUT, cached in a non-agent-writable dir. I was wrong to rate this blocking.

## 🟡 Non-blocking — defense-in-depth asymmetry (downgraded from my 🔴)

`group-config.ts` refuses to inject a group/world-writable local file (`statSync` + `if ((st.mode & 0o022) !== 0)`, with comments calling unsanitized GROUP.md a trusted-injection sink). `group-md-cache.ts readDisk()` has no equivalent perm/symlink check. Since `data/` is gateway-owned this is not reachable as an exploit, but mirroring the sibling's cheap `0o022` refusal (and an `lstat` symlink guard) on the cache read would be reasonable defense-in-depth for parity. Optional, not blocking.

## 🟡 Major-leaning operability — cache never refreshes (converges with Octo-PR)

`GroupMdCache` has no TTL and `invalidate()` has no wired caller in this PR (documented as a separate work item). `resolveGroupInstructions` serves a cached entry (memory or disk) before ever calling `getGroupMd`, and the disk copy survives restart. Net effect with `serverMd` ON: once a group's GROUP.md is fetched and cached, an operator editing it server-side will never see the change propagate — not even across a restart. For a feature whose entire premise is server-driven instructions, that is a real operability gap that will surprise operators. Mitigated by flag-off-by-default and explicit P2-A scoping, so I rate it major-not-blocker — but the config docs should at minimum warn that with no refresh wired, a cached GROUP.md is effectively frozen until manual cache removal.

## ✅ Highlights

- Never-lose local fallback on every failure mode (404 / empty / network / timeout) — thoroughly tested.
- `getGroupMd` reuses the SSRF-validated client; `groupNo` is `encodeURIComponent`'d on the wire and `isSafeGroupNo`-guarded (rejects `..`/slashes) for disk paths — no SSRF / path traversal.
- Error strings carry only path/status/body — no Bearer token leak.
- Server content byte-bounded (`boundInstructions`, 16KiB) like local files; thread routing resolves the parent group number correctly for the API + cache key while the local fallback keeps the full composite channelId.

## Verdict

Downgrading from REQUEST_CHANGES (my security 🔴 were a false premise) to a COMMENT: the remaining substantive item is the cache-refresh operability gap, which Octo-PR rates major. If the project gates on that as blocking, defer to that; from my side there is no security blocker. The refresh gap should be fixed or documented before the feature is enabled by default.

boris-clark:
## Rework: addressed the 🔴 trusted-cache poisoning + 🟡 staleness

Thanks for the three-way review. Both blockers are fixed in `d31d570`.

### 🔴 Trusted-injection poisoning — chose direction (c): memory-only cache

Of the three suggested directions, I picked **(c) in-memory only, no disk persistence**, because it removes the poisoning *surface* outright rather than trying to police it:

- The resolved GROUP.md is injected into the system prompt as a **trusted** block (same channel as the local `groupConfigDir` file). A local file earns that trust from OS permissions; a disk cache here **cannot**, because the gateway runs the agent under `bypassPermissions` with `Bash`/`Write`, so the agent — drivable by untrusted chat — can write any file the gateway user owns, including a cache file. The `0o022` check is useless (agent == gateway user). That made the previous `<dataDir>/group-md-cache` a chat-injection → trusted-prompt vector that **survived restart**.
- Now the cache is a process-local `Map` with **no filesystem backing** (all `fs` usage dropped from `group-md-cache.ts`). The only way content can reach the trusted channel is a live, authenticated `getGroupMd` over the bot token against the SSRF-validated `apiUrl` (server-side `bot_admin`-gated). Server content therefore stands on its **own** trust root (authenticated fetch), not on the local file's OS-permission trust, and there is no on-disk artifact to forge or persist.
- Tradeoff: no cross-restart durability. Acceptable — a cold start just re-fetches, and the local file remains the offline fallback. The win is eliminating the persistence vector entirely.

I considered (a)/(b) but both keep a forgeable on-disk artifact and then try to re-validate or down-rank it at read time, which is more code and a larger surface to get wrong; (c) is the smallest, most auditable guarantee.

### 🟡 Staleness — TTL backstop

Added `serverMdTtlMs` (default `DEFAULT_GROUP_MD_TTL_MS` = 5 min). A cached entry past its TTL reads as a miss → re-fetch, so an operator's server-side GROUP.md edit takes effect even before item B's event-driven refresh lands. Does not conflict with B (B can call `invalidate()` for instant refresh).

### Verification
- New/updated tests: TTL expiry + disable, no-durable-backing guarantee (a fresh instance shares no state), TTL-driven re-fetch; existing server-first / fallback (404 / empty / network) / thread parent-group routing coverage retained.
- Full suite green (60 files / 1156 tests), `type-check` + `lint --max-warnings 0` clean. Scope still A only (no B/C wiring; `updateGroupMd` remains an unwired client function).

Jerry-Xin (approved):
Re-review at head d31d57034afe. Both substantive findings from the prior cycle are resolved, and the security concern was addressed by removing the disk surface entirely. APPROVE.

## What changed since my last review

1. Cache is now IN-MEMORY ONLY — `group-md-cache.ts` no longer imports `fs` at all (no writeFileSync / readFileSync / mkdirSync / disk path). The author eliminated the durable on-disk copy rather than perm-guarding it, with the rationale baked into config.example.json: "cached IN MEMORY ONLY (never on disk — it is injected as a trusted system prompt, and a disk cache the gateway user could write would be a poisoning vector)." This structurally removes the disk-tampering surface I had (incorrectly) rated blocking and the perm-asymmetry I noted as defense-in-depth — there is no disk read path left to harden. Good call.

2. TTL staleness backstop added — `serverMdTtlMs` (default `DEFAULT_GROUP_MD_TTL_MS` = 5 min). `get()` returns undefined once `now() - storedAt >= ttlMs`, so an expired read falls through to a fresh `getGroupMd` fetch. A non-positive TTL disables expiry. This resolves the cache-never-refreshes operability major (mine + Octo-PR's): an operator's server-side GROUP.md edit now propagates within the TTL window even before the event-driven refresh (item B) lands.

## Verified no regression

- `resolveGroupInstructions` ordering intact: cache.get (TTL-aware) → server-first `getGroupMd` → never-lose `local()` fallback on every failure (404 / empty / network / timeout).
- `getGroupMd` still over the SSRF-validated `apiUrl`; `groupNo` `isSafeGroupNo`-guarded + `encodeURIComponent` on the wire.
- Server content still byte-bounded via `boundInstructions` (16KiB) like local files.
- Flag-off default preserves byte-identical pre-P2 local behavior.

## Tests

New head adds deterministic TTL coverage with an injectable clock: expiry-at-TTL, non-positive-TTL-disables, invalidate-clears, default-TTL-positive, and re-fetch-after-expiry. Plus the retained server-success / 404→fallback / empty→fallback / network-error→fallback / cache-reuse / thread parent-group routing cases.

## Verdict: APPROVE

The whole prior review cycle is cleanly resolved — the security premise was correctly retired by going in-memory-only, and the refresh gap is fixed with a TTL. No blocking items remain. Optional future polish (non-blocking): request coalescing on a cold-cache thundering herd, and wiring the event-driven `invalidate()` (item B) so edits propagate faster than the TTL.

OctoBoooot (approved):
## Re-review (delta): #172 — cache redesigned to memory-only + TTL

**Verdict:** Approve. Both my cache-never-refreshes major and the disk-poisoning 🔴 (raised by Jerry-Xin/Steve) are resolved — and by a strictly better design than I suggested.

Delta `302497b21a7c` → `d31d57034afe`:

- ✓ **Disk layer removed entirely** — `GroupMdCache` is now in-memory only. This resolves the disk-poisoning concern at the root: there's no on-disk artifact for a chat-driven `Bash`/`Write` (agent runs under `bypassPermissions`, agent == gateway user) to plant and have re-injected as trusted across restart. The SECURITY doc-comment correctly explains why the sibling's `0o022` perm-check can't protect a cache when agent == gateway user — which is exactly the asymmetry that made this a real finding rather than the "defensible/same-authority" framing I gave it in round 1. Credit to Jerry-Xin/Steve on that; my original trust analysis was too lenient.
- ✓ **TTL staleness backstop** — `get()` evicts and misses on entries older than `ttlMs` (default 5 min, configurable via `serverMdTtlMs`), so the resolver re-fetches and an operator's server-side edit now propagates. That closes my major. Cache-first ordering is preserved but no longer permanent.
- ✓ Tests cover the new behavior with an injectable clock: TTL expiry at the boundary, non-positive-TTL-disables-expiry, no-state-across-instances (no durable backing), and the resolver's re-fetch-after-expiry path.

The security posture I verified in round 1 still holds (SSRF-validated `getJson`, `encodeURIComponent` + safe-slug, byte-bounded content, never-lose local fallback, no token in error strings), and the memory-only move doesn't regress any of it.

### Out of scope (still open, non-blocking)
- `updateGroupMd` (PUT) remains restored-but-uncalled — intentional per the body; keep it uncalled until the write-back work item wires it with the bot_admin check.
- Cold start re-fetches (no cross-restart durability) — an accepted, documented tradeoff; the local file is the offline fallback. Fine.
- Event-driven invalidation (item B) is the planned follow-up on top of the TTL backstop; the two compose cleanly.

yujiawei (changes requested):
# Code Review — PR #172 (cc-channel-octo)

Reviewed at head `302497b` against merge-base `87ae921`. Local `type-check`, `lint --max-warnings 0`, `build`, and the full test suite (**60 files / 1153 tests, all green**, +21 new) all pass — the PR's testing claims check out.

This is a well-structured, well-documented, correctly-scoped phase: P2-A (server fetch + cache + flag) only; PUT write-back and event-driven refresh are deferred to separate work items and the `updateGroupMd` client is intentionally unwired. No over-build, no missing scope. The local-file path is preserved byte-identically when `serverMd` is off (default).

I'm requesting changes on **one** security finding before merge; everything else below is non-blocking hardening.

---

## 1. Specification / scope

**Result: ✅ matches the stated P2-A scope.**

- Server `getGroupMd` (`GET /v1/bot/groups/{groupNo}/md`) restored and routed through the shared `getJson` helper — same timeout / Bearer / int64-safe parse / error format. ✅
- `updateGroupMd` restored as a client function only, deliberately not wired to any write path. ✅ (matches the "PUT trigger out of scope" claim)
- `serverMd` flag added, default off → pure local-file behavior. ✅
- Thread routing: server call + cache key use the **parent** `groupNo` via `extractParentGroupNo`, while the local fallback keeps the full `channelId`. ✅
- No out-of-scope additions.

One semantic mismatch to flag against the PR's own wording (see P1 below): the resolver is described as "server-first," but `resolveGroupInstructions` is in fact **cache-first** — a populated cache entry is returned before any server contact (`src/group-md.ts:79-82`). That is a deliberate design choice for a frozen system prompt, but it interacts badly with where the cache lives.

---

## 2. Code quality

**Result: Changes-Requested (1 × P1).**

### 🔴 P1 — durable cache is a trusted-prompt source living in agent-writable space, with the sibling's permission guard dropped

`src/index.ts:219`, `src/group-md.ts:79-82`, `src/group-md-cache.ts:118-143`

The cached GROUP.md is injected **unsanitized** as a trusted `[Group instructions]` block — same trust level as an operator file. The companion module `group-config.ts` is explicit about the threat model (module header + the `0o022` refusal at `src/group-config.ts:108-119`): under the shipped defaults (`allowedTools:'*'`, `bypassPermissions`) the agent has `Bash`/`Write` and "can write ABSOLUTE paths anywhere the gateway user can write." That is exactly why the operator instruction dir is required to live **outside** any agent-writable area, and why a group/world-writable instruction file is refused.

This PR's cache violates both of those protections:

1. It is created at `<dataDir>/group-md-cache` (`src/index.ts:219`) — inside the bot's own data subtree, which **is** writable by the gateway user, i.e. reachable by the agent under the documented default threat model.
2. `readDisk` (`src/group-md-cache.ts:118`) reads `<groupNo>.md` with **no** ownership / mode check — the `(st.mode & 0o022)` refusal that `group-config.ts:114` applies to trusted instruction files is absent here.
3. Because the resolver is cache-first (`src/group-md.ts:79`), a poisoned cache file is served **without ever contacting the server**.

Concrete attack on the repo's own documented assumptions: a compromised turn in group A writes `<dataDir>/group-md-cache/<groupB>.md`; group B's next turn injects that attacker-controlled text as trusted instructions, server bypassed. The local-file path this replaces is specifically hardened against the equivalent move; the new path is strictly weaker (trusted source, agent-writable location, no perm check, read before the server).

The default-off flag limits blast radius, but `serverMd=on` is the intended production path, so the sink ships live the moment the flag is flipped. Please address before merge, e.g.:
- apply the same `0o022` + ownership refusal on cache reads (fail closed) **and/or** locate the cache outside agent-writable space, and
- don't treat a cache hit as trusted without freshness; at minimum revalidate or treat the disk copy as untrusted-until-verified.

### P2 — non-blocking hardening

- **Response size bound happens after the full body is buffered + parsed, and raw content is persisted.** `getJson` does `await resp.text()` then `JSON.parse` (`src/octo/api.ts:286`) with no pre-read cap; `boundInstructions` only bounds the **returned** string (`src/group-md.ts:90`), while the cache stores `md.content` raw (`src/group-md.ts:93`) and writes it unbounded to disk (`src/group-md-cache.ts:148`). A buggy/hostile server reply can drive unbounded memory + an unbounded `.md` on disk despite `MAX_GROUP_CONFIG_BYTES`. Suggest a `Content-Length`/stream cap before buffering, and cache the bounded string.
- **Non-atomic cache write.** `writeDisk` writes `.md` then `.meta.json` directly (`src/group-md-cache.ts:148-154`); a crash mid-write leaves a truncated `.md`, and `readDisk` trusts the `.md` even when `.meta.json` is missing/corrupt — so a corrupted-but-present file is served indefinitely. Use temp-file + `rename` (the repo already uses this pattern elsewhere).
- **`version` is persisted but never used for freshness.** Two concurrent cold misses can race and the slower/older response can overwrite the newer in memory/disk (`src/group-md.ts:89-98`). Compare versions on `set`.
- **No in-flight fetch coalescing.** Concurrent cold-cache messages for one group each fire an independent `getGroupMd` + disk write (cache stampede). A per-`groupNo` in-flight promise map would dedupe. Low impact given typical group counts; noting for completeness.
- **Synchronous fs on the hot path.** `readFileSync`/`writeFileSync`/`mkdirSync` block the event loop (`src/group-md-cache.ts`). This matches the existing `group-config.ts` convention, so it's consistency-neutral, but worth revisiting if group fan-out grows.
- **Error body echoed into log message.** `updateGroupMd` throws `${text || resp.statusText}` (`src/octo/api.ts:741`) — could surface upstream diagnostics in logs. Pre-existing convention across the whole client (`api.ts:81/284/534`) and `updateGroupMd` is currently unwired, so nit-level only.

### Confirmed NOT issues
- `isSafeGroupNo` (`src/group-md-cache.ts:48`) + `encodeURIComponent` on the API path adequately prevent path traversal / cache-dir escape — the unsafe-id test confirms no file is written outside the dir. The concern that the charset is "too strict" and silently disables caching for exotic IDs is speculative: Octo group numbers are numeric/hex (the tests use `grp001` / a 32-char hex id), all within the allowed set. Defense-in-depth nit at most.
- `dataDir` is per-bot namespaced (`botDataDir`) and the cache instance is per-bot, so there is no cross-bot cache bleed.
- `apiUrl` is SSRF-validated at config load (`isAllowedApiUrl`), so the fetch target is constrained.

---

## 3. Items worth a human security pass (this PR is security-classified)
- The server-side authorization model for who may read/write a group's GROUP.md (cannot be verified from this repo).
- Whether the production deployment's sandbox actually keeps the agent out of `<dataDir>` — the P1 above assumes the repo's documented default threat model where it does not.
- The eventual `group_md_updated`/delete invalidation path (deferred work item) is what makes the no-expiry cache safe to keep long-lived; until it lands, a stale or once-poisoned entry persists across restarts.

---

**Overall verdict: REQUEST_CHANGES** — Spec ✅, but the P1 trusted-cache sink must be addressed (or explicitly risk-accepted by a human security owner) before this is safe to merge with `serverMd` enabled. The P2 list is hardening that can follow.

Jerry-Xin:
@clawsweeper re-review — the blocking changes have been addressed in the latest commit. PR#171: seq=0 sentinel data-loss fixed via `message_seq > 0` guard in both the partial unique index and the dedup migration (commit b843c58). PR#172: trust-poisoning fixed by switching the server-md cache to in-memory-only (no disk artifact), plus a TTL refresh. Please re-review the current head and clear the prior CHANGES_REQUESTED.

Jerry-Xin:
@clawsweeper re-review at head `d31d57034afe`. The P1 trust-poisoning finding from the prior review (against `302497b`) is fixed in this newer commit: the server-md cache is now in-memory-only (all fs/disk persistence removed from group-md-cache.ts, no on-disk artifact), so server content can no longer be poisoned and persisted across restarts; a TTL refresh was added. Suite green (1156 tests). The outstanding CHANGES_REQUESTED reviewed an older SHA; please re-review the current head and clear it.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#172 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo #172
**Head SHA:** d31d57034afeb4242f0522feddee62ada3bdde74
**Title:** feat(group-md): server-first GROUP.md fetch with cache and local fallback
**Scope:** security_sensitive (automated review)
**Files changed:** 8 (+752/-9)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Feature flag defaults off | ✅ | `src/config.ts` L102 `serverMd?: boolean` (optional, undefined = off); `src/group-md.ts` L86 `if (!serverMd \|\| !cache) return local()` |
| Memory-only cache (no disk persistence) | ✅ | `src/group-md-cache.ts` — class uses only `Map<string, StoredEntry>`, no fs imports. Test "persists nothing across instances" guards regression. |
| TTL staleness backstop | ✅ | `src/group-md-cache.ts` L88–92 `if (this.ttlMs > 0 && this.now() - stored.storedAt >= this.ttlMs)` evicts. Default 5 min. |
| Never-lose local fallback | ✅ | `src/group-md.ts` L108–114 catch-all → `local()`; L99 empty content → `local()`. |
| Thread parent-group routing | ✅ | `src/group-md.ts` L90 `extractParentGroupNo(channelId)` for server call; L84 `local()` passes full channelId to `loadGroupConfig`. |
| Backward compat (flag off = old behavior) | ✅ | `src/group-md.ts` L86 returns `loadGroupConfig(groupConfigDir, channelId)` — same call the old `index.ts` made. |
| SSRF validation on apiUrl | ✅ | `src/config.ts` L397 references `isAllowedApiUrl` in `url-policy.ts`; apiUrl validated at config load. |
| Bearer auth on API calls | ✅ | `src/octo/api.ts` getGroupMd → getJson L278 `Authorization: Bearer`; updateGroupMd L733 explicit header. |
| URL-encoding of groupNo path segment | ✅ | `src/octo/api.ts` L706 `encodeURIComponent(params.groupNo)`. |
| Byte-bound on server content | ✅ | `src/group-md.ts` L56–63 `boundInstructions` caps at `MAX_GROUP_CONFIG_BYTES` (16 KiB), handles mid-codepoint slicing. |
| Config merge & per-bot override | ✅ | `src/config.ts` L373–374 `mergeConfig`; L607–608 `resolveBotConfigs` — both `serverMd` and `serverMdTtlMs`. |
| Test coverage (real paths, not hand-fed) | ✅ | 21 new tests: fetch mock + real tmpdir for local files. Covers flag gating, server-first, 404/empty/network fallback, TTL, thread routing, cache isolation. |

## 2. Findings

### P2-1: `api.ts` uses plain `fetch()` — no redirect guard on API calls

- **Location:** `src/octo/api.ts` L275 (`getJson`), L730 (`updateGroupMd`)
- **Diff-scope:** Amplified (new code follows existing pattern; `getGroupMd` and `updateGroupMd` are new but use the same `fetch()` as `postJson`/`getJson` — consistent with all other API functions).
- **Detail:** The codebase has `fetchWithRedirectGuard()` in `url-policy.ts` for URL validation on redirects, but `api.ts` uses plain `fetch()`. If the SSRF-validated apiUrl returned a 302, the redirect target would not be re-validated. `getGroupMd` inherits this — it does not introduce it.
- **Risk:** Low — apiUrl is validated at boot, server is unlikely to redirect. Pre-existing pattern across all API functions.
- **Recommendation:** Consider migrating `getJson`/`postJson` to `fetchWithRedirectGuard` in a follow-up for defense-in-depth. Not blocking.

### P2-2: `updateGroupMd` exported but unwired — dead code

- **Location:** `src/octo/api.ts` L720–746
- **Diff-scope:** New (function exists but is not called from any code path).
- **Detail:** The function is exported and the PR body explicitly states "intentionally NOT wired into any write-back path." When eventually connected, the caller must verify server-side `bot_admin` authorization. Client has no content-size guard on `params.content`.
- **Recommendation:** Acceptable as a staged restore. Add a brief JSDoc `@deprecated` or `@internal` annotation to signal it should not be called without the P2-B write-back wiring.

### P2-3: Server-side `boundInstructions` truncation detection less precise than local file

- **Location:** `src/group-md.ts` L56–63 vs `src/group-config.ts` L121–136
- **Diff-scope:** New (server path is new; local file is unchanged).
- **Detail:** `loadConfigFile` reads `MAX_GROUP_CONFIG_BYTES + 1` bytes to detect truncation. `boundInstructions` checks `byteLength > MAX` then slices to `MAX`. If server content is exactly 16384 bytes, `boundInstructions` does NOT truncate (no marker appended), while the local path reads 16385 and truncates (marker appended). Off-by-one in byte-bound semantics.
- **Impact:** Negligible — both produce valid output within the same byte budget. Consistency nit.

### P2-4: Thundering herd on cold start for same-group concurrent messages

- **Location:** `src/group-md.ts` L93–114
- **Diff-scope:** New (new async resolver path).
- **Detail:** Multiple concurrent messages for the same groupNo all see a cache miss and each independently fetches from the server. The last `cache.set()` wins. Correct behavior (content is idempotent), but adds unnecessary server load on cold start.
- **Recommendation:** Consider an in-flight dedup (e.g., store a `Promise<GroupMd>` in the cache on first fetch, resolve for all waiters). Low priority — acceptable for initial landing.

## 3. Suggestions

1. **Redirect guard migration (follow-up):** Move `getJson`/`postJson` to `fetchWithRedirectGuard` for consistent SSRF defense across all API paths.
2. **In-flight dedup (follow-up):** A simple promise-based dedup in `GroupMdCache` would prevent thundering-herd fetches.
3. **`updateGroupMd` annotation:** Mark as `@internal` or add a JSDoc note that it requires P2-B write-back wiring before use.

## 4. Additional Observations

- The second commit (d31d570) correctly addressed the trusted-prompt poisoning vector identified in automated review review: disk cache → memory-only. The security reasoning is well-documented in `group-md-cache.ts` header comments and the commit message.
- `isSafeGroupNo` in `group-md-cache.ts` is defense-in-depth (no filesystem access). Consistent with `group-config.ts` `isSafeId`. ✅
- `config.example.json` documents both `serverMd` and `serverMdTtlMs` with security rationale. ✅
- Per-bot override works correctly: `false ?? true = false` (override wins via `??`). ✅

## 5. Data-Flow Tracing

| Consumer | Data | Upstream Source | Flows Correctly? |
|----------|------|-----------------|------------------|
| `resolveGroupInstructions` → return value | server GROUP.md content | `getGroupMd` → `getJson` → `fetch(GET /v1/bot/groups/{groupNo}/md)` with Bearer auth, SSRF-validated apiUrl | ✅ Non-2xx throws → catch → local fallback. 200+empty → `boundInstructions('')` → undefined → local fallback. 200+content → `boundInstructions(content)` → returned. |
| `resolveGroupInstructions` → cache key | groupNo | `extractParentGroupNo(channelId)` — splits on first `____` separator | ✅ Thread channelId `groupNo____shortId` → `groupNo`. Plain group → unchanged. |
| `resolveGroupInstructions` → local fallback | channelId (full composite) | `loadGroupConfig(groupConfigDir, channelId)` — receives full channelId, preserves thread short-id routing | ✅ Thread channelId routes to `<shortId>.md` first, then legacy `<groupNo>____<shortId>.md`. |
| `GroupMdCache.get` → stored entry | content, version, updated_at, updated_by | `GroupMdCache.set` ← `getGroupMd` response fields | ✅ Entry stored with `storedAt` for TTL. Expired entries evicted on read. |
| `boundInstructions` → trimmed text | server `content` string | `md.content` from `getJson<GroupMd>` response | ✅ `parseOctoJson` parses JSON with int64 protection. Content passes through trim + byte-bound. |
| `handleMessage` → `groupInstructions` | resolved instruction text | `resolveGroupInstructions({groupConfigDir, serverMd, apiUrl, botToken, channelId, cache})` | ✅ Only called when `isGroup` is true. Result injected into `sessionOpts.groupInstructions`. |
| `index.ts` → `groupMdCache` | cache instance | `new GroupMdCache(config.serverMdTtlMs ?? DEFAULT_GROUP_MD_TTL_MS)` — per-bot, created in `startBot` | ✅ Passed to both WS and cron-fire `handleMessage` paths. |

## 6. Blindspot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths touched. The GET (`getGroupMd`) is wired in; the PUT (`updateGroupMd`) is explicitly unwired — no parity obligation.
- **C2 — Control-flow ordering / nested reuse:** CLEAR. `boundInstructions` is called in two places (cache-hit L92, server-fetch L97) with identical semantics. No double-application or ordering risk. `isSafeGroupNo` regex is anchor-bounded (`^...$`), no backtracking risk.
- **C3 — Authorization boundary ≠ capability boundary:** CLEAR. The `getGroupMd` endpoint uses bot-token auth; server-side is described as `bot_admin`-gated. The client doesn't expose any new capability — it only reads. `updateGroupMd` is dead code (not wired).
- **C4 — Authorization lifecycle / container-member cascade:** N/A. This is a client-side read of server state. Authorization cascade is a server-side concern for the GROUP.md endpoint.
- **C5 — Build ≠ runtime:** N/A. Standard Node.js server module. No packaging, extension, or CLI output involved.
- **C6 — Governance/policy document consistency:** N/A. No governance documents changed.

## 7. Cross-Round Blocker Re-check (R6)

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The security-sensitive design choices are sound: memory-only cache eliminates the trusted-prompt poisoning vector, feature flag defaults off ensures safe rollout, TTL provides a staleness backstop, and thread routing is correctly preserved. Four P2 observations are non-blocking suggestions for follow-up iterations.

yujiawei (approved):
# Code Review — PR #172 (cc-channel-octo)

Re-review at head `d31d570` against merge-base `87ae921`. In a clean checkout, local `type-check`, `lint --max-warnings 0`, `build`, and the full test suite (**60 files / 1156 tests, all green**, +21 new for this feature) all pass — the PR's testing claims hold.

**Verdict: APPROVE.** The single blocking finding from the prior cycle is resolved, and it was resolved the right way: rather than hardening a disk cache, the cache surface was removed entirely.

---

## 1. Specification / scope — ✅

Correctly scoped to P2-A (server fetch + in-memory cache + feature flag). No under-build, no over-build:

- `getGroupMd` (`GET /v1/bot/groups/{groupNo}/md`) restored and routed through the shared `getJson` helper — same timeout / Bearer / int64-safe parse / error format as the rest of the client (`src/octo/api.ts`).
- `updateGroupMd` restored as a **client function only** — it appears solely in its own definition and its unit test, and is not called from any production path. No PUT write-back is wired. ✅ (matches "PUT trigger out of scope").
- No event-driven refresh beyond the `invalidate()` hook, which has no caller yet (reserved for the deferred item B). ✅
- `serverMd` flag defaults **off/undefined** → pure local-file behavior; flag-off path is byte-identical to the prior `loadGroupConfig` behavior. ✅
- Thread routing: the server call + cache key use the **parent** `groupNo` via `extractParentGroupNo`, while the local fallback keeps the full `channelId` so its thread short-id routing is preserved. ✅

## 2. Code quality — Approved

### Prior blocking finding — RESOLVED
The previous cycle blocked on a **durable on-disk cache** (`<dataDir>/group-md-cache/<groupNo>.md`) that was read back and injected unsanitized as a trusted `[Group instructions]` system-prompt block — a chat-injection → trusted-prompt poisoning vector that survived restart, sitting in agent-writable space with the sibling's `0o022` permission guard dropped and served cache-first (server bypassed).

This head removes the vector at its root: `src/group-md-cache.ts` is now a **memory-only** `Map` with a TTL and has **zero filesystem imports**. Content can enter the cache only through a live, authenticated `getGroupMd` over the bot token against the SSRF-validated `apiUrl` (`src/group-md.ts:99-110`) — there is no forgeable on-disk artifact a chat-driven `Bash`/`Write` could plant. A regression test (`group-md.test.ts:97-109`) pins the memory-only invariant: a fresh cache instance shares no state, so nothing durable can be poisoned across restart. The local-file fallback remains the offline path and keeps its `0o022` group/world-writable refusal (`src/group-config.ts:114-119`). Cache keys stay guarded by `isSafeGroupNo` (`src/group-md-cache.ts:54-56`), and the cache instance is created in `startBot` and threaded only into `handleMessage` — never handed to the agent process.

The added TTL (default 5 min, `serverMdTtlMs` override) also addresses the earlier "cache never refreshes" concern: an expired read is a miss that forces a re-fetch, so an operator's server-side edit eventually takes effect even before event-driven invalidation lands.

### Non-blocking (P2 / hardening — follow-up, not merge-blocking)
- **Raw `md.content` cached unbounded, re-bounded on each hit.** `src/group-md.ts` stores the full server `md.content` in the cache entry while returning only the `boundInstructions`-trimmed copy; every subsequent cache hit re-runs `boundInstructions` on the raw content. It's correct and idempotent, but a large server payload lingers in memory for the TTL and is re-processed each turn. Caching the already-bounded string would be cleaner. Note this pairs with the fact that `getJson` buffers + parses the whole body with no pre-read size cap, so `MAX_GROUP_CONFIG_BYTES` bounds only what's injected, not what's read/held. Low impact for a memory-only cache with a TTL; worth tidying when the write-back/refresh work lands.

## 3. Items worth a human security pass (this PR is security-classified)
- The **server-side authorization model** for who may read/write a group's GROUP.md is the real trust root now (the client trusts an authenticated `getGroupMd`); it cannot be verified from this repo and should be confirmed operationally before `serverMd` is enabled in production.
- Because the cache is memory-only with no event-driven invalidation yet, a stale entry lives up to one TTL after a server-side edit — acceptable as a documented backstop, but the deferred invalidation item is what makes longer-lived caching safe.

---

**Overall verdict: APPROVE** — Spec ✅ and Quality Approved. The prior P1 poisoning sink is eliminated by design (memory-only, authenticated-source-only), no new P0/P1 introduced, all checks green. The remaining item is P2 hardening that can follow.

PR #173, opened 2026-06-30
feat(group-md): event-driven GROUP.md cache refresh (P2-B)

boris-clark:
## Summary

Part of #88 (P2). Stage 2 item **B — server GROUP.md change events drive a cache refresh**.

`session-router` currently drops every system event with a single `if (msg.payload.event) return null;`. This change splits that guard by `event.type`:

- A **GROUP.md change** event invalidates that group's entry in the in-memory server GROUP.md cache, so the next turn re-fetches the authoritative copy.
- Every **other** system event (group join/leave, etc.) is dropped unchanged — non-md events must never be mistaken for an md event.

The event payload is **never** trusted as the new content. The event is only a signal to drop the cached entry; the existing resolver then re-fetches over the authenticated bot token against the SSRF-validated `apiUrl` (the sole trusted path, identical to the P2-A design). The worst a forged event can do is force a redundant authenticated re-fetch of the real value — never inject content. This keeps the same memory-only / no-poisoning guarantees as the cache layer.

The affected group is taken from the event's `group_no` (these events arrive on a system/DM channel, so the arriving `channel_id` is the sender, not the group), falling back to the channel only on the mention-free-group path where a group event can reach this point.

## Dependency on A (stacked PR)

This branches off **#172** (`feat/server-group-md-fetch-cache`, A's P2-A cache + `invalidate()` hook) — **not** `main` — and must merge **after** #172. The cache-layer commits (`302497b`, `d31d570`) shown in this PR's range belong to #172; the only new commit here is the event-refresh change. Review base should be A's branch.

## Provisional event type (⚠️ needs calibration against a real event)

XIN-173 confirmed the event **shape** (`{ type, version, updated_by, group_no, short_id }`) but **not** the exact `event.type` literal the server emits for a GROUP.md change. The default literal (`group_md_updated`) is named after the design and is **overridable via `serverMdEventTypes`**, so it can be calibrated once a real event is captured — without a code change. The `⚠️ PROVISIONAL` note in `group-md-events.ts` tracks this until confirmed.

## Rollback

Gated behind the existing `serverMd` feature flag (flag off → cache never populated → the event path is a harmless no-op).

## Tests

New `session-router` cases cover both branches of the split:

- md event → cache for that group is invalidated (re-fetch next turn), still returns `null`, no reply sent;
- join/leave (and typeless) event → dropped, cache untouched;
- `serverMd` off / no cache wired → no-op;
- mention-free-group fallback to the channel group when `group_no` is absent;
- `serverMdEventTypes` override (calibration seam).

Full suite green (`npm test`: 1164 passing), lint clean (`--max-warnings 0`), build clean.

### DIFF
--- a/src/session-router.ts
+++ b/src/session-router.ts
@@ -7,6 +7,9 @@ import type { BotMessage, MentionEntity } from './octo/types.js';
 import { ChannelType, MessageType } from './octo/types.js';
 import { sendMessage } from './octo/api.js';
 import { isAuthenticCronFire } from './cron-fire-marker.js';
+import { extractParentGroupNo } from './octo/channel-id.js';
+import type { GroupMdCache } from './group-md-cache.js';
+import { isGroupMdUpdateEvent } from './group-md-events.js';
 
 export interface RouteResult {
   sessionKey: string;
@@ -64,11 +67,20 @@ export class SessionRouter {
    */
   private readonly knownBotUids = new Set<string>();
 
-  constructor(config: Config, robotId: string, ownerUid = '') {
+  /**
+   * P2-B: in-memory server GROUP.md cache (A's P2-A cache). Held so a GROUP.md
+   * change event can invalidate the affected group's entry, forcing the next
+   * turn to re-fetch the authoritative copy. Optional — omitted (or with
+   * `serverMd` off) the event path is a harmless no-op.
+   */
+  private readonly groupMdCache?: GroupMdCache;
+
+  constructor(config: Config, robotId: string, ownerUid = '', groupMdCache?: GroupMdCache) {
     this.config = config;
     this.robotId = robotId;
     this.ownerUid = ownerUid;
     this.knownBotUids.add(robotId);
+    this.groupMdCache = groupMdCache;
   }
 
   /** G14: register another known bot uid (future multi-bot support). */
@@ -335,6 +347,51 @@ export class SessionRouter {
     return isAuthenticCronFire(msg.payload);
   }
 
+  /**
+   * P2-B: react to a server system event that signals a GROUP.md change by
+   * invalidating that group's cached server GROUP.md, so the next turn re-fetches
+   * the authoritative copy over the authenticated bot token (never trusting the
+   * event payload as the new content — see group-md-events.ts for the
+   * forged-event poisoning rationale). Non-md system events (join/leave, etc.)
+   * match nothing here and fall through to the caller's `return null`, so they
+   * are dropped exactly as before.
+   *
+   * Gated on `serverMd`: with the flag off the cache is never populated, so there
+   * is nothing to invalidate. Best-effort and never throws — a malformed event
+   * must not wedge the router (the caller drops the event regardless).
+   *
+   * The group is identified from the event's `group_no`: these events are
+   * delivered on a system/DM channel (a group event on the group's own channel
+   * dies at the mention gate above, never reaching here — XIN-173), so the
+   * arriving `channel_id` is the sender, not the group. The channel is only a
+   * fallback when `group_no` is absent AND the event happens to have arrived on a
+   * group-like channel (e.g. a mention-free group, the one group path that
+   * reaches this point). Using the event-supplied `group_no` is safe despite the
+   * event being untrusted: invalidation is non-destructive — the worst a forged
+   * event can do is force a redundant authenticated re-fetch of the real value,
+   * never inject content (the new GROUP.md is always re-fetched, never read from
+   * the event body — see group-md-events.ts).
+   */
+  private handleGroupMdEvent(msg: BotMessage): void {
+    if (!this.config.serverMd || !this.groupMdCache) return;
+    const event = msg.payload.event;
+    if (!isGroupMdUpdateEvent(event, this.config.serverMdEventTypes)) return;
+
+    let groupNo = event?.group_no ?? '';
+    if (groupNo === '' && this.isGroupLike(msg.channel_type)) {
+      groupNo = extractParentGroupNo(msg.channel_id ?? '');
+    }
+    if (groupNo === '') return;
+
+    // invalidate() is itself total (validates the id, never throws); the guard
+    // here keeps a future non-total cache from breaking the drop path.
+    try {
+      this.groupMdCache.invalidate(groupNo);
+    } catch {
+      /* best-effort: a refresh hiccup must never block dropping the event */
+    }
+  }
+
   private async processMessage(msg: BotMessage, key: string): Promise<RouteResult | null> {
     // Skip messages from self.
     if (msg.from_uid === this.robotId) return null;
@@ -394,8 +451,16 @@ export class SessionRouter {
       }
     }
 
-    // Skip system events (group join/leave, etc.) — no user-facing reply needed.
-    if (msg.payload.event) return null;
+    // System events (group join/leave, GROUP.md change, etc.) never produce a
+    // user-facing reply, so they always route to `null`. P2-B: before dropping,
+    // a GROUP.md change event is dispatched to invalidate that group's cached
+    // server GROUP.md (forcing the next turn to re-fetch the authoritative copy);
+    // every other system event (join/leave/...) is dropped unchanged, exactly as
+    // before — non-md events must NOT be mistaken for an md event.
+    if (msg.payload.event) {
+      this.handleGroupMdEvent(msg);
+      return null;
+    }
 
     // Rate limit check BEFORE non-text check — prevents DM spam of non-text
     // messages from bypassing rate limiting entirely.
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -10,6 +10,8 @@ import type { BotMessage } from '../octo/types.js';
 import { ChannelType, MessageType } from '../octo/types.js';
 import type { Config } from '../config.js';
 import { sendMessage } from '../octo/api.js';
+import { GroupMdCache } from '../group-md-cache.js';
+import type { GroupMdEntry } from '../group-md-cache.js';
 
 const ROBOT_ID = 'bot-001';
 
@@ -970,5 +972,156 @@ describe('SessionRouter — thread (CommunityTopic) session isolation [#88]', ()
     const a = makeMsg({ channel_type: ChannelType.CommunityTopic, channel_id: `${GROUP}____aaa`, from_uid: 'u1' });
     const b = makeMsg({ channel_type: ChannelType.CommunityTopic, channel_id: `${GROUP}____bbb`, from_uid: 'u1' });
     expect(router.sessionKey(a)).not.toBe(router.sessionKey(b));
+  });
+});
+
+// --- P2-B: server GROUP.md change events drive a cache refresh ---
+//
+// GROUP.md change events are delivered on a system/DM channel (a group event on
+// the group's own channel dies at the mention gate before reaching the event
+// branch — XIN-173), so the group identity travels in `event.group_no`, not the
+// arriving channel_id. The event.type literal is PROVISIONAL (group-md-events.ts)
+// — these tests use the default literal and an explicit override to lock in BOTH
+// the routing (md event → invalidate; everything else → dropped, never
+// invalidated) and the config-override seam, so calibrating the literal later
+// needs no test rewrite.
+describe('SessionRouter — GROUP.md event-driven cache refresh (P2-B)', () => {
+  const GROUP = 'group-abc';
+
+  function makeEntry(): GroupMdEntry {
+    return { content: '# cached', version: 1, updated_at: null };
+  }
+
+  function makeRouter(cache?: GroupMdCache, overrides?: Partial<Config>): SessionRouter {
+    // serverMd on by default here — the refresh path is gated on it.
+    return new SessionRouter(makeConfig({ serverMd: true, ...overrides }), ROBOT_ID, '', cache);
+  }
+
+  // A GROUP.md change event as it really arrives: on a DM channel, with the
+  // affected group carried in event.group_no.
+  function mdEvent(overrides?: Partial<BotMessage>): BotMessage {
+    return makeMsg({
+      channel_type: ChannelType.DM,
+      channel_id: 'dm-peer',
+      from_uid: 'user-1',
+      payload: { type: MessageType.Text, content: '', event: { type: 'group_md_updated', group_no: GROUP } },
+      ...overrides,
+    });
+  }
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('a GROUP.md update event invalidates that group\'s cache (next turn re-fetches) and still returns null', async () => {
+    const cache = new GroupMdCache();
+    cache.set(GROUP, makeEntry());
+    const router = makeRouter(cache);
+
+    const result = await router.route(mdEvent());
+
+    // The event itself never produces a reply.
+    expect(result).toBeNull();
+    expect(sendMessage).not.toHaveBeenCalled();
+    // The cached entry is dropped (keyed by event.group_no, not the DM channel),
+    // so the resolver re-fetches the authoritative copy on the next turn.
+    expect(cache.get(GROUP)).toBeUndefined();
+  });
+
+  it('a join/leave (non-md) system event is dropped WITHOUT invalidating the cache', async () => {
+    const cache = new GroupMdCache();
+    cache.set(GROUP, makeEntry());
+    const router = makeRouter(cache);
+
+    const join = mdEvent({
+      payload: { type: MessageType.Text, content: '', event: { type: 'group_member_join', group_no: GROUP } },
+    });
+    const result = await router.route(join);
+
+    expect(result).toBeNull();
+    expect(sendMessage).not.toHaveBeenCalled();
+    // Non-md events must NOT be mistaken for an md event — cache is untouched.
+    expect(cache.get(GROUP)).toEqual(makeEntry());
+  });
+
+  it('an event with no type is dropped and never invalidates', async () => {
+    const cache = new GroupMdCache();
+    cache.set(GROUP, makeEntry());
+    const router = makeRouter(cache);
+
+    const result = await router.route(
+      mdEvent({ payload: { type: MessageType.Text, content: '', event: { group_no: GROUP } } }),
+    );
+
+    expect(result).toBeNull();
+    expect(cache.get(GROUP)).toEqual(makeEntry());
+  });
+
+  it('an md event with no group_no on a DM channel is a no-op (no group to target)', async () => {
+    const cache = new GroupMdCache();
+    cache.set(GROUP, makeEntry());
+    const router = makeRouter(cache);
+
+    const result = await router.route(
+      mdEvent({ payload: { type: MessageType.Text, content: '', event: { type: 'group_md_updated' } } }),
+    );
+
+    expect(result).toBeNull();
+    expect(cache.get(GROUP)).toEqual(makeEntry());
+  });
+
+  it('with serverMd off the cache is never touched (rollback flag)', async () => {
+    const cache = new GroupMdCache();
+    cache.set(GROUP, makeEntry());
+    const router = makeRouter(cache, { serverMd: false });
+
+    const result = await router.route(mdEvent());
+
+    expect(result).toBeNull();
+    expect(cache.get(GROUP)).toEqual(makeEntry());
+  });
+
+  it('no cache wired → md event is a harmless no-op (still returns null)', async () => {
+    const router = makeRouter(undefined);
+    const result = await router.route(mdEvent());
+    expect(result).toBeNull();
+  });
+
+  it('falls back to the channel group when an md event arrives on a mention-free group with no group_no', async () => {
+    const cache = new GroupMdCache();
+    cache.set(GROUP, makeEntry());
+    // Mention-free is the one group path that reaches the event branch (the
+    // mention gate otherwise drops un-mentioned group messages). A composite
+    // thread channel_id normalizes to its parent group.
+    const router = makeRouter(cache, { mentionFreeGroups: [`${GROUP}____thread-1`] });
+
+    const result = await router.route(
+      mdEvent({
+        channel_type: ChannelType.CommunityTopic,
+        channel_id: `${GROUP}____thread-1`,
+        payload: { type: MessageType.Text, content: '', event: { type: 'group_md_updated' } },
+      }),
+    );
+
+    expect(result).toBeNull();
+    expect(cache.get(GROUP)).toBeUndefined();
+  });
+
+  it('the md event literal is overridable via serverMdEventTypes (calibration seam)', async () => {
+    const cache = new GroupMdCache();
+    cache.set(GROUP, makeEntry());
+    // The provisional default literal no longer matches; a real captured literal does.
+    const router = makeRouter(cache, { serverMdEventTypes: ['group.md.changed'] });
+
+    // Default literal is now ignored.
+    await router.route(mdEvent());
+    expect(cache.get(GROUP)).toEqual(makeEntry());
+
+    // The configured literal triggers invalidation.
+    const real = mdEvent({
+      payload: { type: MessageType.Text, content: '', event: { type: 'group.md.changed', group_no: GROUP } },
+    });
+    await router.route(real);
+    expect(cache.get(GROUP)).toBeUndefined();
   });
 });
--- a/src/config.ts
+++ b/src/config.ts
@@ -107,6 +107,16 @@ export interface Config {
    * `DEFAULT_GROUP_MD_TTL_MS` (5 min). Only meaningful when `serverMd` is true.
    */
   serverMdTtlMs?: number;
+  /**
+   * P2-B: `event.type` literal(s) the server emits when a group's GROUP.md
+   * changes. On such an event the in-memory GROUP.md cache for that group is
+   * invalidated so the next turn re-fetches the authoritative copy (never
+   * trusting the event payload — see group-md-events.ts). PROVISIONAL: the
+   * exact literal is not yet confirmed from a real captured event, so this is
+   * overridable here to calibrate without a code change. Defaults to
+   * `DEFAULT_GROUP_MD_EVENT_TYPES`. Only meaningful when `serverMd` is true.
+   */
+  serverMdEventTypes?: string[];
   sdk: {
     model?: string;
     /**
@@ -279,6 +289,7 @@ type PartialConfig = {
   groupConfigDir?: string;
   serverMd?: boolean;
   serverMdTtlMs?: number;
+  serverMdEventTypes?: string[];
   sdk?: Partial<Config['sdk']>;
   rateLimit?: Partial<Config['rateLimit']>;
   context?: Partial<Config['context']>;
@@ -372,6 +383,7 @@ function mergeConfig(base: Config, override: PartialConfig): Config {
     groupConfigDir: override.groupConfigDir ?? base.groupConfigDir,
     serverMd: override.serverMd ?? base.serverMd,
     serverMdTtlMs: override.serverMdTtlMs ?? base.serverMdTtlMs,
+    serverMdEventTypes: override.serverMdEventTypes ?? base.serverMdEventTypes,
     sdk: {
       ...base.sdk,
       ...(override.sdk ?? {}),
@@ -606,6 +618,7 @@ export function resolveBotConfigs(config: Config): Config[] {
       groupConfigDir: perBotFile.groupConfigDir ?? config.groupConfigDir,
       serverMd: perBotFile.serverMd ?? config.serverMd,
       serverMdTtlMs: perBotFile.serverMdTtlMs ?? config.serverMdTtlMs,
+      serverMdEventTypes: perBotFile.serverMdEventTypes ?? config.serverMdEventTypes,
       sdk: {
         ...config.sdk,
         ...(perBotFile.sdk ?? {}),
--- a/src/group-md-events.ts
+++ b/src/group-md-events.ts
@@ -1,0 +1,52 @@
+/**
+ * GROUP.md server-event classification (P2-B) — event-driven cache refresh.
+ *
+ * The server emits system events on the inbound socket as `payload.event`
+ * (shape `{ type, version, updated_by, group_no, short_id }`, see
+ * octo/types.ts MessagePayload.event). When the operator edits a group's
+ * server GROUP.md, the server reports it as one of these events; on receipt we
+ * INVALIDATE the in-memory GROUP.md cache for that group so the next turn
+ * re-fetches the authoritative copy (see group-md-cache.ts `invalidate`).
+ *
+ * SECURITY — why invalidate-then-refetch, never trust the event payload:
+ *   The event arrives over the same untrusted channel as chat, so a forged
+ *   `group_md_updated` event must NOT be able to inject content. We never read
+ *   the new GROUP.md from the event body; we only use it as a signal to drop
+ *   the cached entry, after which the resolver re-fetches over the
+ *   authenticated bot token against the SSRF-validated apiUrl (the sole trusted
+ *   path, identical to P2-A). The worst a forged event can do is force a
+ *   redundant authenticated re-fetch of the real value — never poisoning.
+ *
+ * ⚠️ PROVISIONAL EVENT TYPE — the exact `event.type` literal the server emits
+ * for a GROUP.md change is NOT yet confirmed from a real captured event
+ * (XIN-173 confirmed the event SHAPE, not the type literal). The default below
+ * is named after the design (`group_md_updated`) and is overridable via
+ * `config.serverMdEventTypes`, so the literal can be calibrated against a real
+ * event WITHOUT a code change once captured. Keep this note until calibrated.
+ */
+
+/**
+ * Provisional `event.type` literals treated as a GROUP.md change. PROVISIONAL —
+ * see the calibration note above. Overridable via `config.serverMdEventTypes`.
+ */
+export const DEFAULT_GROUP_MD_EVENT_TYPES: readonly string[] = ['group_md_updated'];
+
+/** The `payload.event` shape (mirrors MessagePayload.event in octo/types.ts). */
+export interface GroupMdEventLike {
+  type?: string;
+  group_no?: string;
+}
+
+/**
+ * True iff this event signals a GROUP.md change and should drive a cache
+ * refresh. All other system events (group join/leave, etc.) return false and
+ * are dropped unchanged by the router. `eventTypes` lets the operator override
+ * the provisional literal(s) without a code change.
+ */
+export function isGroupMdUpdateEvent(
+  event: GroupMdEventLike | undefined,
+  eventTypes: readonly string[] = DEFAULT_GROUP_MD_EVENT_TYPES,
+): boolean {
+  if (!event || typeof event.type !== 'string' || event.type === '') return false;
+  return eventTypes.includes(event.type);
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -272,7 +272,10 @@ async function startBot(config: ReturnType<typeof loadConfig>, multi: boolean): 
     }
 
     // --- Session router ---
-    const router = new SessionRouter(config, gateway.botId, gateway.ownerUid);
+    // P2-B: the router holds the GROUP.md cache so a server GROUP.md change
+    // event invalidates the affected group's entry (re-fetched authoritatively
+    // next turn — never trusting the event payload). No-op when serverMd is off.
+    const router = new SessionRouter(config, gateway.botId, gateway.ownerUid, groupMdCache);
 
     // --- Active handler tracking (Q6: in-flight drain on shutdown) ---
     const activeHandlers = new Set<Promise<void>>();


src/session-router.ts
```diff
@@ -394,8 +451,16 @@ export class SessionRouter {
...
-    // Skip system events (group join/leave, etc.) — no user-facing reply needed.
-    if (msg.payload.event) return null;
+    // System events (group join/leave, GROUP.md change, etc.) never produce a
+    // user-facing reply, so they always route to `null`. P2-B: before dropping,
+    // a GROUP.md change event is dispatched to invalidate that group's cached
+    // server GROUP.md (forcing the next turn to re-fetch the authoritative copy);
+    // every other system event (join/leave/...) is dropped unchanged, exactly as
+    // before — non-md events must NOT be mistaken for an md event.
+    if (msg.payload.event) {
+      this.handleGroupMdEvent(msg);
```
OctoBoooot, line 461: **[question]** `handleGroupMdEvent` runs here, after `isSupportedChannel` (drops system channel_type 8 per #68), the `looksLikeBot` DM gate (~line 419), blocklist, and the mention gate. The doc says these events "arrive on a system/DM channel." Two reachability cases the new DM-with-human-sender tests don't cover: (1) if the server delivers on a SYSTEM channel (type 8), `isSupportedChannel` returns false → handler never runs → invalidation silently never fires; (2) if delivered on DM from a bot-like sender (`*_bot` / known-bot uid), the `looksLikeBot` gate drops it first. Could you confirm the real channel_type + sender uid (or mark provisional like `event.type`)? If events arrive on a system channel, this may need to hook before `isSupportedChannel`. Non-blocking — flag-off default + #172's TTL backstop still bound staleness — but if unreached in prod, P2-B is a no-op and only the 5-min TTL refreshes.

OctoBoooot (commented):
## Review: feat(group-md): event-driven GROUP.md cache refresh (P2-B) (#173)

**Verdict:** Approve with comments — trust model is correct and the event-split is clean; one reachability question worth calibrating alongside the already-flagged provisional event type.

CI is green. PR is a **draft**, **stacked on #172** (review base is A's branch, not main; must merge after #172). `needs-human-review` + `size/XL` applied.

Scope note: the diff range includes #172's cache commits (`302497b`, `d31d570`) — already reviewed and APPROVED at #172's head. I reviewed only the **new** P2-B surface: `group-md-events.ts` + the `session-router.ts` event-split + wiring.

The security design is right, and I verified it rather than taking the body's word:
- **Event is signal-only.** `isGroupMdUpdateEvent` reads only `event.type` to decide whether to invalidate; the new content is never read from the event body. `handleGroupMdEvent` calls `cache.invalidate(groupNo)` and nothing else — the resolver then re-fetches over the authenticated bot token against the SSRF-validated `apiUrl`. So a forged `group_md_updated` event can at most force a redundant authenticated re-fetch of the real value; it cannot inject content. This preserves #172's memory-only / no-poisoning guarantee.
- **Non-md events unchanged.** The `if (msg.payload.event)` guard now dispatches to `handleGroupMdEvent` then still `return null`, so join/leave/typeless events drop exactly as before — tested both branches.
- **Total / never-throws.** `handleGroupMdEvent` is gated on `serverMd` + cache presence, wraps `invalidate` in try/catch, and the caller drops the event regardless. A malformed event can't wedge the router.

### Question
- `src/session-router.ts` `handleGroupMdEvent` placement — it runs late in `processMessage`, after `isSupportedChannel` (drops system channel_type 8 per #68), `streamOn`, the DM/group blocklist, and the `looksLikeBot` gate (line ~419). The doc-comment says these events "arrive on a system/DM channel." Two reachability cases I can't resolve without a real captured event (same uncertainty as the provisional `event.type`):
  1. If the server delivers GROUP.md events on a **system channel** (e.g. type 8), `isSupportedChannel` returns false and the handler is never reached — the invalidation silently never fires.
  2. If it delivers on DM but from a **bot-like sender uid** (`*_bot` or a known-bot uid), the `looksLikeBot` gate at line 419 drops it before `handleGroupMdEvent`.
  The new tests use `ChannelType.DM` with a non-bot sender, so they exercise the happy path but not these gates. Could you confirm (or note as provisional, like the event type) the actual channel_type and sender uid these events arrive on? If they come on a system channel, `handleGroupMdEvent` may need to hook before `isSupportedChannel` rather than after the mention gate. Not a blocker — the feature is flag-off by default and the TTL backstop (#172) still bounds staleness even if the event path never fires — but if the event never reaches the handler in production, P2-B is a silent no-op and only the 5-min TTL refreshes.

### Praise
- The invalidate-then-refetch design is exactly the right trust boundary for an untrusted-channel signal, and the SECURITY doc-comment in `group-md-events.ts` states the threat model precisely (forged event → redundant authenticated re-fetch, never poisoning). This is the same discipline that made #172's memory-only cache correct, carried through consistently.
- `serverMdEventTypes` as a config-overridable calibration seam for the provisional `event.type` literal is the right call — it lets the literal be fixed against a real event without a code change, and the `⚠️ PROVISIONAL` note is honest about what's confirmed (shape) vs not (type literal).
- Taking `group_no` from the event with a channel fallback only on the mention-free-group path, and documenting why the untrusted `group_no` is safe (invalidation is non-destructive), shows the author traced the actual delivery path rather than assuming.

- Coverage-sniff flagged `group-md-events.ts` / `group-md-cache.ts` / `api.ts` — false positives; `session-router.test.ts` covers the event classification + routing, and #172's suites cover the cache + API.
- No rate-limit on the invalidation path: a flood of forged md events would each force one authenticated re-fetch. Bounded (re-fetch is cheap, server-authenticated, and the cache re-populates), so not a finding — noting in case P2-C wants a debounce.

Jerry-Xin (dismissed):
**APPROVE** — the event-driven GROUP.md cache refresh (P2-B) is sound. The event is treated strictly as a signal to drop a cached entry, never as a content source, so the design preserves #172's no-poisoning guarantees. CI is green and the change is well-scoped.

Scope note: the diff range against `main` shows 11 files, but only the latest commit (`656f207`) is this PR — 5 files (`session-router.ts`, `group-md-events.ts`, `config.ts`, `index.ts`, and the router tests). The other 6 files (`group-md-cache.ts`, `group-md.ts`, `octo/api.ts`, two test files, `config.example.json`) are the stacked #172 cache layer and are unchanged here. No unrelated risky changes ride along.

### ✅ Highlights
- **Event trust boundary handled correctly.** The event only triggers `invalidate(groupNo)`, whose sole effect is deleting one in-memory `Map` entry. A forged `group_md_updated` event cannot inject content — the next `resolveGroupInstructions` always re-fetches the authoritative copy over the authenticated bot token against the SSRF-validated `apiUrl`. The worst a forged event achieves is a redundant authenticated re-fetch of the real value; there is no cross-group poisoning path (invalidation is non-destructive). `group_no` is additionally constrained by `isSafeGroupNo` inside `invalidate`, so a crafted id is a safe no-op.
- **Refresh correctness.** After `invalidate(groupNo)` clears the entry, the resolver hits a cache miss and goes server-first, then re-caches — no stale-after-invalidate.
- **Never crashes the message loop.** `handleGroupMdEvent` is gated on `serverMd` + a wired cache, classifies via `isGroupMdUpdateEvent` (typeless/non-md events return false), and wraps `invalidate` in try/catch. Malformed events (no `type`, no `group_no`) degrade to a no-op and still route to `null`.
- **No #172 regression.** Cache stays memory-only (no `fs`/disk import reintroduced), the never-lose local-file fallback is intact, TTL is unchanged, and the SSRF-validated `apiUrl` + `Bearer` fetch path is reused. Error strings do not leak the token.
- **Tests** cover both branches of the system-event split: md event → invalidate + `null` + no reply; join/leave and typeless events → dropped, cache untouched; missing `group_no` → no-op; `serverMd` off and no-cache-wired rollback paths; mention-free-group channel fallback; and the `serverMdEventTypes` calibration override.

### 💬 Non-blocking
- **Thundering herd (pre-existing).** There is still no in-flight coalescing in `resolveGroupInstructions`: concurrent cold-cache turns for the same group right after an invalidate can each fire a `getGroupMd`. This predates this PR (noted 🟡 in #172) and is not worsened structurally, but event-driven invalidation makes the cold-cache window slightly more frequent. Worth a request-dedup follow-up.
- **Integration test gap.** The new tests assert the routing/classification and the `invalidate` side effect, but none drives the full `event → invalidate → next resolveGroupInstructions re-fetch` chain (or concurrent post-invalidate fetch) end to end. The unit coverage is good; an integration case would lock the contract.
- **Provisional event literal.** The `group_md_updated` type literal is acknowledged as unconfirmed and is overridable via `serverMdEventTypes`, which is the right seam — just remember to calibrate (and drop the PROVISIONAL note) once a real event is captured.

yujiawei (approved):
# Code Review — PR #173 (cc-channel-octo)

Reviewed the single new commit `656f207` against `main`; the cache-layer commits (`302497b`, `d31d570`) belong to the stacked base #172 and are out of scope here. Build, lint (`--max-warnings 0`), and the full suite (1164 passing, incl. the 9 new P2-B cases) are green locally on this head.

## 1. Spec compliance — PASS

This implements P2 item B ("server GROUP.md change events drive a cache refresh"):

- **Under-build: none.** The refresh is wired end-to-end — `index.ts:221` builds one `GroupMdCache`, passes the same instance to the router (`index.ts:278`) and to the resolver (`index.ts:821`); the event branch invalidates and the resolver re-fetches on the next turn.
- **Over-build: none material.** The one new config key (`serverMdEventTypes`) directly serves the acknowledged provisional-literal calibration need.
- **Deviation: none.** The design contract holds: the event body is never read as content; it is only a signal to drop the cached entry, after which the resolver re-fetches over the authenticated bot token (`group-md.ts:100` `getGroupMd`). A forged event can at worst force a redundant authenticated re-fetch of the real value — never inject content. Gated behind `serverMd` (off by default), so rollback is a no-op.

## 2. Code quality — Approved

The security-critical property checked out: `invalidate()` is a total, charset-validated (`isSafeGroupNo`, `group-md-cache.ts:54`) `Map.delete` that never throws; an unsafe forged `group_no` silently no-ops; no new disk persistence; no content path through the event body. No P0/P1 defect found. The following are **P2 (non-blocking)** — worth addressing but none gates merge.

### P2 — Runtime assumptions are unverified by design (the load-bearing risk)

Three coupled assumptions are admittedly **provisional** (the code says so) and none is confirmed against a real captured event:

1. **Delivery channel.** `processMessage` drops any `channel_type` not in `{DM=1, Group=2, CommunityTopic=5}` at `session-router.ts:403` (`isSupportedChannel`) **before** the event branch at `session-router.ts:460`. If the server actually emits GROUP.md change events on a system/command channel (e.g. the `channel_type 8 "systemcmdonline"` case from #68), the event is hard-dropped and the whole feature is a silent no-op. The code/tests assume DM delivery; this is plausible but unproven.
2. **Event-type literal.** `DEFAULT_GROUP_MD_EVENT_TYPES = ['group_md_updated']` (`group-md-events.ts:32`) is marked `PROVISIONAL`. If wrong, `isGroupMdUpdateEvent` returns false and invalidation never fires. Mitigated by the `serverMdEventTypes` config override (a genuine calibration seam).
3. **Key namespace.** The resolver caches under `extractParentGroupNo(channelId)` (`group-md.ts:85,109`), while the event path invalidates under `event.group_no` (`session-router.ts:380`). Nothing reconciles the two, and `isSafeGroupNo` is a charset gate, not a canonicalizer. If the server's `group_no` is not byte-identical to the group's `channel_id`, `invalidate()` no-ops on a missing key and the refresh silently never lands. Unlike (2), this has no config escape hatch.

In all three cases the failure mode is **graceful degradation, not breakage**: the independent 5-minute TTL backstop (`group-md-cache.ts:47`, constructed regardless of events) still bounds staleness, so a server-side GROUP.md edit still propagates within ~5 min. That is why this is P2 and not a blocker — but it also means the feature could be 100% inert in production with no visible symptom. **For a human to verify before this path is relied upon as the primary (sub-TTL) refresh mechanism:** capture one real GROUP.md-change event and confirm (a) its `channel_type` reaches the event branch, (b) its `event.type` literal, and (c) `event.group_no === extractParentGroupNo(<that group's channel_id>)`.

### P2 — No observability on the refresh path

`handleGroupMdEvent` logs nothing on success, on a no-literal-match drop, or on an invalidate against an absent key — and this file otherwise logs notable events (`session-router.ts:209,221,649`). Combined with the TTL masking any miss, there is no way to tell "event refresh is firing" from "event refresh never fires, TTL does all the work." Suggest a debug log on both the successful-invalidate path and on a `payload.event` that reached the branch but matched no literal; this is also what makes literal calibration observable.

### P2 — Authorization / ordering hardening (security-sensitive)

A normal-user **DM** carrying `payload.event` reaches `handleGroupMdEvent` (it is not group-like, so the mention gate at `session-router.ts:438` is skipped) and `event.group_no` is honored verbatim with no check tying the sender to that group — so a peer with only DM access can evict an arbitrary group's cached entry. The event branch also runs **before** the rate limiter (`session-router.ts:460` vs `465+`), so these are unthrottled. Impact is bounded and non-destructive: each forged event is one O(1) `Map.delete`, no fan-out, and N forged events between two legitimate target turns coalesce into at most one redundant authenticated re-fetch — the same operation the TTL already performs. Hence P2, not a DoS. Worth hardening (honor `group_no` only for a sender that belongs to the group, or require an authenticated system sender, and/or gate the event path under the limiter). **Flagging for human attention given the `security_sensitive` classification**, even though no amplification or poisoning vector exists.

### P2 — Minor

- `serverMdEventTypes` is read via a whole-file `JSON.parse` type assertion (`config.ts`) with no field validation; a misconfigured string value (instead of array) would degrade `.includes()` to substring matching. Operator-owned/trusted input, so low risk — a one-line `Array.isArray` guard at the merge site closes it.
- `config.example.json` documents `serverMd`/`serverMdTtlMs` but not the new `serverMdEventTypes` key.
- The new tests are tautological on the primary path: cache is seeded and the event fired with the **same** `group_no` literal, so they prove wiring but cannot catch the key-namespace mismatch in P2(3). An integration test that populates via `resolveGroupInstructions(channelId)` then fires the matching event would pin it.

## Verdict

The change is correct, narrowly scoped, well-tested for the branches it can observe, and security-sound (invalidate-then-refetch, flag-gated, never trusts the payload). All findings are advisory P2. **Approving.** The unverified-runtime-assumption cluster (delivery channel + event-type literal + `group_no`↔`channel_id`) is the one thing that must be confirmed against a real captured event before treating event-driven refresh as the primary sub-TTL mechanism; until then the TTL backstop keeps behavior safe.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#173 Review Report (automated review / automated review)

**PR:** https://github.com/Mininglamp-OSS/cc-channel-octo/pull/173
**Title:** feat(group-md): event-driven GROUP.md cache refresh (P2-B)
**Head SHA:** `656f20773ff1039c4621199ae727e19bfd0a04ab`
**Scope:** 11 files, +1042/−13 (stacked on #172 — cache + resolver commits included in diff range)

---

## 1. 验证结论

| Item | Status | Evidence |
|------|--------|----------|
| Event → cache invalidation (never trust payload) | ✅ | `session-router.ts:377–391`, `group-md-events.ts:44–52` — event is signal-only, content always re-fetched via `getGroupMd` over Bearer auth |
| In-memory-only cache (no disk poisoning surface) | ✅ | `group-md-cache.ts:66–108` — `Map<string, StoredEntry>`, no fs calls; test `persists nothing across instances` guards regression |
| Feature flag gating (serverMd off = unchanged behavior) | ✅ | `group-md.ts:93`, `session-router.ts:378` — both early-return when flag off; test `flag OFF → reads the local file only` confirms |
| Non-md events dropped unchanged | ✅ | `isGroupMdUpdateEvent` (`group-md-events.ts:44–52`) matches only `eventTypes` array; non-matching events fall through to `return null` |
| TTL staleness backstop | ✅ | `group-md-cache.ts:88–93` — `now() - storedAt >= ttlMs` evicts; test `expires an entry once it is older than the TTL` |
| Thread parent-group routing | ✅ | `group-md.ts:96` — `extractParentGroupNo(channelId)` strips `____shortId`; `channel-id.ts:34–37` uses `indexOf` (first occurrence) |
| Never-lose local fallback | ✅ | `group-md.ts:113–119` — catch block falls back to `loadGroupConfig`; tests cover 404, empty, network error |
| Byte-bound on server content | ✅ | `group-md.ts:50–58` — `MAX_GROUP_CONFIG_BYTES` (16KiB) from `group-config.ts:36`, mid-codepoint trim safe |
| Safe groupNo validation | ✅ | `group-md-cache.ts:54–56` — `/^[a-zA-Z0-9._-]+$/`, rejects `.` and `..` |
| Tests comprehensive | ✅ | 1164 passing per PR body; new tests cover all split branches, fallback paths, flag states, override seam |
| Stacked PR dependency on #172 | ⚠️ | PR targets `main` but description states it must merge after #172; commits `302497b`, `d31d570` belong to #172 |

## 2. 发现的问题（P0/P1/P2）

### P2-1: `updateGroupMd` exported but unused (dead code)

**Severity:** P2 (maintainability)
**Diff-scope:** **new** — introduced by this PR (restored alongside `getGroupMd`).
**Location:** `src/octo/api.ts:720–746`

`updateGroupMd` is exported and tested but not called from any production code path. The PR description explicitly notes: _"this client function is restored alongside getGroupMd, but is intentionally NOT wired into any write-back path here — the PUT trigger chain is owned by a separate work item."_

**Impact:** Low — dead export with tests, no runtime effect. Acceptable as forward-restore if a future PR will consume it. The test file (`group-md-api.test.ts`) exercises it independently.

### P2-2: No runtime validation of `config.serverMdEventTypes`

**Severity:** P2 (config robustness)
**Diff-scope:** **new** — introduced by this PR.
**Location:** `src/config.ts:114–122` (type), `src/session-router.ts:379` (consumption)

`serverMdEventTypes` is typed as `string[]` in the `Config` interface but never validated at config load time. If a malformed config file supplies a non-array value, `eventTypes.includes(event.type)` in `isGroupMdUpdateEvent` could throw or produce unexpected results.

**Mitigating factors:** TypeScript types + JSON schema validation (if present) should catch this at deploy time. The default parameter in `isGroupMdUpdateEvent` handles `undefined` correctly. An empty array `[]` silently disables event matching (intentional override behavior, not a bug).

### P2-3: Comment inaccuracy about mention-gate behavior

**Severity:** P2 (documentation accuracy)
**Diff-scope:** **new** — comment added by this PR.
**Location:** `session-router.ts:362–363`

Comment states: _"a group event on the group's own channel dies at the mention gate above, never reaching here — XIN-173"_. This is true for **regular** groups but NOT for **mention-free groups** (`mentionFreeGroups`), where the event passes through and reaches `handleGroupMdEvent`. The code correctly handles both paths (the fallback `extractParentGroupNo(msg.channel_id)` covers the mention-free case), and tests cover it (`falls back to the channel group when an md event arrives on a mention-free group`). The comment is slightly misleading but the implementation is correct.

---

## 3. 建议

1. **P2-1**: Consider deferring `updateGroupMd` export to the PR that actually wires the write-back path, reducing the dead-code surface. Alternatively, add a `// @internal` JSDoc tag or `/* istanbul ignore next */` if keeping it.
2. **P2-2**: Add a runtime guard in `loadConfig` or `mergeConfig`: `if (override.serverMdEventTypes && !Array.isArray(override.serverMdEventTypes)) throw ...` — defense-in-depth for config files that bypass TypeScript.
3. **P2-3**: Amend the comment to note the mention-free-group exception: _"dies at the mention gate above (unless the group is in `mentionFreeGroups`)"_.

---

## 4. 额外发现

### 4a. Minor async overhead when `serverMd` is off (informational, not a finding)

`resolveGroupInstructions` is `async` (returns `Promise<string | undefined>`), replacing the synchronous `loadGroupConfig` call. When `serverMd` is off (the common case), the resolver returns `local()` synchronously but wrapped in a resolved Promise. The overhead is negligible for message-processing throughput but is a micro-architectural change from the prior synchronous path. `handleMessage` is already `async`, so no caller-side impact.

### 4b. `getJson` returns `{} as T` for empty response body

In `api.ts:287`, when the GET response body is empty, `getJson` returns `{} as T`. In `group-md.ts`, this is handled correctly: `md?.content ?? ''` produces `''`, which `boundInstructions` rejects as empty, falling through to the local fallback. No bug, but the implicit empty-object contract is fragile — a future consumer might not check for missing fields.

---

## 5. 数据流回溯（每个被消费数据 → 上游来源 → 是否真的流到消费点）

| Consumed Data | Upstream Source | Flows Correctly? | Evidence |
|---|---|---|---|
| `msg.payload.event` in `handleGroupMdEvent` | WS JSON parse → `BotMessage.payload.event` (`types.ts:75–80`) | ✅ Yes | `processMessage` guards `msg.payload.event` truthy at L460 before calling `handleGroupMdEvent`; payload is never undefined at that point |
| `event.type` in `isGroupMdUpdateEvent` | Same event object → `MessagePayload.event.type: string` | ✅ Yes | Guard: `typeof event.type !== 'string' || event.type === ''` rejects missing/non-string/empty |
| `event.group_no` in `handleGroupMdEvent` | Same event object → `MessagePayload.event.group_no?: string` | ✅ Yes | Guard: `event?.group_no ?? ''` handles undefined; empty string short-circuits to fallback or early return |
| `groupNo` → `cache.invalidate(groupNo)` | From `event.group_no` or `extractParentGroupNo(channel_id)` | ✅ Yes | `isSafeGroupNo` regex validates before `Map.delete`; worst case is deleting a non-existent key (no-op) |
| `cache.get(groupNo)` → `cached.content` | `cache.set(groupNo, entry)` stores `md.content` from server | ✅ Yes | Entry stored after successful `getGroupMd` fetch; `get` validates `isSafeGroupNo` and TTL before returning |
| `md.content` → `boundInstructions` → `sessionOpts.groupInstructions` | `getGroupMd` → `getJson<GroupMd>` → `parseOctoJson` → `md.content: string` | ✅ Yes | Bearer auth on SSRF-validated `apiUrl`; byte-bounded at 16KiB; trimmed; empty rejected → local fallback |
| `channelId` → `extractParentGroupNo` → API URL | `msg.channel_id` from WS → `extractParentGroupNo` → `/v1/bot/groups/{groupNo}/md` | ✅ Yes | `encodeURIComponent(groupNo)` prevents path traversal in URL; `isSafeGroupNo` additionally validates cache key |

---

## 6. 盲点 checklist（R5 — security_sensitive PR, 全项必填）

### C1 — 双路径 parity
**N/A for this PR's core logic.** The event-handler path (invalidate on event) and the TTL path (expire on staleness) are **complementary freshness mechanisms**, not symmetric create/delete pairs. The cache itself has symmetric `set`/`get`/`invalidate` operations, all validated by `isSafeGroupNo`.

The cache is created once per `startBot` invocation and shared by reference between the `SessionRouter` (write/invalidate path) and `handleMessage` → `resolveGroupInstructions` (read/set path). Both paths receive the same `groupMdCache` instance (`index.ts:220,275,293,314`). ✅ Consistent.

### C2 — control-flow ordering / 嵌套复用
**Clear.** `handleGroupMdEvent` is called at exactly one site (`session-router.ts:462`), after the mention gate. The ordering is:
1. Self-message drop → 2. Channel support check → 3. Stream update drop → 4. DM blocklist/bot guard → 5. Group bot guard → 6. Mention gate → 7. **Event check** (handleGroupMdEvent + return null) → 8. Rate limit → 9. Non-text check

No nested or repeated invocation of `handleGroupMdEvent` or `isGroupMdUpdateEvent`. The `isGroupMdUpdateEvent` classifier is a pure function with no side effects — safe under any ordering.

`boundInstructions` is called in two places in `group-md.ts` (cached path L100, server-fetch path L108). Both apply the same bounding to the same type of content. No double-bounding or ordering issue.

### C3 — 授权边界 ≠ 能力边界
**Clear.** The GROUP.md fetch uses the bot's authenticated token against the SSRF-validated `apiUrl`. Authorization (which groups this bot can read) is enforced server-side (`bot_admin-gated` per the comments). The client-side gateway has no additional authorization boundary to enforce.

The event-driven invalidation does NOT read or write content — it only deletes a Map key. Even a forged event can only trigger a redundant authenticated re-fetch. The `group_no` from the event is used only as a cache key, not as an authorization token.

### C4 — 授权生命周期 / 容器-成员状态级联
**N/A.** This PR does not modify any authorization logic. The bot token's validity is managed by the server. If the token is revoked, `getGroupMd` will throw, and the resolver falls back to the local file — correct degradation.

### C5 — build/note 通过 ≠ 运行期路径正确
**N/A.** No build artifacts, browser extensions, CLI packaging, or relative-URL changes in this PR. All paths are server-side Node.js with absolute `apiUrl`.

### C6 — 治理/策略/安全文档自洽性
**N/A.** No governance or policy documents modified. `config.example.json` adds a comment documenting `serverMd` behavior; the comment is consistent with the implementation.

---

## 7. 跨轮 blocker 复检（R6 — 首次 review, N/A）

This is the first review of this PR. No prior blockers to re-check.

---

## Summary

This is a well-designed, well-implemented feature with strong security thinking throughout:

- **Invalidate-then-refetch** ensures forged events cannot inject content
- **Memory-only cache** eliminates disk-based poisoning vectors
- **Feature flag** provides clean rollback
- **Never-lose fallback** preserves local-file behavior on any server failure
- **Comprehensive tests** cover all branches of the split event handler, cache lifecycle, and resolver priority chain

No P0 or P1 findings. Three P2 observations (dead `updateGroupMd` export, missing runtime config validation, slightly misleading comment) are informational and non-blocking.

---

**[Octo-Q] verdict: APPROVE** — no P0/P1 findings. Three P2 observations are non-blocking suggestions for follow-up cleanup.

Jerry-Xin (commented):
Addendum to my APPROVE — crediting a higher-value open item Octo-PR raised that my review missed. This does not change my verdict (still no blocker), but it deserves a calibration note alongside the already-flagged provisional event.type.

## ⚠️ Reachability of `handleGroupMdEvent` — same provisional-assumption risk as the event.type literal

I verified the guard ordering in `processMessage` (session-router.ts). `handleGroupMdEvent` is dispatched at line ~461, but only AFTER these earlier guards, any of which returns null first:

1. `from_uid === robotId` (self-skip)
2. `!isSupportedChannel(channel_type)` → null — allows ONLY `DM` or group-like. Octo's system/command channel (channel_type 8 "systemcmdonline", the #68 case) is dropped here.
3. `streamOn` → null
4. DM blocklist
5. `channel_type === DM && looksLikeBot(from_uid) && !isAllowedBot` → null (uid ending `_bot` or in knownBotUids)
6. group blocklist
7. group mention gate (non-mentioned, non-cron, non-mention-free) → null

So P2-B's event-driven refresh only works if the real server GROUP.md-change event arrives on a DM or group-like channel AND from a sender uid that is not bot-looking/blocklisted. Two concrete ways it silently no-ops:

- If the server emits the change event on a **system channel (type 8)**, guard #2 drops it before the handler — event-driven refresh never fires, P2-B degrades to the 5-min TTL only.
- If the event arrives on **DM from a bot-like/system sender uid**, guard #5 drops it.

The new tests construct the event as `ChannelType.DM` with a human-looking `from_uid`, which sidesteps both guard #2 and guard #5 — so the green tests do not actually demonstrate the production event path is reachable. This is the same class of unverified assumption the author already, correctly, flagged for `event.type` (shape confirmed via XIN-173, but not the real type literal / channel / sender captured from a live event).

Not a blocker: `serverMd` is flag-off by default and the TTL is a correct backstop, so the worst case is graceful degradation to ≤5-min staleness, not breakage or any security issue. But the channel + sender assumptions deserve the same "PROVISIONAL until calibrated against a real captured event" note that `event.type` already carries — and ideally a test (or a captured-event fixture) that exercises the handler on the actual channel_type/sender the server uses, rather than DM-with-human-sender. Credit to Octo-PR for surfacing this; I under-weighted it in my initial pass.

My verdict is unchanged: APPROVE (no blocking or major items at this head).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# cc-channel-octo PR#173 Review Report
**PR:** Mininglamp-OSS/cc-channel-octo #173
**Title:** feat(group-md): event-driven GROUP.md cache refresh (P2-B)
**Head SHA:** `656f20773ff1039c4621199ae727e19bfd0a04ab`
**Scope:** Stacked on #172 (P2-A cache + resolver). This review covers the combined diff (P2-A + P2-B).

---

| Item | Status | Evidence |
|------|--------|----------|
| In-memory-only cache (no disk) | ✅ | `group-md-cache.ts` L44: `private readonly mem = new Map<string, StoredEntry>()` — no fs calls anywhere in the class. Test `persists nothing across instances` guards against future durable backing. |
| Feature flag gating (serverMd off = unchanged) | ✅ | `group-md.ts` L86: `if (!serverMd \|\| !cache) return local()`. `session-router.ts` `handleGroupMdEvent` L380: `if (!this.config.serverMd \|\| !this.groupMdCache) return`. |
| Event never trusted as content | ✅ | `handleGroupMdEvent` only calls `cache.invalidate(groupNo)` — a Map.delete. No content from `event` is ever stored or injected. |
| Authenticated re-fetch on cache miss | ✅ | `group-md.ts` L96: `getGroupMd({ apiUrl, botToken, groupNo, signal })` → `getJson` in `api.ts` L267–289 uses `Bearer ${botToken}` against SSRF-validated `apiUrl`. |
| Non-md events dropped unchanged | ✅ | `isGroupMdUpdateEvent` in `group-md-events.ts` L46–49: only matches `event.type` in the configured literal list. Join/leave events have different type strings → returns false → no invalidation. |
| Never-lose local fallback preserved | ✅ | `group-md.ts` L108–113: catch block returns `local()`. Three separate test cases (404, empty, network error) verify. |
| Thread parent-group routing | ✅ | `group-md.ts` L90: `extractParentGroupNo(channelId)` — `channel-id.ts` L33: splits on first `____`. Test `thread channelId → server fetch uses the PARENT groupNo` verifies. |
| Byte-bound on server content | ✅ | `group-md.ts` `boundInstructions` L50–58: identical logic to `loadConfigFile` truncation in `group-config.ts`. Same `MAX_GROUP_CONFIG_BYTES` (16 KiB). |
| Safe groupNo validation | ✅ | `group-md-cache.ts` L55: `isSafeGroupNo` regex `^[a-zA-Z0-9._-]+$` rejects `.`/`..`/path traversal. Applied in `get`, `set`, `invalidate`. |
| Config propagation (mergeConfig + resolveBotConfigs) | ✅ | `config.ts` L384–386 (mergeConfig), L619–621 (resolveBotConfigs): `serverMd`, `serverMdTtlMs`, `serverMdEventTypes` propagated at all three levels (Config, PartialConfig, per-bot override). |
| Test suite coverage | ✅ | 8 new session-router tests covering both split branches, flag-off, no-cache, mention-free fallback, event-type override. 10 new resolver tests. 6 new API tests. |

## 2. Findings

### F1 (P2-nit): `updateGroupMd` exported but unused

**File:** `src/octo/api.ts` L727–746
**Diff-scope:** New (this PR restores the function).
**Assessment:** The PUT endpoint client is restored alongside GET but intentionally not wired into any write path (comment at L720: "the PUT trigger chain is owned by a separate work item"). This is dead code today but explicitly scoped for a future PR. Not a blocker — acceptable as a co-restoration with the GET endpoint.

### F2 (P2-nit): No observability logging on cache invalidation

**File:** `src/session-router.ts` `handleGroupMdEvent` L380–393
**Diff-scope:** New (this PR).
**Assessment:** The invalidation is silent (no log line). While `console.error` is used in `group-md.ts` for server fetch failures, the event-driven invalidation path has no equivalent. In production, operators would have no way to confirm that a GROUP.md edit event was received and acted upon. A debug-level log (`[cc-channel-octo] group-md: invalidated cache for ${groupNo}`) would aid troubleshooting. Not a blocker — the invalidation itself is correct.

## 3. Suggestions

1. **Observability:** Add a `console.debug` or `console.log` line in `handleGroupMdEvent` when invalidation succeeds. Operators debugging GROUP.md propagation latency need to confirm the event was received.
2. **Provisional event type tracking:** The `⚠️ PROVISIONAL` note in `group-md-events.ts` is well-documented. Consider creating a tracked issue (if not already) to calibrate the literal against a real captured event, so this doesn't linger indefinitely.

## 4. Extra Findings

- **Cache stampede resistance:** A forged event flood could force repeated re-fetches, but the re-fetch rate is bounded by the turn rate (one fetch per group per turn), and the attacker must be a conversation participant. Acceptable residual risk.
- **`GroupMdCache` created unconditionally:** Even when `serverMd` is off, `index.ts` creates the cache. This is documented as "cheap, only populated when serverMd is on" — an empty Map costs negligible memory. Acceptable design choice.
- **`signal` not forwarded from `handleMessage`:** `resolveGroupInstructions` accepts an optional `AbortSignal` but `handleMessage` doesn't pass one. The `getJson` helper uses its own `AbortSignal.timeout(30_000)` default. No issue — there's no external cancellation source in the current call chain.

## 5. Data Flow Tracing

### Flow A: Event → cache invalidation
```
msg.payload.event (from WebSocket)
  → SessionRouter.processMessage() [session-router.ts L451]
    → if (msg.payload.event) { handleGroupMdEvent(msg); return null; }
      → isGroupMdUpdateEvent(event, config.serverMdEventTypes) [group-md-events.ts L46]
        → checks event.type ∈ eventTypes list
      → groupNo = event.group_no ?? '' [session-router.ts L384]
        → fallback: extractParentGroupNo(msg.channel_id) if isGroupLike [L386]
      → cache.invalidate(groupNo) [L391]
        → isSafeGroupNo(groupNo) [group-md-cache.ts L55]
          → mem.delete(groupNo) [L99]
```
**Verification:** Every step is guarded. The event type literal check prevents non-md events from reaching invalidation. The `isSafeGroupNo` regex prevents path traversal. The `invalidate()` method is total (never throws). Data flows correctly from event → groupNo → cache deletion. ✅

### Flow B: Cache miss → server re-fetch → system prompt injection
```
resolveGroupInstructions({ ..., cache }) [group-md.ts L80]
  → cache.get(groupNo) → undefined (miss) [L93]
  → getGroupMd({ apiUrl, botToken, groupNo }) [L96]
    → getJson(apiUrl, botToken, `/v1/bot/groups/${encodeURIComponent(groupNo)}/md`) [api.ts L704]
      → fetch(url, { method: 'GET', headers: { Authorization: Bearer } }) [L275]
      → parseOctoJson(text) [L288]
    → returns GroupMd { content, version, updated_at, updated_by }
  → boundInstructions(md.content) [L97]
    → byte-bound to MAX_GROUP_CONFIG_BYTES (16 KiB) [L50–58]
  → cache.set(groupNo, entry) [L100]
  → return bounded text [L104]
    → sessionOpts.groupInstructions [index.ts L818]
      → injected as trusted system prompt block
```
**Verification:** Content enters the trusted system-prompt channel ONLY through this authenticated fetch path. The `apiUrl` is the server's own URL from bot registration (not user input). The `botToken` is the bot's authenticated token. `encodeURIComponent` on `groupNo` prevents URL injection. The byte-bound matches the local file truncation. ✅

### Flow C: Local fallback on server failure
```
resolveGroupInstructions → try { getGroupMd } catch (err) { return local() }
  → loadGroupConfig(groupConfigDir, channelId) [group-config.ts L76]
    → isSafeId check → file read → byte-bound → permission check (0o022)
```
**Verification:** The local fallback path is identical to the pre-P2 code path. The `loadGroupConfig` function retains its own security checks (safe id, file permissions). ✅

## 6. Blind-spot Checklist (R5 — security_sensitive, all items mandatory)

### C1 — Double-path parity (symmetric paths)
**Clear.** This PR adds a single event-handling path (invalidation). There is no symmetric create/delete or subscribe/unsubscribe pair introduced. The cache has `set`/`get`/`invalidate` — the set/get pair is symmetric and both apply `isSafeGroupNo`. The event path is a one-way signal (invalidate only), not a bidirectional protocol.

**Guard/tool self-coverage:** No guards or lint tools are introduced that claim universal coverage.

### C2 — Control-flow ordering / nested reuse
**Clear.** `isGroupMdUpdateEvent` is called exactly once (in `handleGroupMdEvent`). `boundInstructions` is called in two non-overlapping branches of `resolveGroupInstructions` (cached path vs fresh fetch path) — no double-application risk. `extractParentGroupNo` is called in two places for the same purpose (group identity extraction). No ordering dependency between these call sites.

**Non-canonical form testing:** `isSafeGroupNo` regex `^[a-zA-Z0-9._-]+$` rejects leading `../`, unicode tricks (non-ASCII chars fail the character class), and empty strings (`+` requires at least one char). `isGroupMdUpdateEvent` rejects undefined event, non-string type, and empty-string type.

### C3 — Authorization boundary ≠ capability boundary
**Clear.** No new endpoints or tools are exposed to external callers. `updateGroupMd` is exported but not wired into any write path — it's a library function, not an exposed capability. The cache is process-local (in-memory Map). The `getGroupMd` call uses the bot's authenticated token against the SSRF-validated server URL — the authorization boundary is the server-side bot_admin gate, which is the same boundary the rest of the API surface relies on.

### C4 — Authorization lifecycle / container-member state cascade
**N/A.** No authorization changes in this PR. The cache invalidation is non-destructive (forces re-fetch, never injects content). No container/member status cascade is relevant.

### C5 — Build/note ≠ runtime path correctness
**Clear.** This is a Node.js gateway, not a browser extension, CLI installable, or packaged artifact. No relative paths, base URL resolution, or environment-dependent behavior that differs between build and runtime. The `apiUrl` is resolved at registration time and remains constant. PR body reports `npm test: 1164 passing, lint clean, build clean`.

### C6 — Governance / policy / security document self-consistency
**N/A.** No governance, security policy, or disclosure documents are modified. The only documentation change is a `_comment_server_md` entry in `config.example.json` explaining the feature flag.

## 7. Cross-round Blocker Re-check (R6)

N/A — this is the first review of this PR.

---

## Summary

This is a well-architected, security-conscious implementation. The core design decision — invalidate-then-refetch rather than trusting the event payload — correctly eliminates the forged-event poisoning vector. The in-memory-only cache eliminates the disk-cache poisoning surface identified in PR#172 review. The feature flag provides clean rollback. The event-type override seam allows calibration without a code change. Test coverage is thorough across both the router event dispatch and the resolver.

No P0 or P1 issues found. Two P2-nit observations (unused `updateGroupMd` export, missing observability logging) are non-blocking.

[Octo-Q] verdict: APPROVE — only P2/nit findings; no P0/P1.

yujiawei (dismissed):
# Code Review — PR #173 (cc-channel-octo)

**Scope reviewed:** the incremental change on top of #172 (base = `feat/server-group-md-fetch-cache`, merge-base `d31d570`). The only new commit here is `656f207`; the cache-layer commits shown in the range belong to #172 and were reviewed there. Head SHA reviewed: `656f207`.

Build, type-check, lint (`--max-warnings 0`), and the full test suite (**1164 passing**, incl. 66 session-router cases) are all green locally on the head SHA.

---

## 1. Spec compliance

**Spec: ✅**

The PR does exactly what it claims and nothing more:

- **Split of the system-event guard** — `src/session-router.ts:451-460`. The old `if (msg.payload.event) return null;` becomes `handleGroupMdEvent(msg); return null;`. A GROUP.md-change event invalidates that group's cache entry; every other system event still routes to `null` unchanged. No user-facing reply is ever produced for an event (verified: `route(...)` returns `null` and `sendMessage` is not called in all new tests).
- **Payload is never trusted as content** — `handleGroupMdEvent` (`session-router.ts:375-393`) only ever calls `groupMdCache.invalidate(groupNo)`. It never reads `content`/`version`/`url` from the event body. The authoritative copy is re-fetched by the existing resolver over the bot token against the SSRF-validated `apiUrl` (unchanged from P2-A). This matches the stated security model.
- **Group identity from `event.group_no`, channel fallback** — `session-router.ts:379-383`. Primary key is `event.group_no`; the channel is used only when `group_no` is absent AND the channel is group-like (the mention-free-group path). Consistent with the resolver's keying (see §2).
- **Provisional event-type literal, overridable** — `src/group-md-events.ts` (`DEFAULT_GROUP_MD_EVENT_TYPES = ['group_md_updated']`, overridable via `config.serverMdEventTypes`); config plumbed through `mergeConfig` and `resolveBotConfigs` (`src/config.ts`). The `⚠️ PROVISIONAL` note is present and honest.
- **Rollback gate** — `handleGroupMdEvent` returns early when `!serverMd || !groupMdCache`, so flag-off / no-cache is a genuine no-op (test covers both).

No over-build (no new feature/flag/field beyond the calibration seam the design calls for) and no deviation from the linked-issue approach.

## 2. Code quality

**Quality: Approved**

Strong points:

- **Cache-key consistency holds.** `invalidate(groupNo)` (`group-md-cache.ts:104`) and the resolver's `cache.set(groupNo, …)` (`group-md.ts:109`) key on the same value: the resolver computes `groupNo = extractParentGroupNo(channelId)`, and the event path uses `event.group_no` directly (correct — the event arrives on a DM/system channel, so the arriving `channel_id` is the sender, not the group) with `extractParentGroupNo(channel_id)` only on the group-like fallback. Both sides pass through the same `isSafeGroupNo` gate, so acceptance is symmetric — no silent-miss mismatch.
- **Total / best-effort.** `isGroupMdUpdateEvent` null-guards `type` and empty-string; `invalidate()` is itself total (validates the id, never throws); the extra `try/catch` around it is defensive for a future non-total cache. A malformed event cannot wedge the router — the caller drops it regardless.
- **Tests lock in both branches** (md → invalidate; join/leave/typeless → untouched), the DM-delivery reality, the mention-free-group channel fallback, the `serverMd`-off rollback, the no-cache no-op, and the `serverMdEventTypes` calibration seam. This is the right coverage for the split.

### Security note (PR is security-sensitive) — verified, not blocking

The forged-event concern is bounded and I confirmed it: `invalidate()` is an O(1) in-memory `Map.delete` with **no network I/O and no filesystem access**; the re-fetch it enables happens only on a subsequent genuine user turn through the existing authenticated resolver. So a forged/replayed `group_md_updated` event cannot inject content and cannot by itself amplify into a fetch storm — the worst case is one redundant authenticated re-fetch of the real value. The design claim holds.

### Items for a human to keep an eye on (non-blocking)

- **P2 (blind spot, shared across reviewers):** whether the inbound gateway scrubs `payload.event` on ordinary user chat frames is *out of this diff*. It doesn't change this PR's verdict (invalidation is non-destructive, as above), but it's the one thing worth a human confirming end-to-end before relying on event-driven refresh in production, since it governs who can trigger invalidations at all.
- **P2 (follow-up, not this PR):** the resolver has no negative-cache / in-flight coalescing, so a group with no server GROUP.md re-fetches (and error-logs) on every turn, and concurrent misses can thunder. This lives entirely in `group-md.ts` / `group-md-cache.ts`, which are **#172's** files — not touched here. Flagging so it lands on #172's follow-up rather than being lost; it is not a #173 blocker.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The change is tightly scoped, the security model is sound and verified, and test coverage matches the split. The provisional `event.type` literal is correctly externalized so real-event calibration needs no code change. Merge after #172, as the PR notes.

Jerry-Xin (dismissed):
**APPROVE** — event-driven GROUP.md cache refresh (P2-B). Re-confirmed at head `656f2077`: code is unchanged and byte-identical to the previously verified state. The server system event only forces `invalidate(groupNo)` on the in-memory Map; a forged event cannot inject content — the next `resolve` re-fetches the authoritative copy over the authenticated, SSRF-validated apiUrl, gated by the `isSafeGroupNo` guard. No #172 regression: cache stays memory-only (no disk import), TTL intact (5min backstop), token not leaked. Handler is best-effort with try/catch and never wedges the message loop; malformed events no-op.

🔴 Blocking: None.

💬 Non-blocking notes:
- Event-handler reachability caveat: `handleGroupMdEvent` IS dispatched in `processMessage` (not dead code), but the production event path on the system-channel / bot-sender guards is not exercised by tests. Low risk because `serverMd` is flag-off by default and the ≤5min TTL is a correct staleness backstop.
- Thundering-herd on concurrent misses is pre-existing (not introduced here).
- Integration test gap: the end-to-end event → invalidate → re-fetch path lacks a dedicated integration test; unit coverage for the cache/guard logic is present.

✅ Highlights:
- Clean separation: invalidation-only design means an untrusted event can at worst force a redundant authenticated re-fetch of the real value — never poison the cache.
- Defensive handler (try/catch, non-throwing, no-op on malformed/absent group_no) keeps the router robust.
- TTL preserved as an independent freshness backstop, so correctness does not depend solely on the event path.

Jerry-Xin (dismissed):
Re-review at head 656f20773ff1 (event-driven GROUP.md cache refresh, P2-B). Prior approvals on this head were auto-dismissed by the merge-base shift as #172 landed underneath; the code is unchanged and all required checks are green, so re-affirming.

Verdict: APPROVE. No blocking issues.

Trust boundary (byte-verified correct): the group_md_updated event is signal-only — it triggers invalidate(groupNo) on the in-memory-only cache and nothing else. A forged/spoofed event can at most force a redundant authenticated re-fetch of the authoritative GROUP.md over the SSRF-validated apiUrl (isSafeGroupNo slug guard, encodeURIComponent); it cannot inject or poison cached content. No regression of the #172 memory-only cache (no fs/disk import, TTL intact, tokens not leaked in error paths). Handler is guarded by try/catch and degrades to a no-op on malformed events, so it never crashes the message loop.

Non-blocking notes:
- Reachability calibration: handleGroupMdEvent is dispatched in processMessage but sits after the isSupportedChannel and bot-like-sender guards, so if the server emits these events on a system channel (type 8) or from a bot-like sender, the handler no-ops and P2-B degrades to the 5-minute TTL backstop. Worth confirming the real production channel_type + sender against a captured-event fixture. Non-blocking because serverMd is flag-off by default and the TTL backstop bounds staleness to <=5 min.
- No in-flight coalescing on concurrent misses (thundering-herd), pre-existing from #172.
- The group_md_updated event literal is provisional (overridable via serverMdEventTypes).

Highlights: clean signal-only invalidation design, memory-only cache preserved, authoritative re-fetch on miss, good fallback/routing coverage.

OctoBoooot (approved):
## Re-review (delta): #173 — rebase, code unchanged

**Verdict:** Approve. New head `d82c75abc0a1` is a pure rebase of `656f20773ff1` (absorbs the now-merged #172 cache layer); the P2-B source is byte-identical to what I already reviewed.

Verified `group-md-events.ts`, `session-router.ts`, `index.ts`, and `config.ts` are byte-for-byte unchanged between the prior head and this one — the diff shrank from 11 files/1042 to 5 files/294 only because #172's stacked commits are now in the base, not because anything in P2-B changed. My prior analysis carries over verbatim:

- Trust boundary correct: the event is signal-only (`isGroupMdUpdateEvent` reads only `type`/`group_no`), the sole effect is `cache.invalidate(groupNo)`, and content is always re-fetched over the authenticated, SSRF-validated path. A forged event can at most force a redundant authenticated re-fetch of the real value — never poison the cache. No #172 regression (still memory-only).
- `handleGroupMdEvent` is gated + try/catch, so a malformed event degrades to a no-op and can't wedge the router.
- CI is green on this head (check-sprint passes; the merge-base drift that auto-dismissed prior approvals is resolved by the rebase).

### Still open (non-blocking, carried from prior review)
- Reachability: `handleGroupMdEvent` runs after `isSupportedChannel` and the `looksLikeBot` gate, so if the server delivers GROUP.md events on a system channel (type 8) or from a bot-like sender, the handler never fires and P2-B degrades to the 5-min TTL backstop. Same class as the already-flagged provisional `event.type` — worth confirming against a real captured event, but flag-off default + TTL fallback keep it non-blocking. Jerry-Xin's re-review 🟡 (production event path untested) converges on this.

Clean to merge on this head.

Jerry-Xin (approved):
✅ **APPROVE** (re-affirm after rebase — pure rebase, no code change)

**This is a pure rebase.** The branch was rebased onto a fresh `main` to clear the earlier merge-base-drift auto-dismissal. Byte-verification confirms none of #173's own work changed:

- All 5 of this PR's files are **byte-identical** to the previously-approved tree (`656f2077`), proven by identical git blob hashes:
  - `src/group-md-events.ts`, `src/session-router.ts`, `src/config.ts`, `src/index.ts`, `src/__tests__/session-router.test.ts` — all unchanged.
- `git diff 656f2077 <new-head>` restricted to those files is **empty**.
- The files that *do* differ between the old and new head (`src/octo/api.ts`, `src/session-store.ts`, and the `e2e`/`session-store`/`thread-api` tests) are **merge-base drift** from the stacked cache PRs pulled in by the rebase — not part of this PR's delta. The true additive delta over the new merge-base is exactly the same 5 files (group-md-events +52, session-router split guard +71, config `serverMdEventTypes` +13, index wiring +5, tests +153).

**Prior APPROVE findings still hold (byte-verified, unchanged):**
- Event is **signal-only**: a GROUP.md-change event calls `invalidate(groupNo)` on the in-memory Map; it never trusts payload content. Next resolve re-fetches the authoritative copy over the authenticated bot token against the SSRF-validated `apiUrl`, gated by the `isSafeGroupNo` guard. Worst case of a forged event is a redundant authenticated re-fetch of the real value — no content injection.
- No regression to the cache layer: memory-only, no disk import, TTL intact, tokens not leaked.
- Malformed events are handled by a try/catch no-op.
- Non-md system events remain dropped unchanged.

**Non-blocking (unchanged from prior review):**
- Handler-reachability caveat: `isSupportedChannel` + bot-sender guards may drop system-channel/bot-originated events; the cache TTL is the backstop.
- Potential thundering-herd on simultaneous invalidation + re-fetch.
- Integration-test gap for the end-to-end server-event → refresh path (unit coverage present).

CI fully green (type-check, ESLint zero-warnings, Vitest, coverage, secret-scan, dependency-review, common-ancestor). Mergeable. Re-affirming APPROVE.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#173 Review Report — feat(group-md): event-driven GROUP.md cache refresh (P2-B)
**Head SHA:** `d82c75abc0a12d88c2bc7b580109dabfef7e1ae6`
**Base:** stacked on PR#172 (P2-A cache + `invalidate()` hook)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| Security model (invalidate-then-refetch, never trust event payload) | ✅ | `group-md-events.ts` L1–27 (design doc); `session-router.ts` `handleGroupMdEvent` never reads event body as content |
| `isGroupMdUpdateEvent` guards on type literal | ✅ | `group-md-events.ts` L46–51: rejects `undefined`, non-string, empty type; checks `eventTypes.includes()` |
| `group_no` resolution (event → channel fallback) | ✅ | `session-router.ts` L384–388: `event.group_no ?? ''` → `extractParentGroupNo(channel_id)` if group-like → `isSafeGroupNo` filter in `cache.invalidate()` |
| `groupMdCache` always in scope at construction site | ✅ | `index.ts` L221: created unconditionally before `SessionRouter` at L278 |
| `serverMd` feature-flag gate | ✅ | `handleGroupMdEvent` L381: `if (!this.config.serverMd || !this.groupMdCache) return;` |
| `invalidate()` is total (never throws) | ✅ | `group-md-cache.ts` L98–101: `isSafeGroupNo` guard + `this.mem.delete()` |
| Config merge for `serverMdEventTypes` | ✅ | `config.ts` L386: `override.serverMdEventTypes ?? base.serverMdEventTypes` — consistent with all other array/scalar fields |
| `serverMdEventTypes` undefined → default types | ✅ | `handleGroupMdEvent` passes `this.config.serverMdEventTypes` (possibly `undefined`) to `isGroupMdUpdateEvent` which defaults to `DEFAULT_GROUP_MD_EVENT_TYPES` |
| Event guard placement (after mention gate, before rate limit) | ✅ | `session-router.ts` L454–458: inside `if (msg.payload.event)` block, after mention gate at L438, before rate limit at L461 |
| Test coverage | ✅ | 9 new tests covering: md event invalidate, non-md drop, no-type drop, no-group_no DM no-op, serverMd off, no cache wired, mention-free group fallback, event type override |

## 2. Findings

### No P0/P1 issues found.

### P2 — Notes

**P2-1: Provisional event type literal (documented, mitigated)**
- The default `group_md_updated` literal is unconfirmed against a real server event. This is clearly documented (`⚠️ PROVISIONAL` in `group-md-events.ts` and PR description) and the `serverMdEventTypes` config override provides a calibration seam that requires no code change.
- **Severity:** informational — acknowledged risk, not a code defect.

**P2-2: Empty `serverMdEventTypes` array silently disables all md events**
- If a user configures `serverMdEventTypes: []`, `isGroupMdUpdateEvent` receives `[]` and `[].includes(type)` is always `false`. The event-refresh path becomes a silent no-op.
- **Diff-scope:** new (this PR adds the config). However, this is consistent with how other config arrays work in the project (empty = disabled) and is not a defect — just a configuration footgun worth a config-validation note in documentation.

## 3. Suggestions

- Consider adding a config validation warning in `loadConfig` / `resolveBotConfigs` if `serverMdEventTypes` is set to an empty array while `serverMd` is `true`. Low priority.

## 4. Additional Observations

- The mention gate correctly filters GROUP.md events on normal group channels (un-mentioned group messages die there). The channel fallback path is only reachable for mention-free groups — a narrow and well-documented edge case.
- Forged-event DoS surface is bounded: invalidation is a Map delete, the re-fetch only happens on the next real turn (not proactively), and the system event path is not rate-limited but also not user-addressable.
- The `try/catch` around `this.groupMdCache.invalidate(groupNo)` is defense-in-depth for future cache implementations — today `invalidate()` is documented total. Acceptable.

## 5. Data-flow Tracing

| Consumed data | Upstream source | Reaches consumption point? |
|---------------|----------------|---------------------------|
| `msg.payload.event.type` | WebSocket inbound → `MessagePayload.event` (types.ts L75) | ✅ → `isGroupMdUpdateEvent` in `handleGroupMdEvent` |
| `event.group_no` | Server event payload (`{ type, version, updated_by, group_no, short_id }`) | ✅ → `groupNo` variable, then `cache.invalidate(groupNo)` |
| `msg.channel_id` (fallback) | Inbound message | ✅ → `extractParentGroupNo()` strips thread suffix for CommunityTopic channels |
| `config.serverMdEventTypes` | Config file → `mergeConfig` / `resolveBotConfigs` | ✅ → `isGroupMdUpdateEvent` second arg (defaults to `DEFAULT_GROUP_MD_EVENT_TYPES` when undefined) |
| `groupMdCache` | `index.ts` L221, created unconditionally | ✅ → SessionRouter constructor L278, stored as `this.groupMdCache` |

## 6. Blindspot Checklist (R5 — security_sensitive PR)

- **C1 双路径 parity:** N/A — single invalidation site, no add/remove or create/delete symmetry at play. The event type match is a single function with one call site.
- **C2 control-flow ordering / 嵌套复用:** Clear — `handleGroupMdEvent` has exactly one call site (`processMessage` L455), no nested/reused invocation. The `isGroupMdUpdateEvent` type check uses `Array.includes()` — no regex or escape-sensitive pattern to probe with non-canonical input.
- **C3 授权边界 ≠ 能力边界:** Clear — cache invalidation is non-destructive (Map delete). The subsequent re-fetch uses the same authenticated bot-token + SSRF-validated apiUrl path as the initial fetch. No new capability is exposed.
- **C4 授权生命周期 / 容器-成员状态级联:** N/A — this is a cache invalidation, not an authorization change. The re-fetch path (outside this PR's scope) uses the same auth chain.
- **C5 build/note 通过 ≠ 运行期路径正确:** N/A — standard TypeScript/Node.js project, no browser extension/CLI/packaging concerns. No relative paths or origin changes.
- **C6 治理/策略/安全文档自洽性:** N/A — code change, not a governance/policy document.

## 7. Cross-round Blocker Re-review (R6)

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. The security model (invalidate-then-refetch, never trust event payload) is sound and well-documented. Data flows are correct, test coverage is thorough, and the implementation follows established project conventions. The two P2 notes (provisional event type literal and empty-array config footgun) are informational and non-blocking.

yujiawei (approved):
# Code Review — PR #173 (cc-channel-octo)

Reviewed at head SHA `d82c75abc0a12d88c2bc7b580109dabfef7e1ae6`, base `main` (this branch sits on top of the already-merged #172, verified: #172's tip is an ancestor of head). Independent, security-sensitive review with two additional advisory perspectives absorbed.

Verification run locally at head: `tsc --noEmit` clean, `eslint --max-warnings 0` clean, full suite green (1172 passing, incl. the 8 new P2-B cases).

## 1. Spec compliance

**Spec: ✅**

The change does exactly what the PR describes and nothing more.

- Not missing: the system-event drop guard is split by `event.type`; a GROUP.md-change event invalidates that group's in-memory cache entry, every other system event is dropped unchanged (`src/session-router.ts:460-462`, `handleGroupMdEvent` at `:376-394`).
- Not over-built: the classifier (`src/group-md-events.ts`), the config seam (`serverMdEventTypes`, plumbed symmetrically through `mergeConfig`/`resolveBotConfigs` in `src/config.ts:397,633`), and the router wiring (`src/index.ts:278`) are all in service of that one behavior. No unrelated flags or endpoints added.
- Not diverging: the event payload is never read as content — invalidate-then-refetch only, re-fetched over the authenticated bot token by the existing resolver. This matches the stated design and the P2-A memory-only trust model.

The provisional `event.type` literal (`group_md_updated`) is correctly flagged as needing calibration and is overridable at runtime, so calibration needs no code change. Good.

## 2. Code quality

**Quality: Approved**

Design is sound: invalidation is non-destructive, so trusting the event-supplied `group_no` cannot inject content — the worst case is a redundant authenticated re-fetch of the real value. The cache key is consistent between the resolver (`extractParentGroupNo(channelId)` for get/set, `src/group-md.ts:85`) and the invalidation path (`event.group_no`, or the same `extractParentGroupNo` fallback, `src/session-router.ts:380-383`). `invalidate()` is total and id-validated. Tests cover both branches of the split plus the flag-off, no-cache, no-group_no, channel-fallback, and override seams.

Three non-blocking findings, none of which requires a change before merge:

**[P2] Numeric `group_no` in a forged/malformed event silently misses invalidation** — `src/session-router.ts:380`. `let groupNo = event?.group_no ?? ''` is not string-coerced. The wire type declares `group_no?: string`, but nothing validates the inbound JSON, so a payload with `"group_no": 123` yields a `number` at runtime. `groupNo === ''` is false, then `GroupMdCache.invalidate(123)` calls `mem.delete(123)` which misses the string key `"123"` (confirmed: the `/^[a-zA-Z0-9._-]+$/` guard even passes because `RegExp.test` coerces to string, so the id check doesn't catch it either). Effect: a genuine GROUP.md change delivered with a numeric `group_no` would not invalidate, and the group serves stale instructions until the 5-min TTL backstop expires. Bounded by TTL, hence P2, but a one-line `String(event?.group_no ?? '')` closes it cleanly.

**[P2] Event-driven invalidation is not rate-limited** — `src/session-router.ts:460-462` runs before the rate-limit check at `:475`. A client that can emit `group_md_updated` events (e.g. over a DM) can keep a target group's cache perpetually invalidated, forcing a re-fetch on every subsequent processed turn. Real amplification is bounded: the event branch itself only does an O(1) `Map.delete` (no network), and the expensive authenticated re-fetch happens on the next *processed* message, which IS rate-limited — so this is at most one extra upstream GET per already-allowed turn, i.e. cache-effectiveness degradation, not an unbounded outbound flood. Note also the event guard sat ahead of rate-limiting before this PR too; P2-B only gives the pre-drop branch a side effect. Worth a per-group invalidation dedup/debounce (e.g. skip if version unchanged) if server events ever arrive on an untrusted path, but not a merge blocker.

**[P2] Channel fallback for `group_no` is effectively dead for a group's own channel** — `src/session-router.ts:381-383`. The `isGroupLike` + `extractParentGroupNo(channel_id)` fallback only runs when `group_no` is absent, but a group-channel event with no @mention and not in `mentionFreeGroups` returns `null` at the mention gate (`:438-452`) before reaching the event branch. This is consistent with the PR's own XIN-173 assumption (these events arrive on a DM/system channel), and the mention-free-group path that the tests exercise does reach it — so the fallback isn't fully dead, just narrow. Fine as documented; flagging so the assumption stays visible if event delivery semantics change.

### Cross-file / coverage notes
- Config schema is not runtime-validated: `serverMdEventTypes` is taken straight from `JSON.parse` with no `Array.isArray` check (`src/config.ts:370`). If an operator sets it to a string instead of an array, `.includes()` degrades to substring matching. Pre-existing pattern for this config file (other fields aren't validated either), so not introduced here — noting for a future hardening pass.
- Not exercised here (out of diff, would want a human eye if event delivery changes): the upstream server's own rate limits on `getGroupMd`, and concurrent invalidate/re-fetch interleaving on the cache.

## 3. Verdict

**APPROVED.** Spec is clean and the security model (invalidate-then-refetch, memory-only trust root) holds. No P0/P1. The three P2s are correctness/robustness polish — the `String()` coercion is the one most worth doing, ideally in this PR, but none blocks merge.

Extra care flagged for the human reviewer per the security-sensitive classification: the two things genuinely worth a manual eye are (a) whether GROUP.md change events can ever reach this bot over a path a non-operator controls (that reframes the unrated-invalidation P2), and (b) confirming the real server `event.type` literal so the provisional default gets calibrated.

PR #174, opened 2026-07-01
feat(group-md): owner-gated GROUP.md write-back

Jerry-Xin:
## Summary

Adds an owner-gated GROUP.md write-back tool so a bot can persist group instruction changes back to the server, completing the read+write GROUP.md chain (server-first fetch/cache on top, write-back on top of that).

## Changes

- `feat(group-md): owner-gated GROUP.md write-back MCP tool` — exposes `update_group_md`, registered per group session at message time, gated on `mdWriteback && isGroup`. Only the bot owner uid may invoke it. Content is capped at 10240 bytes (full-replace semantics).
- Builds on the server-first GROUP.md fetch + memory-only cache with TTL (read path) so read and write share one source of truth.

## Behavior verified (real environment)

- Owner-origin write-back on a real group: owner @-triggers the bot, the tool calls the server PUT, server GROUP.md is updated with `updated_by = <bot id>` and the version bumps. Confirmed by reading the server GROUP.md after the write.
- Owner gate: non-owner triggers are rejected (no PUT).
- Byte cap enforced at 10240.

Part of #88

### DIFF
--- a/src/group-md-tool.ts
+++ b/src/group-md-tool.ts
@@ -1,0 +1,135 @@
+/**
+ * P2-C: GROUP.md write-back tool — an in-process MCP server letting the agent
+ * persist an updated GROUP.md back to the server. The tool surfaces to the model
+ * as `mcp__group_md__update_group_md`.
+ *
+ * The server is built PER TURN (`createGroupMdToolServer`) with the current
+ * message's channel coords + the bot owner uid, so:
+ *  - the write targets the PARENT group of the channel the agent is in
+ *    (`extractParentGroupNo` — a thread shares its parent group's GROUP.md);
+ *  - invocation is GATED to the bot owner (registerBot.owner_uid). The group's
+ *    octo_tag token has server-side write permission, but the agent is driven by
+ *    untrusted IM users, so this owner gate — not LLM judgment, and independent
+ *    of the token's group-role permission — is what stops a prompt-injected
+ *    agent from rewriting the operator's trusted GROUP.md from any chat.
+ *
+ * Concurrency, the byte ceiling, and the cache refresh are owned by the shared
+ * {@link GroupMdWriteback} coordinator (group-md-writeback.ts); this layer is
+ * only the owner-gate policy + the MCP surface.
+ */
+
+import { z } from 'zod';
+import { createSdkMcpServer, tool } from '@anthropic-ai/claude-agent-sdk';
+import { extractParentGroupNo } from './octo/channel-id.js';
+import {
+  GroupMdContentTooLargeError,
+  MAX_GROUP_MD_CONTENT_BYTES,
+  type GroupMdWriteback,
+} from './group-md-writeback.js';
+
+/** MCP server name; the tool surfaces as `mcp__group_md__update_group_md`. */
+export const GROUP_MD_TOOL_SERVER_NAME = 'group_md';
+
+/** Raw coords of the session invoking the tool — gates the call + targets the group. */
+export interface GroupMdSessionCoords {
+  /** Full channelId (may be a `<groupNo>____<shortId>` thread composite). */
+  channelId: string;
+  fromUid: string;
+  fromName?: string;
+}
+
+/** Shared deps the tool needs to perform a write-back. */
+export interface GroupMdToolDeps {
+  writeback: GroupMdWriteback;
+  apiUrl: string;
+  botToken: string;
+}
+
+function jsonResult(value: unknown): { content: Array<{ type: 'text'; text: string }> } {
+  return { content: [{ type: 'text', text: JSON.stringify(value, null, 2) }] };
+}
+function errResult(msg: string): { content: Array<{ type: 'text'; text: string }>; isError: true } {
+  return { content: [{ type: 'text', text: `Error: ${msg}` }], isError: true };
+}
+
+/**
+ * Build the GROUP.md tool DEFINITIONS for one agent turn. Exported separately
+ * from the server so tests can invoke the handler directly. `coords` targets the
+ * group + supplies the caller uid; `ownerUid` is the owner gate.
+ */
+export function buildGroupMdTools(
+  deps: GroupMdToolDeps,
+  coords: GroupMdSessionCoords,
+  ownerUid: string,
+) {
+  const isOwner = coords.fromUid === ownerUid && ownerUid !== '';
+
+  return [
+    tool(
+      'update_group_md',
+      'Persist new GROUP.md content for THIS group on the server (it becomes the ' +
+        "group's trusted operator instructions on the next turn). `content` is the " +
+        'FULL replacement document, not a diff. Hard limit: 10240 bytes UTF-8. Only ' +
+        'the bot owner may call this; a non-owner request is rejected. Last write ' +
+        'wins server-side — compose the complete updated document, do not assume a ' +
+        'concurrent edit merged.',
+      {
+        content: z
+          .string()
+          .describe('Full replacement GROUP.md document (≤10240 bytes UTF-8).'),
+      },
+      async (args) => {
+        try {
+          if (!isOwner) {
+            return errResult('Only the bot owner can update GROUP.md.');
+          }
+          const groupNo = extractParentGroupNo(coords.channelId);
+          if (!groupNo) {
+            return errResult('Could not resolve a group number from this channel.');
+          }
+          // Surface a friendly over-limit message before the coordinator throws
+          // (it re-checks as the authoritative boundary; this is just for UX).
+          const bytes = Buffer.byteLength(args.content, 'utf-8');
+          if (bytes > MAX_GROUP_MD_CONTENT_BYTES) {
+            return errResult(
+              `content is ${bytes} bytes, over the ${MAX_GROUP_MD_CONTENT_BYTES}-byte ` +
+                `UTF-8 limit — trim it before writing (the server would reject it).`,
+            );
+          }
+          const res = await deps.writeback.writeBack({
+            apiUrl: deps.apiUrl,
+            botToken: deps.botToken,
+            groupNo,
+            content: args.content,
+          });
+          return jsonResult({
+            updated: { groupNo: res.groupNo, version: res.version, bytes: res.bytes },
+          });
+        } catch (err) {
+          if (err instanceof GroupMdContentTooLargeError) {
+            return errResult(
+              `content is ${err.bytes} bytes, over the ${MAX_GROUP_MD_CONTENT_BYTES}-byte UTF-8 limit.`,
+            );
+          }
+          return errResult(err instanceof Error ? err.message : String(err));
+        }
+      },
+    ),
+  ];
+}
+
+/**
+ * Build the GROUP.md write-back MCP server for one agent turn. `coords` targets
+ * the group + supplies the caller uid; `ownerUid` is the owner gate.
+ */
+export function createGroupMdToolServer(
+  deps: GroupMdToolDeps,
+  coords: GroupMdSessionCoords,
+  ownerUid: string,
+) {
+  return createSdkMcpServer({
+    name: GROUP_MD_TOOL_SERVER_NAME,
+    version: '1.0.0',
+    tools: buildGroupMdTools(deps, coords, ownerUid),
+  });
+}
--- a/config.example.json
+++ b/config.example.json
@@ -11,6 +11,9 @@
 
   "_comment_server_md": "P2-A: set serverMd=true to fetch each group's instructions from the server GROUP.md API (GET /v1/bot/groups/{groupNo}/md) first, falling back to the local groupConfigDir file on any failure (404 / network / no content). The fetched copy is cached IN MEMORY ONLY (never on disk — it is injected as a trusted system prompt, and a disk cache the gateway user could write would be a poisoning vector). serverMdTtlMs (default 300000 = 5 min) bounds how stale a cached entry may get before re-fetch. Default serverMd=false = pure local-file behavior (flip off to roll back).",
 
+  "_comment_md_writeback": "P2-C: set mdWriteback=true to give the agent an in-process MCP tool (mcp__group_md__update_group_md) that writes GROUP.md back to the server (PUT /v1/bot/groups/{groupNo}/md). Invocation is owner-gated (only the bot owner uid may call it); content is capped at 10240 bytes UTF-8 locally; write-backs to the same group are serialized in-process (the server has no compare-and-swap, so this prevents this gateway from racing itself — cross-source last-write-wins still applies). Default false = no write-back tool (GROUP.md is read-only from the gateway). Independent of serverMd.",
+
+
   "sdk": {
     "_comment_anthropic_base_url": "Q1: Optional self-hosted Claude API gateway base URL. Forwarded to the SDK subprocess (scoped). Must be https:// (or http://localhost for dev) — SSRF-validated at boot. Omit to use Anthropic's public endpoint.",
     "_comment_allowed_tools": "Q2: either '*' (allow every tool the SDK exposes) or an explicit string[] whitelist.",
--- a/src/__tests__/group-md-tool.test.ts
+++ b/src/__tests__/group-md-tool.test.ts
@@ -1,0 +1,157 @@
+/**
+ * P2-C: GROUP.md write-back tool tests (group-md-tool.ts).
+ *
+ * Drives the tool handler directly via buildGroupMdTools (the MCP server keeps
+ * tools private). The SDK's `tool`/`createSdkMcpServer` are mocked to plain
+ * passthroughs so the module loads without the real SDK.
+ *
+ * Covers the acceptance gates at the policy layer:
+ *   - owner-gate: only the bot owner may write (non-owner / empty-owner rejected,
+ *     no server call);
+ *   - ≤10240-byte UTF-8 rejection surfaces a clean error (no server call);
+ *   - a successful write hits A's updateGroupMd client and refreshes the cache;
+ *   - a thread channelId resolves to its PARENT group number;
+ *   - concurrent tool calls for the same group serialize through the shared
+ *     coordinator (no overlap, no lost write).
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+vi.mock('@anthropic-ai/claude-agent-sdk', () => ({
+  createSdkMcpServer: (opts: { name: string }) => ({ type: 'sdk', name: opts.name, instance: {} }),
+  tool: (name: string, description: string, inputSchema: unknown, handler: unknown) => ({ name, description, inputSchema, handler }),
+}));
+
+import {
+  buildGroupMdTools,
+  createGroupMdToolServer,
+  GROUP_MD_TOOL_SERVER_NAME,
+  type GroupMdSessionCoords,
+  type GroupMdToolDeps,
+} from '../group-md-tool.js';
+import { GroupMdWriteback, MAX_GROUP_MD_CONTENT_BYTES, type UpdateGroupMdFn } from '../group-md-writeback.js';
+import { GroupMdCache } from '../group-md-cache.js';
+
+const OWNER = 'owner-uid';
+const API = 'https://api.example.com';
+const TOKEN = 'bot-token';
+
+const delay = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
+
+function coords(over: Partial<GroupMdSessionCoords> = {}): GroupMdSessionCoords {
+  return { channelId: 'grp-1', fromUid: OWNER, fromName: 'Owner', ...over };
+}
+
+function getTool(deps: GroupMdToolDeps, c: GroupMdSessionCoords, owner = OWNER) {
+  const t = buildGroupMdTools(deps, c, owner).find((x) => x.name === 'update_group_md');
+  if (!t) throw new Error('tool update_group_md not found');
+  return t as { name: string; handler: (args: { content: string }, extra: unknown) => Promise<{ content: Array<{ text?: string }>; isError?: boolean }> };
+}
+function text(r: { content: Array<{ text?: string }> }): string {
+  return r.content.map((c) => c.text ?? '').join('');
+}
+
+describe('group-md-tool', () => {
+  let cache: GroupMdCache;
+  let putFn: ReturnType<typeof vi.fn<UpdateGroupMdFn>>;
+  let deps: GroupMdToolDeps;
+
+  beforeEach(() => {
+    cache = new GroupMdCache();
+    let version = 0;
+    putFn = vi.fn<UpdateGroupMdFn>(async () => ({ version: ++version }));
+    deps = { writeback: new GroupMdWriteback(cache, putFn), apiUrl: API, botToken: TOKEN };
+  });
+
+  it('createGroupMdToolServer builds an MCP server named "group_md"', () => {
+    const s = createGroupMdToolServer(deps, coords(), OWNER);
+    expect(GROUP_MD_TOOL_SERVER_NAME).toBe('group_md');
+    expect((s as { name: string }).name).toBe('group_md');
+  });
+
+  it('owner write: PUTs the content and refreshes the cache', async () => {
+    const r = await getTool(deps, coords()).handler({ content: 'new persona' }, {});
+    expect(r.isError).toBeFalsy();
+    expect(putFn).toHaveBeenCalledWith(
+      expect.objectContaining({ apiUrl: API, botToken: TOKEN, groupNo: 'grp-1', content: 'new persona' }),
+    );
+    expect(text(r)).toContain('"version": 1');
+    expect(cache.get('grp-1')).toEqual({ content: 'new persona', version: 1, updated_at: null });
+  });
+
+  it('non-owner write is rejected and never calls the server', async () => {
+    const r = await getTool(deps, coords({ fromUid: 'rando' })).handler({ content: 'x' }, {});
+    expect(r.isError).toBe(true);
+    expect(text(r)).toMatch(/only the bot owner/i);
+    expect(putFn).not.toHaveBeenCalled();
+    expect(cache.get('grp-1')).toBeUndefined();
+  });
+
+  it('empty owner uid gates everyone out (no owner → unusable)', async () => {
+    const r = await getTool(deps, coords(), '').handler({ content: 'x' }, {});
+    expect(r.isError).toBe(true);
+    expect(putFn).not.toHaveBeenCalled();
+  });
+
+  it('over-limit content is rejected with a clear error and no server call', async () => {
+    const tooBig = 'a'.repeat(MAX_GROUP_MD_CONTENT_BYTES + 1);
+    const r = await getTool(deps, coords()).handler({ content: tooBig }, {});
+    expect(r.isError).toBe(true);
+    expect(text(r)).toMatch(/over the 10240-byte/i);
+    expect(putFn).not.toHaveBeenCalled();
+    expect(cache.get('grp-1')).toBeUndefined();
+  });
+
+  it('content exactly at the limit is accepted', async () => {
+    const atLimit = 'a'.repeat(MAX_GROUP_MD_CONTENT_BYTES);
+    const r = await getTool(deps, coords()).handler({ content: atLimit }, {});
+    expect(r.isError).toBeFalsy();
+    expect(putFn).toHaveBeenCalledTimes(1);
+  });
+
+  it('a thread channelId writes to its PARENT group number', async () => {
+    const r = await getTool(deps, coords({ channelId: 'grp-7____thread9' })).handler({ content: 'doc' }, {});
+    expect(r.isError).toBeFalsy();
+    expect(putFn).toHaveBeenCalledWith(expect.objectContaining({ groupNo: 'grp-7' }));
+    expect(cache.get('grp-7')).toMatchObject({ content: 'doc' });
+  });
+
+  it('surfaces the client error when the PUT fails (cache untouched)', async () => {
+    putFn.mockRejectedValueOnce(new Error('Octo API PUT failed (400): err.server.bot_api.content_too_large'));
+    const r = await getTool(deps, coords()).handler({ content: 'doc' }, {});
+    expect(r.isError).toBe(true);
+    expect(text(r)).toMatch(/PUT failed/);
+    expect(cache.get('grp-1')).toBeUndefined();
+  });
+
+  it('serializes concurrent tool calls for the same group through the shared coordinator', async () => {
+    let active = 0;
+    let maxActive = 0;
+    let version = 0;
+    const order: string[] = [];
+    const slow: UpdateGroupMdFn = async (p) => {
+      active++;
+      maxActive = Math.max(maxActive, active);
+      await delay(10);
+      order.push(p.content);
+      active--;
+      return { version: ++version };
+    };
+    const shared = new GroupMdWriteback(cache, slow);
+    const d: GroupMdToolDeps = { writeback: shared, apiUrl: API, botToken: TOKEN };
+
+    // Two separate per-turn tool instances (as in production) borrowing the same
+    // shared coordinator — concurrency must still be contained.
+    const tA = getTool(d, coords());
+    const tB = getTool(d, coords());
+    const results = await Promise.all([
+      tA.handler({ content: 'A' }, {}),
+      tB.handler({ content: 'B' }, {}),
+    ]);
+
+    expect(results.every((r) => !r.isError)).toBe(true);
+    expect(maxActive).toBe(1);
+    expect(order).toHaveLength(2); // both writes executed; none lost
+    expect(cache.get('grp-1')?.version).toBe(2);
+  });
+});
--- a/src/__tests__/group-md-writeback.test.ts
+++ b/src/__tests__/group-md-writeback.test.ts
@@ -1,0 +1,184 @@
+/**
+ * P2-C: GROUP.md write-back coordinator tests (group-md-writeback.ts).
+ *
+ * Covers the four XIN-201 contract obligations at the mechanism layer:
+ *   - ≤10240-byte UTF-8 cap rejected LOCALLY (no server PUT) — byte, not char,
+ *     counting on the boundary;
+ *   - successful PUT writes the new content/version back into A's in-memory
+ *     cache (so the next resolve serves what we wrote, not a stale/refetched copy);
+ *   - per-groupNo serialization: concurrent write-backs to the SAME group never
+ *     overlap and never lose a write (no-CAS / last-write-wins is contained
+ *     within this gateway), while different groups still run concurrently;
+ *   - a failed PUT leaves the cache untouched and does not wedge the lock.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import {
+  GroupMdWriteback,
+  GroupMdContentTooLargeError,
+  MAX_GROUP_MD_CONTENT_BYTES,
+  type UpdateGroupMdFn,
+} from '../group-md-writeback.js';
+import { GroupMdCache } from '../group-md-cache.js';
+
+const API = 'https://api.example.com';
+const TOKEN = 'bot-token';
+
+const delay = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
+
+/** A version-incrementing fake PUT client (server-style monotonic counter). */
+function versioningUpdateFn(): { fn: UpdateGroupMdFn; calls: () => number } {
+  let version = 0;
+  let calls = 0;
+  const fn: UpdateGroupMdFn = async () => {
+    calls++;
+    version++;
+    return { version };
+  };
+  return { fn, calls: () => calls };
+}
+
+describe('GroupMdWriteback', () => {
+  it('writes content back to the cache with the server-assigned version', async () => {
+    const cache = new GroupMdCache();
+    const { fn } = versioningUpdateFn();
+    const wb = new GroupMdWriteback(cache, fn);
+
+    const res = await wb.writeBack({ apiUrl: API, botToken: TOKEN, groupNo: 'g1', content: 'hello' });
+
+    expect(res).toEqual({ groupNo: 'g1', version: 1, bytes: 5 });
+    expect(cache.get('g1')).toEqual({ content: 'hello', version: 1, updated_at: null });
+  });
+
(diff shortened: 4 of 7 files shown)


src/group-md-tool.ts
```diff
@@ -0,0 +1,135 @@
...
+ * Build the GROUP.md tool DEFINITIONS for one agent turn. Exported separately
+ * from the server so tests can invoke the handler directly. `coords` targets the
+ * group + supplies the caller uid; `ownerUid` is the owner gate.
+ */
+export function buildGroupMdTools(
+  deps: GroupMdToolDeps,
+  coords: GroupMdSessionCoords,
+  ownerUid: string,
+) {
+  const isOwner = coords.fromUid === ownerUid && ownerUid !== '';
```
OctoBoooot, line 65: **[minor]** the owner gate correctly authenticates *who triggered the turn* (`coords.fromUid === ownerUid`), but the module doc-comment claims it stops a prompt-injected agent "from any chat." On the owner's OWN turn, the prompt includes `groupContextBlock` (index.ts:452-461 — other members' recent messages, `sanitizePromptBody`'d, which strips role/section markers but isn't a semantic injection defense). So a non-owner can plant instructions in group chat that steer the owner's agent to call this tool with attacker-influenced content when the owner next triggers the bot. Residual, not a blocker (requires owner trigger; full-replace + version bump is visible/recoverable; flag-off default). Suggest scoping the comment's guarantee to non-owner turns, and consider an audit log (updated_by + content hash) or owner-confirmation for this write in a later phase.

OctoBoooot (commented):
## Review: feat(group-md): owner-gated GROUP.md write-back (#174)

**Verdict:** Approve with comments — the owner-gate + serialization design is sound; one residual prompt-injection nuance to document, and CI must go green first.

CI is red — `check-sprint / check-sprint` is failing (project-board gate, not code). Vitest / ESLint / type-check pass. Verdict can't be APPROVE until that's resolved, but there are no code blockers.

Reviewed at elevated rigor: this is an auth-gated **write** that mutates operator-trusted GROUP.md, stacked on #172/#173 (new code is `group-md-tool.ts` + `group-md-writeback.ts`; the rest is the reviewed stack). `needs-human-review` + `size/XL` applied.

The security-critical parts hold up:
- **Owner gate is independent of LLM judgment and of the token's server-side permission.** `isOwner = coords.fromUid === ownerUid && ownerUid !== ''`, checked in the tool handler before any PUT. A non-owner turn is rejected with no server call; empty owner locks everyone out. Both are tested. This is the right place for the gate (policy in the tool layer, mechanism in the coordinator).
- **Per-groupNo serialization** via the shared `GroupMdWriteback` promise-chain lock prevents this gateway from racing itself on read-modify-write; the cross-source last-write-wins limitation (no server CAS) is honestly documented as accepted, not hidden.
- **Byte cap enforced locally before the PUT** (10240 UTF-8), with the coordinator re-checking as the authoritative boundary — defense in depth, tested at and over the limit.
- **No new durable-trust surface** — the cache it refreshes is the same memory-only cache from #172.

### Question / Minor
- `src/group-md-tool.ts` doc-comment — the gate is described as what "stops a prompt-injected agent from rewriting the operator's trusted GROUP.md **from any chat**." That's true for a *non-owner's* turn (tool rejects). But on the **owner's own turn**, the agent's prompt includes `groupContextBlock` — other members' recent messages (index.ts:452-461, sanitized by `sanitizePromptBody`, which strips role/section markers but is not a semantic injection defense). So a non-owner can plant instructions in group chat, and when the owner later triggers the bot, the owner's agent — with the write-back tool available — could be steered by that planted content to rewrite GROUP.md with attacker-influenced text. The gate authenticates *who triggered*, not *what the content reflects*. Residual, not a blocker: it requires the owner to trigger, `sanitizePromptBody` blunts the surface, full-replace + version bump is visible/recoverable, and it's flag-off by default. Suggest softening the comment to scope the guarantee to non-owner turns, and consider whether a high-value write like this warrants an explicit owner confirmation step or an audit log of `updated_by` + content hash in a later phase. Would settle it: is the owner expected to review GROUP.md content the agent composes before it's persisted, or is agent-authored content trusted once the owner triggers?

### Praise
- Splitting policy (owner gate, tool layer) from mechanism (serialization + byte cap + cache sync, coordinator) is clean, and the shared-instance-per-bot note ("a per-turn instance would defeat the lock") shows the concurrency model was thought through, not bolted on.
- The XIN-201 "measured, NOT inferred" server-contract facts (hard 10240 cap, no CAS) are exactly the kind of verified-not-assumed grounding these GROUP.md PRs have been strong on — and the no-CAS caveat is surfaced rather than silently accepted.
- Test coverage hits the load-bearing security cases directly: non-owner → no PUT, empty-owner → locked out, thread → parent group, concurrent same-group serialization, PUT-fails → cache untouched.

- Coverage-sniff flagged `group-md-cache.ts` / `api.ts` — false positives (covered by the #172/#173 suites + new `group-md-tool`/`group-md-writeback` tests).
- Stacked on #172/#173 — merge order applies.

Jerry-Xin (commented):
This PR is in scope for `cc-channel-octo` and the implementation looks ready to merge; I found no blocking issues.

💬 Non-blocking

- 🟡 Warning: Consider adding explicit write-back guidance to the security prompt near [src/agent-bridge.ts](src/agent-bridge.ts:91), mirroring the scheduled-task warning. The owner gate is enforced in code, but `update_group_md` mutates trusted server-side instructions, so it would be safer to tell the runtime to use it only when the current owner message explicitly asks.

- 🔵 Suggestion: Add README coverage for `serverMd`, `serverMdTtlMs`, and `mdWriteback`. The config table currently documents `groupConfigDir` and `sdk.cron` around [README.md](README.md:203), but the new user-facing flags are only in `config.example.json`.

- 🔵 Suggestion: Add an integration test for the case where both `sdk.cron` and `mdWriteback` are enabled, asserting both MCP servers are present. The current merge at [src/index.ts](src/index.ts:870) looks correct, but a regression test would protect this composition.

✅ Highlights

- Owner gating, byte-limit enforcement, per-group serialization, and cache refresh are well covered.
- The cache remains memory-only, which preserves the trusted prompt boundary.
- Verified locally: `npm run type-check`, `npm run lint`, and full `npm test` all pass.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# cc-channel-octo PR#174 Review Report
**PR:** feat(group-md): owner-gated GROUP.md write-back
**Head SHA:** `3faa2cf7557c286ac63ac8a0e098aaabad0b452d`
**Files:** 12 changed (+1422 / −9)

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Owner gate correctness | ✅ | `group-md-tool.ts:66` — `coords.fromUid === ownerUid && ownerUid !== ''`. Empty-owner defense verified. Test: `group-md-tool.test.ts:94`. |
| Memory-only cache (no disk poisoning) | ✅ | `group-md-cache.ts` — `Map<string, StoredEntry>` only. No fs calls. Test: `group-md.test.ts:113-120` (fresh instance shares no state). |
| Byte cap enforcement (UTF-8, not char) | ✅ | `group-md-writeback.ts:127` — `Buffer.byteLength(content, 'utf-8')`. Multi-byte test: `group-md-writeback.test.ts:102-113`. |
| Per-groupNo write serialization | ✅ | `group-md-writeback.ts:107-119` — promise-chain lock. Concurrent test: maxActive=1 verified (`group-md-writeback.test.ts:117-139`). |
| Server-first + local fallback chain | ✅ | `group-md.ts:88-120` — try/catch with local fallback on any error. Tests cover 404, network, empty, flag-off, no-cache. |
| Thread routing (parent groupNo) | ✅ | `channel-id.ts:31-34` — `extractParentGroupNo`. Tests: both read (`group-md.test.ts:243-257`) and write (`group-md-tool.test.ts:116-120`). |
| isGroup gating (read + write) | ✅ | `index.ts:386,822,859` — both paths behind same `isGroup` var (Group ∪ CommunityTopic). DM default on missing channel_type. |
| Cron path parity | ✅ | `index.ts:320` — cron `onFire` passes both `groupMdCache` and `groupMdWriteback`. |
| Config wiring (flag + TTL + writeback) | ✅ | `config.ts:287-323,381-383,616-618` — all three new fields in PartialConfig, mergeConfig, resolveBotConfigs. |
| Test coverage | ✅ | 4 new test files (732 lines). API, tool policy, write-back mechanism, resolver all covered. |

## 2. Findings

### P2-1 — Stale JSDoc on `updateGroupMd` claims it's "NOT wired into any write-back path"

**Severity:** P2 (documentation inaccuracy)
**Diff-scope:** Pre-existing in this PR — the function was added in this same diff with a misleading comment.

`src/octo/api.ts:716-718`:
```
* NOTE (P2-A scope): this client function is restored alongside getGroupMd, but
* is intentionally NOT wired into any write-back path here — the PUT trigger
* chain is owned by a separate work item.
```

This contradicts reality: `group-md-writeback.ts:1` imports `updateGroupMd` and uses it as the default `updateFn`. The function **is** wired into the write-back path in this same PR. The "separate work item" language is stale.

**Suggestion:** Update the JSDoc to reflect that `updateGroupMd` is consumed by `GroupMdWriteback` in this PR.

### P2-2 — `resolveGroupInstructions` does not forward `signal` to `getGroupMd`

**Severity:** P2 (resource waste on aborted turns, not a correctness bug)
**Diff-scope:** New code in this PR.

`src/group-md.ts:96`:
```typescript
const md = await getGroupMd({ apiUrl, botToken, groupNo, signal });
```

Wait — on re-reading: `signal` **is** in the params destructuring (`group-md.ts:81`) and **is** forwarded on line 96. Let me re-verify…

Actually, the `signal` parameter IS present in the destructured params and IS passed to `getGroupMd`. I retract this finding on closer inspection — the signal forwarding is correct.

### P2-3 — `updateGroupMd` uses raw `fetch` instead of a shared `postJson`/`putJson` helper

**Severity:** P2 (maintenance risk)
**Diff-scope:** New code in this PR.

`src/octo/api.ts:720-747` — `updateGroupMd` duplicates the fetch/auth/timeout/error pattern from `postJson` (lines 68-88) rather than calling through a shared helper. Currently the patterns are byte-identical, but any future change to `postJson` (retry logic, telemetry, timeout tuning) would silently not apply to the PUT path.

**Suggestion:** Either extract a `putJson` helper mirroring `postJson`, or add a comment explicitly noting the intentional duplication and the maintenance coupling.

## 3. Suggestions

1. **Fix the stale JSDoc** on `updateGroupMd` (P2-1 above) — one-line change.
2. Consider a `putJson` helper to reduce the `updateGroupMd` ↔ `postJson` duplication.
3. The `MAX_GROUP_CONFIG_BYTES` (16 KiB, read path) vs `MAX_GROUP_MD_CONTENT_BYTES` (10 KiB, write path) asymmetry is intentional but could confuse future maintainers. A brief comment in `group-md-writeback.ts` noting the deliberate difference would help.

## 4. Additional Observations

- **Security posture is strong.** Memory-only cache avoids the disk poisoning vector identified in review #172. Owner gate with empty-uid defense is correct. Byte-counting (not char-counting) on the limit is right. `isSafeGroupNo` defense-in-depth on cache keys mirrors the filesystem path's `isSafeId`.
- **Write-back lock is correctly process-scoped.** `GroupMdWriteback` is constructed ONCE per bot (index.ts:228) and shared across turns — the per-groupNo serialization actually spans concurrent agent turns as intended. The documented cross-source last-write-wins caveat is an accepted server limitation.
- **Feature-flag safety.** Both `serverMd` and `mdWriteback` default to `false`/undefined. When off, the code paths are byte-identical to pre-PR behavior (`resolveGroupInstructions` returns `local()` immediately, write-back tool is never registered).
- **Cron path parity confirmed.** `handleMessage` called from cron `onFire` (index.ts:320) receives both `groupMdCache` and `groupMdWriteback`. The cron message's `from_uid` comes from `task.fromUid` (cron-scheduler.ts:44), so the owner gate works correctly for cron-fired messages.

## 5. Data Flow Trace

| Consumed data | Upstream source | Reaches consumer? |
|---|---|---|
| `ownerUid` (gate) | `router.getOwnerUid()` ← `gateway._ownerUid` ← `registerBot().owner_uid` | ✅ Yes. Empty-string defense (`&& ownerUid !== ''`) prevents open gate when registration hasn't completed. |
| `groupNo` (API + cache key) | `extractParentGroupNo(channelId)` — splits on `____` separator | ✅ Yes. For plain groups returns channelId unchanged; for threads returns parent groupNo. Edge: empty channelId → empty groupNo → `isSafeGroupNo` rejects cache store; server 404 → local fallback. |
| `content` (write-back) | Agent tool call → `args.content` → `Buffer.byteLength` check → `updateFn` → cache | ✅ Yes. Byte cap enforced before server call. Cache refreshed only after successful PUT. Failed PUT leaves cache untouched. |
| `fromUid` (owner gate) | `msg.from_uid` from server-pushed `BotMessage` | ✅ Yes. Server-assigned, not user/agent-controllable. Captured at tool-server creation time per turn — no stale closure risk. |
| `config.mdWriteback` (feature flag) | `config.ts` → mergeConfig → resolveBotConfigs | ✅ Yes. Gated at `index.ts:859`. Default off → no tool registered. |
| `config.serverMd` (feature flag) | `config.ts` → mergeConfig → resolveBotConfigs | ✅ Yes. Gated at `group-md.ts:83`. Default off → pure local file, unchanged behavior. |

## 6. Blindspot Checklist (R5)

- **C1 — Symmetric paths (add↔remove, create↔delete):** N/A. This PR adds a new write-back path; no symmetric removal.
- **C2 — Control-flow ordering / nested reuse:** CLEAR. The write-back flow is linear: owner gate → byte check → `writeBack` → `withLock` → `updateFn` → cache refresh. No double-application or ordering hazards. The `boundInstructions` function is only called on the read path, never nested with the write path.
- **C3 — Authorization boundary ≠ capability boundary:** CLEAR. The MCP tool is only registered for group channels (`isGroup` gate). The owner uid gate is evaluated per-turn from `msg.from_uid` (server-assigned). The bot token has server-side `bot_admin` permission, but the owner gate prevents any non-owner chat user from triggering a write through the agent.
- **C4 — Authorization lifecycle / container-member state cascade:** N/A. No auth hierarchy changes. The owner uid is a flat identity check, not a role in a container.
- **C5 — Build ≠ runtime path:** N/A. No build artifacts, extensions, or CLI packaging changes. Pure TypeScript source additions.
- **C6 — Governance / policy document consistency:** N/A. No governance or policy documents changed.

N/A — automated review review, no prior round.

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. Three P2 notes (stale JSDoc, raw-fetch duplication, limit asymmetry documentation) are all maintainability improvements, not blockers. The security model is well-designed (memory-only cache, owner gate with empty-uid defense, byte-level cap, per-groupNo serialization), test coverage is thorough (4 files / 732 lines covering API, tool policy, write mechanism, and resolver), and feature-flag safety ensures zero behavior change for existing deployments.

yujiawei (dismissed):
# Code Review — PR #174 (cc-channel-octo)

**Scope reviewed:** owner-gated GROUP.md write-back (12 files, +1422/−9) at head `3faa2cf`, diffed against `origin/main` (merge-base `87ae921`). Typecheck clean; full suite green locally (1173 tests). Reviewed line-by-line plus two independent cross-family analysis passes; findings below are the union after verification.

**Verdict: APPROVED.** No blocking (P0/P1) issues survive verification. The core trust model is sound: the owner gate is fail-closed and driven by server-delivered `from_uid` (not an LLM/tool argument), the cache is memory-only so untrusted chat can never reach the trusted system-prompt channel via an on-disk artifact, the per-`groupNo` promise-chain lock correctly serializes same-group writes, and the byte cap is enforced (UTF-8) both in the tool and authoritatively in the coordinator. The items below are all advisory (P2) — worth considering but none should block merge.

## Part 1 — Spec compliance

**Spec: ✅**

- **Missing:** none that violate the PR's stated scope. One parity nicety is absent — see P2-6 below (no boot-time warning when `mdWriteback` is on but the bot has no `owner_uid`, unlike the `sdk.cron` precedent at `index.ts:254`). The PR never promised this, so it is not a spec miss, only a convention gap.
- **Over-build:** none. `update_group_md` is gated exactly on `mdWriteback && isGroup` (`index.ts:859`) and owner-only (`group-md-tool.ts:65,83`); content is full-replace, capped at 10240 bytes as claimed. No extra endpoints/flags/side effects beyond the stated P2-A read path + P2-C write path.
- **Deviation:** none functional. One stale doc comment contradicts the code within the same commit — see P2-7.

## Part 2 — Code quality

**Quality: Approved** (all findings advisory)

**P2-1 — SSRF-on-redirect: new PUT/GET use plain `fetch()` (default `redirect: follow`).** `src/octo/api.ts` (`updateGroupMd` ~L720-746, `getGroupMd` via `getJson` L267-289) do not route through `fetchWithRedirectGuard`/`assertPublicUrl` the way the inbound/media paths do (`inbound.ts:635`, `media-inbound.ts:78`, which drop `Authorization` on host change). `apiUrl` is boot-validated (`isAllowedApiUrl`, `config.ts:445`) and operator-controlled, so exploitation would require the trusted Octo server itself to emit a 3xx to a private/loopback target, which default `fetch` would follow and re-send the Bearer token to. This is an inherited pattern (all `getJson`/`postJson` calls share it), but the PR does add two new un-guarded call sites. Defense-in-depth, not a live user-reachable SSRF.

**P2-2 — `groupNo` validation asymmetry: PUT fires before any `isSafeGroupNo` check.** `writeBack` (`group-md-writeback.ts:123-142`) passes `groupNo` (from `extractParentGroupNo`, which does no charset validation) straight to the PUT at L131, while `cache.set` guards with `isSafeGroupNo` (`group-md-cache.ts:95-97`). Not exploitable — `api.ts:727` wraps `groupNo` in `encodeURIComponent`, so no path traversal, and the request stays on the trusted host (an unsafe id just 404s). `channel_id` is server-delivered. Recommendation: add a single `isSafeGroupNo` guard at the top of `writeBack` (or in the tool right after `extractParentGroupNo`) so the network call and the cache write share one boundary.

**P2-3 — Trust model note (accepted risk, no code change).** The owner gate is the entire trust boundary and it holds: a prompt-injected *non-owner* agent cannot forge `coords.fromUid`, and the empty-owner case is explicitly closed (`ownerUid !== ''`, `group-md-tool.ts:65`) and tested. The residual risk is owner *self*-injection (owner asks the bot to save attacker-pasted text) — but that is the same power the owner already has by editing GROUP.md directly, so it is not a new privilege escalation. Flagging only so the audit trail records that the gate is load-bearing.

**P2-4 — Unlocked read-refresh can transiently clobber a newer locked write.** The read path's `cache.set` (`group-md.ts:109`) runs outside the `GroupMdWriteback` per-`groupNo` lock. Interleaving: a read's `getGroupMd` returns version N; a write commits N+1 under the lock; the read's continuation then overwrites the cache back to N. Bounded and self-healing — the server already holds N+1 (no data loss), TTL re-fetches, and only this gateway's in-memory copy is briefly stale with still-*authentic* content (no poisoning). This is an in-process instance of the cross-source last-write-wins already documented as accepted (`group-md-writeback.ts:14-21`). Optional: make `cache.set` skip when an existing entry has a strictly higher version.

**P2-5 — Cache stores coerced version, result returns raw version.** `group-md-writeback.ts:138` caches `typeof version === 'number' ? version : 0` but L141 returns the raw `version`; the tool echoes the raw value to the model (`group-md-tool.ts:106`). The declared return type is `{ version: number }` and `parseOctoJson` is an unchecked cast, so a malformed server body could in theory diverge, but `version` gates no control flow (no CAS). Cosmetic/internal-consistency nit.

**P2-6 — Empty-body PUT caches `version: 0`.** `updateGroupMd` returns `{ version: 0 }` on an empty 2xx body (`api.ts:743-745`); that 0 is then cached and reported to the model. Not a correctness/data-loss bug — the *content* cached is correct, and no read path decides on `version` (0 is already the in-band "unknown" sentinel used by the GET fallback). Observability-only; an empty 2xx body is not expected per the documented contract.

**P2-7 (spec/maintainability) — No boot-time warning for `mdWriteback` + empty `owner_uid`.** `sdk.cron` warns loudly when the bot has no owner (`index.ts:254-261`) because the owner gate can then never pass and every call is silently rejected. `mdWriteback` has the identical fail-closed gate (`group-md-tool.ts:65,83`) but no equivalent boot warning. Parity/observability nicety consistent with the codebase's own convention.

**P2-8 (maintainability) — Stale doc comment on `updateGroupMd`.** `api.ts:715-717` still says the function is "intentionally NOT wired into any write-back path here — the PUT trigger chain is owned by a separate work item." This same PR (P2-C) *is* that work item and wires it (`group-md-writeback.ts:31,91,131` → tool → `index.ts:859`). Update the comment to point at the `group-md-writeback.ts` coordinator to avoid a future stale-context trap.

## Notes on refuted / non-issues

- **Read-vs-write cap truncation is NOT a bug.** Read cap `MAX_GROUP_CONFIG_BYTES = 16384` (`group-config.ts:33`) > write cap `MAX_GROUP_MD_CONTENT_BYTES = 10240` (`group-md-writeback.ts:39`), so a valid ≤10240-byte write can never be truncated by the read path. The two caps are consistent by construction.
- **Owner-gate bypass / untrusted-chat → trusted-prompt poisoning: closed.** Verified end-to-end.
- **`withLock` Map growth / cleanup race:** the `this.tails.get(key) === tail` identity check correctly avoids deleting a newer tail; a rejected PUT does not wedge later writes (chain tail swallows rejection). Correct.
- **Thread parent-group resolution:** `extractParentGroupNo` (`channel-id.ts:31-34`) correctly targets the parent group for a `<groupNo>____<shortId>` thread channelId.

## Coverage / blind spots

Assessed from the diff + surrounding source at head SHA. Could not exercise at runtime: (a) real server behavior of `PUT /v1/bot/groups/{groupNo}/md` (empty-body / non-`{version}` responses, actual 400 on oversize — the XIN-201 contract is taken as given); (b) production memory footprint of the unbounded-by-group-count in-memory cache under a very large number of distinct active groups (mechanism noted in P2-4 neighborhood; content is capped and eviction is TTL-on-`get`, so growth is bounded by concurrently-active group count, not runaway); (c) redirect behavior of the live `apiUrl` host (P2-1 assumes the operator endpoint does not 3xx to internal addresses).

One analysis leg in the cross-check produced an off-target result (its conclusion referenced changes not present in this PR) and was discarded rather than absorbed; the findings above rest on the line-level pass plus the on-target cross-family pass, both independently verified against the diff.

OctoBoooot (approved):
## Re-review (delta): #174 — rebase, write-back code unchanged; CI now green

**Verdict:** Approve. New head `420a69772ce7` rebases onto the merged #172/#173 base; #174's own write-back code is byte-identical to my prior review, and the CI gate that was blocking is now green.

Delta isolation:
- `src/group-md-tool.ts` and `src/group-md-writeback.ts` — **byte-identical** to `3faa2cf7557c`. The owner-gate + serialization + byte-cap logic I reviewed at auth-tier rigor is unchanged.
- `src/index.ts` — the only source change is P2-B's SessionRouter wiring (`new SessionRouter(..., groupMdCache)`), which is #173 landing in the rebased base, not new #174 code.
- File count dropped 12→7 because #172/#173 are now in the base rather than stacked in the diff.

Standing conclusion (carried from review 4606555465): owner gate is sound and unspoofable — `fromUid === ownerUid && ownerUid !== ''`, enforced before the PUT, `fromUid` from the server-authenticated WS envelope, `ownerUid` from `registerBot.owner_uid`; non-owner rejected with no PUT, empty owner fail-closed, all tested. Per-groupNo serialization lock, local 10240-byte cap before the PUT, cache sync after, no new disk surface.

Merge-gate status:
- ✓ **CI green** — `check-sprint` now passes (was the red gate blocking merge last round).
- ✓ Non-author formal APPROVE — this review (viewer OctoBoooot ≠ author Jerry-Xin), plus yujiawei's.

### Still open (non-blocking, carried forward)
- Owner gate authenticates *who triggers*, not *what the content reflects*: on the owner's own turn, planted group-context messages (`groupContextBlock`, `sanitizePromptBody`'d but not injection-proof) could steer the agent's write-back content. Residual — requires owner trigger, full-replace + version bump is visible/recoverable, flag-off default. Worth a follow-up (scope the doc-comment's guarantee to non-owner turns; consider an audit log of `updated_by` + content hash), doesn't gate merge.

Clean to merge on this head.

Jerry-Xin (commented):
Re-review at rebased head `420a6977` (approve-equivalent conclusion). CI is now fully green including check-sprint; the previously noted merge gates are resolved.

## Summary

This is a re-review after the branch was rebased onto current `main`. I isolated PR #174's true additive delta over the new merge-base (`4ca07a0..420a6977`) rather than the inflated 15-file compare-against-old-head list, most of which is merge-base drift from the stacked P2-B work (`serverMdEventTypes`, router-held GROUP.md cache invalidation) that has since landed on `main`.

**#174's own delta is 7 files:** `config.example.json`, `src/config.ts`, `src/group-md-tool.ts`, `src/group-md-writeback.ts`, `src/index.ts`, and the two write-back test files.

**The auth-critical write-back core is byte-identical to the previously reviewed tree** (`3faa2cf`):
- `src/group-md-tool.ts` — identical blob
- `src/group-md-writeback.ts` — identical blob
- `src/__tests__/group-md-tool.test.ts` — identical blob
- `src/__tests__/group-md-writeback.test.ts` — identical blob

`src/config.ts` and `src/index.ts` show as changed only because the rebase carried the stacked `serverMdEventTypes` / router-cache lines through them; diffing `3faa2cf..420a6977` on those two files shows exclusively that P2-B carry-over — no change to P2-C write-back wiring or the owner gate.

## Owner-gate re-verification (still sound)

The load-bearing property here is that GROUP.md write-back is a trusted-prompt-injection surface, so the gate must be server-side and unspoofable. Re-confirmed at the new head:

- Owner gate `fromUid === ownerUid && ownerUid !== ''` enforced in the MCP tool handler as the first statement, before any resolve/cap/PUT.
- Fail-closed on empty owner: empty `ownerUid` forces `isOwner=false`, so a bot with no registered owner is unusable for write-back (covered by test "empty owner uid gates everyone out").
- Non-owner request returns an error and never reaches `writeBack()` (no server call) — covered by test "non-owner write is rejected and never calls the server".
- `ownerUid` comes from the server-issued registerBot owner uid via the router; `fromUid` comes from the decrypted WS envelope, not a client-supplied body.
- Content byte-capped at 10240 UTF-8 before the PUT (handler pre-check plus the coordinator's authoritative re-check).
- Per-`groupNo` serialization lock and cache refresh live in the shared write-back coordinator; no new on-disk surface (it refreshes the existing in-memory cache on success).

## Non-blocking follow-up (unchanged)

The gate authenticates WHO triggers the write (owner uid, unspoofable), but not WHAT the content reflects: on the owner's own turn, planted group context could in principle steer write-back content. This remains a non-blocking follow-up — consider adding explicit write-back guidance to the security prompt so the runtime treats `update_group_md` content as owner-authored intent rather than echoing injected context.

## Verdict

Approve-equivalent — no blocking issues; the auth axis re-verifies sound at the rebased head and the write-back core is byte-unchanged from the prior approved review. This is posted as a COMMENT rather than a formal APPROVE because of the same-author restriction (self-approve is blocked). A fresh non-author formal APPROVE at `420a6977` is required to merge: the earlier non-author approval may have been invalidated by the rebase to the new head.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#174 Review Report

**PR:** feat(group-md): owner-gated GROUP.md write-back
**Head SHA:** `420a69772ce74d8acf032aebdf3b8175050a916a`
**Repo:** Mininglamp-OSS/cc-channel-octo

---

## 1. 验证结论

- ✅ Owner gate correctly mirrors proven cron-tool pattern (`coords.fromUid === ownerUid && ownerUid !== ''`) — closure-captured at tool-build time, per-turn (`src/group-md-tool.ts:68`, `src/cron-tool.ts` same pattern)
- ✅ Byte cap (10240 UTF-8) enforced at BOTH tool layer (`src/group-md-tool.ts:87-92`) AND coordinator (`src/group-md-writeback.ts:125-128`), using `Buffer.byteLength` (not `.length`)
- ✅ Per-groupNo serialization via promise-chain lock is correct: `prev.then(fn, fn)` runs regardless of prior outcome; tail swallows rejections so a failed write does not wedge later writes (`src/group-md-writeback.ts:105-118`)
- ✅ Cache refresh on successful write: `cache.set(groupNo, { content, version, updated_at: null })` matches `GroupMdEntry` interface (`src/group-md-writeback.ts:132-136`)
- ✅ Feature flag `mdWriteback` is off by default, properly merged in all config paths (mergeConfig, resolveBotConfigs, PartialConfig) (`src/config.ts:397, 634`)
- ✅ MCP server spread preserves existing servers (cron): `{ ...(sessionOpts?.mcpServers ?? {}), [GROUP_MD_TOOL_SERVER_NAME]: groupMdServer }` (`src/index.ts:875-876`)
- ✅ `extractParentGroupNo` correctly handles both plain group and thread channelIds (`src/octo/channel-id.ts`)
- ✅ `groupMdWriteback` parameter is optional in `handleMessage` signature — backward-compatible (`src/index.ts:385`)
- ✅ Tests cover owner-gate, empty-owner, byte limit (including multi-byte UTF-8), thread routing, concurrency serialization, cross-group concurrency, and lock recovery after failure

## 2. 发现的问题

### P2 — sessionOpts spread inconsistency (pre-existing, amplified)

**Diff-scope:** new/amplified/pre-existing? → **Pre-existing pattern, this PR does it correctly.**

The cron block at `src/index.ts:~862` assigns `sessionOpts.mcpServers` directly (no spread), while the group-md block spreads existing servers. This means the ordering dependency is fragile — if a future tool is added after group-md without spreading, it would clobber the group-md server. This PR's spread is the correct pattern; the cron block should be updated to match. Not blocking.

### P2 — AbortSignal not forwarded from tool to coordinator (new, minor)

**Diff-scope:** new.

The tool handler at `src/group-md-tool.ts:94-99` calls `deps.writeback.writeBack({...})` without passing `signal`. The `GroupMdWriteParams` interface accepts an optional `signal`, but the tool never provides one. This means if the agent session is cancelled mid-write, the in-flight PUT continues to completion and updates the cache. For a write-back tool with a ~30s timeout, this is low-risk (the write should be fast, and the cache update is idempotent), but forwarding the signal would be more correct for graceful cancellation.

### P2 — Read/write cache race within single process (acknowledged limitation)

**Diff-scope:** pre-existing (cross-source last-write-wins documented in PR).

`resolveGroupInstructions` (read path) and `writeBack` (write path) can both call `cache.set` concurrently without coordination. If a read fetches stale data while a write is in-flight, the read's `cache.set` could overwrite the write's fresher cache entry. Narrow window; TTL self-heals within 5 minutes. The PR documents this as an accepted limitation. Not blocking.

## 3. 建议

1. **Consider forwarding AbortSignal** from the tool handler to `writeBack` for graceful cancellation on session teardown.
2. **Consider updating the cron block** to use the same spread pattern as the group-md block (`{ ...(sessionOpts?.mcpServers ?? {}), ... }`) for defensive consistency.

## 4. 额外发现

- The `isSafeGroupNo` guard in `GroupMdCache.set()` provides defense-in-depth against malformed group numbers extracted from channelIds.
- The `withLock` cleanup (`tails.delete(key)`) correctly prevents unbounded Map growth across many groups.
- Test quality is high — multi-byte UTF-8 byte-counting test (`✓` × 3414 = 10242 bytes) specifically validates the byte-vs-char boundary.

## 5. 数据流回溯

| Consumed Data | Source | Flows Correctly? |
|---|---|---|
| `msg.from_uid` → `coords.fromUid` → `isOwner` check | Octo server (trusted) | ✅ Yes — server-controlled, not user-forgeable |
| `router.getOwnerUid()` → `ownerUid` gate | `registerBot` → `gateway.ownerUid` → `SessionRouter` | ✅ Yes — trusted registration flow |
| `coords.channelId` → `extractParentGroupNo()` → `groupNo` | `msg.channel_id` (Octo server) → string parsing | ✅ Yes — `indexOf('____')` correctly handles both plain and thread IDs |
| `args.content` (LLM output) → byte check → PUT → cache | LLM → tool handler → coordinator → API → cache | ✅ Yes — double-checked (tool + coordinator), PUT result written to cache |
| `PUT response.version` → cache entry | `updateGroupMd` returns `{ version: number }` → type-guarded (`typeof version === 'number' ? version : 0`) | ✅ Yes |
| `cache.set(groupNo, entry)` → `resolveGroupInstructions` consumption | Cache entry matches `GroupMdEntry` interface (`content, version, updated_at, updated_by?`) | ✅ Yes — `updated_by` is optional |

## 6. 盲点 checklist（R5）

- **C1 双路径 parity (含守卫/lint 工具自身等价写法覆盖):** N/A — no symmetric add/remove or subscribe/unsubscribe paths in this PR. The write-back tool is a single-direction operation (write to server).
- **C2 control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿:** Clear. The byte cap is checked at two layers (tool + coordinator), preventing TOCTOU within the lock. The owner gate is closure-captured at tool-build time, not per-invocation, so re-invocation within the same turn is safe (fromUid is fixed per turn).
- **C3 授权边界 ≠ 能力边界 (谁能到达 tool):** Clear. The tool is only registered when `mdWriteback && isGroup`. The owner gate checks `msg.from_uid` (server-controlled) against `router.getOwnerUid()` (from registration). The LLM cannot bypass the owner gate — it's checked in the handler before any server call. The bot token has server-side write permission, but the owner gate at the tool layer is the policy boundary preventing unauthorized invocation.
- **C4 授权生命周期 / 容器-成员状态级联:** N/A — this is a gateway-side MCP tool, not a server-side authorization endpoint. Container/member state cascade is a server concern.
- **C5 build/note 通过 ≠ 运行期路径正确:** N/A — this is a pure Node.js process, no browser extension/CLI/packaging/relative URL concerns.
- **C6 治理/策略/安全文档自洽性:** N/A — config.example.json is operator documentation, not a governance/policy document.

## 7. 跨轮 blocker 复检

---

**[Octo-Q] verdict: APPROVE**

This PR is well-architected with clean separation of concerns (tool layer = policy, coordinator = mechanism). The owner gate correctly mirrors the proven cron-tool pattern. Byte cap is double-checked. Concurrency serialization is sound. Cache refresh on write is correct. Feature flag provides safe rollback. Tests comprehensively cover all acceptance gates including edge cases (multi-byte UTF-8, concurrent same-group writes, lock recovery after failure, thread→parent group routing). No P0/P1 issues found.

yujiawei (approved):
# Code Review — PR #174 (cc-channel-octo)

**Scope:** owner-gated GROUP.md write-back — an in-process MCP tool (`mcp__group_md__update_group_md`) that PUTs a group's GROUP.md back to the server, gated to the bot owner, byte-capped, and per-group serialized. Reviewed at head `420a6977`, diff against merge-base `4ca07a03`.

**Result:** ✅ Builds clean (`tsc --noEmit`), all 17 new tests pass (`group-md-writeback.test.ts`, `group-md-tool.test.ts`), lint clean. No blocking defects found.

---

## 1. Spec compliance — ✅

- **Owner-gate present & fails closed.** `buildGroupMdTools` computes `isOwner = coords.fromUid === ownerUid && ownerUid !== ''` (`src/group-md-tool.ts:65`), checked at the top of the handler before any group resolution or PUT. `ownerUid` flows from the server-authoritative `registerBot.owner_uid` (`gateway.ts:99` → `index.ts:288` → `SessionRouter.getOwnerUid`), and `msg.from_uid` is the decrypted routing-envelope sender, not chat-controllable — so the prompt-injection boundary against rewriting the trusted system prompt holds. Empty-owner correctly locks everyone out.
- **10240-byte cap enforced full-replace** at both layers against the same constant (`MAX_GROUP_MD_CONTENT_BYTES`), counted in UTF-8 bytes (not chars), rejected locally before the server is hit. The write cap (10240) sits below the read cap (`MAX_GROUP_CONFIG_BYTES = 16384`), so a written doc is never truncated on read-back.
- **Gated on `config.mdWriteback && isGroup`** (`src/index.ts:862`), DM correctly excluded.
- No **missing**, **over-build**, or hard **deviation** vs. the stated feature.

## 2. Code quality — Approved

Owner-gate, concurrency lock, and cache-refresh were independently traced and adversarially verified. The per-groupNo promise-chain lock (`group-md-writeback.ts:withLock`) genuinely serializes same-group writes (synchronous `tails.set` before return; JS single-threaded), runs different groups concurrently, does not wedge on a rejected PUT, and cleans up the Map without a race. The coordinator is constructed once and shared across turns (`index.ts:230`), so the lock and the cache it refreshes are process-wide — the per-turn part is only the MCP server, which correctly borrows the shared instance. No P0/P1.

### Suggestions (P2 — non-blocking)

- **`mdWriteback=true` + `serverMd=false` is a write-only dead end for *this* gateway.** The write persists to the server correctly, but the read path short-circuits to the local file (`src/group-md.ts:83`) and never consults the cache the write refreshed — so this gateway keeps injecting the local GROUP.md and the owner's write appears to have "no effect" until `serverMd` is also enabled. It fails closed (no data loss; the server copy is correct and any `serverMd`-on consumer sees it), which is why this is advisory. But the config doc frames the flags as fully independent; consider either (a) requiring `serverMd` for write-back to take local effect, or (b) a note in the tool/config docs that write-back only feeds back into *this* gateway's prompt when `serverMd` is on.
- **No boot-time owner-uid warning, unlike cron.** `sdk.cron` logs a loud warning when enabled but the bot has no `owner_uid` (`index.ts:254`), because the gate can then never pass. `mdWriteback` has the same silent-disable failure mode but no equivalent warning — worth adding for parity.
- **Abort signal not threaded into the PUT.** `GroupMdWriteParams` supports `signal`, but the tool handler (`group-md-tool.ts:99`) never passes the turn's abort signal, so a write-back cannot be cancelled with the turn. Minor.
- **`updateGroupMd` returns `{version: 0}` on an empty 2xx body** (`api.ts:768`), which the coordinator then caches and reports as success. Harmless (a 2xx means the write was accepted) but `version: 0` shown to the owner is misleading.
- **The cron `mcpServers` assignment doesn't spread existing servers** (`index.ts:852`); it works today only because the write-back block runs after it and *does* spread. Latent clobber risk if the two blocks are ever reordered — a small robustness nit.

## 3. Overall verdict — APPROVE

Spec ✅ AND quality Approved. Security boundary (owner-gate against untrusted chat rewriting a trusted prompt) is sound; concurrency and cache handling are correct; tests are meaningful and green. The five items above are all P2 polish, none blocking.

PR #175, opened 2026-07-01
feat(prompt): identify the sender on the current-message anchor and unify recent-messages format

JianwenXu:
### Problem

The `[Current message — respond to this ONLY]` body carried no sender label, so an agent that shares context with other participants had no reliable way to tell who spoke the current message. On top of that, the `[Recent group messages]` block rendered the wire `fromName` even when it was missing or echoed the uid, so human users often appeared as `uid(uid)：` instead of their real display name.

Concretely, a group participant reading its own prompt would see:

```
[Recent group messages]
...
ada105087fbe...(ada105087fbe...)：earlier line

[Current message — respond to this ONLY]
fix something
```

No way to know who "current message" belongs to; and even in the history block, the display name never resolves for human users because the wire `fromName` equals the uid.

### Change

This PR unifies both surfaces on `name(uid)：` and threads real display-name resolution through the write path and both renderers.

- **`src/prompt-safety.ts`**: add `formatSenderLabel(uid, name?)` → returns `"name(uid)"` when a real name is present, else `"uid"`, else `"unknown"`. Sanitized so a hostile display name can't forge a section marker (`[...]`).
- **`src/group-context.ts`**: add `GroupContext.resolveDisplayName(channelId, uid, wireName?)`. Priority: `refreshMembers` roster → wire `fromName` (only when non-empty AND not an echo of the uid) → `undefined`. Called from:
  - `pushMessage` — write-time upgrade so new rows land with the correct name in the DB
  - `buildContext` (in-memory renderer)
  - `buildContextSince` (DB renderer, from `group_messages` rows via new `from_uid` column already in schema)
- **`src/index.ts`**: prefix the current-message body with `${formatSenderLabel(uid, name)}：` in group channels only; DMs are unchanged.

After the change the same participant sees:

```
[Recent group messages]
...
许建文(ada105087fbe...)：earlier line

[Current message — respond to this ONLY]
许建文(ada105087fbe...)：fix something
```

### Design notes

- `resolveDisplayName` returns `undefined` rather than a bare uid so each caller decides its own fallback via `formatSenderLabel`.
- Renderers use nullish-coalescing on top of the resolver (`resolveDisplayName(...) ?? m.fromName`) so behaviour is unchanged when no better name is known — minimum-surprise for existing consumers.
- Sanitization in `formatSenderLabel` strips `[]` and control characters so a crafted display name can't forge headers like `[Current message ...]` inside the prompt.

### Tests

- `+5` in `prompt-safety.test.ts` for `formatSenderLabel` (uid-only, name+uid, sanitization, empty name, hostile input)
- `+6` in `group-context.test.ts` for `resolveDisplayName` (roster hit, wire fallback, wire-echoes-uid guard, missing roster, no wire, priority ordering)
- `+3` in `group-context.test.ts` for renderer roster fallback and pushMessage write-time upgrade
- Existing `group-context` and `e2e` assertions updated to the new `name(uid)：` format

Type-check clean. Full suite delta 0 vs `main` baseline (15 pre-existing GROUP.md filesystem-test failures unchanged).

### Verification

Deployed to a live cc-channel-octo running two Claude-Code participants in the same Octo group; both confirmed they now see `许建文(ada105087fbe...)：` on the current-message anchor and the same format in the recent-messages block. Before the change, one participant reported "current message has only uid, no display name" and "human user rendered as uid(uid)"; after the change both are resolved.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,7 +16,7 @@ import { OctoGateway } from './gateway.js';
 import { SessionRouter } from './session-router.js';
 import { GroupContext } from './group-context.js';
 import { queryAgent } from './agent-bridge.js';
-import { sanitizeDisplayName, escapeSectionMarkers, sanitizePromptBody } from './prompt-safety.js';
+import { sanitizeDisplayName, escapeSectionMarkers, sanitizePromptBody, formatSenderLabel } from './prompt-safety.js';
 import type { SessionCtx } from './cwd-resolver.js';
 import { cleanupExpiredCwds, resolveMemoryDir, resolveSessionCwd } from './cwd-resolver.js';
 import { StreamRelay } from './stream-relay.js';
@@ -625,10 +625,30 @@ export async function handleMessage(
       // persists the raw user content without the quote prefix to avoid prefix
       // duplication on conversation replay.
       //
-      // The final user message is assembled AFTER history is built (below), so
-      // the one-time history block + group-context delta can be prepended and
-      // the whole payload capped together. See the assembly near queryAgent.
-      const userBody = quotePrefix + bodyText;
+      // Sender-identity prefix (groups only): the `[Current message — respond to
+      // this ONLY]` anchor previously carried an anonymous body, so a shared-group
+      // Claude Code / Claw agent could not tell WHICH member sent the request
+      // being replied to (the anchor block itself has no sender field, unlike
+      // `[Recent group messages]` where each line already prefixes `name(uid)：`).
+      // We prepend `name(uid)：` — sanitized via formatSenderLabel — to align the
+      // current-message identity semantics with the historical block, so the
+      // agent has a single, uniform convention for identifying speakers across
+      // the whole prompt. DM channels stay anonymous: there's exactly one human
+      // party, adding the prefix would be pure noise. The prefix is persisted
+      // via appendUser below too — otherwise a session-replay would drop identity
+      // from history while keeping it on current, defeating the whole point.
+      //
+      // Resolve the displayName via GroupContext so the wire's from_name (which
+      // is optional and, when present, sometimes just echoes the uid) is
+      // upgraded to the roster displayName learned via refreshMembers. Without
+      // this, human users whose wire payload lacks from_name render as
+      // `uid：body` instead of `name(uid)：body`, breaking the promised uniform
+      // shape.
+      const resolvedName = isGroup && msg.channel_id
+        ? groupContext.resolveDisplayName(msg.channel_id, msg.from_uid, msg.from_name)
+        : undefined;
+      const senderPrefix = isGroup ? `${formatSenderLabel(msg.from_uid, resolvedName ?? msg.from_name)}：` : '';
+      const userBody = quotePrefix + senderPrefix + bodyText;
       // G4: Backfill history from API when local cache is empty for groups.
       // Only triggered on first interaction with a group (cold start) to avoid
       // duplicate API calls; checked via a sentinel marker stored in-memory.
--- a/src/index.ts
+++ b/src/index.ts
@@ -647,7 +647,7 @@ export async function handleMessage(
       const resolvedName = isGroup && msg.channel_id
         ? groupContext.resolveDisplayName(msg.channel_id, msg.from_uid, msg.from_name)
         : undefined;
-      const senderPrefix = isGroup ? `${formatSenderLabel(msg.from_uid, resolvedName ?? msg.from_name)}：` : '';
+      const senderPrefix = isGroup ? `${formatSenderLabel(msg.from_uid, resolvedName)}：` : '';
       const userBody = quotePrefix + senderPrefix + bodyText;
       // G4: Backfill history from API when local cache is empty for groups.
       // Only triggered on first interaction with a group (cold start) to avoid
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -473,10 +473,13 @@ describe('E2E smoke tests', () => {
     const msg = makeGroupMsg('What is this code?', true);
     await simulateMessage(msg, config, store, router, groupContext, streamRelay);
 
-    // queryAgent was called with user message in user role
+    // queryAgent was called with user message in user role. The body is prefixed
+    // with the sender label `name(uid)：` in group channels so the agent can
+    // identify the speaker across shared-context participants (unified with the
+    // `[Recent group messages]` format).
     expect(queryAgent).toHaveBeenCalledTimes(1);
     const userMsg2 = (queryAgent as ReturnType<typeof vi.fn>).mock.calls[0][0] as string;
-    expect(userMsg2).toBe('What is this code?');
+    expect(userMsg2).toBe('TestUser(user-001)：What is this code?');
 
     // Output was sent to group channel
     expect(sendMessage).toHaveBeenCalled();
@@ -542,7 +545,7 @@ describe('E2E smoke tests', () => {
 
     // But message IS cached in group context
     const ctx = groupContext.buildContext(GROUP_CHANNEL);
-    expect(ctx).toContain('TestUser：Just chatting');
+    expect(ctx).toContain('TestUser(user-001)：Just chatting');
   });
 
   // --- 4. Non-text message: G1 routes it through to the agent ---
@@ -647,13 +650,13 @@ describe('E2E smoke tests', () => {
     // system-prompt arg. The prior chatter is present; the current message is NOT
     // echoed back into the [Recent group messages] block.
     const userMsg3 = (queryAgent as ReturnType<typeof vi.fn>).mock.calls[0][0] as string;
-    expect(userMsg3).toContain('Alice：Previous message');
+    expect(userMsg3).toContain('Alice(other-user)：Previous message');
     expect(userMsg3).toContain('My question'); // the actual question (the body) is present
     // The [Recent group messages] context block (everything up to the body) must
     // not include the current message — only prior chatter.
     const recentIdx = userMsg3.indexOf('[Recent group messages]');
     const contextBlock = userMsg3.slice(recentIdx, userMsg3.lastIndexOf('My question'));
-    expect(contextBlock).toContain('Alice：Previous message');
+    expect(contextBlock).toContain('Alice(other-user)：Previous message');
     expect(contextBlock).not.toContain('My question');
   });
 
--- a/src/__tests__/group-context.test.ts
+++ b/src/__tests__/group-context.test.ts
@@ -81,8 +81,10 @@ describe('GroupContext', () => {
     ctx.pushMessage('ch1', 'u2', 'Bob', 'world', 2000);
     const output = ctx.buildContext('ch1');
     expect(output).toContain('[Recent group messages]');
-    expect(output).toContain('Alice：hello');
-    expect(output).toContain('Bob：world');
+    // New format: `name(uid)：content` — identity semantics aligned with the
+    // current-message anchor so an agent has ONE way to identify speakers.
+    expect(output).toContain('Alice(u1)：hello');
+    expect(output).toContain('Bob(u2)：world');
     // Alice should come before Bob (chronological)
     expect(output.indexOf('Alice')).toBeLessThan(output.indexOf('Bob'));
   });
@@ -250,8 +252,8 @@ describe('GroupContext', () => {
 
     // Should restore messages and produce context
     const context = ctx2.buildContext('ch1');
-    expect(context).toContain('Alice：msg1');
-    expect(context).toContain('Bob：msg2');
+    expect(context).toContain('Alice(u1)：msg1');
+    expect(context).toContain('Bob(u2)：msg2');
   });
 
   // --- Q16: loadAllFromDb ---
@@ -273,9 +275,9 @@ describe('GroupContext', () => {
 
     // Messages should be restored
     const context1 = ctx2.buildContext('ch1');
-    expect(context1).toContain('Alice：hello');
+    expect(context1).toContain('Alice(u1)：hello');
     const context2 = ctx2.buildContext('ch2');
-    expect(context2).toContain('Bob：world');
+    expect(context2).toContain('Bob(u2)：world');
   });
 });
 
@@ -299,8 +301,8 @@ describe('GroupContext consumption cursor', () => {
     ctx.pushMessage('ch1', 'u1', 'Alice', 'one', 1);
     ctx.pushMessage('ch1', 'u2', 'Bob', 'two', 2);
     const first = ctx.buildContextSince('ch1', 0);
-    expect(first.text).toContain('Alice：one');
-    expect(first.text).toContain('Bob：two');
+    expect(first.text).toContain('Alice(u1)：one');
+    expect(first.text).toContain('Bob(u2)：two');
     expect(first.lastId).toBeGreaterThan(0);
 
     // Advance the cursor; a new message arrives.
@@ -308,9 +310,9 @@ describe('GroupContext consumption cursor', () => {
     ctx.pushMessage('ch1', 'u3', 'Carol', 'three', 3);
     const second = ctx.buildContextSince('ch1', ctx.getContextCursor('ch1'));
     // Only the new message is included; the old ones are not re-shown.
-    expect(second.text).toContain('Carol：three');
-    expect(second.text).not.toContain('Alice：one');
-    expect(second.text).not.toContain('Bob：two');
+    expect(second.text).toContain('Carol(u3)：three');
+    expect(second.text).not.toContain('Alice(u1)：one');
+    expect(second.text).not.toContain('Bob(u2)：two');
     expect(second.lastId).toBeGreaterThan(first.lastId);
   });
 
@@ -346,14 +348,14 @@ describe('GroupContext consumption cursor', () => {
     // A small budget that fits ~2 short lines. With the old ASC fetch + newest-
     // within-budget loop the cursor would still advance past everything, dropping
     // recent lines; the newest-first fetch makes the budget keep the latest ones.
-    const small = new GroupContext(adapter, 40);
+    const small = new GroupContext(adapter, 60);
     for (let i = 1; i <= 6; i++) {
       small.pushMessage('ch1', `u${i}`, 'A', `m${i}`, i);
     }
     const out = small.buildContextSince('ch1', 0);
     // The most-recent message must be present; the oldest must be the one dropped.
-    expect(out.text).toContain('A：m6');
-    expect(out.text).not.toContain('A：m1');
+    expect(out.text).toContain('A(u6)：m6');
+    expect(out.text).not.toContain('A(u1)：m1');
     // Cursor advances past the whole delta (highest existing id), so dropped-oldest
     // lines are never re-shown on a later turn.
     expect(out.lastId).toBe(small.getMaxMessageId('ch1'));
@@ -368,7 +370,7 @@ describe('GroupContext consumption cursor', () => {
     }
     const out = ctx.buildContextSince('ch1', 0);
     // Newest message reaches the model.
-    expect(out.text).toContain('A：m130');
+    expect(out.text).toContain('A(u)：m130');
     // lastId reflects the true max in the channel, not just the last fetched row.
     expect(out.lastId).toBe(ctx.getMaxMessageId('ch1'));
   });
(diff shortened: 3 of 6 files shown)


OctoBoooot (commented):
## Review: feat(prompt): identify the sender on the current-message anchor and unify recent-messages format (#175)

**Verdict:** Approve with comments — the sanitization is sound; one description-vs-code mismatch to reconcile, and CI must go green first.

CI is red — `check-sprint / check-sprint` is failing (project-board gate, not code). type-check/lint/vitest pass. Verdict can't be APPROVE until that clears; no code blockers.

Reviewed at prompt-injection rigor since a user-controlled display name now flows into the `[Current message]` anchor and every `[Recent group messages]` line. Enumerating the contexts the name reaches:
- Anchor body prefix (index.ts:650) → `formatSenderLabel` → both fields through `sanitizeDisplayName`.
- In-memory renderer (group-context.ts buildContext) → `formatSenderLabel`.
- DB renderer (buildContextSince) → `formatSenderLabel`.
- Write-time `pushMessage` → `sanitizeDisplayName` chain.

All four go through `sanitizeDisplayName`, and `NAME_UNSAFE_RE = /[[\]\r\n\v\f  ]/g` strips `[`, `]`, and every line-break variant — so a hostile name can't forge a `[section]` marker or inject a newline to fake a turn/anchor boundary. The `(uid)` wrapper is new but adds only literal parens, which aren't structural markers. Sanitization coverage is complete across the embedding sites.

### Minor / Question
- **PR body vs code mismatch on persistence.** The body states the sender prefix "is persisted via appendUser below too — otherwise a session-replay would drop identity from history." But `store.appendUser` (index.ts:695) stores `userContent` (the raw payload, line 583), NOT `userBody` (line 651, which carries `senderPrefix`). So the prefix is on the live turn's `userBody` (and thus in the resumed SDK session) but is NOT written to the session store. Identity in the store-replay fallback path comes from a different mechanism — `appendUser`'s `from_name` arg + the group-context `name(uid)：` renderer — not from the stored body. The behavior is defensible (SDK session carries the prefixed body; group-context renders identity separately), but the body's stated rationale describes a persistence path the code doesn't implement. Either the prefix should be stored (use `userBody`/a prefixed record like the existing `historyRecord` split), or the body/comment should be corrected to say identity in history comes from the renderer, not the stored prefix. Worth settling because the "otherwise session-replay drops identity" justification is load-bearing for the design and currently doesn't match `appendUser`'s argument.

### Praise
- `formatSenderLabel` routing both `uid` AND `name` through `sanitizeDisplayName` (not just the name) closes the raw-uid-as-fallback re-injection that PR #128's review caught — the uid is user-adjacent too, and this doesn't trust it.
- `resolveDisplayName`'s wire-echoes-uid guard (`fromName !== fromUid`) is the right call: it prevents the `uid(uid)：` degenerate rendering the PR set out to fix, and it's tested.
- Renderers consult `resolveDisplayName` as a pure read (learning is separate) so an old row written before the roster caught the real name still renders with the current display name once `refreshMembers` populates — a nice consistency property, and the comment says why render stays side-effect-free.

- `check-sprint` red is the same project-board gate as the rest of this batch.
- 15 pre-existing GROUP.md filesystem-test failures are noted in the body as unchanged-vs-main — out of scope for this PR.

src/index.ts
```diff
@@ -625,10 +625,30 @@ export async function handleMessage(
...
+      // is optional and, when present, sometimes just echoes the uid) is
+      // upgraded to the roster displayName learned via refreshMembers. Without
+      // this, human users whose wire payload lacks from_name render as
+      // `uid：body` instead of `name(uid)：body`, breaking the promised uniform
+      // shape.
+      const resolvedName = isGroup && msg.channel_id
+        ? groupContext.resolveDisplayName(msg.channel_id, msg.from_uid, msg.from_name)
+        : undefined;
+      const senderPrefix = isGroup ? `${formatSenderLabel(msg.from_uid, resolvedName ?? msg.from_name)}：` : '';
+      const userBody = quotePrefix + senderPrefix + bodyText;
```
OctoBoooot, line 651: **[minor]** the PR body says this prefix "is persisted via appendUser below too — otherwise a session-replay would drop identity from history." But `store.appendUser` (line 695) stores `userContent` (raw payload, line 583), not `userBody` — so `senderPrefix` reaches the live turn's `userBody` (and the resumed SDK session) but is NOT written to the session store. Identity in the store-replay fallback comes from `appendUser`'s `from_name` arg + the group-context `name(uid)：` renderer, a different path than the body claims. Behavior is defensible, but either store the prefixed form (cf. the existing `historyRecord`/`bodyText` split) or correct the body/comment so the stated rationale matches what `appendUser` actually persists.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# cc-channel-octo PR#175 — automated review Review Report
**Head SHA:** `dcda077e6d1249e2f9b199a0a4649028893943f4`
**PR:** feat(prompt): identify the sender on the current-message anchor and unify recent-messages format

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| `formatSenderLabel` correctness | ✅ | `prompt-safety.ts:121-124` — both uid and name go through `sanitizeDisplayName`; returns `name(uid)` or bare `uid` |
| `resolveDisplayName` priority chain | ✅ | `group-context.ts:168-175` — roster → wire name (non-empty, non-uid-echo) → `undefined` |
| Sanitization at every consumer | ✅ | All 4 call sites verified: `pushMessage` (explicit), `buildContext`, `buildContextSince`, `index.ts` (all via `formatSenderLabel` which sanitizes internally) |
| SQL schema alignment | ✅ | `group_messages` table already has `from_uid TEXT NOT NULL`; `selectMessagesSince` now selects it |
| `pushMessage` write-time upgrade | ✅ | `group-context.ts:187-196` — resolves name before storing, then calls `learnMember` to update roster |
| DM path unaffected | ✅ | `index.ts:647-650` — `isGroup` guard ensures DMs get `senderPrefix = ''` |
| Test coverage | ✅ | 14 new tests (5 `formatSenderLabel` + 6 `resolveDisplayName` + 3 renderer/pushMessage); existing assertions updated to `name(uid)：` format |
| Prompt-injection defense | ✅ | `NAME_UNSAFE_RE` strips `[]\r\n\v\f\u0085\u2028\u2029`; `escapeSectionMarkers` provides defense-in-depth at block level |

## 2. Findings

### No P0/P1 Findings

After tracing every data flow from wire payload → `resolveDisplayName` → `formatSenderLabel` → rendered prompt, and applying the severity rubric R1–R4:

- **No path becomes unavailable or produces user-visible incorrect data.** The change is additive: it adds sender identity where none existed before and improves name resolution. All fallback chains degrade gracefully to bare uid or `'unknown'`.
- **No diff-scope violation.** All behavioral changes (new format, sender prefix) are intentional and tested.

### P2 — Minor Observations

**P2-1: Per-line character cost increase reduces history line count within fixed budget**
- **Diff-scope:** new (introduced by this PR's format change)
- The `name(uid)：` format is ~5-20 chars longer per line than the prior `name：` format. With `maxContextChars` unchanged, fewer history lines fit within the budget. Test budget adjusted from 40→60 (`group-context.test.ts:349`). This is expected and not a regression — the identity information is worth the cost — but operators with tight context budgets may notice slightly shorter history windows.
- **Suggestion:** No action needed; noted for awareness. If context budget becomes a concern, consider a separate tuning PR.

**P2-2: Double sanitization in `pushMessage` (harmless, idempotent)**
- **Diff-scope:** new
- `resolveDisplayName` sanitizes wire names internally, then `pushMessage` applies `sanitizeDisplayName(resolvedName)` again. This is idempotent and correct, but slightly wasteful. Not worth changing — the explicit sanitization at the boundary is good defensive practice.

## 3. Suggestions

- None blocking. The code is clean, well-documented, and the test suite thoroughly covers the new behavior including edge cases (uid-echo guard, channel scoping, hostile name sanitization, roster-after-write rendering).

## 4. Additional Observations

- The design decision to make `resolveDisplayName` read-only (no `learnMember` call) and let `pushMessage` handle learning separately is architecturally sound — it prevents renderer paths from mutating state during reads.
- The `formatSenderLabel` mirroring openclaw-channel-octo's `buildSenderPrefix` ensures cross-adapter consistency for agents that see both channel types.

## 5. Data Flow Trace

| Consumer | Data | Upstream Source | Runtime Path Verified |
|----------|------|----------------|----------------------|
| `pushMessage.safeName` | `resolvedName` | `resolveDisplayName` → roster (from `refreshMembers` API) or wire `fromName` | ✅ roster populated before push via `refreshMembers` in `handleMessage`; fallback chain to sanitized uid |
| `buildContext` line | `displayName` | `resolveDisplayName(channelId, m.fromUid, m.fromName) ?? m.fromName` | ✅ roster hit returns current name; cached `m.fromName` (already sanitized at write time) as fallback |
| `buildContextSince` line | `displayName` | `resolveDisplayName(channelId, rows[i].from_uid, rows[i].from_name) ?? rows[i].from_name` | ✅ `from_uid` now selected from DB (column exists in schema); same fallback logic |
| `index.ts` senderPrefix | `resolvedName ?? msg.from_name` | `resolveDisplayName(msg.channel_id, msg.from_uid, msg.from_name)` | ✅ `refreshMembers` called before this point in `handleMessage`; DM guard (`isGroup`) prevents prefix |
| `formatSenderLabel` output | sanitized `name(uid)` | `sanitizeDisplayName` on both uid and name | ✅ strips `[]\r\n\v\f\u0085\u2028\u2029`, caps at `MAX_DISPLAY_NAME_LEN` |

- **C1 — Dual-path parity:** N/A. No add/remove or subscribe/unsubscribe symmetry in this change.
- **C2 — Control-flow ordering / nested reuse:** Clear. `formatSenderLabel` and `resolveDisplayName` are simple linear functions with no nesting risk. Called at 4 sites; all apply sanitization consistently.
- **C3 — Authorization boundary:** N/A. No permission/jail/tool changes.
- **C4 — Authorization lifecycle / container-member cascade:** N/A. No auth changes.
- **C5 — Build ≠ runtime:** N/A. Node.js server, no packaging/extension concerns.
- **C6 — Governance/policy docs:** N/A. Code change, no policy documents.

## 7. Cross-round Blocker Recheck (R6)

---

[Octo-Q] verdict: **APPROVE**

No P0/P1 findings. The change cleanly resolves the stated problem (anonymous current-message anchor + inconsistent name resolution in recent-messages block) with proper sanitization, well-tested fallback chains, and no regressions. The two P2 observations (budget impact, double sanitization) are expected consequences of the format change and do not affect correctness or safety.

Jerry-Xin (dismissed):
This PR is in scope and the direction is right (a shared `formatSenderLabel(uid, name)` choke point, sanitizing both fields, and threading real display-name resolution through the write path and both renderers). Sanitization is solid: `formatSenderLabel` routes both name and uid through `sanitizeDisplayName` (strips `[` `]` and all line/paragraph separators, caps length), so a hostile display name cannot forge a section marker or break into a new line, and the recent-messages block still passes through `sanitizePromptBody` (`escapeSectionMarkers` + `escapeRoleLabels`) before it enters the prompt. Verified empirically. However, the core behavioral goal — never render `uid(uid)：` for a human — is not actually achieved on the common path, and the write path can poison the roster cache.

**🔴 Blocking**

- 🔴 The `fromName === fromUid` guard is defeated by the fallback chain, so the PR can still render `uid(uid)：…` — the exact bug it claims to fix. In `src/group-context.ts` `resolveDisplayName()` correctly returns `undefined` for an exact uid echo, but `pushMessage()` then falls back to `sanitizeDisplayName(fromName)` and stores that uid as `from_name`. The renderers (`buildContext`, `buildContextSince`) do `resolveDisplayName(...) ?? m.fromName`, and the current-message path in `src/index.ts` does `formatSenderLabel(msg.from_uid, resolvedName ?? msg.from_name)`. When wire `from_name === from_uid === "u1"` and the roster has no better entry (cold start, or a uid the roster hasn't learned), the chain resolves back to `"u1"` and `formatSenderLabel("u1","u1")` emits `u1(u1)：body`. Reproduced by inlining the actual helpers: stored `from_name = "u1"`, rendered label = `u1(u1)：body`. The PR description explicitly promises this case renders the real name; it does not unless `refreshMembers` has already populated the roster.

- 🔴 `pushMessage()` can persist a uid echo as an authoritative roster name and then block a later real wire name. On a first message with `fromName === fromUid`, `safeName` is the uid and `learnMember(channelId, fromUid, safeName)` stores `u1 -> "u1"` in the roster. Afterwards, if the wire supplies a real name (e.g. `"Alice"`) before any roster refresh, `resolveDisplayName()` returns the cached `"u1"` (roster is priority 1) instead of the real wire name — reproduced: `resolveDisplayName(roster{u1:"u1"}, "u1", "Alice") === "u1"`. This violates the documented priority (the roster is supposed to hold real names, not uid echoes) and can keep both current-message and recent-message rendering stuck at `u1(u1)` until an external roster update overwrites it.

Suggested fix: treat a uid echo as "no name" everywhere, not just on the wire input — (1) do not `learnMember` a uid-only fallback (only learn a value that differs from the uid after sanitization), and (2) apply the same "non-empty AND not a uid echo after sanitization" predicate to the roster value inside `resolveDisplayName` so a poisoned `u1 -> "u1"` entry does not win over a real wire name.

**💬 Non-blocking**

- 🟡 Test gap: the new tests cover "uid echo later corrected by an explicit `learnMember('Alice')`", but not the failing transitions above. Add: `pushMessage('u1','u1')` then `pushMessage('u1','Alice')` (expect it to upgrade, not stay `u1(u1)`); a `buildContext`/`buildContextSince` render assertion for the cold-start uid-echo case (expect it NOT to contain `u1(u1)：`); and current-message formatting when `msg.from_name === msg.from_uid`.
- 🟡 Could not run the targeted suite in the review checkout (`vitest` not installed in the mirror); logic was verified by reproducing the helpers directly rather than via the test runner.
- 🟡 `sanitizeDisplayName` does not strip a fullwidth colon `：`, so a name like `A：x` renders `A：x(uid)：body`. Not an injection (newlines are stripped, so it stays one line and cannot forge a separate sender row), just a cosmetic ambiguity — noting for awareness.

**✅ Highlights**
- Shared `formatSenderLabel` helper with defense-in-depth sanitization on both fields is the right structural move, and mirrors the openclaw-channel-octo convention.
- Selecting `from_uid` in the DB renderer (`buildContextSince`) is the correct structural change to make recent-message identity explicit.
- Sender attribution is correct: the current-message prefix uses the message's own `from_uid`/`from_name`, and each recent-message row carries its own uid — no off-by-one / cross-attribution.

Note: CI `check-sprint` is failing because the PR description has no `Closes #<issue>` link (a sprint/board gate, not a code/compile failure); please add the linked issue so the merge gate can pass.

src/index.ts
```diff
@@ -625,10 +625,30 @@ export async function handleMessage(
...
+      // Resolve the displayName via GroupContext so the wire's from_name (which
+      // is optional and, when present, sometimes just echoes the uid) is
+      // upgraded to the roster displayName learned via refreshMembers. Without
+      // this, human users whose wire payload lacks from_name render as
+      // `uid：body` instead of `name(uid)：body`, breaking the promised uniform
+      // shape.
+      const resolvedName = isGroup && msg.channel_id
+        ? groupContext.resolveDisplayName(msg.channel_id, msg.from_uid, msg.from_name)
+        : undefined;
+      const senderPrefix = isGroup ? `${formatSenderLabel(msg.from_uid, resolvedName ?? msg.from_name)}：` : '';
```
OctoBoooot: **[major]** `resolvedName ?? msg.from_name` defeats `resolveDisplayName`'s deliberate `undefined` return and re-creates `uid(uid)`. In the echo case (roster miss AND `from_name === from_uid`), `resolveDisplayName` returns undefined to mean "emit bare uid," but `?? msg.from_name` re-supplies the uid, so `formatSenderLabel(uid, uid)` → `uid(uid)` — the exact rendering this PR fixes. Failing input: group message, sender not in roster (refreshMembers failed/absent), wire `from_name` echoes uid → anchor renders `uid(uid)：body`; both group-context renderers regress identically via `?? m.fromName`. Fix: pass `resolvedName` directly (drop `?? fromName`) — `resolveDisplayName` already returns a real wire name at priority 2 and undefined otherwise, so `formatSenderLabel`'s `name ? name(uid) : uid` yields bare uid. Add a roster-miss + echo regression test asserting `uid：` not `uid(uid)：`. Credit Jerry-Xin.

OctoBoooot (dismissed):
## Review (follow-up): #175 — `?? fromName` fallback re-introduces `uid(uid)` (credit Jerry-Xin)

**Updated verdict:** Request changes — Jerry-Xin's first 🔴 is real and I missed it in my prior COMMENT. Upgrading.

### Major
- `src/index.ts:650` (and both renderers in `src/group-context.ts` — buildContext, buildContextSince) — the `resolvedName ?? msg.from_name` fallback defeats `resolveDisplayName`'s deliberate `undefined` return and re-creates the exact `uid(uid)` rendering this PR set out to fix.

  Walk: `resolveDisplayName` returns `undefined` in the echo case (roster miss AND `from_name === from_uid`) — by design, to mean "no real name, emit bare uid." But `resolvedName ?? msg.from_name` overrides that `undefined` with `msg.from_name`, which *is* the uid in the echo case. `formatSenderLabel(uid, uid)` then sees a non-empty name → returns `uid(uid)`.

  Failing input: group message where the sender isn't resolved via roster (refreshMembers failed / member absent from roster) and the wire `from_name` echoes the uid (the exact case the PR body calls out as common for human users) → the anchor renders `uid(uid)：body`, and the `[Recent group messages]` lines do the same via their identical `?? m.fromName`.

  Fix: pass `resolvedName` directly to `formatSenderLabel`, dropping `?? msg.from_name` / `?? m.fromName`. `resolveDisplayName` already returns a *real* wire name at priority 2 (non-empty AND not an echo), and returns `undefined` precisely so `formatSenderLabel`'s own `name.length > 0 ? name(uid) : uid` produces the bare `uid`. The `?? fromName` only ever re-adds what `resolveDisplayName` deliberately rejected — harmless for empty, but the uid-echo is the bug. This matches the PR's own stated design ("resolveDisplayName returns undefined rather than a bare uid so each caller decides its own fallback via formatSenderLabel") — the `?? fromName` is what breaks it.

  Add a regression test: roster-miss + `from_name === from_uid` → asserts bare `uid：` (not `uid(uid)：`) at both the anchor and the recent-messages block.

### On Jerry-Xin's other points
- 🔴 #2 (pushMessage persists the uid-echo name, "blocking real name update") — same root, but weaker: the renderers re-resolve via `resolveDisplayName(channelId, m.fromUid, m.fromName)` at render time, so once `refreshMembers` populates the roster the stored uid IS upgraded to the real name. It only stays wrong while the roster lacks the uid — which is the same condition as the major above. Fixing the fallback + not persisting a uid-as-name (store `undefined`/keep resolving at render) covers both.
- 🟡 full-width colon (`：`) not stripped — minor: the label separator is `：`, and a name containing it is a display ambiguity, not a structural forge (renderers don't split on it). Worth a follow-up, not blocking.

My COMMENT (4607363203) verified sanitization coverage and caught the appendUser persistence-doc mismatch, but I missed that the `?? fromName` fallback re-introduces `uid(uid)` — the PR's core goal — in the roster-miss/echo path. Credit Jerry-Xin. The sanitization analysis and the appendUser note still stand.

yujiawei (approved):
# Code Review — PR #175 (cc-channel-octo)

Reviewed at head `dcda077e6d1249e2f9b199a0a4649028893943f4` against `main` (merge-base `6401c0b`). Type-check clean, full suite green (1179 passed, 60 files), including the +14 new tests. This review absorbs an independent line-level pass and a whole-repo cross-family pass; both were consulted and reconciled below.

## 1. Spec compliance

The stated goal — put a `name(uid)：` label on the current-message anchor (groups only) and unify the `[Recent group messages]` block on the same shape, with real display-name resolution threaded through the write path and both renderers — is fully and only implemented.

- **Under-build:** none. `formatSenderLabel` (prompt-safety), `resolveDisplayName` (group-context), the `pushMessage` write-time upgrade, both renderers (`buildContext` in-memory + `buildContextSince` DB), and the `index.ts` anchor prefix are all present.
- **Over-build:** none. DMs are correctly left anonymous (`senderPrefix` gated on `isGroup`). `from_uid` was already in the `group_messages` schema on `main`, so no migration was smuggled in.
- **Deviation:** none functional. One **documentation** deviation, see Q1 below.

**Spec: PASS**

## 2. Code quality

No P0/P1. Three P2 items, all advisory — none blocks merge.

### P2 — Q1: the "prefix is persisted" comment is inaccurate (doc-only)

`src/index.ts:637-639` states the sender prefix is persisted "via appendUser below too — otherwise a session-replay would drop identity from history." It is not: `store.appendUser(sessionKey, userContent, ...)` at `src/index.ts:695` passes the **raw** `userContent`, while the prefixed `userBody` (`src/index.ts:651`) flows only into `assembleUserMessage` for the live LLM turn.

Importantly, the **code is correct** — this is a comment bug, not a logic bug. History replay renders identity through `SessionStore.renderTurn` as `[user <from_name>]: <body>` (`src/session-store.ts:323`), and `src/index.ts:624` explicitly documents that the raw body is persisted to avoid prefix duplication on replay. Persisting `userBody` instead would double-stamp identity (`[user 许建文]: 许建文(uid)：…`). So keep the code; fix the comment to say the anchor prefix is a live-turn-only concern and history identity is carried by the stored `from_name` + `renderTurn`.

### P2 — Q2: residual `uid(uid)：` when roster misses AND wire echoes the uid (consensus)

Both review legs independently flagged this. `resolveDisplayName` deliberately returns `undefined` when `fromName === fromUid` (`src/group-context.ts:161`), but every call site then does `resolveDisplayName(...) ?? m.fromName` / `?? msg.from_name` (`src/group-context.ts:374,467`, `src/index.ts:650`), which resurrects the very uid-echo the resolver rejected and hands two identical strings to `formatSenderLabel` → renders `uid(uid)：`.

Scope is narrow: it needs a roster miss (name not yet learned via `refreshMembers`/`learnMember`) **and** a wire payload whose `from_name` equals `from_uid`. It is cosmetic and is not a regression (main rendered a bare `uid：` in the same case). Cleanest fix is a guard inside `formatSenderLabel`: collapse to bare `uid` when the sanitized name equals the sanitized uid. That closes it at the single choke point regardless of caller fallbacks.

### P2 — Q3: `resolveDisplayName` returns the roster name unsanitized (consensus, latent)

`src/group-context.ts:162` returns `rosterName` directly; the wire-fallback branch below it is wrapped in `sanitizeDisplayName`. Every current consumer (`formatSenderLabel`, `pushMessage`) re-sanitizes downstream, so there is **no live injection** today. But the new `prompt-safety.test.ts` comment frames the resolver as returning "an already-safe value [that] keeps every consumer honest," and that contract is not actually upheld for the roster branch. If a future caller trusts the return value directly, a roster name carrying `]`/newline (roster comes from the Octo members API — a trust boundary) could forge a section marker. Either sanitize the roster branch here, or sanitize once at the `learnMember`/`refreshMembers` ingestion point so the roster is safe-by-construction. Note the write path already leans this way (`pushMessage` sanitizes into `safeName`), but `refreshMembers` upserts `m.name` into `memberMap`/DB **without** sanitizing (`src/group-context.ts:275-277`), so ingestion-time is the more complete fix.

## 3. Verdict

**APPROVE.** Spec is exactly met, security posture is sound (the two-site sanitization choke point holds for all live paths), tests are meaningful and green. Q1 is a stale comment, Q2/Q3 are narrow, non-regressing hardening. Recommend addressing at least Q1 (comment) and Q2 (`formatSenderLabel` self-collapse) in a follow-up, and folding Q3 into the roster-ingestion sanitization for defense-in-depth — none are merge blockers.

### Coverage / blind spots
- Verified: type-check, full 1179-test suite, both renderers, DM gating, `from_uid` pre-existing on `main`, replay path in `session-store`, `pushMessage`/`refreshMembers` roster writes.
- Not verified (out of diff): live Octo `/groups/<id>/members` response shape (the real injection risk for Q3 depends on whether the API can return bracket-bearing names), and downstream agent system-prompt assembly beyond `assembleUserMessage`.

Jerry-Xin (approved):
Re-review at fc2320d1c9ae. Both blockers from my prior review (dcda077e) are fixed at the root and covered by regression tests — clearing to APPROVE.

Verdict: APPROVE.

Resolved blockers (verified against the diff):
1. uid(uid) on uid-echo cold-start — the `resolvedName ?? fromName` fallback that re-injected the uid is removed at all three sites: the current-message anchor (index.ts) and both group-context renderers (buildContext + buildContextSince). resolveDisplayName now returns undefined on a uid echo and formatSenderLabel emits the bare uid, exactly as the design intended.
2. Roster poisoning — resolveDisplayName now guards the roster-value branch with `rosterName !== fromUid`, so a poisoned uid-echo entry no longer shadows a later real wire name. Root cause also closed: pushMessage only calls learnMember when safeName is a genuine displayName (`safeName !== fromUid && safeName !== "unknown"`), so a uid echo is never written as an authoritative roster name in the first place. The wire-fallback branch is likewise hardened with `safe !== fromUid`.

Tests: new regression coverage asserts bare-uid rendering on cold-start echo (both in-memory and DB paths), that a poisoned roster entry is ignored in favor of a subsequent real name, and that pushMessage does not lock the roster to a uid echo — precisely the transition that was previously untested.

Sanitization/injection remains sound (unchanged): name and uid both route through sanitizeDisplayName; recent-messages bodies through escapeRoleLabels/escapeSectionMarkers.

Non-blocking:
- sanitizeDisplayName still does not strip the fullwidth colon; cosmetic only (no line break = no forged row).
- check-sprint CI is red, but it is the sprint-board process gate (missing a `Closes #<issue>` link), not a compile/test failure — type-check/eslint/vitest are green. Please add the issue link to satisfy that gate before merge.

yujiawei (approved):
# Code Review — PR #175 (cc-channel-octo) — re-review at head fc2320d

Re-reviewed at head `fc2320d1c9aeb7c5f6f63a21ef1ebc47b4d6bd92`. A prior approval was posted against `dcda077`; a follow-up commit (`fc2320d`, "stop the `?? fromName` fallback from re-introducing uid(uid)") landed during review, so this supersedes it and is the authoritative verdict for the current head. Type-check clean; changed-file suite green (119 passed, +4 new regression tests over the prior head). Approvals do not auto-carry across a new head — this re-verdict is against `fc2320d` specifically.

## What changed since dcda077

The follow-up commit resolves the residual `uid(uid)：` issue that both review legs flagged:
- `resolveDisplayName` now rejects a roster entry that echoes the uid (`group-context.ts:168`, `rosterName !== fromUid`) and a wire name that sanitizes back to the uid.
- `pushMessage` no longer learns a uid-echo (or `'unknown'`) as an authoritative roster name, so a real wire name arriving later is no longer blocked by a poisoned roster entry.
- All three call sites dropped the `?? fromName` fallback (`group-context.ts:384,482`, `index.ts:650`), so `formatSenderLabel` receives `undefined` and correctly emits the bare `uid` instead of `uid(uid)`.
- +4 regression tests lock the behavior (cold-start echo → bare uid in both renderers; poisoned-roster resolver ignore; pushMessage does-not-lock).

Verified against the current tree, not the commit message.

## 1. Spec compliance

Goal — `name(uid)：` on the current-message anchor (groups only) plus a unified `[Recent group messages]` format with real display-name resolution through the write path and both renderers — is fully and only implemented.

- **Under-build:** none.
- **Over-build:** none. DMs stay anonymous (`isGroup` gate); `from_uid` pre-existed in the `group_messages` schema on `main`, so no migration was added.
- **Deviation:** none functional. One documentation inaccuracy (Q1 below).

**Spec: PASS**

## 2. Code quality

No P0/P1. Two P2 items remain, both advisory — neither blocks merge.

### P2 — Q1: "prefix is persisted" comment is inaccurate (doc-only, unchanged by the fix)

`src/index.ts:637-639` still claims the sender prefix is persisted "via appendUser below too." It is not: `appendUser(sessionKey, userContent, ...)` at `src/index.ts:695` passes the **raw** `userContent`; the prefixed `userBody` flows only into `assembleUserMessage` for the live turn. The **code is correct** — persisting `userBody` would double-stamp identity on replay (`[user 许建文]: 许建文(uid)：…`), and `index.ts:624` documents the raw-body-on-replay rule; history identity is carried by the stored `from_name` + `SessionStore.renderTurn` (`session-store.ts:323`). Fix the comment, not the code.

### P2 — Q3: `resolveDisplayName` returns the roster name unsanitized (latent, unchanged by the fix)

`src/group-context.ts:168` now guards `rosterName !== fromUid` but still `return rosterName` unsanitized, while the wire branch below is wrapped in `sanitizeDisplayName`. No live injection today — every consumer (`formatSenderLabel`, `pushMessage`) re-sanitizes — but the test comment frames the resolver as returning an "already-safe value," a contract the roster branch does not uphold. Note `refreshMembers` also upserts `m.name` into the roster without sanitizing (`group-context.ts:275-277`), so the roster (fed by the Octo members API, a trust boundary) can hold bracket/newline-bearing names. Cleanest fix: sanitize at the `learnMember`/`refreshMembers` ingestion point so the roster is safe-by-construction and the resolver's stated contract becomes true.

## 3. Verdict

**APPROVE** at head `fc2320d`. Spec exactly met; the two-site sanitization choke point holds for all live paths; the `uid(uid)` edge case (Q2) is fixed and regression-tested. Q1 (stale comment) and Q3 (defense-in-depth roster sanitization) are non-blocking follow-ups.

### Coverage / blind spots
- Verified: type-check, changed-file suite (119 tests + 4 new), both renderers, DM gating, `from_uid` pre-existing on `main`, replay path in `session-store`, `pushMessage`/`refreshMembers` roster writes, the full dcda077→fc2320d delta.
- Not verified (out of diff): live Octo `/groups/<id>/members` response shape (the real risk surface for Q3), and downstream agent system-prompt assembly beyond `assembleUserMessage`.

OctoBoooot (commented):
## Re-review (delta): #175 — uid(uid) major fixed; CI red is the only thing left

**Verdict:** Approve with comments — my `uid(uid)` major (and the deeper roster-poisoning vector) is fully resolved. Can't stamp APPROVE only because `check-sprint` CI is red (project-board gate, not code).

Delta `dcda077e6d12` → `fc2320d1c9ae` addresses every vector, including one beyond my original finding:

- ✓ `index.ts:650` — `?? msg.from_name` dropped; passes `resolvedName` directly. My exact fix.
- ✓ Both renderers (`buildContext`, `buildContextSince`) — `?? m.fromName` dropped; pass `resolveDisplayName(...)` directly, so an unresolved sender emits bare `uid：` via `formatSenderLabel`'s undefined-path.
- ✓ `resolveDisplayName` roster branch — now guards `rosterName !== fromUid`, so a poisoned uid-echo roster entry can't outrank a later real wire name (the roster-poisoning 🔴 — credit Steve/Jerry-Xin for pushing past my anchor-only framing to the roster layer).
- ✓ `pushMessage` — no longer `learnMember`s a uid-echo or `'unknown'` (`if (safeName !== fromUid && safeName !== 'unknown')`), fixing poisoning at the write source.

Regression tests cover exactly the failing inputs I asked for and more: cold-start echo → bare `uid：` (not `uid(uid)`) at both renderers, poisoned-roster-ignored-when-real-name-arrives, and no-learn-uid-echo. That's the roster-miss + echo matrix pinned.

Sanitization coverage (verified in round 1) is unchanged and still complete — all name paths route through `sanitizeDisplayName`.

### Remaining merge gate (not a code finding)
- `check-sprint / check-sprint` is red — that must go green before merge/APPROVE. The code test jobs (Vitest/type-check/ESLint) aren't reporting on this head yet; confirm they pass too. Once CI is green, this is a clean APPROVE from my side — no code concerns remain.

### Still open (non-blocking, carried)
- PR-body persistence wording (prefix "persisted via appendUser") — `appendUser` still stores `userContent`, not the prefixed body; identity in history comes from the renderer. Doc-accuracy nit, doesn't affect behavior.
- Full-width colon `：` not stripped by `sanitizeDisplayName` — display ambiguity, not a structural forge. yujiawei's suggestion to sanitize `resolveDisplayName`'s roster return is reasonable defense-in-depth (roster names come from the members API, lower risk), non-blocking.

Credit to Jerry-Xin for the original catch and to the author for a fix that went to the root (roster) rather than just patching the anchor.

PR #176, opened 2026-07-01
feat(thread-md): inject a thread's own THREAD.md, not the parent GROUP.md (#88 P3-1)

boris-clark:
## Summary

`#88` **P3-1** — a CommunityTopic thread now injects **only its own THREAD.md** and never the parent group's GROUP.md. Previously `resolveGroupInstructions` collapsed every channelId to its parent `groupNo` via `extractParentGroupNo`, so with server-md reads enabled a thread message was injected the parent group's server GROUP.md — violating the redline that a thread and its parent group are mutually exclusive instruction sources.

## Changes

- **`resolveGroupInstructions` split on channelId shape into two mutually-exclusive branches**:
  - **Plain group** — unchanged: server-first GROUP.md, in-memory cache keyed by `groupNo`, never-lose local `<groupNo>.md` fallback (byte-for-byte pre-P3 behavior).
  - **Thread** — resolves its own THREAD.md via a new `getThreadMd` (`GET /v1/bot/groups/{groupNo}/threads/{shortId}/md`), a **separate** in-memory `ThreadMdCache` keyed by the composite `<groupNo>::<shortId>`, and a never-lose local `<shortId>.md` fallback. The thread branch **never reads or writes the group GROUP.md cache and never calls `getGroupMd`.**
- **The thread/group split is unconditional** (it is the bug fix, shipped as the default). A new `threadMd` flag (default off) gates only the new server-read capability; with it off a thread still resolves its local `<shortId>.md`, which is already the correct non-stacking behavior. `threadMd` is independent of `serverMd`.
- **Composite cache key** `groupNo::shortId` is defensive double-insurance: `shortId` is a globally unique 19-digit snowflake today (a bare key would not collide), but the composite key ensures one group's THREAD.md can never be served for a same-`shortId` thread under a different parent group if the id scheme ever changes. A code comment declares the "shortId is globally unique" assumption. `::` is outside the safe-id charset, so it can never appear inside a validated key component.

## Thread server-md endpoint verification

The proposal noted the thread server-md endpoint might not exist upstream. **Verified against the live backend that it does** and is symmetric to the group md endpoint:

```
GET /v1/bot/groups/{groupNo}/threads/{shortId}/md
→ 200 { "content": "...", "version": 49, "updated_at": "...", "updated_by": "..." }
```

Same payload shape as `GET /v1/bot/groups/{groupNo}/md`. The thread branch therefore does a **server-first read of THREAD.md** (behind `threadMd`), reusing the existing `getJson` timeout / SSRF-validated `apiUrl` / Bearer auth / int64-safe parse chain. No follow-up work item is needed for the read path.

## Security

Same trust model as P2-A GROUP.md: the resolved THREAD.md is injected as a trusted system-prompt block, so the `ThreadMdCache` is **in-memory only** (no disk) — the only path content can enter the trusted channel is a live, authenticated `getThreadMd`, never a forgeable on-disk artifact. TTL-bounded staleness backstop shared with the group cache. The thread cache is kept decoupled from group md-change events (P2-B), so a thread never swallows a group event and vice versa.

## Testing

- New `src/__tests__/thread-md.test.ts` (18 tests): `ThreadMdCache` composite-key store/read, cross-group isolation, TTL, invalidate, unsafe-id rejection, no durable backing; thread-branch flag gating, server-first, cache reuse, TTL re-fetch, 404 / empty / network local fallback, malformed channelId, and a **spy assertion that the thread branch never touches the group GROUP.md cache**.
- Core regression added in `group-md.test.ts`: a thread with `serverMd` on + a parent GROUP.md present asserts the parent GROUP.md is **not** injected and the group cache is never read/written.
- `getThreadMd` endpoint/auth/shape tests in `group-md-api.test.ts`.
- Regression baseline (group-md, group-md-api, group-md-tool, group-md-writeback, thread-api, channel-id, group-config) unchanged; **full suite 1210 → 1231 passing, zero group-path regressions.** Lint (0 warnings), type-check, and build all clean.

Part of #88

### DIFF
--- a/src/__tests__/group-md-api.test.ts
+++ b/src/__tests__/group-md-api.test.ts
@@ -8,7 +8,7 @@
  */
 
 import { describe, it, expect, vi, beforeEach, afterAll } from 'vitest';
-import { getGroupMd, updateGroupMd, type GroupMd } from '../octo/api.js';
+import { getGroupMd, getThreadMd, updateGroupMd, type GroupMd, type ThreadMd } from '../octo/api.js';
 
 const fetchMock = vi.fn();
 const originalFetch = globalThis.fetch;
@@ -75,6 +75,39 @@ describe('getGroupMd', () => {
   it('throws on a 500', async () => {
     fetchMock.mockResolvedValueOnce(new Response('boom', { status: 500, statusText: 'Internal Server Error' }));
     await expect(getGroupMd({ ...BASE, groupNo: GROUP })).rejects.toThrow(/failed \(500\)/);
+  });
+});
+
+const SHORT = '2071488441815666688';
+
+describe('getThreadMd', () => {
+  it('GETs /v1/bot/groups/{groupNo}/threads/{shortId}/md with Bearer auth and parses the payload', async () => {
+    const payload: ThreadMd = {
+      content: 'Thread-only rules.',
+      version: 49,
+      updated_at: '2026-07-01T00:00:00Z',
+      updated_by: 'op-uid',
+    };
+    fetchMock.mockResolvedValueOnce(mockJsonResponse(payload));
+
+    const md = await getThreadMd({ ...BASE, groupNo: GROUP, shortId: SHORT });
+
+    const { url, init } = call();
+    expect(url).toBe(`${BASE.apiUrl}/v1/bot/groups/${GROUP}/threads/${SHORT}/md`);
+    expect(init.method).toBe('GET');
+    expect(authHeader(init)).toBe(`Bearer ${BASE.botToken}`);
+    expect(md).toEqual(payload);
+  });
+
+  it('url-encodes both the groupNo and shortId path segments', async () => {
+    fetchMock.mockResolvedValueOnce(mockJsonResponse({ content: '', version: 0, updated_at: null, updated_by: '' }));
+    await getThreadMd({ ...BASE, groupNo: 'a/b', shortId: 'c/d' });
+    expect(call().url).toBe(`${BASE.apiUrl}/v1/bot/groups/a%2Fb/threads/c%2Fd/md`);
+  });
+
+  it('throws on a 404 (no THREAD.md set) so the caller can degrade to local', async () => {
+    fetchMock.mockResolvedValueOnce(new Response('not found', { status: 404, statusText: 'Not Found' }));
+    await expect(getThreadMd({ ...BASE, groupNo: GROUP, shortId: SHORT })).rejects.toThrow(/failed \(404\)/);
   });
 });
 
--- a/src/__tests__/group-md.test.ts
+++ b/src/__tests__/group-md.test.ts
@@ -249,39 +249,47 @@ describe('resolveGroupInstructions', () => {
     expect(out).toBeUndefined();
   });
 
-  it('thread channelId → server fetch uses the PARENT groupNo (P1 routing preserved)', async () => {
+  it('🔴 thread channelId + serverMd on + parent has server GROUP.md → does NOT inject the parent GROUP.md (core P3 fix)', async () => {
+    // The bug (XIN-224): the old resolver collapsed a thread to its parent
+    // groupNo and, with serverMd on, injected the parent's server GROUP.md.
+    // After the fix a thread is mutually exclusive from its parent group.
     const channelId = `${GROUP}____tid123`;
-    fetchMock.mockResolvedValueOnce(mockMd('server rules'));
+    // Any fetch that DID happen would return parent GROUP.md — its presence in
+    // the output is exactly the regression we guard against.
+    fetchMock.mockResolvedValue(mockMd('parent GROUP rules'));
     const cache = new GroupMdCache();
 
     const out = await resolveGroupInstructions({
       groupConfigDir: cfgDir,
-      serverMd: true,
+      serverMd: true, // group flag on…
+      // …but threadMd off → thread stays local; parent GROUP.md must never leak.
       ...BASE,
       channelId,
       cache,
     });
 
-    expect(out).toBe('server rules');
-    // URL is keyed by the parent group number, NOT the composite channelId.
-    expect(call0Url()).toBe(`${BASE.apiUrl}/v1/bot/groups/${GROUP}/md`);
+    expect(out).toBeUndefined(); // no thread-local file, and NO parent GROUP.md
+    // The thread branch must not touch the group GROUP.md endpoint at all.
+    expect(fetchMock).not.toHaveBeenCalled();
+    // …and it must never read or write the parent-group cache.
+    expect(cache.get(GROUP)).toBeUndefined();
   });
 
-  it('thread channelId → local fallback still routes by the thread short-id file', async () => {
+  it('thread channelId → local fallback routes by the thread short-id file, no server call', async () => {
     // Thread's own short-id instruction file (loadGroupConfig new semantics).
     writeFileSync(join(cfgDir, 'tid123.md'), 'thread-local rules');
-    fetchMock.mockResolvedValueOnce(new Response('nope', { status: 404, statusText: 'Not Found' }));
     const cache = new GroupMdCache();
 
     const out = await resolveGroupInstructions({
       groupConfigDir: cfgDir,
-      serverMd: true,
+      serverMd: true, // group flag on, but a thread ignores it
       ...BASE,
       channelId: `${GROUP}____tid123`,
       cache,
     });
 
     expect(out).toBe('thread-local rules');
+    expect(fetchMock).not.toHaveBeenCalled();
   });
 });
 
--- a/src/__tests__/thread-md.test.ts
+++ b/src/__tests__/thread-md.test.ts
@@ -1,0 +1,301 @@
+/**
+ * Tests for the THREAD.md server cache + thread-branch resolver (P3-1).
+ *
+ * Covers:
+ *   - ThreadMdCache: in-memory store/read keyed by the COMPOSITE `groupNo::shortId`,
+ *     TTL expiry, invalidate, path-safe components, cross-group isolation (same
+ *     shortId under different parents never collide), and no durable backing.
+ *   - resolveGroupInstructions thread branch: mutual exclusion from the group
+ *     GROUP.md (the core XIN-224 fix), threadMd flag gating, server-first
+ *     preference, never-lose local `<shortId>.md` fallback, cache reuse, TTL
+ *     re-fetch, and the invariant that the thread branch NEVER reads or writes
+ *     the group GROUP.md cache / endpoint.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach, afterAll } from 'vitest';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { GroupMdCache, ThreadMdCache } from '../group-md-cache.js';
+import { resolveGroupInstructions } from '../group-md.js';
+
+const fetchMock = vi.fn();
+const originalFetch = globalThis.fetch;
+
+let cfgDir: string;
+
+beforeEach(() => {
+  fetchMock.mockReset();
+  globalThis.fetch = fetchMock as unknown as typeof fetch;
+  cfgDir = mkdtempSync(join(tmpdir(), 'thread-md-cfg-'));
+});
+
+afterEach(() => {
+  rmSync(cfgDir, { recursive: true, force: true });
+});
+
+afterAll(() => {
+  globalThis.fetch = originalFetch;
+});
+
+function mockMd(content: string, version = 1): Response {
+  return new Response(
+    JSON.stringify({ content, version, updated_at: '2026-07-01T00:00:00Z', updated_by: 'op' }),
+    { status: 200, headers: { 'Content-Type': 'application/json' } },
+  );
+}
+
+const BASE = { apiUrl: 'https://test.example.com', botToken: 'bf_test' };
+const GROUP = 'grp001';
+const SHORT = '2071488441815666688'; // a real-shaped snowflake shortId
+const CHANNEL = `${GROUP}____${SHORT}`;
+
+/** A controllable clock for deterministic TTL tests. */
+function fakeClock(start = 1_000_000) {
+  let t = start;
+  return { now: () => t, advance: (ms: number) => { t += ms; } };
+}
+
+function call0Url(): string {
+  return (fetchMock.mock.calls[0] as [string, RequestInit])[0];
+}
+
+// ─── ThreadMdCache ─────────────────────────────────────────────────────────
+
+describe('ThreadMdCache', () => {
+  it('stores and reads an entry keyed by the composite groupNo::shortId', () => {
+    const cache = new ThreadMdCache();
+    cache.set(GROUP, SHORT, { content: 'hi', version: 3, updated_at: null });
+    expect(cache.get(GROUP, SHORT)).toEqual({ content: 'hi', version: 3, updated_at: null, updated_by: undefined });
+  });
+
+  it('isolates the same shortId under different parent groups (composite key double-insurance)', () => {
+    const cache = new ThreadMdCache();
+    cache.set('groupA', SHORT, { content: 'A rules', version: 1, updated_at: null });
+    cache.set('groupB', SHORT, { content: 'B rules', version: 1, updated_at: null });
+    expect(cache.get('groupA', SHORT)?.content).toBe('A rules');
+    expect(cache.get('groupB', SHORT)?.content).toBe('B rules');
+  });
+
+  it('expires an entry once it is older than the TTL (staleness backstop)', () => {
+    const clock = fakeClock();
+    const cache = new ThreadMdCache(1000, clock.now);
+    cache.set(GROUP, SHORT, { content: 'x', version: 1, updated_at: null });
+
+    clock.advance(999);
+    expect(cache.get(GROUP, SHORT)?.content).toBe('x'); // still fresh
+
+    clock.advance(1); // now exactly at TTL → expired
+    expect(cache.get(GROUP, SHORT)).toBeUndefined();
+  });
+
+  it('a non-positive TTL disables expiry', () => {
+    const clock = fakeClock();
+    const cache = new ThreadMdCache(0, clock.now);
+    cache.set(GROUP, SHORT, { content: 'x', version: 1, updated_at: null });
+    clock.advance(10 * 365 * 24 * 3600 * 1000);
+    expect(cache.get(GROUP, SHORT)?.content).toBe('x');
+  });
+
+  it('invalidate clears the entry', () => {
+    const cache = new ThreadMdCache();
+    cache.set(GROUP, SHORT, { content: 'x', version: 1, updated_at: null });
+    cache.invalidate(GROUP, SHORT);
+    expect(cache.get(GROUP, SHORT)).toBeUndefined();
+  });
+
+  it('rejects an unsafe groupNo or shortId (never stored)', () => {
+    const cache = new ThreadMdCache();
+    cache.set('../escape', SHORT, { content: 'evil', version: 1, updated_at: null });
+    cache.set(GROUP, '../escape', { content: 'evil', version: 1, updated_at: null });
+    expect(cache.get('../escape', SHORT)).toBeUndefined();
+    expect(cache.get(GROUP, '../escape')).toBeUndefined();
+  });
+
+  it('persists nothing across instances — a fresh cache shares no state (no durable backing)', () => {
+    const a = new ThreadMdCache();
+    a.set(GROUP, SHORT, { content: 'durable?', version: 9, updated_at: '2026-07-01T00:00:00Z', updated_by: 'op' });
+    expect(a.get(GROUP, SHORT)?.content).toBe('durable?');
+
+    const b = new ThreadMdCache();
+    expect(b.get(GROUP, SHORT)).toBeUndefined();
+  });
+});
+
+// ─── resolveGroupInstructions — thread branch ────────────────────────────────
+
+describe('resolveGroupInstructions (thread branch)', () => {
+  it('threadMd OFF → reads the local <shortId>.md only, never hits the server', async () => {
+    writeFileSync(join(cfgDir, `${SHORT}.md`), 'thread-local rules');
+    const threadCache = new ThreadMdCache();
+
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: false,
+      ...BASE,
+      channelId: CHANNEL,
+      threadCache,
+    });
+
+    expect(out).toBe('thread-local rules');
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it('threadMd ON but no thread cache wired → local <shortId>.md only, no server call', async () => {
+    writeFileSync(join(cfgDir, `${SHORT}.md`), 'thread-local rules');
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: CHANNEL,
+    });
+    expect(out).toBe('thread-local rules');
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it('threadMd ON → server THREAD.md wins over the local file (server-first) and uses the thread md endpoint', async () => {
+    writeFileSync(join(cfgDir, `${SHORT}.md`), 'thread-local rules');
+    fetchMock.mockResolvedValueOnce(mockMd('server THREAD rules'));
+    const threadCache = new ThreadMdCache();
+
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: CHANNEL,
+      threadCache,
+    });
+
+    expect(out).toBe('server THREAD rules');
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+    expect(call0Url()).toBe(`${BASE.apiUrl}/v1/bot/groups/${GROUP}/threads/${SHORT}/md`);
+  });
+
+  it('caches the server fetch — a second call serves from cache (no second fetch)', async () => {
+    fetchMock.mockResolvedValueOnce(mockMd('server THREAD rules'));
+    const threadCache = new ThreadMdCache();
+    const args = { groupConfigDir: cfgDir, threadMd: true, ...BASE, channelId: CHANNEL, threadCache };
+
+    expect(await resolveGroupInstructions(args)).toBe('server THREAD rules');
+    expect(await resolveGroupInstructions(args)).toBe('server THREAD rules');
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+  });
+
+  it('re-fetches after the cache entry expires (TTL backstop picks up edits)', async () => {
+    const clock = fakeClock();
+    const threadCache = new ThreadMdCache(1000, clock.now);
+    const args = { groupConfigDir: cfgDir, threadMd: true, ...BASE, channelId: CHANNEL, threadCache };
+
+    fetchMock.mockResolvedValueOnce(mockMd('old rules'));
+    expect(await resolveGroupInstructions(args)).toBe('old rules');
+
+    clock.advance(1000); // expire the cached entry
+    fetchMock.mockResolvedValueOnce(mockMd('new rules'));
+    expect(await resolveGroupInstructions(args)).toBe('new rules');
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+  });
+
+  it('server 404 → falls back to the local <shortId>.md (fallback never lost)', async () => {
+    writeFileSync(join(cfgDir, `${SHORT}.md`), 'thread-local rules');
+    fetchMock.mockResolvedValueOnce(new Response('nope', { status: 404, statusText: 'Not Found' }));
+    const threadCache = new ThreadMdCache();
+
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: CHANNEL,
+      threadCache,
+    });
+
+    expect(out).toBe('thread-local rules');
+  });
+
+  it('server reachable but empty content → local fallback', async () => {
+    writeFileSync(join(cfgDir, `${SHORT}.md`), 'thread-local rules');
+    fetchMock.mockResolvedValueOnce(mockMd('   '));
+    const threadCache = new ThreadMdCache();
+
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: CHANNEL,
+      threadCache,
+    });
+
+    expect(out).toBe('thread-local rules');
+  });
+
+  it('server network error → local fallback, never throws', async () => {
+    writeFileSync(join(cfgDir, `${SHORT}.md`), 'thread-local rules');
+    fetchMock.mockRejectedValueOnce(new Error('ECONNREFUSED'));
+    const threadCache = new ThreadMdCache();
+
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: CHANNEL,
+      threadCache,
+    });
+
+    expect(out).toBe('thread-local rules');
+  });
+
+  it('no server THREAD.md and no local file → undefined', async () => {
+    fetchMock.mockResolvedValueOnce(new Response('nope', { status: 404, statusText: 'Not Found' }));
+    const threadCache = new ThreadMdCache();
+
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: CHANNEL,
+      threadCache,
+    });
+
+    expect(out).toBeUndefined();
+  });
+
+  it('🔴 thread branch NEVER reads or writes the group GROUP.md cache (mutual exclusion)', async () => {
+    // Spy on the group cache. The thread branch must not touch it — a thread's
+    // instructions are its own THREAD.md, never the parent group's GROUP.md.
+    const groupCache = new GroupMdCache();
+    const getSpy = vi.spyOn(groupCache, 'get');
+    const setSpy = vi.spyOn(groupCache, 'set');
+    const threadCache = new ThreadMdCache();
+    fetchMock.mockResolvedValueOnce(mockMd('server THREAD rules'));
+
+    await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      serverMd: true, // group flag on…
+      threadMd: true, // …thread flag on
+      ...BASE,
+      channelId: CHANNEL,
+      cache: groupCache,
+      threadCache,
+    });
+
+    expect(getSpy).not.toHaveBeenCalled();
+    expect(setSpy).not.toHaveBeenCalled();
+    // The only fetch was the thread md endpoint, not the group md endpoint.
+    expect(call0Url()).toBe(`${BASE.apiUrl}/v1/bot/groups/${GROUP}/threads/${SHORT}/md`);
+  });
+
+  it('malformed thread channelId (no shortId) → local fallback, never fetches a group-scoped path', async () => {
+    fetchMock.mockResolvedValue(mockMd('should not be used'));
+    const threadCache = new ThreadMdCache();
+
+    const out = await resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: `${GROUP}____`, // trailing separator, empty shortId
+      threadCache,
+    });
+
+    expect(out).toBeUndefined();
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+});
--- a/src/config.ts
+++ b/src/config.ts
@@ -127,6 +127,17 @@ export interface Config {
    * server-first READ path), so an operator can enable reads without writes.
    */
   mdWriteback?: boolean;
+  /**
+   * P3-1 feature flag: when true, a THREAD (CommunityTopic) channel resolves its
+   * THREAD.md server-first (GET /v1/bot/groups/{groupNo}/threads/{shortId}/md,
+   * cached in-memory keyed by `groupNo::shortId`) before its local `<shortId>.md`
+   * file. Off (default) → a thread resolves ONLY its local file, which is already
+   * the correct behavior (a thread injects its own THREAD.md and never stacks the
+   * parent group's GROUP.md). So this flag gates the NEW server-read capability,
+   * NOT the thread/group split — that split is unconditional. Independent of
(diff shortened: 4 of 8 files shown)


OctoBoooot (approved):
## Review: feat(thread-md): inject a thread's own THREAD.md, not the parent GROUP.md (#88 P3-1) (#176)

**Verdict:** Approve — the thread/group mutual-exclusion redline is enforced both structurally and by test; clean, well-covered, CI green. PR is a draft; `needs-human-review` + `size/XL` applied.

Reviewed at elevated rigor (trusted-instruction injection + the explicit redline that a thread must never stack its parent's GROUP.md). Verified the invariant rather than trusting the description:

- **Mutual exclusion is structural, not just intended.** `resolveGroupInstructions` routes on `isThreadChannelId` at the single entry point; the thread branch (`resolveThreadInstructions`) never references the group `cache` or `getGroupMd` — it can't touch the parent GROUP.md by construction. This is the correct fix for the XIN-224 bug (old code collapsed every channelId to `extractParentGroupNo`).
- **And it's pinned by a spy test** — `thread-md.test.ts:261` asserts `groupCache.get`/`set` are never called on a thread resolve. Structural + behavioral proof of the same invariant; a future refactor that reintroduces the stacking breaks the test.
- **Same memory-only trust model as P2-A/#172.** `ThreadMdCache` is in-memory only (no `fs`), TTL-bounded; content only enters the trusted channel via authenticated `getThreadMd` (reuses the SSRF-validated `getJson` — timeout / Bearer / int64-safe parse). No forgeable on-disk artifact.
- **Composite key `groupNo::shortId`** with both components `isSafeGroupNo`-validated, and `::` outside the safe-id charset so it can't be forged into a component to cross a key boundary. Cross-group isolation is tested (same shortId under two parents doesn't collide).
- **`isGroup` includes `CommunityTopic`** (index.ts:398), so threads actually reach the resolver — the routing isn't dead.
- **Never-lose local fallback preserved** — thread branch passes the full `channelId` to `loadGroupConfig`, which routes a thread to its own `<shortId>.md` (the #169 behavior), never the parent `<groupNo>.md`; 404/empty/network/malformed-channelId all degrade to local, each tested.

### Praise
- The spy-test assertion (`getSpy`/`setSpy` not called) is exactly how to lock a "never touches X" invariant so it can't silently regress — better than relying on the branch structure alone.
- `threadMd` gated independently of `serverMd`, with the flag off still yielding the correct non-stacking local behavior — so the bug fix ships as the default and the flag only gates the new server-read capability. That separation is the right call and clearly documented.
- The composite-key rationale is honest: it states shortId is a globally-unique snowflake today (bare key wouldn't collide) and frames the composite as defensive double-insurance against a future per-group id scheme — not overclaiming a present bug.
- Server endpoint verified against the live backend (payload shape symmetric to GROUP.md) rather than assumed — the P2-A pattern of grounding the contract before building on it, carried through.

- Coverage-sniff clean on the new files; `thread-md.test.ts` (18) + group-md regression + `getThreadMd` api tests cover the surface directly.
- Event-driven THREAD.md invalidation (the `invalidate` hook exists, symmetric to GroupMdCache) is presumably a later phase like P2-B was for groups — not wired here, which is fine for P3-1's read+cache scope.
- Stacked on the merged GROUP.md work; standard merge-order care.

mochashanyao (approved):
[Octo-Q · automated review]

---

# cc-channel-octo PR#176 Review Report
**PR:** https://github.com/Mininglamp-OSS/cc-channel-octo/pull/176
**Head SHA:** `01cd61faa255cdb210e848ef7af6ce387bd18e06`
**Scope:** feat(thread-md): inject a thread's own THREAD.md, not the parent GROUP.md (#88 P3-1)

---

## 1. 验证结论

| Item | Status | Evidence |
|------|--------|----------|
| Thread/group mutual exclusion | ✅ | `resolveGroupInstructions` routes on `isThreadChannelId()` (src/group-md.ts L86-90); thread branch never destructures `cache` or calls `getGroupMd` |
| `getThreadMd` API symmetry | ✅ | Reuses `getJson` helper (src/octo/api.ts L761-778) — same Bearer auth, timeout, int64-safe parse, `encodeURIComponent` as `getGroupMd` |
| ThreadMdCache composite key | ✅ | `groupNo::shortId` after `isSafeGroupNo` on both halves; `::` outside `[a-zA-Z0-9._-]` charset → unambiguous separator (src/group-md-cache.ts L147-153) |
| Flag independence (`threadMd` vs `serverMd`) | ✅ | Thread branch reads `threadMd` only (not `serverMd`); group branch reads `serverMd` only. Verified: thread with `serverMd=true` + `threadMd=off` → local only, no server call |
| Cache wiring in index.ts | ✅ | `threadMdCache` created per-bot (L230), passed to both `handleMessage` call sites (inbound L311, cron L332), forwarded to resolver (L867) |
| Local fallback correctness | ✅ | `loadGroupConfig(groupConfigDir, channelId)` routes thread to `<shortId>.md` first, then legacy `<groupNo>____<shortId>.md` fallback — never `<groupNo>.md` (src/group-config.ts L67-78) |
| Test coverage | ✅ | 18 new tests in thread-md.test.ts + updated regression in group-md.test.ts + getThreadMd API tests in group-md-api.test.ts; 1210→1231 passing per PR description |
| Security model preserved | ✅ | In-memory only cache, no disk persistence, same SSRF-validated apiUrl + authenticated bot token chain |

## 2. 发现的问题

### P2-1: Thread cache lacks event-driven invalidation (new capability, acknowledged gap)

**Diff-scope:** new — `ThreadMdCache` is introduced by this PR. The `invalidate(groupNo, shortId)` method exists and is tested, but is not wired to any event source.

**Detail:** `SessionRouter.handleGroupMdEvent()` (session-router.ts L375-391) invalidates only `groupMdCache` on GROUP.md change events. There is no equivalent path for `ThreadMdCache`. When the server emits a thread-md-updated event, the thread cache stays stale until TTL expiry (default 5min).

**Severity rationale (P2, not P1):** The TTL backstop ensures eventual consistency within the documented staleness window. Stale THREAD.md does not produce user-visible error data or make a working path unavailable — it shows a slightly outdated version for up to 5 minutes, identical to the GROUP.md pre-P2-B behavior. The PR description explicitly notes "The thread cache is kept decoupled from group md-change events (P2-B)" — suggesting the author is aware and considers this future work.

**Suggestion:** Add a follow-up to wire `ThreadMdCache.invalidate()` to the server event path (either a new `thread_md_updated` event type, or a combined handler that checks the event's `short_id` field). The `invalidate()` method and its tests are already in place.

### P2-2: No thread-md write-back path (out of scope, noted)

**Diff-scope:** pre-existing — `group-md-writeback.ts` only handles GROUP.md writes. No `updateThreadMd` API or writeback coordinator exists.

**Severity rationale (P2/note):** Explicitly out of PR scope (title says "inject" = read path only). Noted for tracking.

## 3. 建议

- **Follow-up (P2-1):** Wire `ThreadMdCache` invalidation to the server event path. The infrastructure is ready (`invalidate()` method + tests), just needs a `handleThreadMdEvent()` in `SessionRouter` and possibly a new `isThreadMdUpdateEvent()` classifier.
- **Nit:** Consider renaming `GroupMdEntry` to `MdEntry` (or similar) since it's now shared by both group and thread caches. Currently `ThreadMdCache.get()` returns `GroupMdEntry` which is semantically confusing.

## 4. 额外发现

- The `::` composite key separator is a sound design choice: since `isSafeGroupNo` rejects any character outside `[a-zA-Z0-9._-]`, and `:` is outside that set, neither component can contain `::`, making the separator unambiguous even if future id schemes change.
- The thread branch's explicit non-destructuring of `cache` (group cache) is a good structural invariant — it makes the mutual exclusion enforceable by code inspection, not just by intent.
- The backward-compat fallback in `loadGroupConfig` (try `<shortId>.md` first, then legacy `<groupNo>____<shortId>.md`) is correct and tested.

## 5. 数据流回溯

|---|---|---|
| `channelId` in resolver | `msg.channel_id ?? ''` (index.ts) → `resolveGroupInstructions({ channelId })` | ✅ |
| `isThreadChannelId` routing | `channelId.includes("____")` (channel-id.ts L50) | ✅ deterministic |
| `groupNo` for thread API call | `extractParentGroupNo(channelId)` → left of first `____` | ✅ |
| `shortId` for thread API call | `extractThreadShortId(channelId)` → right of first `____`, null if empty | ✅ (null → local fallback) |
| `getThreadMd` response | Server `GET /v1/bot/groups/{groupNo}/threads/{shortId}/md` via `getJson` → Bearer auth + SSRF-validated apiUrl + timeout + int64-safe parse | ✅ |
| `threadCache.get(groupNo, shortId)` | Composite key `${groupNo}::${shortId}` after both validated by `isSafeGroupNo` | ✅ |
| Local fallback content | `loadGroupConfig(groupConfigDir, channelId)` → thread path reads `<shortId>.md` then `<groupNo>____<shortId>.md`, never `<groupNo>.md` | ✅ |
| `threadMdCache` in handleMessage | Created per-bot in `startBot()`, passed through both inbound and cron-fire paths | ✅ |

## 6. 盲点 checklist (R5 — security_sensitive PR)

- **C1 双路径 parity (含守卫/lint 工具自身等价写法覆盖):** getThreadMd ↔ getGroupMd symmetric (same `getJson`, same `encodeURIComponent`, same auth). `resolveThreadInstructions` structurally mirrors `resolveGroupBranch`. `ThreadMdCache` mirrors `GroupMdCache` with composite key. ✅ Parity holds.
- **C2 control-flow ordering / 嵌套复用 + 安全控件非 canonical 试穿:** `boundInstructions` called exactly once per resolution in each branch — no double-application risk. `isSafeGroupNo` regex `/^[a-zA-Z0-9._-]+$/` rejects all non-canonical forms (path traversal, unicode, empty). ✅
- **C3 授权边界 ≠ 能力边界:** `getThreadMd` uses same bot token + SSRF-validated `apiUrl` as `getGroupMd`. Server endpoint should enforce bot authorization for the specific group/thread (PR states "verified against the live backend"). No new capability boundary. ✅
- **C4 授权生命周期 / 容器-成员状态级联:** N/A — read-only instruction injection path, no auth state changes.
- **C5 build/note 通过 ≠ 运行期路径正确:** N/A — no build artifacts, relative paths, CLI changes, or extension packaging.
- **C6 治理/策略/安全文档自洽性:** N/A — code PR, not governance doc.

## 7. 跨轮 blocker 复检 (R6)

---

**[Octo-Q] verdict: APPROVE**

No P0 or P1 findings. The thread/group mutual exclusion is correctly implemented by construction (the thread branch structurally cannot touch the group cache). The security model (in-memory only, authenticated fetch, SSRF-validated URL) is preserved. Two P2 items noted (thread cache event invalidation not wired, thread write-back out of scope) — both are acknowledged scope items with infrastructure ready for follow-up.

Jerry-Xin (approved):
APPROVE — thread/group mutual exclusion is byte-verified correct. This is the right fix for the P3-1 redline (a thread injects its own THREAD.md and never stacks the parent group's GROUP.md), the trust-boundary carryover from #172/#173/#174 holds, and coverage is thorough.

## Scoping correctness (the load-bearing axis) ✅

- `resolveGroupInstructions` routes unconditionally on `isThreadChannelId(channelId)`: a thread goes to `resolveThreadInstructions`, a plain group to `resolveGroupBranch`. The split is the bug fix and is NOT flag-gated (only the new server-read is gated by `threadMd`).
- The thread branch never calls `getGroupMd` and never touches the group `GroupMdCache` — confirmed by both the code path and the spy assertion in `thread-md.test.ts` (`getSpy`/`setSpy` not called). No parent-GROUP.md leak into a thread.
- Cross-thread / cross-group isolation: `ThreadMdCache` is keyed by the COMPOSITE `groupNo::shortId`, not a bare `groupNo` and not a bare `shortId`. Two threads in the same group get distinct keys (different shortId); the same shortId under two parents cannot collide (different groupNo). No groupNo-only collision path exists — the group cache and thread cache are separate Map instances with different key shapes.
- `::` is outside the `isSafeGroupNo` charset (`/^[a-zA-Z0-9._-]+$/`), so a validated component can never contain the separator and cannot be forged to cross a key boundary.

## Trust-boundary carryover (#172/#173/#174) ✅

- `ThreadMdCache` is in-memory only (a `Map`), no disk persistence — verified by the "no durable backing" test. Same rationale as `GroupMdCache`: the only path into the trusted system-prompt channel is a live authenticated fetch, never a forgeable on-disk artifact.
- Server-first ordering preserved: fresh cache hit → server `getThreadMd` → local fallback.
- TTL shared with the group policy (`DEFAULT_GROUP_MD_TTL_MS`), expired read is a miss → re-fetch.
- `getThreadMd` reuses the existing `getJson` chain: SSRF-validated `apiUrl`, Bearer auth, int64-safe parse, and the error string carries only the status code (no token leak).

## id-validation on the new thread id ✅

The new `shortId` folded into the URL path and the Map key is validated the same way `groupNo` is: `ThreadMdCache.key()` runs BOTH `groupNo` and `shortId` through `isSafeGroupNo` and returns null (store/read reject) on either being unsafe — covered by the "rejects an unsafe groupNo or shortId" test (`../escape` rejected). In `getThreadMd` both path segments are `encodeURIComponent`-escaped (test asserts `a/b` → `a%2Fb`, `c/d` → `c%2Fd`). No path-traversal / unsafe-Map-key surface introduced.

## Fallback behavior ✅

A thread with no THREAD.md never falls back to the parent GROUP.md — it degrades to its own local `<shortId>.md` (via `loadGroupConfig`, which is already thread-aware and routes by shortId), then to undefined ("no custom instructions"). 404 / empty / network error all degrade to local and never throw. Malformed `<groupNo>____` (empty shortId) degrades to local with no server call. All four paths are tested.

## Sanitization ✅

Injected THREAD.md goes through the same `boundInstructions` as GROUP.md: 16 KiB byte cap with mid-codepoint-safe truncation, whitespace trim, empty → undefined. Identical bounding to the group path.

## Non-blocking notes

🟡 `resolveThreadInstructions` guards `!shortId` but not `!groupNo`. A malformed `____shortId` (empty parent) has the cache correctly reject the empty-groupNo key, but the code still proceeds to `getThreadMd` with an empty `groupNo`, producing a request to `.../groups//threads/<shortId>/md`. Harmless (it 404s and degrades to local), but consider guarding `!groupNo` symmetrically with `!shortId` to avoid the wasted bad call.

🟡 `mdWriteback` remains exposed in thread sessions and writes the parent `GROUP.md`. After this PR a thread no longer reads that file, so an owner-triggered writeback from inside a thread updates instructions that do not apply to the current conversation — a UX/semantics mismatch (owner-gated, not a security blocker). Worth clarifying or disabling for thread channels until THREAD.md writeback exists.

🔵 Two new comments contain non-English wording. For consistency across the codebase, keep source comments entirely in English.

## Highlights

- The resolver split is correctly unconditional and the thread branch provably never touches the group cache or the group endpoint (spy-asserted).
- `ThreadMdCache` is memory-only, TTL-bounded, composite-keyed — matches the stated trust + isolation model exactly.
- New API client reuses the authenticated `getJson` path and URL-encodes both path segments.
- Strong test coverage: composite-key store/read, cross-group isolation, TTL, invalidate, unsafe-id rejection, no durable backing, flag gating, server-first, cache reuse, TTL re-fetch, 404/empty/network fallback, malformed channelId, and the group-cache-untouched spy assertion.

CI is green (type-check, ESLint zero-warnings, Vitest, Coverage, secret-scan, dependency-review, check-sprint all pass).

ploy-elison (approved):
APPROVED — the redline lands cleanly, the group path is untouched, and the test coverage backs every claim.

## What I verified (diff, not the description)

**1. Group GROUP.md path — byte-for-byte unchanged.** The old `resolveGroupInstructions` body is now `resolveGroupBranch`, reached only for non-thread channelIds. Diffing the extracted branch against `origin/main:src/group-md.ts` shows an identical body (server-first + in-memory cache keyed by groupNo + never-lose local `<groupNo>.md`). P2-A/B/C are not touched: `session-router.ts` group-md event invalidation and the P2-C per-groupNo writeback lock are unchanged.

**2. Mutual exclusion is structural, not just documented.** The thread branch never imports or receives the group `cache`, never calls `getGroupMd`, and keys a separate `ThreadMdCache`. This is enforced by construction and covered by the `🔴 thread branch NEVER reads or writes the group GROUP.md cache` test — it spies on `GroupMdCache.get/set` with **both** `serverMd` and `threadMd` on and asserts neither is called, and that the only fetch is the thread endpoint. The core regression test (`serverMd` on + parent GROUP.md present → parent GROUP.md not injected, group cache untouched) is present in `group-md.test.ts`, replacing the old test that asserted the now-removed buggy behavior.

**3. Composite key is safe.** `ThreadMdCache.key` validates both halves with `isSafeGroupNo` (`^[a-zA-Z0-9._-]+$`), which excludes `:`, so `::` can never appear inside a component and the separator is unambiguous. Cross-group isolation (same shortId under different parents) is tested.

**4. Flag semantics match the claim.** `threadMd` (default off) gates only the new server read; the thread/group split itself is unconditional. Verified: `threadMd` off + `serverMd` on → thread stays on its local `<shortId>.md`, no fetch, no parent GROUP.md, group cache untouched. `loadGroupConfig` was already thread-aware, so the local path was never stacking — the bug only ever existed on the server path, and that is what this fixes.

**5. Thread server read reuses the group security chain.** `getThreadMd` goes through the same `getJson` (Bearer auth, timeout, int64-safe parse, same error format) against the same SSRF-validated `apiUrl`; both path segments are `encodeURIComponent`-ed (tested). `ThreadMdCache` is in-memory only — the "no durable backing across instances" test confirms no disk trust surface is added.

Full suite green locally: **1231 passing** (1210 → 1231, +21), lint 0 warnings, type-check and build clean. `isGroup` includes `CommunityTopic`, so the thread branch is actually reached in production.

## Non-blocking follow-ups (out of this PR's read-path scope)

- **Write-side is still keyed to the parent group.** With `mdWriteback` on, a thread turn (a CommunityTopic is `isGroup`) is handed the `group_md` write tool, and `group-md-tool.ts` resolves its target via `extractParentGroupNo(channelId)` — i.e. a thread could still *write* the parent group's GROUP.md even though it can no longer *inject* it. This is pre-existing, gated behind `mdWriteback` (default off), and not a regression from this PR, but it is the same redline on the write axis. Worth a P3-2 to make the write side thread-aware too.
- `group-md-tool.ts` header comment still reads "a thread shares its parent group's GROUP.md", which now contradicts the P3 redline — update when the write side is addressed.
- Thread cache has no event-driven invalidation (TTL backstop only). Fine for P3-1 and documented; a thread-md-update event hook is the natural P3-2 companion to the write side.

None of these regress existing behavior or block the injection fix, so approving.

yujiawei (approved):
# Code Review — PR #176 (cc-channel-octo)

**Verdict: APPROVE.** This is a well-scoped, well-tested bug fix. A CommunityTopic thread now resolves only its own THREAD.md and can no longer be injected the parent group's GROUP.md. No blocking (P0/P1) issues found. A few non-blocking suggestions below.

Reviewed at head SHA `01cd61faa255cdb210e848ef7af6ce387bd18e06` (the single commit on this branch; the other commits shown against `main` are already-merged siblings #174/#175). Verified locally: `type-check`, `lint` (0 warnings), and the full suite (**1231 passing**) are all green.

## 1. Spec compliance

The change implements exactly what the linked issue's P3-1 item requires, and nothing beyond it:

- **Correct core fix.** `resolveGroupInstructions` (`src/group-md.ts:98`) now routes on `isThreadChannelId` into two mutually-exclusive branches. The thread branch (`resolveThreadInstructions`, `src/group-md.ts:175`) never calls `getGroupMd` and never touches the group `GroupMdCache`, so the previous behavior — collapsing every channelId to its parent `groupNo` and injecting the parent's server GROUP.md — is gone. The regression is locked down by a spy assertion (`src/__tests__/thread-md.test.ts:252`) and by the rewritten `group-md.test.ts:252`.
- **Group branch is behavior-preserving.** `resolveGroupBranch` (`src/group-md.ts:113`) is the pre-P3 logic verbatim (server-first + in-memory cache keyed by `groupNo` + never-lose local `<groupNo>.md`). Only comments changed.
- **No over-build.** The new `threadMd` flag defaults off and gates *only* the new server-read capability; with it off, a thread still resolves its local `<shortId>.md`, which is already the correct non-stacking behavior. The thread/group split ships unconditionally as the fix. Flag is independent of `serverMd`, wired through `mergeConfig`/`resolveBotConfigs` symmetrically (`src/config.ts`).
- **Local fallback routing verified.** `loadGroupConfig` already routes a thread channelId to `<shortId>.md` (with a legacy whole-composite fallback), so a thread never inherits `<groupNo>.md` on the local path either (`src/group-config.ts`).

## 2. Code quality

**Approved.** Notable strengths:

- **Composite cache key `<groupNo>::<shortId>`** is sound. Both halves are validated by `isSafeGroupNo` (charset `[a-zA-Z0-9._-]`), so `::` can never appear inside a validated component — the separator is unambiguous and cannot be forged to cross a key boundary. Cross-group isolation is tested (`thread-md.test.ts:72`).
- **Trust model consistent with P2-A.** `ThreadMdCache` is in-memory only (no disk), matching the durable-cache-poisoning rationale for the group cache: server content reaches the trusted system-prompt channel only via a live, authenticated `getThreadMd` over the bot token against the SSRF-validated `apiUrl`, never a forgeable on-disk artifact.
- **`getThreadMd`** (`src/octo/api.ts:731`) reuses the shared `getJson` (timeout, Bearer auth, int64-safe parse) and URL-encodes both path segments; endpoint/auth/404 behavior is tested.
- **Fallback discipline.** Every failure mode (404 / network / timeout / empty / malformed channelId) degrades to the never-lose local `<shortId>.md`. A malformed `<groupNo>____` (empty shortId) short-circuits to local before any fetch (`group-md.ts:191`).

### Non-blocking suggestions (P2 / nits)

- **P2 — no negative caching on 404/empty (server-read path only).** When `threadMd` is on, a thread with no server THREAD.md re-fetches (and logs `console.error`) on every message until content appears, since only non-empty responses are cached (`group-md.ts:200-217`). This is not new — the group branch behaves identically — and it is gated behind a default-off flag, so it does not block. But threads are far more numerous than groups, so before enabling `threadMd` broadly, consider caching a negative/empty sentinel (TTL-bounded) to avoid a per-message fetch + log storm.
- **P2 — TTL-only staleness for thread edits.** `ThreadMdCache.invalidate()` exists but is not wired to any server md-change event; the group cache gets event-driven invalidation (P2-B in `session-router.ts:375`), threads rely solely on the ~5-min TTL backstop. Acceptable (same state the group cache was in at its P2-A stage, and the endpoint symmetry suggests a thread md-change event may not exist upstream yet), but worth a follow-up if operators expect edits to take effect promptly.
- **Nit — empty `groupNo` not short-circuited.** The thread branch guards `!shortId` but not an empty `groupNo`; a channelId like `____<shortId>` would issue a malformed `/v1/bot/groups//threads/.../md` that 404s and degrades to local. Harmless (safe degrade), and Octo does not emit such ids, but a symmetric `if (!groupNo) return local()` would be tidier.

## 3. Things a human may want to double-check (security-sensitive)

- **Backend endpoint contract.** The PR states `GET /v1/bot/groups/{groupNo}/threads/{shortId}/md` was verified against the live backend and returns `{content, version, updated_at, updated_by}`. Worth a human confirming this endpoint's auth scope matches the group md endpoint (same bot-token / permission model) so a thread read can't cross a tenant boundary.
- **Read/write asymmetry (out of scope, informational).** The owner-gated write-back tool (sibling #174, `group-md-tool.ts`) still resolves `extractParentGroupNo(channelId)`, so an owner invoking `update_group_md` *from inside a thread* writes the **parent group's** GROUP.md — while reads from that thread are now isolated to THREAD.md. Owner-gated so not exploitable, but the read/write target divergence for threads may surprise operators; consider a follow-up to either target THREAD.md on write or reject write-back from a thread channel.

## Summary

The mutual-exclusion invariant is correct and well-guarded by tests, the cache keying and flag gating are consistent with the intended behavior, and the change is behavior-preserving for plain groups. Suggestions above are non-blocking.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #176 (cc-channel-octo)

## Summary

This PR fixes XIN-224: when `serverMd` was enabled, a CommunityTopic thread's channelId was collapsed to its parent groupNo via `extractParentGroupNo`, causing the resolver to fetch and inject the parent group's GROUP.md instead of the thread's own THREAD.md. The fix splits `resolveGroupInstructions` into two mutually exclusive branches routed by `isThreadChannelId` — threads resolve their own THREAD.md (server-first behind a new `threadMd` flag, with a never-lose local `<shortId>.md` fallback) and plain groups resolve their GROUP.md exactly as before. A new `ThreadMdCache` (in-memory, TTL-bounded, composite-keyed by `groupNo::shortId`) and `getThreadMd` API client mirror the existing group infrastructure.

The thread/group split is unconditional (the bug fix); the `threadMd` flag gates only the new server-read capability. 18 new tests plus a regression guard in `group-md.test.ts` cover the mutual-exclusion invariant, cache mechanics, and edge cases.

## Verification

Static analysis only at head `01cd61faa255cdb210e848ef7af6ce387bd18e06`; build and tests not executed in this environment.

- ✅ **Mutual exclusion structurally enforced** — `resolveGroupInstructions` at `src/group-md.ts:99` routes on `isThreadChannelId`; the thread branch at `:170-230` never reads `params.cache` (group cache) and never calls `getGroupMd`. Verified by code inspection and the spy-based test at `src/__tests__/thread-md.test.ts:260-282`.
- ✅ **Security model consistent** — `ThreadMdCache` is in-memory only (`src/group-md-cache.ts:131`), `getThreadMd` reuses the existing `getJson` chain with SSRF-validated `apiUrl` + Bearer auth (`src/octo/api.ts:762-775`). No new disk surface.
- ✅ **Composite cache key safe** — `::` separator is outside the `isSafeGroupNo` charset (`/^[a-zA-Z0-9._-]+$/`), so neither `groupNo` nor `shortId` can contain `::`. Key collision across groups is structurally impossible.
- ✅ **Local fallback correct** — thread branch passes the full `channelId` to `loadGroupConfig` (`src/group-md.ts:183`), which routes to `<shortId>.md` first then `<groupNo>____<shortId>.md` as a legacy fallback (`src/group-config.ts:72-80`). Never falls back to `<groupNo>.md`.
- ✅ **Byte cap shared** — `boundInstructions` at `src/group-md.ts:60-68` applies `MAX_GROUP_CONFIG_BYTES` (16 KiB) to both server GROUP.md and THREAD.md content.
- ✅ **Config wiring complete** — `threadMd` flag added to `Config` (`src/config.ts:130-140`), `PartialConfig` (`:315`), `mergeConfig` (`:411`), and `resolveBotConfigs` (`:648`). Per-bot override works.
- ✅ **Input validation** — `extractThreadShortId` returns `null` for empty shortId (`<groupNo>____`), and the thread branch falls back to local on `!shortId` (`src/group-md.ts:194`). `isSafeGroupNo` validates both key components in `ThreadMdCache.key()`.
- ✅ **URL encoding** — both `groupNo` and `shortId` are `encodeURIComponent`'d in `getThreadMd` (`src/octo/api.ts:770-771`), matching `getGroupMd`.
- ✅ **Test coverage** — spy assertion confirms thread branch never touches group cache (`thread-md.test.ts:260-282`); regression guard confirms thread + serverMd on + parent GROUP.md present does not inject parent content (`group-md.test.ts:252-275`).
- ✅ **Cron path wired** — `threadMdCache` is passed to `handleMessage` on both the inbound (`src/index.ts:311`) and cron fire (`:332`) paths.

## Findings

No P0/P1 issues. One P2 freshness observation below.

### P2 — Thread cache has no event-driven invalidation; `ThreadMdCache.invalidate()` is unused (`src/group-md-cache.ts:183`)

`GroupMdCache` is wired into `SessionRouter.handleGroupMdEvent` (`src/session-router.ts:380-390`) so a `group_md_updated` socket event immediately drops the stale cached entry and forces a re-fetch on the next turn. `ThreadMdCache` has an `invalidate(groupNo, shortId)` method but it is never called from any production code path — `threadMdCache` is not passed to `SessionRouter`, and no `thread_md_updated` event handler exists. The thread cache relies solely on TTL expiry (5 minutes by default) for freshness.

This is acknowledged in the PR body ("The thread cache is kept decoupled from group md-change events (P2-B)") and the `invalidate()` API is prepared for future wiring. However, an operator who edits a thread's THREAD.md on the server will see stale instructions for up to 5 minutes, while GROUP.md edits are picked up near-instantly via the event path. Not a correctness issue — the TTL is the documented staleness backstop — but worth tracking as a follow-up if thread-md event types are confirmed server-side.

## Human-verify

1. The PR states the thread server-md endpoint (`GET /v1/bot/groups/{groupNo}/threads/{shortId}/md`) was "verified against the live backend." This is not independently verifiable from the diff alone. Confirm the endpoint exists and returns the documented `{content, version, updated_at, updated_by}` shape in the deployment target. This is **not** a merge blocker — the code correctly degrades to local fallback on 404 or any error.

2. Whether the server emits a `thread_md_updated` event (and if so, what the exact `event.type` literal is) is not confirmed. The `invalidate()` hook is ready but unwired. This is **not** a merge blocker — flagging for future P2-B follow-up.

- `isThreadChannelId` uses `____` separator detection; a plain groupNo cannot contain `____` (the `isSafeGroupNo` charset excludes underscores), so the routing check is unambiguous.
- `extractThreadShortId` correctly returns `null` for `<groupNo>____` (empty shortId), and the thread branch degrades to local rather than making a malformed server call.
- `GroupMdEntry` type is reused for both caches — the shape is identical (`{content, version, updated_at, updated_by?}`), and the test assertions match.
- `loadGroupConfig` thread routing (`src/group-config.ts:72-80`) correctly prefers `<shortId>.md` then falls back to `<groupNo>____<shortId>.md` for legacy deployments.
- `handleMessage` receives `threadMdCache` as an optional parameter (`src/index.ts:394`) and passes it through to `resolveGroupInstructions` (`:866`). No new required-parameter breakage for existing callers.
- `getThreadMd` throws on non-2xx (matching `getGroupMd`), and the thread branch catches and falls back to local — the never-lose invariant is preserved.

## Data Flow Tracing

| Consumed data | Source | Verified |
|---|---|---|
| `groupNo` (thread branch) | `extractParentGroupNo(channelId)` — splits on first `____` | ✅ flows to `getThreadMd` + `threadCache.get/set` |
| `shortId` (thread branch) | `extractThreadShortId(channelId)` — splits on first `____` | ✅ flows to `getThreadMd` + `threadCache.get/set` |
| `md.content` (server THREAD.md) | `getThreadMd` → `getJson` → authenticated GET | ✅ flows through `boundInstructions` → `groupInstructions` in system prompt |
| `threadMd` flag | `config.threadMd` → `resolveBotConfigs` per-bot merge | ✅ gates server-read in `resolveThreadInstructions` |
| `threadCache` | `new ThreadMdCache()` in `startBot` → passed through `handleMessage` | ✅ consumed by `resolveThreadInstructions` |
| `local()` fallback | `loadGroupConfig(groupConfigDir, channelId)` — full composite | ✅ routes to `<shortId>.md` via `group-config.ts:72-80` |

## Blind Spot Checklist (C1–C6)

- **C1 — Dual-path parity**: Thread branch mirrors group branch structurally (cache get/set/invalidate, server-first + local fallback, feature flag gating, byte-bounded content). `invalidate()` exists but is unwired — flagged as P2 above, not a parity defect.
- **C2 — Control-flow ordering / nested reuse**: `resolveGroupInstructions` routes exclusively via `isThreadChannelId` — no double-call or ordering risk. The two branches share no mutable state.
- **C3 — Auth boundary ≠ capability boundary**: `getThreadMd` uses the same bot-token + SSRF-validated `apiUrl` auth as `getGroupMd`. No new auth surface.
- **C4 — Auth lifecycle / container-member state cascade**: N/A. No auth changes.
- **C5 — Build/note ≠ runtime correctness**: N/A. No build/packaging changes.
- **C6 — Governance/docs self-consistency**: N/A. No governance changes.

## Cross-round Blocker Re-review

## Verdict: APPROVED

Well-structured fix for XIN-224 with clean mutual exclusion between thread and group branches, consistent security model (in-memory-only cache, authenticated fetches), and thorough test coverage including a spy-based assertion that the thread branch never touches the group cache. The one P2 (no event-driven thread cache invalidation) is an acknowledged scope deferral, not a correctness issue — the TTL backstop bounds staleness to 5 minutes.

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. The thread/group mutual exclusion is structurally sound, the security model is consistent with the existing group cache, and the test suite guards the core invariant. The P2 (unwired `ThreadMdCache.invalidate()`) is non-blocking.

yujiawei (approved):
# Code Review — PR #176 (cc-channel-octo)

Independent review at head SHA `01cd61faa255cdb210e848ef7af6ce387bd18e06`. Verified against a fresh checkout: `type-check`, `lint` (0 warnings), and the full suite (**1231 tests passing**) all clean locally. This PR is security-relevant (trusted system-prompt injection), so items below flagged 🔎 are worth a human's manual eye.

## Verdict

**APPROVE.** The read/injection path — the actual scope of this change — is correct, well-tested, and free of P0/P1 defects. The thread/group split is enforced structurally, the cache is memory-only as claimed, and the SSRF/auth chain is reused correctly. Findings below are all P2 (advisory/follow-up); two of them are pre-enablement action items that should be closed before `threadMd` / `mdWriteback` are turned on in a thread-bearing deployment.

---

## 1. Spec compliance — ✅

All four P3-1 requirements are implemented, and nothing beyond scope was added:

- **Thread resolves its own THREAD.md** — `resolveGroupInstructions` routes on `isThreadChannelId` (`src/group-md.ts:103`) into `resolveThreadInstructions` (`src/group-md.ts:178-230`), which calls `getThreadMd` against `GET /v1/bot/groups/{groupNo}/threads/{shortId}/md` and falls back to the local `<shortId>.md`.
- **Never the parent GROUP.md** — the thread branch destructures only `threadCache` (`src/group-md.ts:181`), never calls `getGroupMd`, and never reads/writes the group `cache`. The regression test at `src/__tests__/group-md.test.ts:252` asserts a thread + `serverMd` on + a parent GROUP.md present yields `undefined`, with `fetchMock` never called and the group cache untouched.
- **Split is unconditional** — the `isThreadChannelId` routing has no flag guard; with `threadMd` off, `src/group-md.ts:189` still returns the local `<shortId>.md` (via `loadGroupConfig`, `src/group-config.ts:72-81`), never the `<groupNo>.md`.
- **Flag gates only the new server read** — `if (!threadMd || !threadCache) return local()` (`src/group-md.ts:189`). `threadMd?: boolean` is default-off (`src/config.ts:140`), merged through the same path as `serverMd`/`mdWriteback` (`src/config.ts:411,648`).

No over-build: exactly one new GET endpoint symmetric to `getGroupMd`, a `ThreadMd` interface with the identical 4 fields, one composite-keyed cache, one boolean flag. No thread PUT/writeback endpoint was added (correctly — P3-1 is read-only).

## 2. Code quality — Approved

Structurally sound: the mutual exclusion is guaranteed by construction (a separate `ThreadMdCache` instance, `src/index.ts:231`, that the thread branch never crosses into the group cache), not merely by convention. The memory-only security model (`src/group-md-cache.ts:110-131`) correctly mirrors `GroupMdCache`: the only path content reaches the trusted prompt block is a live authenticated `getThreadMd`, never a forgeable on-disk artifact. `encodeURIComponent` on both path segments + Bearer auth + `AbortSignal.timeout` are inherited from `getJson`, and `channelId` originates from the server-issued `msg.channel_id`, not chat body — so an untrusted user cannot redirect the fetch. 404 / empty / network all degrade to the never-lose local fallback and never throw.

### P2 findings (non-blocking)

**P2-a 🔎 — Write-back path leaves the mutual-exclusion redline half-enforced (pre-existing, owner-gated).**
`src/index.ts:896` injects the `update_group_md` tool whenever `mdWriteback && isGroup`, and `isGroup` is true for a CommunityTopic thread (`src/index.ts:398`). The tool resolves its target with `extractParentGroupNo(coords.channelId)` (`src/group-md-tool.ts:86`), which for a thread collapses to the **parent** groupNo, then PUTs `/v1/bot/groups/{groupNo}/md`. So the read side is now mutually exclusive but the write side is not: an agent driven inside a thread can overwrite the parent group's GROUP.md. This is **not introduced by this PR** (the write-back path and `isGroup` are untouched by the diff), is **owner-gated** (`src/group-md-tool.ts:83` — only `fromUid === ownerUid`), and sits behind the separate default-off `mdWriteback` flag — so it is not a blocker for this read-path change. But it maps directly to the stated redline, so it should be closed in a follow-up: either skip the write-back tool when `isThreadChannelId(channelId)`, or add a thread-scoped write endpoint. **Action item before enabling `mdWriteback` on a thread-bearing deployment.**

**P2-b 🔎 — `ThreadMdCache` growth is unbounded across the process lifetime.**
Entries are only evicted lazily inside `.get()` for the *same* key (`src/group-md-cache.ts:168`). Threads are ephemeral and numerous; an abandoned thread's entry is never re-`get()`'d and so lingers until process exit. `GroupMdCache` shares this model but groups are a small bounded set, so it's benign there — for threads it is a slow leak. Bounded per entry (≤16 KiB) and only populated when `threadMd` is on (default off), so it's not a blocker now, but a long-running instance at scale wants an LRU cap or an active TTL sweep. **Action item before enabling `threadMd` in high-traffic deployments.**

**P2-c — Expected 404s are logged at `console.error`.**
`src/group-md.ts:225` logs on every catch, including the common "thread has no THREAD.md" 404. With `threadMd` on and most threads lacking custom instructions, this is log noise that can drown real errors. Consider demoting the 404 case to debug/silent while keeping network/timeout errors at error level. (The group branch has the same pattern at `src/group-md.ts:156`.)

**P2-d — `getThreadMd` fetch path doesn't gate its components through `isSafeGroupNo`.**
The cache validates both `groupNo` and `shortId` with `isSafeGroupNo` before keying (`src/group-md-cache.ts:154`), but the fetch passes them straight into the URL with only `encodeURIComponent`. Not exploitable (`channelId` is server-issued; `encodeURIComponent` neutralizes `/ ? #`; a stray segment 404s → local), and it's the same latent asymmetry `getGroupMd` already has, so no regression. For belt-and-suspenders symmetry, short-circuiting to `local()` when either component is unsafe would make the "validated component" invariant hold on both the cache and the wire.

**P2-e — `ThreadMdCache.invalidate()` and event-driven refresh are unwired.**
`ThreadMdCache.invalidate()` (`src/group-md-cache.ts:183`) has no production caller, and the router's `handleGroupMdEvent` (`src/session-router.ts:375`) only invalidates the group cache — the `threadCache` is never handed to `SessionRouter`. A server-side THREAD.md edit is therefore only picked up after TTL (~5 min), not promptly like GROUP.md. Out of scope for P3-1 (the event schema doesn't even parse a thread short-id yet), but noting the asymmetry for the follow-up that wires thread event refresh.

## 3. Coverage / blind spots

- Both an independent long-context pass and a line-level pass converged on P2-a (the write-back gap) — high confidence it is real; they diverged only on severity, adjudicated to P2 here because it is pre-existing, owner-gated, and outside this PR's read-only scope.
- One advisor flagged a possible `TypeError` on `md.content` inside the `if (bounded)` block — **refuted**: `bounded` derives from `md?.content ?? ''`, so the block is unreachable when `md` is nullish, and `getJson` returns `{}` (never `undefined`) on an empty body. Not a defect.
- Not verifiable from the diff (asserted by the PR author + tests, worth one human confirmation 🔎): the **live backend actually exposes** `GET /v1/bot/groups/{groupNo}/threads/{shortId}/md` with the stated payload shape; and `ChannelType.CommunityTopic` is the value the gateway actually stamps on `msg.channel_type` for threads (upstream of this diff).

PR #179, opened 2026-07-02
fix(group): backfill member display names so mentions render names not uids

caster-Q:
## 目标

@人显示人名、@bot 显示 bot 名,永远不出现裸 uid 串。本 PR 是 cc-channel-octo 侧的一半(另一半是 OpenClaw 插件的 isValidOutboundUid 冷缓存问题,独立)。

## 诊断(已核准:不是 API 字段问题)

成员 API 的 `/members` **确实返回真名(字段就是 `name`)**。裸 uid 的成因在 cc 本地这一侧:

1. 本地 roster 没被可靠地填上这些名字(现象:多个频道 name==uid),而经 `/user/info` 按 uid 回填真名的 helper `fetchAndLearnUser` **在消息流里从未被调用**——真名只能靠偶发的 wire `from_name`(常缺失)。helper 实为死代码。
2. 即便被调到,`if (cached) return cached` 会把 `name==uid` 占位符(wire 无名 / DB 遗留行)当成已解析名字短路,**恰好对最需要回填的成员不 fetch**。

> 注:这是 cc 本地 roster 的填充问题,**不是** `/members` API 不返回 name。

## 改动

- 入站群消息:roster/wire 都没真名时,构建发送者前缀前做一次 `/user/info` 回填 → 前缀与 outbound `@mention` 反查用的 nameToUid 都带上真人名。
- 缓存命中仅认真实名字(`cached && cached !== uid`);`/user/info` 回显 uid 也忽略。
- `fetchUserInfo` 语义分流:确定性 miss(404 / 有响应无名)返 null 走满 TTL 负缓存;瞬时失败(网络/超时/非404)throw,只做 60s 短 backoff,避免一次抖动冻结裸 uid 一小时。
- 负缓存按 (channel, uid):成功即删、超上限清理过期项,长运行有界。

## 校验

codex review:round1 fail(3 findings)→ 改 → round2 verdict=pass 全 resolved,无 pre-push blocker。全量 1217 测试通过。纯 prompt 无关,无字段映射改动。


OctoBoooot (commented):
## Review: fix(group): backfill member display names so mentions render names not uids (#179)

**Verdict:** Approve with comments — clean fix for both root causes, consistent with the #175 `!== uid` treatment; one latency tradeoff worth an explicit decision. Held at COMMENT only because governance CI is red with no out-of-band sign-off.

CI: sole red check is `check-sprint` — governance CI (rule 11), maintainer-cleared not code. No sign-off, so §8.1 → COMMENT. Code CI green (type-check/vitest 1216). `mergeStateStatus` stays BLOCKED until `check-sprint` clears.

Verified both root causes are actually fixed:
- **Dead helper wired in** — `fetchAndLearnUser` now has a real call site (index.ts:664, group inbound, only when roster+wire both miss). Previously it was test-only, as the body claims — confirmed no prior caller.
- **Short-circuit fixed** — `if (cached && cached !== uid) return cached` no longer treats a `name===uid` placeholder as a hit, so the members that render as bare uid (the ones that needed backfill) now reach the `/user/info` fetch. This matches the `!== uid` guard `resolveDisplayName`/`pushMessage` got in #175 — consistent identity semantics across the module.
- **Negative cache is correct** — stamps `lastUserInfoAttempt` BEFORE the request (so failures are paced, not just successes), keyed `channel uid`, TTL aligned with the 1h roster refresh; the response-echoes-uid case is guarded too. Both new tests pin the placeholder-not-a-hit and no-refetch-within-TTL behaviors.

### Question
- `src/index.ts:664` — the backfill is an `await`-blocking `/user/info` round-trip (up to the 5s timeout) on the inbound path, on first sight of an unnamed sender. The author explicitly asks reviewers to weigh this vs fire-and-forget. My read: awaiting is defensible — it makes the *current* message's prefix + the nameToUid map correct immediately (fire-and-forget would render the first message with a bare uid and only fix the next one), the negative cache bounds it to one attempt per hour per unnamed uid, and 5s is a ceiling not the norm. But if inbound latency is more sensitive than first-prefix correctness for your deployment, fire-and-forget is the alternative. Flagging as the decision the author requested, not a defect — either choice is reasonable.

### Praise
- Fixing the `cached !== uid` short-circuit is the load-bearing insight: the placeholder-as-hit bug is exactly why the dead helper wouldn't have helped even once wired in, and the fix reuses the same `!== uid` placeholder semantics established in #175 rather than inventing a new rule.
- Stamping the negative-cache timestamp *before* the request (not after success) is the correct detail — it paces failing lookups too, so a persistently-nameless member can't hammer `/user/info` every message. Easy to get backwards.

- `lastUserInfoAttempt` grows unbounded with member count (never pruned) — author disclosed this, and it mirrors the existing `lastRefresh` map, so it's consistent prior art. Acceptable; a shared TTL-sweep for both maps would be a reasonable future cleanup, not this PR's job.
- Coverage-sniff flagged `index.ts` — the new group-context logic is tested (2 cases); the index.ts change is a straight-line awaited call in the already-covered handleMessage path.

Jerry-Xin (changes requested):
Relevance gate passed: this PR is in scope for `cc-channel-octo`, but it should not merge while a TypeScript source file is committed as binary.

**🔴 Blocking**
🔴 Critical: [src/group-context.ts](/Users/mlamp/.openclaw/tmp/pr-review-ph8y0pbq/repo/src/group-context.ts:38) and [src/group-context.ts](/Users/mlamp/.openclaw/tmp/pr-review-ph8y0pbq/repo/src/group-context.ts:373) contain literal NUL bytes. Git now reports `src/group-context.ts` as binary (`Bin 26608 -> 28339 bytes`), hiding the actual code diff from normal review and many text-based tools. The runtime separator choice is fine, but the source should use an escaped form, for example `\u0000`, or a named separator constant written without an embedded control byte.

**💬 Non-blocking**
🟡 Warning: I could not run the project checks locally because `node_modules` is absent: `vitest`, `eslint`, and `tsc` were not found. I did confirm the repo script is `npm run type-check`, not `typecheck`.

🔵 Suggestion: Consider an integration test around `handleMessage` proving the first inbound group turn with missing `from_name` awaits `fetchAndLearnUser` and uses the resolved name in the current-message prefix. The `GroupContext` unit coverage is good, but the new wiring in [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-ph8y0pbq/repo/src/index.ts:670) is the behavioral point of the PR.

**✅ Highlights**
The cache guard change correctly avoids treating `name === uid` as resolved, and the negative cache prevents repeated `/user/info` calls for the same unresolved sender. The backfill is also placed before current-message prompt assembly, so successful lookups affect the immediate sender label.

OctoBoooot (changes requested):
## Review (follow-up): #179 — literal NUL bytes in group-context.ts (credit Jerry-Xin)

**Updated verdict:** Request changes — Jerry-Xin's 🔴 is real and I missed it. My first pass noted the "Binary files differ" diff line but failed to investigate it; that was the error. Upgrading.

### Major
- `src/group-context.ts` contains **two literal NUL (`\x00`) bytes** in committed source — verified by decoding the blob at `ed905e55a1bc`:
  - **line 373 (executable code):** `` const key = `${channelId}\x00${uid}`; `` — the negative-cache map key uses a NUL as the `channelId`/`uid` separator.
  - **line 38 (comment):** `` keyed by `channelId\x00uid` `` — same corruption in the doc.

  Why it's a real defect even though CI is green: a NUL inside a template literal is a valid string character, so it compiles and the tests pass — the key is still unique/consistent, so behavior "works." But:
  1. The source file now reads as **binary** to git/GitHub (hence "Binary files … differ" in the diff) — it can't be diffed, blamed, or inline-reviewed, and greps/editors/tooling that stop at NUL will choke on it. That's a maintainability regression on a core file.
  2. It's almost certainly an accidental encoding/paste artifact, not intent — the comment shows the author meant a readable delimiter, and every other composite key in this codebase uses a visible separator (e.g. `${groupNo}::${shortId}` in ThreadMdCache, `${channelId} ${uid}` in the existing `lastRefresh` path).

  Fix: replace both `\x00` with the intended visible separator (a space matches the existing `lastUserInfoAttempt`/`lastRefresh` convention, or `::` to match ThreadMdCache) and re-save the file as clean UTF-8 so it stops registering as binary. A test that the key is well-formed isn't really possible (any separator "works"), so the guard here is: reviewers should be able to see the diff — confirm `git diff` renders it as text after the fix.

My COMMENT (4614346102) verified the backfill logic and the `!== uid` fix correctly, but I saw "Binary files a/src/group-context.ts and b/src/group-context.ts differ" and moved on instead of decoding the blob to find out *why*. The why is embedded NUL bytes — a real finding. Credit Jerry-Xin. The logic assessment (both root causes fixed, negative-cache-stamped-before-request, latency tradeoff) still stands; this is orthogonal and must be fixed before merge.

### On Jerry-Xin's other notes
- 🟡 missing node_modules locally (can't run vitest/eslint/tsc) — that's a reviewer-environment issue, not a defect in the PR; CI ran these green. Not on the author.
- 🟡 add a handleMessage integration test for the from_name-missing → fetchAndLearnUser path — fair coverage suggestion, non-blocking.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #179 (cc-channel-octo)

## Summary

This PR fixes two root causes that caused group-chat mentions and sender prefixes to render raw uids instead of human display names. First, `fetchAndLearnUser` — the `/user/info` backfill helper — was dead code with zero production callers; the PR wires it into `handleMessage` in `src/index.ts` so a one-shot name lookup runs when neither the roster nor the wire payload carries a real display name. Second, the existing cache-hit guard `if (cached) return cached` treated `name === uid` placeholder entries (seeded from DB or wire-echo) as resolved, short-circuiting backfill for exactly the members that needed it most. The fix adds a `cached !== uid` check, a negative-cache TTL (1h) to prevent burst `/user/info` storms from unnamed members, and a `uid`-echo guard on the API response itself. Two new tests cover the placeholder-bypass and negative-cache paths. Overall, a well-reasoned fix with clear design rationale in the PR body.

## Verification

- ✅ **Diff scope** — 3 files, +41/-1 lines. Production changes confined to `fetchAndLearnUser` internals (`src/group-context.ts`) and a single call-site wiring (`src/index.ts:668-678`). Two new test cases (`src/__tests__/group-context.test.ts`).
- ✅ **Short-circuit fix correctness** — `if (cached && cached !== uid) return cached` at `src/group-context.ts:365` correctly distinguishes genuine display names from uid-echo placeholders. Mirrors the same `!== fromUid` pattern already present in `resolveDisplayName` at `:168`.
- ✅ **Negative-cache stamp-before-request** — `lastUserInfoAttempt.set(key, now)` at `:376` runs before the `await fetchUserInfo` at `:379`, so a failed request is also paced (not just successful misses). Prevents retry storms on transient API failures.
- ✅ **API error handling preserved** — `fetchUserInfo` in `src/octo/api.ts:346` already has `AbortSignal.timeout(5000)` and try/catch returning null on any failure. The wrapping try/catch in `fetchAndLearnUser` at `:381-383` is defense-in-depth.
- ✅ **nameToUid map populated** — `learnMember` (called on successful backfill at `:382`) updates both `memberMap` and `nameToUidByChannel`, so outbound `@mention` resolution via `resolveMentions` in `src/stream-relay.ts:223` benefits from the newly learned name.
- ✅ **No self-review** — PR author is not the review bot.

## Findings

No P0/P1 issues. One P2 and one nit below.

### P2 — Null byte embedded in source file (`src/group-context.ts:373`)

The negative-cache key template literal at `src/group-context.ts:373` contains a literal null byte (`\x00`) between `${channelId}` and `${uid}`, hex-confirmed at file offset 17307. A second null byte appears in the adjacent comment at offset 1660 (`keyed by \`channelId uid\``). This causes git to treat the entire file as binary — `git diff` without `-a` shows `Binary files differ` rather than a text diff, GitHub's web UI cannot render inline diffs for this file, and `git grep` skips it by default. Functionally the key works (null is a valid JS string character and a collision-free separator), but this is almost certainly an editor artifact that will surprise every future reviewer and tool that touches this file.

```ts
// Fix: replace the null byte with a regular space or explicit separator
const key = `${channelId} ${uid}`;  // was: `${channelId}\x00${uid}`
```

### Nit — Negative-cache TTL expiry path untested (`src/__tests__/group-context.test.ts:244`)

The new negative-cache test covers the within-TTL suppression (immediate retry returns undefined without a second API call) but not the after-TTL case where `USER_INFO_RETRY_MS` (1h) has elapsed and a fresh `/user/info` call should fire. The expiry logic at `src/group-context.ts:374-376` is a simple timestamp comparison, so risk is low. A `vi.useFakeTimers()` + `vi.advanceTimersByTime(USER_INFO_RETRY_MS + 1)` assertion would close the gap.

## Data Flow Trace

- `msg.from_uid` → `resolveDisplayName()` → returns `undefined` when roster has uid-echo or no entry, and wire `from_name` is missing/uid-echo → triggers `fetchAndLearnUser()`.
- `fetchAndLearnUser()` → checks `memberMap` (skips uid-echo) → checks `lastUserInfoAttempt` negative-cache → calls `/user/info` API (5s timeout) → on success: `learnMember()` updates `memberMap` + `nameToUid` → returns real name.
- On miss: returns `undefined` → `formatSenderLabel(msg.from_uid, undefined)` → bare uid prefix. Best-effort, not a regression from prior behavior.
- `nameToUid` map updated on success → `resolveMentions` in `src/stream-relay.ts` can resolve `@Name` in outbound replies using the newly learned mapping.

## Blind-Spot Checklist

- **C1 (dual-path parity)** — N/A. No paired add/remove or subscribe/unsubscribe paths touched.
- **C2 (control-flow ordering)** — Clear. `fetchAndLearnUser` has a single call-site in production (`src/index.ts:671`). No nested/reused call chains.
- **C3 (authorization boundary)** — Clear. Uses existing `config.botToken` for `/user/info`, no new endpoint exposure.
- **C4 (authorization lifecycle)** — N/A. No auth/status changes.
- **C5 (build/runtime parity)** — N/A. No build/packaging changes.
- **C6 (governance/docs)** — N/A. No governance/doc changes.

## Cross-Round Blocker Recheck (R6)

- `!uid` early-return guard at `src/group-context.ts:359` prevents empty-uid lookups.
- `info.name !== uid` guard at `:382` prevents the `/user/info` endpoint echoing the uid as the name from being learned as a genuine display name.
- `const resolvedName` → `let resolvedName` change at `src/index.ts:662` is correctly scoped; the second assignment at `:671` is guarded by `!resolvedName` (only fires when resolveDisplayName returned undefined).
- Negative-cache map (`lastUserInfoAttempt`) grows per `(channelId, uid)` pair — same pattern as `lastRefresh`, bounded by member count, acceptable.
- Blocking `await` on the hot path is a deliberate design choice (PR body explicitly calls it out); the 5s `AbortSignal.timeout` bounds worst-case latency.

## Verdict: APPROVED

Solid fix with well-identified root causes and proportionate scope. The null byte artifact in `src/group-context.ts` should be cleaned up before merge (it breaks git/tooling diff rendering), but it has no runtime correctness impact. No P0/P1 blockers.

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. The null byte in the negative-cache key is a P2 code-hygiene fix (breaks git binary detection, no runtime impact). The short-circuit fix, negative-cache design, and production wiring are all correct.

yujiawei (approved):
# Code Review — PR #179 (cc-channel-octo)

**Head SHA:** `ed905e55a1bc9dcf09f14fec248445876feb7d31` · base `main` (merge-base `688f404`)

## Summary
Wires up the previously dead `fetchAndLearnUser` backfill so group sender prefixes and outbound `@mention` resolution render human display names instead of bare uids, fixes a short-circuit bug where a `name === uid` placeholder was treated as a resolved cache hit, and adds a per-`(channel, uid)` negative cache to pace the miss path. Focused change (+41/-1, 3 files).

**Verification (at head SHA):** `type-check` clean · `lint --max-warnings 0` clean · full suite **1216 passed** (incl. 2 new group-context tests). Build MERGEABLE.

## 1. Spec compliance
**Spec: ✅**
- Missing: none — the three stated changes (wiring, guard fix, negative cache) are all present and correctly scoped.
- Over-build: none — no unrequested flags/fields; the two new tests cover exactly the placeholder-hit and negative-cache-miss cases described.
- Deviation: none — the `name === uid` placeholder handling is applied consistently in `fetchAndLearnUser` (cache read at `group-context.ts:368`, endpoint echo at `:382`), matching the existing `resolveDisplayName` convention (`:178`).

The synchronous stamp-before-await ordering (`group-context.ts:375-377`, set at `:377` before the `await` at `:380`) correctly collapses concurrent bursts from the same uid into a single request — a subsequent handler running its synchronous check-and-set after the first can only observe the already-written stamp. No cache stampede, and the fetch is double-guarded (`fetchAndLearnUser` try/catch at `:379-388`, plus `fetchUserInfo` catching all failures to `null` at `octo/api.ts:370-373`), so a network/timeout error cannot abort `handleMessage`.

## 2. Code quality
**Quality: Approved**

No P0/P1. The items below are P2 suggestions — none blocks merge.

**[P2] Negative cache conflates transient failures with a real miss** (`group-context.ts:373-388`)
`fetchUserInfo` collapses 404, non-OK (5xx), and network/timeout into `null` (`octo/api.ts:358-373`). Because the stamp is written before the request, a transient 500 or a 5s timeout suppresses any retry for the full 1h TTL, so a member can keep rendering as a bare uid for up to an hour after a blip. This is graceful degradation (self-heals on the next window or a roster refresh), not a correctness bug — but consider a shorter retry TTL for the transport-failure case vs. a confirmed "no name / 404".

**[P2] Blocking await on the inbound path** (`index.ts:670-677`)
The first unresolved sender per `(channel, uid)` adds a synchronous round trip (5s cap) before prompt assembly. You flagged this tradeoff explicitly in the PR description; noting it as accepted. The negative cache bounds repeats but not the first miss. A fire-and-forget variant (render this turn with uid, apply the learned name next turn) would remove the latency at the cost of one-turn staleness — reasonable to defer.

**[P2] Unsanitized name at the learn boundary** (`group-context.ts:382-383`)
`learnMember` stores `info.name` verbatim into `memberMap`/`nameToUid`/SQLite. This is safe in practice — render always passes through `formatSenderLabel` (which sanitizes), the DB write is parameterized, and `nameToUid` is used only as a lookup key gated by `isMember` — and it is **consistent with the existing `refreshMembers` path** (`:298`), which also stores the API name raw. Not a regression; worth a length-cap/sanitize only if you want to harden both paths together.

**[P2] `lastUserInfoAttempt` grows monotonically** (`group-context.ts:373-377`)
No prune on success, roster removal, or TTL sweep. As you noted in the PR description, it is bounded to the same order as `memberMap` (per `(channel, uid)`), so this is acceptable for the current usage; a departed-member prune hook would keep it tidy on very long-lived processes.

## 3. Overall verdict
**APPROVE** — Spec ✅ and Quality Approved. All findings are non-blocking P2 suggestions, two of which you already surfaced as accepted tradeoffs.

## 4. Advisory coverage
Two independent adversarial legs plus my own diff re-read. One cross-family leg had its view of `src/group-context.ts` truncated to a binary blob (tooling/encoding artifact in the diff payload, not a code defect) so its two flagged concurrency/error-handling risks were adjudicated directly against the source and found already handled (stamp-before-await; double try/catch). The line-level leg's four points are the P2s above.

yujiawei (changes requested):
# Code Review — PR #179 (cc-channel-octo)

**Head SHA:** `ed905e55a1bc9dcf09f14fec248445876feb7d31` · base `main` (merge-base `688f404`)

> Correcting my earlier approval on this PR: a re-check confirmed a blocking defect I initially missed — two literal NUL bytes committed in `src/group-context.ts` (independently flagged by other reviewers). Updated verdict below.

## Summary
Wires up the previously dead `fetchAndLearnUser` backfill so group sender prefixes and outbound `@mention` resolution render human display names instead of bare uids, fixes a short-circuit bug where a `name === uid` placeholder was treated as a resolved cache hit, and adds a per-`(channel, uid)` negative cache to pace the miss path. The *logic* is sound; the *blocker* is a file-encoding artifact, orthogonal to the logic.

**Verification (at head SHA):** `type-check` clean · `lint --max-warnings 0` clean · full suite **1216 passed** (incl. 2 new group-context tests). The NUL bytes are valid string characters, so they compile and tests pass — CI green does not catch this.

## 1. Spec compliance
**Spec: ✅**
- Missing: none — wiring, `!== uid` guard fix, and negative cache are all present and correctly scoped.
- Over-build: none — no unrequested flags/fields; the two new tests cover exactly the placeholder-hit and negative-cache-miss cases described.
- Deviation: none in behavior — placeholder handling is consistent with the existing `resolveDisplayName` convention (`group-context.ts:178`).

## 2. Code quality
**Quality: Changes-Requested**

### P1 (blocking) — Two literal NUL (`\x00`) bytes committed in `src/group-context.ts`
- **`src/group-context.ts:373`** (executable): `` const key = `${channelId}\x00${uid}`; `` — the negative-cache map key uses a NUL as the separator.
- **`src/group-context.ts:38`** (comment): `` keyed by `channelId\x00uid` `` — same byte in the doc comment.

The string is unique and consistent, so runtime behavior is correct and tests pass. But committing NUL into a TypeScript source file makes git/GitHub classify `src/group-context.ts` as **binary** (`Bin 26608 -> 28339 bytes` in the diff), which:
1. Hides the actual code diff from normal review — it can't be diffed, blamed, or inline-commented, and grep/editors/text tools that stop at NUL choke on it. That is a maintainability regression on a **core** file.
2. Is almost certainly an accidental paste/encoding artifact, not intent: the comment clearly meant a readable delimiter, and the sibling `lastRefresh` map in this same file already uses a visible separator — `` `${channelId} ${uid}` ``.

**Fix:** replace both `\x00` with the intended visible separator (a single space to match the existing `lastRefresh`/`lastUserInfoAttempt` convention, or `::` to match `ThreadMdCache`) and re-save as clean UTF-8. Confirm the fix by verifying `git diff` renders `src/group-context.ts` as text (not "Binary files differ") at the new head.

### P2 (non-blocking) — noted for the same revision
- **Negative cache conflates transient failures with a real miss** (`group-context.ts:373-388`): `fetchUserInfo` collapses 404, 5xx, and network/timeout into `null` (`octo/api.ts:358-373`). Since the stamp is written before the request, a transient 500 or a 5s timeout suppresses retry for the full 1h TTL, so a member can render as a bare uid for up to an hour after a blip. Self-heals on the next window/roster refresh — consider a shorter retry TTL for transport failures vs. a confirmed 404/"no name".
- **Blocking await on the inbound path** (`index.ts:670-677`): first unresolved sender per `(channel, uid)` adds a round trip (5s cap) before prompt assembly. You surfaced this tradeoff in the description; awaiting is defensible (current turn's prefix + nameToUid become correct immediately, negative cache bounds repeats). Fire-and-forget is the alternative if inbound latency matters more than first-prefix correctness. Accepted as-is.
- **Unsanitized name at the learn boundary** (`group-context.ts:382-383`): `learnMember` stores `info.name` verbatim. Safe in practice (render always goes through `formatSenderLabel`, DB write parameterized, `nameToUid` is a lookup key gated by `isMember`) and consistent with `refreshMembers:298` which also stores the API name raw — not a regression; harden both paths together only if desired.
- **`lastUserInfoAttempt` grows monotonically** (`group-context.ts:373-377`): no prune on success/roster removal/TTL sweep. Bounded to the order of `memberMap` and disclosed in the description; mirrors the existing `lastRefresh` map — acceptable, a shared TTL sweep would be a future cleanup.

### Positive
- Fixing the `cached !== uid` short-circuit is the load-bearing insight — the placeholder-as-hit bug is why the helper wouldn't have helped even once wired in, and the fix reuses the `!== uid` semantics already established in the module.
- Stamping the negative-cache timestamp **before** the request (not after success) is the correct detail — it paces failing lookups too, and the synchronous set-before-await also collapses concurrent bursts from the same uid into a single request (no stampede). Easy to get backwards.

## 3. Overall verdict
**REQUEST_CHANGES** — Spec ✅ but Quality Changes-Requested. The AND gate fails on the P1: a core TypeScript source file must not be committed as binary. The fix is a one-character change per site; once `src/group-context.ts` diffs as text again this is ready to re-approve — the backfill logic itself is correct.

## 4. Advisory coverage
Two independent adversarial legs plus my own diff re-read. One cross-family leg saw `src/group-context.ts` as an opaque binary blob and flagged exactly that ("accidental encoding change") — a signal I under-weighted on first pass; decoding the blob confirmed the NUL bytes are the cause. The line-level leg's four points are the P2s above. Lesson applied: "Binary files differ" on a source file is a finding to decode, never a tooling artifact to wave off.

caster-Q:
### Qflow ④ 双 reviewer 终审:两方均 PASS

**codex-review**(持久 session,2 轮):round1 fail(C1 瞬时失败毒化负缓存 / C2 map 无界 / C3 源码字面 NUL)→ 修 → round2 verdict=pass,三条全 resolved,无 pre-push blocker。

**cc-review**(持久 session,2 轮,标准走 superpowers:requesting-code-review):round1 PASS,提 1 Important(fetchUserInfo throw/null 契约无直接单测)+ 3 Minor → 补测试+改注释 → round2 verdict=PASS,Important #1 与 Minor #4 均 RESOLVED。Minor #2/#3 经双方判为非阻塞设计权衡,接受并备忘。

全量 1222 测试通过(含新增 fetchUserInfo 契约直测 5 条)。

PR #186, opened 2026-07-16
fix(session-router): cancel timed-out agent turns

### LINKED ISSUE #184: Dispatch timeout releases the session lock without cancelling the active Agent SDK query
## Summary

The dispatch timeout introduced by #141 / #142 releases the per-session lock but explicitly leaves the timed-out handler and Claude Agent SDK query running in the background.

This prevents a permanently wedged queue, but it also allows the next message for the same session to start another turn while the previous turn — including any tools and subagents it created — may still be active. The old and new turns can therefore overlap, continue consuming upstream resources, and race on shared session state.

A July 2026 production incident showed that this accepted trade-off has a substantially larger blast radius than a delayed duplicate reply.

## Affected versions

- Incident build: based on `v1.0.3` (`1340db74ab549230ce8a0b480159b6ef9a2c5987`)
- Verified still present on `main` at `4a55e52068ed2fc4e4c1a39373360a1257f5ec33`

Current implementation:

https://github.com/Mininglamp-OSS/cc-channel-octo/blob/4a55e52068ed2fc4e4c1a39373360a1257f5ec33/src/session-router.ts#L178-L231

The source explicitly documents:

> we do NOT cancel the in-flight turn — the SDK query keeps running to completion in the background; we only unblock the queue

## Production evidence

The affected deployment used `dispatchTimeoutMs: 3600000`. Sanitized gateway logs contain four timeout events for the same session:

```text
session-router: dispatch hung past 3600000ms, releasing session lock (session=<redacted>)
session-router: dispatch hung past 3600000ms, releasing session lock (session=<redacted>)
session-router: dispatch hung past 3600000ms, releasing session lock (session=<redacted>)
session-router: dispatch hung past 3600000ms, releasing session lock (session=<redacted>)
```

The wider incident contained extensive Agent/subagent activity. Because the current timeout path does not cancel or isolate the losing `handlerPromise`, a timeout can release the lock while that execution tree is still generating model/tool activity.

This is **not** a claim that cc-channel-octo duplicated billing records. The risk is that multiple real upstream executions can remain active after a logical turn has timed out.

## Why this matters

After turn A times out:

1. `Promise.race()` settles on `timeoutError`.
2. The session lock is released.
3. Turn B is allowed to start for the same session.
4. Turn A remains live in the background.
5. A and B can concurrently:
   - invoke the model and tools;
   - create or continue subagents;
   - deliver output to the same channel;
   - update the stored SDK session ID;
   - append assistant history;
   - update reply sequence state.

PR #142 review comments already identified the late-write/session-pointer race. The production incident adds concrete resource-consumption and billing impact to that known race.

## Minimal reproduction

A deterministic test can use fake timers and a handler whose completion is externally controlled:

1. Route message A with a handler that does not resolve.
2. Advance time past `dispatchTimeoutMs`.
3. Verify the timeout reply is sent and the session lock is released.
4. Route message B for the same session.
5. Verify B starts while A is still active.
6. Resolve A after B has started.
7. Observe that A can still emit callbacks or mutate state after it has logically timed out.

The current implementation intentionally permits step 5 and does not guard step 7.

## Expected behavior

A timed-out turn should be cancelled or made incapable of producing any further externally visible or persistent side effects before a newer same-session turn proceeds.

Possible implementation directions:

- propagate an abort/interrupt handle from `SessionRouter` into `queryAgent()`;
- call the Agent SDK interruption mechanism when the timeout fires;
- mark the turn stale immediately, even if upstream cancellation is best-effort;
- ignore all late stream events, replies, session-ID callbacks, and persistence writes from a stale turn;
- propagate cancellation to subagents created by that root turn where the SDK permits it;
- prevent a newer same-session turn from overlapping until cancellation is acknowledged, or fully isolate the stale turn from shared state.

The exact SDK cancellation mechanism is left to maintainer judgment.

## Acceptance criteria

- [ ] Timeout triggers cancellation/interruption of the associated Agent SDK query.
- [ ] A timed-out turn cannot send late replies or tool-progress events.
- [ ] A timed-out turn cannot overwrite a newer SDK session ID or append out-of-order history.
- [ ] Descendant work is cancelled or isolated from shared state.
- [ ] A regression test covers a handler that continues after the timeout.
- [ ] A regression test covers a late resolve and a late reject after timeout.
- [ ] Normal fast turns and the per-session serialization contract remain unchanged.

## Relationship to existing work

This is a production-impact follow-up to #141 / #142, not a report that the original deadlock fix failed. The original fix achieved its stated availability goal; the request here is to close the explicitly deferred cancellation and stale-turn isolation gap.

ningzimu:
## Summary

- abort timed-out dispatches through the Claude Agent SDK instead of only releasing the session lock
- propagate cancellation to Octo requests and stream delivery so stale turns cannot send partial replies or update session state
- fence delayed group-roster, G4 backfill, and GROUP.md/THREAD.md fetch results before they mutate shared caches, cursors, or history
- preserve the default 30-second HTTP timeout when a caller cancellation signal is also supplied
- keep a bounded 3-second cleanup grace before releasing the queue when a handler ignores cancellation

## Verification

- `npm run type-check`
- `npx vitest run src/__tests__/group-md.test.ts src/__tests__/thread-md.test.ts src/__tests__/api-timeout.test.ts src/__tests__/e2e.test.ts`
- `npm run lint`
- `npm run build`
- `npm test` (66 files, 1,289 tests)

## Notes

The bounded grace preserves the queue-unblocking behavior introduced by #141. If a handler ignores the abort signal, the queue is still released after the grace period, while production reply, context, instruction-cache, and session-state side effects are fenced by the same signal.

Closes #184

### DIFF
--- a/src/__tests__/agent-bridge-query.test.ts
+++ b/src/__tests__/agent-bridge-query.test.ts
@@ -448,6 +448,19 @@ describe('queryAgent', () => {
     expect(mockQuery.mock.calls[0][0].options.settingSources).toEqual(['project']);
   });
 
+  it('forwards the dispatch AbortController to the SDK', async () => {
+    mockQuery.mockReturnValue(createMockStream([
+      { type: 'assistant', message: { content: [{ type: 'text', text: 'hi' }] } },
+    ]));
+    const abortController = new AbortController();
+
+    for await (const _ of queryAgent('t', makeConfig(), undefined, undefined, { abortController })) {
+      void _;
+    }
+
+    expect(mockQuery.mock.calls[0][0].options.abortController).toBe(abortController);
+  });
+
   it('reports the SDK session_id once via onSessionId', async () => {
     mockQuery.mockReturnValue(createMockStream([
       { type: 'assistant', session_id: 'sid-xyz', message: { content: [{ type: 'text', text: 'a' }] } },
--- a/src/__tests__/api-timeout.test.ts
+++ b/src/__tests__/api-timeout.test.ts
@@ -6,7 +6,14 @@
  */
 
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { getGroupMembers, postJson } from "../octo/api.js";
+import {
+  getGroupMembers,
+  getUploadCredentials,
+  leaveThread,
+  postJson,
+  updateGroupMd,
+  updateThreadMd,
+} from "../octo/api.js";
 
 // Mock the global fetch before any postJson calls.
 const mockFetch = vi.fn();
@@ -186,6 +193,72 @@ describe("GET request cancellation", () => {
       groupNo: "group-1",
       signal: controller.signal,
     });
+
+    const options = mockFetch.mock.calls[0][1] as RequestInit;
+    const effectiveSignal = options.signal as AbortSignal;
+    expect(effectiveSignal).not.toBe(controller.signal);
+    expect(effectiveSignal.aborted).toBe(false);
+
+    controller.abort(new Error("dispatch timed out"));
+    expect(effectiveSignal.aborted).toBe(true);
+  });
+});
+
+describe("direct API request cancellation", () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+    mockFetch.mockResolvedValue({
+      ok: true,
+      json: () => Promise.resolve({
+        bucket: "bucket",
+        region: "region",
+        key: "key",
+        credentials: {
+          tmpSecretId: "id",
+          tmpSecretKey: "key",
+          sessionToken: "token",
+        },
+        startTime: 1,
+        expiredTime: 2,
+      }),
+      text: () => Promise.resolve('{"version": 1}'),
+    });
+  });
+
+  const requests: Array<[string, (signal: AbortSignal) => Promise<unknown>]> = [
+    ["getUploadCredentials", (signal) => getUploadCredentials({
+      apiUrl: "https://api.example.com",
+      botToken: "test-token",
+      filename: "file.txt",
+      signal,
+    })],
+    ["leaveThread", (signal) => leaveThread({
+      apiUrl: "https://api.example.com",
+      botToken: "test-token",
+      groupNo: "group-1",
+      shortId: "thread-1",
+      signal,
+    })],
+    ["updateGroupMd", (signal) => updateGroupMd({
+      apiUrl: "https://api.example.com",
+      botToken: "test-token",
+      groupNo: "group-1",
+      content: "rules",
+      signal,
+    })],
+    ["updateThreadMd", (signal) => updateThreadMd({
+      apiUrl: "https://api.example.com",
+      botToken: "test-token",
+      groupNo: "group-1",
+      shortId: "thread-1",
+      content: "rules",
+      signal,
+    })],
+  ];
+
+  it.each(requests)("%s combines caller cancellation with the default timeout", async (_name, request) => {
+    const controller = new AbortController();
+    await request(controller.signal);
 
     const options = mockFetch.mock.calls[0][1] as RequestInit;
     const effectiveSignal = options.signal as AbortSignal;
--- a/src/__tests__/api-timeout.test.ts
+++ b/src/__tests__/api-timeout.test.ts
@@ -6,7 +6,7 @@
  */
 
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { postJson } from "../octo/api.js";
+import { getGroupMembers, postJson } from "../octo/api.js";
 
 // Mock the global fetch before any postJson calls.
 const mockFetch = vi.fn();
@@ -165,5 +165,34 @@ describe("postJson default timeout", () => {
     expect(result).toBeDefined();
     expect(result!.message_id).toBe("123456789012345678");
     expect(result!.data).toBe("ok");
+  });
+});
+
+describe("GET request cancellation", () => {
+  beforeEach(() => {
+    mockFetch.mockReset();
+  });
+
+  it("combines caller cancellation with the default timeout", async () => {
+    mockFetch.mockResolvedValue({
+      ok: true,
+      text: () => Promise.resolve('{"members": []}'),
+    });
+
+    const controller = new AbortController();
+    await getGroupMembers({
+      apiUrl: "https://api.example.com",
+      botToken: "test-token",
+      groupNo: "group-1",
+      signal: controller.signal,
+    });
+
+    const options = mockFetch.mock.calls[0][1] as RequestInit;
+    const effectiveSignal = options.signal as AbortSignal;
+    expect(effectiveSignal).not.toBe(controller.signal);
+    expect(effectiveSignal.aborted).toBe(false);
+
+    controller.abort(new Error("dispatch timed out"));
+    expect(effectiveSignal.aborted).toBe(true);
   });
 });
--- a/src/__tests__/api-timeout.test.ts
+++ b/src/__tests__/api-timeout.test.ts
@@ -2,7 +2,7 @@
  * Tests for postJson default timeout behavior (Q2 fix).
  *
  * Verifies that postJson applies a 30s default timeout when no signal is provided,
- * and respects caller-provided signals when explicitly passed.
+ * and combines it with caller-provided cancellation signals.
  */
 
 import { describe, it, expect, vi, beforeEach } from "vitest";
@@ -35,7 +35,7 @@ describe("postJson default timeout", () => {
     expect((options.signal as AbortSignal).aborted).toBe(false);
   });
 
-  it("uses caller-provided signal when explicitly passed", async () => {
+  it("combines caller cancellation with the default timeout", async () => {
     mockFetch.mockResolvedValue({
       ok: true,
       text: () => Promise.resolve('{"result": "ok"}'),
@@ -52,7 +52,12 @@ describe("postJson default timeout", () => {
 
     expect(mockFetch).toHaveBeenCalledTimes(1);
     const options = mockFetch.mock.calls[0][1] as RequestInit;
-    expect(options.signal).toBe(controller.signal);
+    const effectiveSignal = options.signal as AbortSignal;
+    expect(effectiveSignal).not.toBe(controller.signal);
+    expect(effectiveSignal.aborted).toBe(false);
+
+    controller.abort(new Error("dispatch timed out"));
+    expect(effectiveSignal.aborted).toBe(true);
   });
 
   it("rejects with abort error when timeout fires", async () => {
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -72,6 +72,7 @@ import { handleMessage } from '../index.js';
 import {
   sendMessage,
   sendReadReceipt,
+  getGroupMembers,
   getChannelMessages,
 } from '../octo/api.js';
 import { ChannelType, MessageType } from '../octo/types.js';
@@ -489,6 +490,85 @@ describe('E2E smoke tests', () => {
     const history = store.buildHistoryPrefix(sessionKey, 40);
     expect(history).toContain('[user TestUser]: What is this code?');
     expect(history).toContain('[assistant bot-001]: Hello from Claude');
+  });
+
+  it('dispatch timeout prevents delayed roster refresh from mutating group context', async () => {
+    const channelId = 'group-cancelled-roster';
+    const timeoutConfig = makeConfig({ dispatchTimeoutMs: 10 });
+    const timeoutRouter = new SessionRouter(timeoutConfig, BOT_ID);
+    let resolveMembers: ((members: Array<{ uid: string; name: string; role: number }>) => void) | undefined;
+    (getGroupMembers as ReturnType<typeof vi.fn>).mockImplementationOnce(
+      () => new Promise((resolve) => { resolveMembers = resolve; }),
+    );
+
+    const processing = simulateMessage(
+      makeGroupMsg('delayed roster', true, { channel_id: channelId }),
+      timeoutConfig,
+      store,
+      timeoutRouter,
+      groupContext,
+      streamRelay,
+    );
+    await vi.waitFor(() => expect(getGroupMembers).toHaveBeenCalledTimes(1));
+    const request = (getGroupMembers as ReturnType<typeof vi.fn>).mock.calls[0][0];
+    expect(request.signal).toBeInstanceOf(AbortSignal);
+
+    await new Promise((resolve) => setTimeout(resolve, 20));
+    expect(request.signal.aborted).toBe(true);
+    resolveMembers?.([{ uid: 'late-user', name: 'Late User', role: 0 }]);
+    await processing;
+
+    expect(groupContext.getName('late-user', channelId)).toBeUndefined();
+    expect(groupContext.buildContextSince(channelId, 0).text).toBe('');
+    expect(groupContext.getContextCursor(channelId)).toBe(0);
+  });
+
+  it('dispatch timeout prevents delayed G4 backfill from persisting stale history', async () => {
+    const channelId = 'group-cancelled-backfill';
+    const timeoutConfig = makeConfig({ dispatchTimeoutMs: 10 });
+    const timeoutRouter = new SessionRouter(timeoutConfig, BOT_ID);
+    let resolveHistory: ((messages: Array<{ from_uid: string; content: string; timestamp: number; message_seq: number }>) => void) | undefined;
+    (getChannelMessages as ReturnType<typeof vi.fn>).mockImplementationOnce(
+      () => new Promise((resolve) => { resolveHistory = resolve; }),
+    );
+
+    const first = simulateMessage(
+      makeGroupMsg('first turn', true, { channel_id: channelId, message_seq: 2 }),
+      timeoutConfig,
+      store,
+      timeoutRouter,
+      groupContext,
+      streamRelay,
+    );
+    await vi.waitFor(() => expect(getChannelMessages).toHaveBeenCalledTimes(1));
+    const request = (getChannelMessages as ReturnType<typeof vi.fn>).mock.calls[0][0];
+    expect(request.signal).toBeInstanceOf(AbortSignal);
+
+    await new Promise((resolve) => setTimeout(resolve, 20));
+    expect(request.signal.aborted).toBe(true);
+    resolveHistory?.([
+      { from_uid: 'old-user', content: 'stale history', timestamp: 1, message_seq: 1 },
+    ]);
+    await first;
+    expect(store.buildHistoryPrefix(channelId, 40)).toBe('');
+
+    (getChannelMessages as ReturnType<typeof vi.fn>).mockResolvedValueOnce([
+      { from_uid: 'old-user', content: 'retry history', timestamp: 1, message_seq: 1 },
+    ]);
+    const retryConfig = makeConfig();
+    const retryRouter = new SessionRouter(retryConfig, BOT_ID);
+    await simulateMessage(
+      makeGroupMsg('second turn', true, { channel_id: channelId, message_seq: 3 }),
+      retryConfig,
+      store,
+      retryRouter,
+      groupContext,
+      streamRelay,
+    );
+
+    expect(getChannelMessages).toHaveBeenCalledTimes(2);
+    expect(store.buildHistoryPrefix(channelId, 40)).toContain('retry history');
+    expect(store.buildHistoryPrefix(channelId, 40)).not.toContain('stale history');
   });
 
   // --- 2b. PR#51 per-session cwd wiring (regression: was uncovered) ---
--- a/src/__tests__/e2e.test.ts
+++ b/src/__tests__/e2e.test.ts
@@ -15,6 +15,7 @@ vi.mock('../octo/api.js', () => ({
   sendTyping: vi.fn().mockResolvedValue(undefined),
   sendReadReceipt: vi.fn().mockResolvedValue(undefined),
   getGroupMembers: vi.fn().mockResolvedValue([]),
+  getGroupMd: vi.fn().mockResolvedValue(undefined),
   // G4 backfill path in the real handleMessage — default to no history.
   getChannelMessages: vi.fn().mockResolvedValue([]),
   getUploadCredentials: vi.fn().mockResolvedValue({
@@ -65,6 +66,7 @@ vi.mock('../media-inbound.js', async (importOriginal) => {
 import { SessionStore } from '../session-store.js';
 import { SessionRouter } from '../session-router.js';
 import { GroupContext } from '../group-context.js';
+import { GroupMdCache } from '../group-md-cache.js';
 import { StreamRelay } from '../stream-relay.js';
 import { createAdapter, type DbAdapter } from '../db-adapter.js';
 import { queryAgent } from '../agent-bridge.js';
@@ -73,6 +75,7 @@ import {
   sendMessage,
   sendReadReceipt,
   getGroupMembers,
+  getGroupMd,
   getChannelMessages,
 } from '../octo/api.js';
 import { ChannelType, MessageType } from '../octo/types.js';
@@ -173,8 +176,19 @@ async function simulateMessage(
   router: SessionRouter,
   groupContext: GroupContext,
   streamRelay: StreamRelay,
+  groupMdCache?: GroupMdCache,
 ): Promise<void> {
-  await handleMessage(msg, config, store, router, groupContext, streamRelay, BOT_ID);
+  await handleMessage(
+    msg,
+    config,
+    store,
+    router,
+    groupContext,
+    streamRelay,
+    BOT_ID,
+    undefined,
+    groupMdCache,
+  );
 }
 
 // --- Tests ---
@@ -569,6 +583,43 @@ describe('E2E smoke tests', () => {
     expect(getChannelMessages).toHaveBeenCalledTimes(2);
     expect(store.buildHistoryPrefix(channelId, 40)).toContain('retry history');
     expect(store.buildHistoryPrefix(channelId, 40)).not.toContain('stale history');
+  });
+
+  it('dispatch timeout prevents delayed GROUP.md fetches from mutating the cache', async () => {
+    const channelId = 'group-cancelled-md';
+    const timeoutConfig = makeConfig({ dispatchTimeoutMs: 10, serverMd: true });
+    const timeoutRouter = new SessionRouter(timeoutConfig, BOT_ID);
+    const groupMdCache = new GroupMdCache();
+    let resolveMd: ((md: { content: string; version: number; updated_at: string; updated_by: string }) => void) | undefined;
+    (getGroupMd as ReturnType<typeof vi.fn>).mockImplementationOnce(
+      () => new Promise((resolve) => { resolveMd = resolve; }),
+    );
+
+    const processing = simulateMessage(
+      makeGroupMsg('delayed group instructions', true, { channel_id: channelId }),
+      timeoutConfig,
+      store,
+      timeoutRouter,
+      groupContext,
+      streamRelay,
+      groupMdCache,
+    );
+    await vi.waitFor(() => expect(getGroupMd).toHaveBeenCalledTimes(1));
+    const request = (getGroupMd as ReturnType<typeof vi.fn>).mock.calls[0][0];
+    expect(request.signal).toBeInstanceOf(AbortSignal);
+
+    await new Promise((resolve) => setTimeout(resolve, 20));
+    expect(request.signal.aborted).toBe(true);
+    resolveMd?.({
+      content: 'stale server rules',
+      version: 1,
+      updated_at: '2026-07-16T00:00:00Z',
+      updated_by: 'operator',
+    });
+    await processing;
+
+    expect(groupMdCache.get(channelId)).toBeUndefined();
+    expect(queryAgent).not.toHaveBeenCalled();
   });
 
   // --- 2b. PR#51 per-session cwd wiring (regression: was uncovered) ---
--- a/src/__tests__/group-md.test.ts
+++ b/src/__tests__/group-md.test.ts
@@ -162,6 +162,31 @@ describe('resolveGroupInstructions', () => {
     expect(call0Url()).toBe(`${BASE.apiUrl}/v1/bot/groups/${GROUP}/md`);
   });
 
+  it('does not cache or return a delayed server result after cancellation', async () => {
+    let resolveFetch: ((response: Response) => void) | undefined;
+    fetchMock.mockImplementationOnce(
+      () => new Promise<Response>((resolve) => { resolveFetch = resolve; }),
+    );
+    const cache = new GroupMdCache();
+    const controller = new AbortController();
+
+    const pending = resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      serverMd: true,
+      ...BASE,
+      channelId: GROUP,
+      cache,
+      signal: controller.signal,
+    });
+    await vi.waitFor(() => expect(fetchMock).toHaveBeenCalledTimes(1));
+
+    controller.abort(new Error('dispatch timed out'));
+    resolveFetch?.(mockMd('stale server rules'));
+
+    await expect(pending).resolves.toBeUndefined();
+    expect(cache.get(GROUP)).toBeUndefined();
+  });
+
   it('caches the server fetch — a second call serves from cache (no second fetch)', async () => {
     fetchMock.mockResolvedValueOnce(mockMd('server rules'));
     const cache = new GroupMdCache();
--- a/src/__tests__/session-router.test.ts
+++ b/src/__tests__/session-router.test.ts
@@ -483,25 +483,31 @@ describe('dispatch timeout (#141)', () => {
       makeConfig({ rateLimit: { maxPerMinute: 100 }, dispatchTimeoutMs: 50 }),
       ROBOT_ID,
     );
-    const completed: number[] = [];
+    const events: string[] = [];
 
-    // Message 1: handler never resolves (simulates a hung agent turn).
+    // Message 1: handler stays active until the router cancels it.
     const p1 = router.routeAndHandle(
       makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'same-user' }),
-      () => new Promise<void>(() => { /* never resolves */ }),
+      (_result, { abortController }) => new Promise<void>((resolve) => {
+        events.push('first-started');
+        abortController.signal.addEventListener('abort', () => {
+          events.push('first-aborted');
+          resolve();
+        }, { once: true });
+      }),
     );
 
     // Message 2: a normal fast handler on the SAME session. Without the
     // timeout it would be blocked forever behind message 1's session lock.
     const p2 = router.routeAndHandle(
       makeMsg({ message_id: '2', channel_type: ChannelType.DM, from_uid: 'same-user' }),
-      async () => { completed.push(2); },
+      async () => { events.push('second-started'); },
     );
 
     await Promise.all([p1, p2]);
 
     // Message 2 ran — the lock was released after message 1 timed out.
-    expect(completed).toEqual([2]);
+    expect(events).toEqual(['first-started', 'first-aborted', 'second-started']);
   });
 
   it('sends a single bounded apology on timeout', async () => {
@@ -512,13 +518,47 @@ describe('dispatch timeout (#141)', () => {
 
     await router.routeAndHandle(
       makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'u1' }),
-      () => new Promise<void>(() => { /* never resolves */ }),
+      (_result, { abortController }) => new Promise<void>((resolve) => {
+        abortController.signal.addEventListener('abort', () => resolve(), { once: true });
+      }),
     );
 
     expect(sendMessage).toHaveBeenCalledTimes(1);
     expect(vi.mocked(sendMessage).mock.calls[0][0]).toMatchObject({
       content: expect.stringContaining('处理超时'),
     });
+  });
+
+  it('still releases the session after the abort grace when a handler ignores cancellation', async () => {
+    vi.useFakeTimers();
+    try {
+      const router = new SessionRouter(
+        makeConfig({ rateLimit: { maxPerMinute: 100 }, dispatchTimeoutMs: 30 }),
+        ROBOT_ID,
+      );
+      const completed: number[] = [];
+      let signal: AbortSignal | undefined;
+
+      const p1 = router.routeAndHandle(
+        makeMsg({ message_id: '1', channel_type: ChannelType.DM, from_uid: 'same-user' }),
+        (_result, { abortController }) => {
+          signal = abortController.signal;
+          return new Promise<void>(() => { /* deliberately ignores cancellation */ });
+        },
+      );
+      const p2 = router.routeAndHandle(
+        makeMsg({ message_id: '2', channel_type: ChannelType.DM, from_uid: 'same-user' }),
+        async () => { completed.push(2); },
+      );
+
+      await vi.advanceTimersByTimeAsync(3_100);
+      await Promise.all([p1, p2]);
+
+      expect(signal?.aborted).toBe(true);
+      expect(completed).toEqual([2]);
+    } finally {
+      vi.useRealTimers();
+    }
   });
 
   it('does not fire for a normal fast handler', async () => {
--- a/src/__tests__/stream-relay.test.ts
+++ b/src/__tests__/stream-relay.test.ts
@@ -297,6 +297,54 @@ describe("StreamRelay", () => {
     expect(countLater).toBe(countAfter);
   });
 
+  it("drops partial output when the dispatch is cancelled", async () => {
+    const controller = new AbortController();
+    async function* cancelledChunks(): AsyncIterable<string> {
+      yield "partial";
+      controller.abort(new Error("dispatch timed out"));
+      throw new Error("cancelled stream");
+    }
+
+    const promise = relay.deliver(
+      CH_ID,
+      CH_TYPE,
+      cancelledChunks(),
+      API_URL,
+      BOT_TOKEN,
+      undefined,
+      undefined,
+      undefined,
+      controller.signal,
+    );
+    await vi.runAllTimersAsync();
+    await expect(promise).resolves.toBeUndefined();
+
+    const sendCalls = mockState.calls.filter((c) => c.fn === "sendMessage");
+    expect(sendCalls).toHaveLength(0);
+  });
+
+  it("forwards the dispatch signal to typing and message requests", async () => {
+    const controller = new AbortController();
+    const promise = relay.deliver(
+      CH_ID,
+      CH_TYPE,
+      asyncChunks(["hello"]),
+      API_URL,
+      BOT_TOKEN,
+      undefined,
+      undefined,
+      undefined,
+      controller.signal,
+    );
+    await vi.runAllTimersAsync();
+    await promise;
+
+    const typingCall = mockState.calls.find((c) => c.fn === "sendTyping");
+    const sendCall = mockState.calls.find((c) => c.fn === "sendMessage");
+    expect(typingCall?.args.signal).toBe(controller.signal);
+    expect(sendCall?.args.signal).toBe(controller.signal);
+  });
+
   it("passes apiUrl and botToken to sendMessage", async () => {
     const chunks = asyncChunks(["test content"]);
     const promise = relay.deliver(CH_ID, CH_TYPE, chunks, API_URL, BOT_TOKEN);
--- a/src/__tests__/thread-md.test.ts
+++ b/src/__tests__/thread-md.test.ts
@@ -171,6 +171,31 @@ describe('resolveGroupInstructions (thread branch)', () => {
     expect(call0Url()).toBe(`${BASE.apiUrl}/v1/bot/groups/${GROUP}/threads/${SHORT}/md`);
   });
 
+  it('does not cache or return a delayed server result after cancellation', async () => {
+    let resolveFetch: ((response: Response) => void) | undefined;
+    fetchMock.mockImplementationOnce(
+      () => new Promise<Response>((resolve) => { resolveFetch = resolve; }),
+    );
+    const threadCache = new ThreadMdCache();
+    const controller = new AbortController();
+
+    const pending = resolveGroupInstructions({
+      groupConfigDir: cfgDir,
+      threadMd: true,
+      ...BASE,
+      channelId: CHANNEL,
+      threadCache,
+      signal: controller.signal,
+    });
+    await vi.waitFor(() => expect(fetchMock).toHaveBeenCalledTimes(1));
+
+    controller.abort(new Error('dispatch timed out'));
+    resolveFetch?.(mockMd('stale thread rules'));
+
+    await expect(pending).resolves.toBeUndefined();
+    expect(threadCache.get(GROUP, SHORT)).toBeUndefined();
+  });
+
   it('caches the server fetch — a second call serves from cache (no second fetch)', async () => {
     fetchMock.mockResolvedValueOnce(mockMd('server THREAD rules'));
     const threadCache = new ThreadMdCache();
--- a/src/agent-bridge.ts
+++ b/src/agent-bridge.ts
@@ -224,6 +224,8 @@ const MAX_SYSTEM_PROMPT_CHARS = 100 * 1024;
  *     caller injects nothing (history already lives in the session).
  *   - `onSessionId`: called with the SDK session id observed for this turn, so
  *     the caller can persist it and resume next time.
+ *   - `abortController`: owned by the dispatch router and forwarded to the SDK
+ *     so timeout cancellation terminates the query and its resources.
  * @yields string chunks of assistant text output
  */
 export async function* queryAgent(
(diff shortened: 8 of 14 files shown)


Jerry-Xin (dismissed):
This PR is relevant to `cc-channel-octo`, but the cancellation fence is incomplete: a timed-out handler can still mutate shared context/history state after the router has released the session queue.

**🔴 Blocking**
- 🔴 Critical: `handleMessage` can update group context after dispatch cancellation. If the timeout fires while awaiting `groupContext.refreshMembers(...)`, the router aborts and releases the lock after the 3s grace, but the old handler can later resume and still run `groupContext.pushMessage(...)` and `groupContext.setContextCursor(...)`. That means a stale timed-out turn can mutate group context and cursor state concurrently with the next turn. Add cancellation propagation/checks around this path, ideally by passing the signal through `refreshMembers`/`getGroupMembers`/`getJson` and checking before any cache/cursor mutation. See [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-_h1aqkkd/repo/src/index.ts:487), [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-_h1aqkkd/repo/src/index.ts:500), [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-_h1aqkkd/repo/src/index.ts:515).

- 🔴 Critical: G4 backfill can persist stale history after timeout. `getChannelMessages` already supports `signal`, but this call does not pass the dispatch signal; if the dispatch times out while the backfill request is in flight, the stale handler can later call `seedHistoryFromApi(...)` before the `signal.aborted` guard at line 739. Pass `signal` into `getChannelMessages` and check `signal.aborted` before seeding/rebuilding history. See [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-_h1aqkkd/repo/src/index.ts:713), [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-_h1aqkkd/repo/src/index.ts:731), [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-_h1aqkkd/repo/src/index.ts:739).

**✅ Highlights**
- The router now owns a per-dispatch `AbortController` and forwards it into the agent SDK path.
- `StreamRelay` correctly drops partial output when cancellation has happened.
- `postJson` now preserves the default API timeout while also honoring caller cancellation.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #186 (cc-channel-octo)

## Summary

This PR upgrades the dispatch timeout mechanism from a "release the lock, let the orphaned turn finish in the background" model to active cancellation via `AbortController`. The timeout now aborts the in-flight agent turn, propagates cancellation through the entire side-effect chain (SDK query, stream delivery, HTTP requests, session-state persistence), and keeps a bounded 3-second grace period before releasing the session lock. This closes the gap where a stale turn could deliver partial replies or corrupt session state after the apology had already been sent.

The design is thorough: the abort signal is threaded through `DispatchContext` → handler → `sendMessage` / `streamRelay.deliver` / `queryAgent` / `postJson`, with `AbortSignal.any` combining dispatch cancellation with per-request HTTP timeouts. Test coverage adds cancellation-aware handlers, grace-period edge cases, and signal-forwarding assertions.

## Verification

- ✅ **Abort lifecycle** — `AbortController` is created in `runHandlerWithTimeout` (`src/session-router.ts:207`), aborted on timeout (`:222`), and the same instance is forwarded via `DispatchContext` to the handler. No leaked controllers.
- ✅ **Signal propagation** — `signal` reaches every user-visible side effect: `sendMessage` (command replies, tool progress, error replies, no-output fallback), `sendReadReceipt`, `queryAgent` (→ SDK), `streamRelay.deliver` (→ typing heartbeat + per-segment sends). Each downstream call passes `signal` to `postJson`, which combines it with `AbortSignal.timeout` via `AbortSignal.any` (`src/octo/api.ts:68-70`).
- ✅ **Persistence fencing** — `store.appendUser` is gated by `if (signal.aborted) return;` at `src/index.ts:739`. `store.appendAssistant` and `setLastBotReplySeq` are gated by the check at `:1003`. `onSessionId` and `onResumeFailed` both check `signal.aborted` before mutating (`:843`, `:849`). Cancelled turns cannot corrupt session state.
- ✅ **Stream-relay abort checks** — `deliver` checks `signal?.aborted` at entry (`:171`), per-chunk (`:193`), post-catch (`:208-210`), post-accumulate (`:213`), per-segment (`:242`), and per-send-catch (`:281`). Typing timer is cleared in `finally` (`:296`).
- ✅ **Grace period preserves #141** — The 3s grace (`DISPATCH_ABORT_GRACE_MS`) races `handlerOutcome` against a timer. If the handler ignores cancellation, the lock still releases. The existing apology (`replySafe`) fires after the grace, independent of the dispatch signal.
- ✅ **Timeout disabled path** — When `dispatchTimeoutMs` is 0 or negative, the handler runs unguarded with a non-aborting `AbortController`. The signal checks in the handler are no-ops. Consistent with the timeout-enabled path.
- ✅ **Backward-compatible test migration** — Existing `routeAndHandle` callers in test files use single-parameter callbacks (`async (result) => ...`), which TypeScript allows for a two-parameter handler type. No test breakage.
- ✅ **Node version** — `AbortSignal.any` requires Node 20.3+; `engines.node` is `>=22` (`package.json:77`). Safe.

Static analysis only at head `9cad98af`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two P2 items below.

### P2 — Grace period extends lock hold by up to 3s (`src/session-router.ts:247`)

When a handler ignores the abort signal, the session lock is held for `dispatchTimeoutMs + DISPATCH_ABORT_GRACE_MS` before the queue advances. For group sessions with concurrent users, queued messages wait up to 3 extra seconds. The trade-off is sound (bounded cleanup vs. bounded delay, preserving #141's guarantee), but operators tuning `dispatchTimeoutMs` may not realize the effective hold time. Consider documenting the additive nature of the grace period in the config schema or operator docs.

Additionally, the grace-period expiry logs at `console.warn` (`:248`). If cooperative handlers routinely use the full grace window for SDK cleanup, this becomes noise. Consider `console.info` for the expected-settle-during-grace path and `console.warn` only when the grace actually expires.

### P2 — `onSessionId` discard on cancelled turns is undocumented (`src/index.ts:853`)

The `signal.aborted` guard in `onSessionId` prevents persisting a session id from a cancelled turn. This is the safer default — a cancelled SDK session may hold partial state, and resuming it on the next turn risks history corruption. However, it also means the next turn pays first-turn setup cost again (history injection, MCP server init). A brief comment at `:841` noting this is deliberate (and why) would prevent a future maintainer from "fixing" it by removing the guard.

- `postJson` signal combination: `AbortSignal.any([signal, timeoutSignal])` correctly fires on either cancellation source. When no caller signal is provided, only the HTTP timeout applies (unchanged behavior).
- `replySafe` apology intentionally omits the dispatch signal — it uses its own HTTP timeout (`DEFAULT_TIMEOUT_MS`), so a cancelled dispatch cannot prevent the user-facing apology from being delivered.
- `teeChunks` generator checks `signal.aborted` per chunk, providing a second cancellation layer even if the SDK generator doesn't immediately terminate.
- Handler outcome conversion (`.then(() => ({ kind: 'completed' }), (error) => ({ kind: 'failed', error }))`) eliminates the unhandled-rejection risk from late-settling handlers — cleaner than the old `.catch(() => {})` pattern.
- Test coverage: the new `'still releases the session after the abort grace when a handler ignores cancellation'` test exercises the grace-period edge case with fake timers. The stream-relay cancellation test verifies no partial output is flushed.
- No dual-path parity issues (C1): both timeout-enabled and timeout-disabled paths pass `DispatchContext` consistently.

## Verdict: APPROVED

Well-engineered cancellation propagation with comprehensive abort fencing across all side-effect paths. The grace period correctly preserves the queue-unblock guarantee from #141. No correctness, data-integrity, or safety issues found. The two P2 items are documentation/observability improvements, not blockers.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Abort signal propagation is comprehensive and correctly fences all persistence and outbound side effects. Grace period preserves #141 queue-unblock guarantee.

yujiawei (approved):
# Code Review — PR #186 (cc-channel-octo)

## 1. Scope / spec compliance

The diff does exactly what the PR describes and nothing more:

- `session-router.ts` — on dispatch timeout it now `abortController.abort(timeoutError)` instead of only unblocking the queue, then waits a bounded `DISPATCH_ABORT_GRACE_MS` (3s) for cooperative cleanup before releasing the lock. The `#141` guarantee (the queue always advances, even if a handler ignores cancellation) is preserved by the grace race.
- `index.ts` — the dispatch `signal` is threaded through the whole handler: read-receipts, tool-progress notices, `onSessionId`/`onResumeFailed` persistence callbacks, `queryAgent`, the tee generator, `streamRelay.deliver`, and every terminal `sendMessage`. Late side effects are fenced with `if (signal.aborted) return`.
- `agent-bridge.ts` — forwards the `AbortController` into the SDK `query()` options (a supported option that stops the query and cleans up resources).
- `octo/api.ts` — composes the caller signal with the existing 30s timeout via `AbortSignal.any`, so cancellation and timeout both apply.
- `stream-relay.ts` — drops partial output and suppresses the expected cancellation error when the signal is aborted, and forwards the signal to typing/message requests.

No functionality added beyond the stated fix; no flags/fields/interfaces introduced that weren't required. **Spec: compliant.**

## 2. Correctness

I traced the abort/grace race and the fencing paths and found no blocking issues:

- **Persistence fencing is correct.** `store.appendUser` (index.ts:739) and `store.appendAssistant` (index.ts:1018, guarded by the `signal.aborted` return at index.ts:1001) are both skipped after an abort, so a cancelled turn never writes a half-turn or a partial assistant reply to the local store. `onSessionId`/`onResumeFailed` are guarded with `!signal.aborted`, so a stale turn can't corrupt the SDK session id.
- **The timeout apology intentionally does not use the aborted signal.** `replySafe` (session-router.ts:734) calls `sendMessage` without the dispatch signal, relying only on the api-layer 30s timeout. That's the right call — passing the aborted signal would make the apology fail instantly. Good.
- **Unhandled-rejection safety is improved.** `handler(...)` is now wrapped in `.then(ok, err)` that converts rejection into a value, so a handler that rejects after the timeout can't surface as an unhandled rejection. This is cleaner than the previous "attach a no-op catch" approach.
- **SDK-abort → stream-relay interaction is safe.** When the SDK query aborts it rejects the stream; stream-relay's `catch` returns early on `signal?.aborted` (stream-relay.ts:210) so it neither flushes the stale partial nor logs the expected cancellation as an error. The agent-bridge stale-resume recovery is not falsely triggered because `isResumeError` only matches resume/tool_result-shaped messages, not an abort reason.

Tests cover the new behavior well: the "handler ignores cancellation still releases after grace" case (fake timers, 3.1s advance), the "drops partial output when cancelled" case, signal forwarding to typing/message, the SDK abortController forwarding, and the combined `AbortSignal.any` behavior. Type-check, lint, build, and the four cited test files all pass locally against head `9cad98a`.

## 3. Non-blocking suggestions (P2)

These are advisory and do not block merge:

- **`stream-relay.ts` typing-heartbeat timer relies solely on `finally` for cleanup.** If the upstream `chunks` generator never honors the abort and never settles, the `for await` loop blocks forever and `finally { clearInterval(typingTimer) }` is never reached, leaking the interval (the callback itself is already fenced with `!signal?.aborted`, so no network spam — just an orphaned no-op timer). This is a pre-existing property, not a regression: this PR strictly improves it by forwarding the AbortController to the SDK so the common hang path now actually cancels and exits the loop. Optional hardening: `signal?.addEventListener('abort', () => clearInterval(typingTimer), { once: true })` to cover the degraded "SDK ignores abort entirely" path that the router's grace window is designed for.
- **`octo/api.ts` `AbortSignal.timeout` retention.** `AbortSignal.timeout(30_000)` keeps a live timer (and, via `AbortSignal.any`, its listeners) for the full 30s even when `fetch` resolves in milliseconds. Under bursty traffic (typing + read-receipts + segments) this accumulates short-lived closures/timers. Minor GC pressure only; a manual `AbortController` + `clearTimeout` in a `finally` would avoid it. `AbortSignal.any` is the correct way to combine caller cancellation with the timeout, so this is an optimization, not a correctness concern.

## 4. Coverage notes (what I could not fully verify)

- I could not confirm from this repo whether the SDK actively tears down the underlying network socket/subprocess on `abortController.abort()` versus only stopping iteration; the fix's resource-cleanup claim depends on the SDK honoring the signal. The bounded 3s grace + lock release makes this safe either way for queue liveness.
- I did not exercise the DB/store layer for async-consistency edge cases where an abort lands mid-write; the code paths I traced fence before each write, so a torn write is not reachable through the reviewed paths.

## Verdict

No P0/P1 issues. Spec-compliant, well-fenced, well-tested. The two P2 items are optional hardening/optimization. Approving.

Jerry-Xin (dismissed):
Summary: The PR is relevant to this repository and the core cancellation design is sound, but one group/thread instruction path is still not cancellation-fenced and can mutate shared state after timeout.

🔴 Blocking

🔴 Critical — `resolveGroupInstructions` is called without the dispatch `signal` at [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/index.ts:910). If a server instruction fetch is slow, the router aborts and releases the session after the grace period, but this fetch continues outside the lock. When it later resolves, [src/group-md.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/group-md.ts:139) and [src/group-md.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/group-md.ts:209) can still populate `GroupMdCache` / `ThreadMdCache` at [src/group-md.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/group-md.ts:148) and [src/group-md.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/group-md.ts:218). That can overwrite a newer refresh or post-timeout invalidation with stale instructions before `handleMessage` finally observes `signal.aborted` at [src/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/index.ts:985). Pass `signal` into `resolveGroupInstructions`, and also fence inside `group-md.ts` after the awaited fetch and before cache mutation/return.

💬 Non-blocking

🟡 Warning — A few lower-level helpers still use `params.signal ?? AbortSignal.timeout(...)`, which drops the default timeout whenever a caller supplies a signal: [src/octo/api.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/octo/api.ts:171), [src/octo/api.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/octo/api.ts:534), [src/octo/api.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/octo/api.ts:802), and [src/octo/api.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/octo/api.ts:847). This PR fixed `postJson`/`getJson`; consider applying the same `AbortSignal.any([signal, timeout])` pattern consistently.

✅ Highlights

The router now aborts timed-out dispatches and preserves a bounded queue release window at [src/session-router.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/session-router.ts:228).

Stream delivery correctly checks cancellation before flushing partial output and before each send at [src/stream-relay.ts](/Users/mlamp/.openclaw/tmp/pr-review-lir9zyh_/repo/src/stream-relay.ts:191).

The new tests cover the main timeout, stream, API cancellation, roster, and history backfill paths well.

yujiawei (changes requested):
# Code Review — PR #186 (cc-channel-octo) — Updated

Superseding my earlier approval. On a closer pass of the handler window that runs **before** the first `signal.aborted` guard (`src/index.ts:739`), the cancellation fence is incomplete relative to this PR's stated goal ("propagate cancellation ... so stale turns cannot ... update session state"). Two session-state mutation paths are reachable by a timed-out, already-released turn.

## 1. Spec compliance

**Spec: ❌ — the cancellation fence omits two session-state paths the PR claims to cover.**

- **Group context (unfenced).** `groupContext.refreshMembers(...)` at `src/index.ts:487` is not passed the dispatch signal (and `refreshMembers` → `getGroupMembers` → `getJson` have no signal parameter threaded), then `groupContext.pushMessage(...)` at `src/index.ts:500` and `groupContext.setContextCursor(...)` at `src/index.ts:515` mutate shared group-context state. All three run before the `signal.aborted` guard at `src/index.ts:739`. If the dispatch times out while awaiting `refreshMembers`, the router releases the lock after the 3s grace and the next turn can start, while the stale turn later resumes and still mutates group context/cursor.
- **G4 history backfill (unfenced).** `getChannelMessages(...)` at `src/index.ts:713` already supports a `signal` (see `src/octo/api.ts:406-429`) but this call site does not pass it, and there is no `signal.aborted` check before `seedHistoryFromApi(...)` at `src/index.ts:731`. A backfill in flight at timeout can persist history for a cancelled turn.

The rest of the diff is spec-accurate: SDK `abortController` forwarding, `AbortSignal.any` timeout composition in `octo/api.ts`, stream-relay partial-output suppression, and the post-query persistence guards (`appendUser`/`appendAssistant`) are all correctly fenced. The gap is specifically the pre-query window above.

## 2. Code quality

The core mechanism is well built and well tested:

- Router now owns a per-dispatch `AbortController`, aborts on timeout, waits a bounded 3s grace, then releases the lock — preserving the `#141` queue-liveness guarantee even if a handler ignores cancellation.
- Handler rejection is converted to a value (`.then(ok, err)`), removing the unhandled-rejection risk from the previous no-op-catch approach.
- `StreamRelay` drops partial output and suppresses the expected cancellation error on abort; signal is forwarded to typing/message requests.
- Type-check, lint, build, and the four cited test files pass against head `9cad98a`.

**Severity of the two spec gaps (independent calibration):** contained, not data-corruption.
- Low reachability at the default `dispatchTimeoutMs` (300s): `getGroupMembers`/`getChannelMessages` each self-bound at the 30s API timeout, far under 300s, and the multi-minute hang risk (the agent query) runs *after* these lines. The race is realistically reachable only when an operator configures `dispatchTimeoutMs` at or below the ~30s per-call ceiling.
- DB guards blunt the worst outcomes even when reached: the cursor upsert is monotonic (`ON CONFLICT ... WHERE excluded.last_id > group_context_cursors.last_id`, so no cursor regression), and message insert is `INSERT OR IGNORE` on a partial unique index over `(session_id, role, message_seq>0)`, so re-persisting real-seq backfill rows is idempotent. Residual: a stale `pushMessage` can still append a duplicate `group_messages` row, and NULL-seq messages are exempt from the idempotency guard.

**Quality: Changes-Requested** (driven by the spec-completeness gap, not by an independent correctness P0).

## 3. Recommended fix (small, local)

- Thread `signal` into `refreshMembers` → `getGroupMembers` → `getJson`, and add `if (signal.aborted) return;` before `pushMessage`/`setContextCursor`.
- Pass `signal` into the backfill `getChannelMessages`, and check `signal.aborted` before `seedHistoryFromApi`.

Both keep the pre-query window consistent with the post-query fencing the PR already implements.

## 4. Coverage notes (what I could not fully verify)

- Whether the SDK actively tears down the underlying subprocess/socket on `abortController.abort()` vs only stopping iteration — the resource-cleanup claim depends on the SDK honoring the signal; the bounded grace + lock release keeps queue liveness safe either way.
- Minor (non-blocking): the stream-relay typing `setInterval` is cleared only in `finally`; if an upstream generator never settles and never honors abort, the timer leaks (the callback itself is fenced, so no network spam). Pre-existing; this PR strictly improves it by making the common hang path cancellable. Optional hardening: clear it on `abort`.
- Minor (non-blocking): `AbortSignal.timeout(30_000)` in `octo/api.ts` retains a live timer/listeners for the full 30s under `AbortSignal.any` even when fetch resolves fast — minor GC pressure under bursty traffic; a manual controller + `clearTimeout` avoids it.

## Verdict: REQUEST_CHANGES

Complete the cancellation fence over the two pre-query session-state paths, then this is good to merge. No other blockers.

yujiawei (dismissed):
# Code Review — PR #186 (cc-channel-octo) — Re-review at head 8122052

Re-reviewing at the new head `8122052` ("fix: fence cancelled dispatch side effects"), which lands on top of the previously-reviewed `9cad98a`. The two blocking items from my prior review are now resolved. Flipping to approve.

## What changed since the last review

My earlier REQUEST_CHANGES flagged two session-state write paths that ran before the first `signal.aborted` guard and were not fenced. Both are now fixed:

- **Group context.** `refreshMembers` takes a `signal`, forwards it to `getGroupMembers` → `getJson` (now combined with the HTTP timeout via `AbortSignal.any`), and checks `signal?.aborted` after the fetch and in its catch. The caller passes `signal` and adds `if (signal.aborted) return;` before `getContextCursor`/`pushMessage`/`setContextCursor` (`src/index.ts:487-490`). A timed-out roster refresh can no longer mutate group context or the cursor.
- **G4 history backfill.** `getChannelMessages` now receives `signal`, and the call site checks `signal.aborted` before `seedHistoryFromApi` (and in the catch), returning early (`src/index.ts:717-746`). Nice detail: it calls `backfilledSessions.delete(backfillKey)` on the aborted path, so a cancelled cold-start backfill retries next turn instead of being permanently marked done.

## Verification at this head

- Type-check, lint, and the full test suite pass locally (66 files, 1282 tests).
- The two new e2e tests exercise exactly the scenarios I flagged: a delayed roster refresh is confirmed not to mutate group context after cancellation, and a delayed G4 backfill is confirmed not to persist stale history — with a follow-up assertion proving the cancelled backfill actually retries on the next turn and that stale history never lands.
- Spot-checked that `getJson` gained the same `AbortSignal.any` composition `postJson` already had, so both HTTP verbs honor caller cancellation plus the default timeout.

## On the first-pass checklist

The first-pass review (against `9cad98a`) suggested APPROVED, but its "Persistence fencing — cancelled turns cannot corrupt session state" item was drawn only from the post-query writes (`appendUser`/`appendAssistant`/`onSessionId`/`onResumeFailed`) and missed the two pre-query paths above. On the original head that conclusion was incomplete. On this head it now holds, because those two paths are fenced.

## Remaining (non-blocking) notes

- The first-pass P2 items still stand as optional polish: document that the effective lock hold is `dispatchTimeoutMs + DISPATCH_ABORT_GRACE_MS` when a handler ignores cancellation, and add a one-line comment noting the `onSessionId` discard on cancelled turns is deliberate.
- Prior P2s unchanged: the stream-relay typing `setInterval` is cleared only in `finally` (fine in practice now that the common hang path cancels), and `AbortSignal.timeout` retains a timer for its full window under `AbortSignal.any` (minor GC pressure). None block merge.

## Verdict: APPROVED

The cancellation fence is now complete across both the pre-query and post-query session-state paths, with targeted tests proving it. No P0/P1 issues.

Jerry-Xin (approved):
Approve — re-review of head `301fdfd3`. The remaining unfenced-mutation blocker from the previous head is now fixed, and the two earlier fixes still hold. Clean one-commit fix on top of `81220529` (ahead 1 / behind 0), so this delta is purely the author's fix.

## 🔴 Blocking
None. The previously-open third blocker is resolved:

- **resolveGroupInstructions unfenced md-cache write (was open at `81220529`)** — FIXED. The call site (`src/index.ts:919`) now passes `signal` into `resolveGroupInstructions` and adds `if (signal.aborted) return;` right after (`:921`). Inside `group-md.ts`, both branches now guard the shared-cache write: after `getGroupMd({…signal})` there is `if (signal?.aborted) return undefined;` before `cache.set(groupNo, entry)`, and after `getThreadMd({…signal})` there is the same guard before `threadCache.set(groupNo, shortId, entry)`. Entry guards and catch-branch guards were also added. A timed-out turn can no longer mutate the process-level GROUP.md/THREAD.md caches.

## 💬 Non-blocking
- **Prior 💬 (api.ts `signal ?? AbortSignal.timeout()`) — RESOLVED.** A `withDefaultTimeout()` helper now uniformly returns `AbortSignal.any([signal, timeoutSignal])` when a caller signal is present, so the default 30s request timeout is preserved instead of dropped. Applied consistently across `postJson`, `getJson`, `getUploadCredentials`, `requestNoBody`, `updateGroupMd`, and `updateThreadMd`.
- Media/file preprocessing helpers (`downloadInboundImage`, `tryResolveFile`) still create their own 30s timeout signals and do not receive the dispatch signal, so a timed-out turn can keep consuming network/disk during preprocessing. Downstream session-store/reply side effects are already fenced, so this is not blocking, but threading the dispatch signal through these helpers would make the cancellation model more complete.
- `getChannelMessages` now propagates the caller signal but still logs each abort as an error before returning `[]`. Since timeout cancellation is now expected behavior, consider suppressing/downgrading that log when `signal?.aborted`.

## ✅ Highlights
- **Regression check passed** — both earlier fixes still hold on this head:
  - Blocker 1 (roster refresh): `groupContext.refreshMembers(…signal)` followed by `if (signal.aborted) return;` (`src/index.ts:488`) before the cursor/context-block work.
  - Blocker 2 (G4 backfill): `getChannelMessages({…signal})` followed by `if (signal.aborted){…return;}` (`:722`) before `seedHistoryFromApi`, plus an abort guard in the catch and `if (signal.aborted) return;` (`:749`) before `store.appendUser`.
- No new sibling of the unfenced-mutation class was introduced — SDK session-id persistence (`:867/:872`), history/store writes, and cursor updates are all fenced behind `signal.aborted` checks.
- Router-side cancellation is coherent: timed-out dispatches are aborted with a bounded cleanup grace before queue release, preserving the #141 queue-unblocking safety net.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #186 (cc-channel-octo)

## Summary

This PR transforms the dispatch timeout from a "release the lock, let the orphan run" model into full cooperative cancellation. An `AbortController` is created per dispatch in `session-router.ts`, its signal is threaded through every async boundary in the handler chain (API calls, SDK query, stream delivery, group-context refresh, GROUP.md/THREAD.md fetches), and abort fences prevent late-arriving results from mutating shared caches, session state, or message history. The timeout race is refactored from rejection-based to value-based (eliminating the unhandled-rejection footgun), and a bounded 3-second grace period preserves the queue-unblock guarantee from #141 even when a handler ignores cancellation. The `withDefaultTimeout` helper in `octo/api.ts` fixes a real bug where passing a caller signal silently dropped the 30-second HTTP timeout.

Overall, this is a well-designed cancellation architecture with thorough test coverage across unit, integration, and E2E levels.

## Verification

Static analysis only at head `301fdfd3ae5f5d1bb8c16e51fe61c8f9dbc28aad`; build and tests not executed in this environment.

- ✅ **Signal propagation completeness** — traced the `AbortController` from `runHandlerWithTimeout` through `DispatchContext` → handler in `index.ts` → every async API call (`sendMessage`, `sendReadReceipt`, `sendTyping`, `getChannelMessages`, `getGroupMembers`, `getGroupMd`, `getThreadMd`) → SDK query via `abortController` option. All call sites in the dispatch path forward the signal. Each API function routes through `withDefaultTimeout` which combines caller signal with the 30s HTTP timeout via `AbortSignal.any`.
- ✅ **Abort fences before state mutations** — verified `signal.aborted` guards precede every shared-state write: `backfilledSessions.delete` rollback (`src/index.ts:725`), `store.appendUser` (`src/index.ts:744`), `store.setSdkSessionId` (`src/index.ts:831`), `store.clearSdkSessionId` (`src/index.ts:836`), `groupContext` member map updates (`src/group-context.ts:268`), `cache.set`/`threadCache.set` for GROUP.md/THREAD.md (`src/group-md.ts:139,214`), `store.appendAssistant` (gated by `stream-relay.deliver` return + `signal.aborted` check at `src/index.ts:1012`).
- ✅ **Stream relay cancellation** — `deliver` checks `signal?.aborted` at entry, in the chunk accumulation loop, after stream error, before segment sending, and in the send error handler. The typing interval timer is properly cleared in `finally` (including on early returns from abort). Partial output is correctly suppressed on cancellation — no stale turn replies.
- ✅ **Timeout race refactor** — value-based `Promise.race` between `handlerOutcome` and `timeoutOutcome` eliminates the old rejection-based unhandled-rejection risk. The handler promise is wrapped to never reject (`then(() => completed, (error) => failed)`), so no `.catch(() => {})` defense-in-depth is needed.
- ✅ **Grace period bounded** — `DISPATCH_ABORT_GRACE_MS = 3_000` provides a bounded cleanup window. If the handler ignores the abort signal, the lock is released after 3s (with a warning log). This preserves the #141 queue-unblock guarantee.
- ✅ **Test coverage** — new tests cover: timeout→abort→handler-settle ordering (session-router), delayed roster/backfill/GROUP.md prevented from mutating state (e2e), partial output suppression (stream-relay), AbortController forwarding to SDK (agent-bridge), cancellation fence in group-md and thread-md resolvers, and `withDefaultTimeout` combining signals (api-timeout).
- ✅ **C1 — paired path parity** — `backfillKey` add↔delete: `backfilledSessions.add` at `:712`, rollback `delete` on abort at `:725` and in catch at `:739`. `onSessionId` set↔clear: both gated by `!signal.aborted` at `:831` and `:836`.
- ✅ **C2 — control-flow ordering** — abort fences are placed after each async operation in the handler chain, consistently before the next state mutation. No double-action risk.

## Findings

No P0/P1 issues. One nit below.

### Nit — `replySafe` apology not cancellable (`src/session-router.ts:253`)

`replySafe` calls `sendMessage` without the dispatch signal. The apology gets its own 30-second `withDefaultTimeout` via the default path. This is intentional (the apology is the router's own response, not part of the cancelled handler), and it is bounded — a stuck Octo API cannot wedge the session permanently. In the absolute worst case the session lock is held for up to ~33 seconds after timeout (3s grace + 30s apology timeout) instead of 3s. Not a defect, but worth being aware of when interpreting timeout-related latency.

- `AbortSignal.any` requires Node.js 20+ — already implied by `AbortSignal.timeout` usage pre-PR.
- `teeChunks` generator in `src/index.ts:993` correctly gates each chunk on `signal.aborted`, preventing stale output from reaching the stream relay.
- `onToolUse` callback in `src/index.ts:831` early-returns on abort, suppressing tool-progress messages after cancellation.
- `getChannelMessages` signal is forwarded to `postJson` → `withDefaultTimeout`, so the G4 backfill HTTP call is both cancellable and timeout-bounded.
- `refreshMembers` throttle (`lastRefresh`) is correctly NOT updated on cancellation — the next non-cancelled call can retry immediately.
- The old `handlerPromise.catch(() => {})` defense-in-depth is correctly removed — the value-based race makes it unnecessary.

## Verdict: APPROVED

Well-structured cancellation propagation with comprehensive abort fencing and thorough test coverage. The signal threading through the dispatch chain is complete, state mutations are properly fenced, and the timeout race refactor eliminates a real unhandled-rejection risk. The nit about `replySafe` is informational only.

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. Cancellation signal propagation is complete across all dispatch-path async boundaries. Abort fences correctly prevent late-arriving results from mutating shared state. The value-based timeout race and bounded grace period are sound. Test coverage spans unit, integration, and E2E levels with realistic cancellation scenarios.

yujiawei (approved):
# Code Review — PR #186 (cc-channel-octo)

Re-review at head `301fdfd3`. This supersedes my earlier request for changes (at head `9cad98a`), whose two blockers are now resolved.

## 1. Scope / spec compliance

**Spec: ✅**

The diff does exactly what the PR describes and nothing beyond it:
- Forwards a per-dispatch `AbortController` to the Claude Agent SDK `query()` so a timed-out turn is actually cancelled, not merely lock-released (`agent-bridge.ts`, `index.ts` sessionOpts).
- Propagates the cancellation signal through Octo requests and stream delivery so a stale turn cannot send partial replies or update session state (`stream-relay.ts`, `octo/api.ts`, `index.ts`).
- Fences delayed group-roster refresh, G4 backfill, and GROUP.md/THREAD.md fetch results before they mutate shared caches, cursors, or history (`group-context.ts`, `group-md.ts`, `index.ts`).
- Preserves the default 30s HTTP timeout when a caller signal is also supplied, via `AbortSignal.any([signal, AbortSignal.timeout(...)])` (`octo/api.ts:21-24`).
- Keeps a bounded 3s cleanup grace before releasing the queue when a handler ignores cancellation (`session-router.ts` `DISPATCH_ABORT_GRACE_MS`), preserving the #141 queue-liveness guarantee.

No over-build (no unrequested flags/fields/endpoints) and no divergence from the stated approach.

### Prior blockers — both resolved

My earlier review flagged two session-state mutation paths reachable by a timed-out, already-released turn, in the handler window *before* the first abort guard. Both are now fenced:

1. **Group context.** `refreshMembers` now takes `signal` (`group-context.ts:249`) and threads it through `getGroupMembers` → `getJson` → `fetch` (`octo/api.ts:333-340`, `275-286`), with an early `if (signal?.aborted) return;` after the await and in the catch. The call site adds `if (signal.aborted) return;` (`index.ts:488`) before `pushMessage`/`setContextCursor`.
2. **G4 history backfill.** `getChannelMessages` now receives `signal` (`index.ts:720`) and an `if (signal.aborted) { backfilledSessions.delete(backfillKey); return; }` guard (`index.ts:722-725`, mirrored in the catch at `741-744`) runs before `seedHistoryFromApi`. The sentinel rollback correctly allows a later legitimate turn to retry backfill.

## 2. Code quality

**Quality: Approved**

The mechanism is well built and well tested:
- The router owns one `AbortController` per dispatch, aborts on timeout, waits a bounded grace, then releases the lock. Handler settlement is converted to a value (`.then(ok, err)`), removing the unhandled-rejection risk of the previous no-op-catch approach (`session-router.ts:219-224`).
- Post-query persistence is consistently fenced: `setSdkSessionId`/`clearSdkSessionId` guard on `!signal.aborted`; `appendUser`/`appendAssistant`, read receipts, and the "no response" fallback all sit behind `signal.aborted` checks.
- `StreamRelay.deliver` returns early on abort at every stage (pre-flight, per-chunk, per-segment, and in both catch blocks), suppresses the expected cancellation error instead of flushing partial output, and forwards the signal to typing/message requests. The typing heartbeat interval is cleared in `finally`, so no timer leaks on the abort-return paths.
- `AbortSignal.any` composition is correct and the environment supports it (Node >=22 per `package.json` engines).

Verified locally at head `301fdfd3`: `tsc --noEmit` clean; the four cancellation-relevant test suites pass (`session-router`, `stream-relay`, `api-timeout`, `agent-bridge-query` — 167 tests). The new tests cover the abort-grace lock release when a handler ignores cancellation, partial-output suppression, signal forwarding to the SDK and to Octo requests, and the `AbortSignal.any` combination.

### Non-blocking notes (P2)

- `AbortSignal.timeout(30_000)` inside `withDefaultTimeout` retains a live 30s timer under `AbortSignal.any` even when `fetch` resolves quickly; under bursty traffic this is minor GC pressure. A manual controller + `clearTimeout` on settle would avoid it. Pre-existing pattern; not introduced by this PR.
- The stream-relay typing `setInterval` is cleared only in `finally`. The callback itself is abort-fenced (no network spam), so this is purely a latent-timer nit. Optional hardening: clear on `abort`. Pre-existing; this PR strictly improves the surrounding path.

## 3. Overall verdict

**APPROVED** — Spec ✅ AND Quality Approved. Both prior blockers are fixed, the cancellation fence is now complete across the pre-query and post-query windows, and the change is well tested. The remaining notes are optional micro-optimizations, not merge blockers.

## 4. Coverage notes

- SDK teardown semantics: I confirmed against `@anthropic-ai/claude-agent-sdk@0.3.162` that `Options.abortController` is typed as `AbortController` (not `abortSignal`), so the forwarding is correct and cancellation reaches the SDK; the SDK documents its own stdin-EOF + ~2s graceful-close before killing the subprocess, which is consistent with the router's 3s grace. Whether the subprocess/socket is fully reclaimed on abort is internal to the SDK; the bounded grace + lock release keeps queue liveness safe regardless.
- Not assessable from the diff alone: production `dispatchTimeoutMs` configuration (the race window shrinks as this approaches the 30s per-call API ceiling), and real-world behavior of a handler that ignores the signal past the grace window (covered by unit test with fake timers, not observed against a live SDK).

OctoBoooot (approved):
## Review: fix(session-router): cancel timed-out agent turns (#186)

**Verdict:** Approve — the timeout-cancellation machinery and all shared-state fences are byte-verified correct at this head. One caveat: the code CI jobs (Vitest/type-check/ESLint/build) are still **pending** — merge must wait for them to report green (only `check-sprint`/label/pr-title-lint have passed so far; none are failing).

Reviewed at high scrutiny (concurrency/lifecycle — the primary risk is a timed-out turn racing shared state against the next turn). I independently traced every fence rather than trusting the summary; all three shared-state write paths the design must guard are correctly fenced:

- **Group roster/context** (`index.ts:487-488`) — `refreshMembers(…, signal)` then `if (signal.aborted) return` *before* the synchronous `pushMessage` + `setContextCursor`. No await between guard and mutation, so on a single event loop the guard is effective.
- **G4 history backfill** (`index.ts:712-749`) — `getChannelMessages(…, signal)` + `signal.aborted` guard (with `backfilledSessions.delete` rollback) before `seedHistoryFromApi`, in the catch, and again before `store.appendUser`.
- **GROUP.md / THREAD.md cache** (`group-md.ts`) — signal threaded through; `signal?.aborted` guard after each `getGroupMd`/`getThreadMd` fetch and *before* every `cache.set` / `threadCache.set` (lines 141→150, 214→223), plus the post-call guard at `index.ts:921`. The fetch→set window a timed-out turn could use to write the process-shared md cache is closed.

Also verified beyond the three fences:
- **SDK session-id persistence** (`index.ts:867/872`) — `onSessionId`/`onResumeFailed` wrapped in `if (!signal.aborted)`, so a stale turn can't clobber the next turn's session id.
- **stream-relay** — signal threaded with guards before typing + delivery, so a cancelled turn can't emit a partial reply.
- **`withDefaultTimeout`** (`octo/api.ts`) — `AbortSignal.any([signal, timeoutSignal])` preserves the 30s default even when a caller signal is supplied, across all four call sites (fixes the earlier `signal ?? AbortSignal.timeout()` that dropped the default when a signal was passed).
- **Router timeout/grace** — both outcomes converted to values (no unhandled rejection after timeout), `abortController.abort()` on timeout, bounded 3s grace preserving #141's queue-unblock guarantee, timers cleared in `finally`, single lock-release path.

Tests cover abort-on-timeout ordering, the single bounded apology, grace-release when a handler ignores cancellation, and no-fire on a fast handler.

## Parity checked
The PR preserves #141's queue-unblock guarantee (it doesn't claim a pure refactor, but this is the behavior most at risk of regressing). Confirmed base→head: the bounded release-after-timeout still fires — `DISPATCH_ABORT_GRACE_MS` (3s) release path runs even when the handler ignores the abort signal (test: "still releases the session after the abort grace"), so the queue can't wedge. The change adds cancellation *on top of* the release, rather than replacing it.

### Praise
- The three shared-state fences are placed exactly where they must be — guard immediately after the async boundary, before any mutation, with no intervening await — and the G4 path even rolls back its `backfilledSessions` sentinel on abort so a cancelled backfill can be retried cleanly next turn.
- Converting the handler outcome to a tagged value (`completed`/`failed`/`timeout`) instead of racing a rejecting promise removes the unhandled-rejection footgun the old `Promise.race` had, and the bounded grace keeps the #141 guarantee intact rather than trading it away for cancellation.

### Note (non-blocking)
- Code CI (Vitest/type-check/ESLint/build) is pending at review time — I'm approving the code on its merits; the merge gate should confirm these land green (a real test failure would flip this to request-changes).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #186 (cc-channel-octo)

## Summary

This PR replaces the old "fire-and-forget" dispatch timeout behavior (where a hung turn kept running in the background after the session lock was released) with proper cancellation via `AbortController`/`AbortSignal`. On dispatch timeout, the router now actively aborts the in-flight handler, and the signal is propagated through all downstream side-effect paths — API calls, stream delivery, group context refresh, GROUP.md/THREAD.md fetches, SDK session persistence, and outbound messaging. A bounded 3-second grace period preserves the #141 queue-unblock guarantee even for handlers that ignore cancellation.

The `withDefaultTimeout` helper in `src/octo/api.ts` is also improved: caller-provided signals are now combined with the 30-second default timeout via `AbortSignal.any`, ensuring API calls always have a safety-net timeout even when a dispatch signal is forwarded.

This is a well-designed, well-tested improvement. No P0/P1 issues found.

## Verification

- ✅ **Signal propagation completeness** — traced the dispatch `AbortController` from `src/session-router.ts:207` through `DispatchContext` → `src/index.ts:422` → every downstream consumer. All 6 production API call sites in `src/octo/api.ts` use `withDefaultTimeout(signal)` which combines caller cancellation with the 30s default. All `sendMessage`, `sendReadReceipt`, `sendTyping`, `refreshMembers`, `getChannelMessages`, and `resolveGroupInstructions` calls receive the signal.
- ✅ **Late side-effect fencing** — verified `signal.aborted` checks before every state-mutating operation: `store.appendUser` (`src/index.ts:748`), `store.appendAssistant` (`:1028`), `store.setSdkSessionId` (`:863`), `store.clearSdkSessionId` (`:869`), `groupMdCache.set` (`src/group-md.ts:140`), `threadCache.set` (`:213`), `backfilledSessions.add/delete` (`src/index.ts:720–741`). Backfill key is correctly rolled back on abort so retry works.
- ✅ **Grace period bounded** — `DISPATCH_ABORT_GRACE_MS = 3_000` at `src/session-router.ts:59` caps the wait for cooperative cleanup. If the handler ignores cancellation, the lock releases after grace + apology. `handlerOutcome` is a value-based promise (never rejects), so no unhandled rejection even if the handler settles late.
- ✅ **Typing heartbeat cleanup** — `setInterval` at `src/stream-relay.ts:178` is guarded by `!signal?.aborted` in the callback, and `clearInterval` in the `finally` block at `:296` always fires for returns inside the try block.
- ✅ **Timeout apology not self-cancelled** — `replySafe` at `src/session-router.ts:256` intentionally does NOT pass the dispatch signal; it uses its own 30s default timeout via `withDefaultTimeout(undefined)`.
- ✅ **Test coverage** — session-router tests verify cooperative abort, grace-period expiry with non-cooperative handler, single apology, and fast-handler passthrough. E2E tests verify roster refresh, G4 backfill, and GROUP.md fetch are correctly fenced on timeout. Stream-relay tests verify partial output is dropped and signal is forwarded. API timeout tests verify `withDefaultTimeout` combines signals for all 6 API entry points.
- ✅ **Backfill retry correctness** — e2e test at `src/__tests__/e2e.test.ts:540` proves stale history is not persisted on abort, and a subsequent turn successfully fetches fresh history.

Static analysis only at head `301fdfd3ae5f5d1bb8c16e51fe61c8f9dbc28aad`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. No correctness, security, or functional concerns.

## Data-Flow Trace

- **Dispatch signal** (`src/session-router.ts:207`): `new AbortController()` created per dispatch → passed via `DispatchContext` to handler callback → destructured in `src/index.ts:422` as `const { signal } = abortController`.
- **API calls**: signal forwarded to `postJson`/`getJson`/`requestNoBody` → `withDefaultTimeout(signal)` → `AbortSignal.any([signal, timeoutSignal])` → `fetch({ signal })`. All 6 API entry points covered.
- **Stream relay**: signal passed to `deliver()` at `src/index.ts:1010` → checked at entry, in chunk loop, after stream error, before each segment send, and forwarded to `sendMessage`/`sendTyping`.
- **Group context**: signal passed to `refreshMembers` at `src/index.ts:487` → forwarded to `getGroupMembers` → checked after API return and in catch block.
- **Group/Thread MD**: signal passed to `resolveGroupInstructions` at `src/index.ts:916` → forwarded to `getGroupMd`/`getThreadMd` → checked at entry, after API return, and in catch blocks.
- **SDK query**: `abortController` passed via `sessionOpts` at `src/index.ts:862` → forwarded to SDK `queryAgent` at `src/agent-bridge.ts:319` → SDK terminates query on abort.
- **Session persistence**: `onSessionId` and `onResumeFailed` callbacks guarded by `!signal.aborted` at `src/index.ts:863` and `:869`.
- **Backfill rollback**: `backfilledSessions.delete(backfillKey)` on abort at `src/index.ts:722` and `:741`, allowing retry on the next turn.

## Blind-Spot Checklist

- **C1 — Dual-path parity**: N/A — no paired add/remove or subscribe/unsubscribe changes.
- **C2 — Control-flow ordering**: Clear — abort checks are at consistent points before each side effect; `withDefaultTimeout` applied uniformly across all API helpers; no non-canonical forms.
- **C3 — Authorization boundary**: N/A — no auth changes.
- **C4 — Authorization lifecycle**: N/A — no auth changes.
- **C5 — Build/runtime**: Clear — `AbortSignal.any` is Node.js 20+; consistent with pre-existing usage in the codebase.
- **C6 — Governance/docs**: N/A — no docs/governance changes.

## Cross-Round Blocker Recheck

N/A — first review round.

- **`AbortSignal.any` compatibility** — `withDefaultTimeout` at `src/octo/api.ts:21` uses `AbortSignal.any([signal, timeoutSignal])`, available in Node.js 20+. Consistent with pre-existing usage in the codebase.
- **`onSessionId` / `onResumeFailed` fencing** — both callbacks at `src/index.ts:863` and `:869` guard against `signal.aborted`, preventing stale session ID persistence after cancellation.
- **`teeChunks` generator abort** — `src/index.ts:993` checks `signal.aborted` on each chunk, stopping the stream before the next yield. Combined with `streamRelay.deliver` checking abort in its chunk loop (`src/stream-relay.ts:192`), both sides fence independently.
- **Group context abort path** — `refreshMembers` at `src/group-context.ts:258` checks abort after the API returns and in the catch block, preventing stale roster mutation.
- **Group/Thread MD abort path** — `resolveGroupBranch` and `resolveThreadInstructions` in `src/group-md.ts` check abort at entry, after API return, and in catch blocks, preventing stale cache writes.
- **Dispatch timeout disabled path** — when `dispatchTimeoutMs <= 0`, the handler runs unguarded with a non-aborted controller; all `signal.aborted` checks evaluate to false, preserving existing behavior.
- **No cross-dispatch interference** — each `runHandlerWithTimeout` invocation creates its own `AbortController`, so concurrent sessions are isolated.

## Verdict: APPROVED

Thorough cancellation design with comprehensive signal propagation and well-targeted test coverage. The grace period correctly bounds cleanup time while preserving the #141 queue-unblock guarantee. No issues found.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Signal propagation is complete across all 6 API entry points and all downstream side-effect paths. Grace period is bounded (3s). Test coverage proves correct behavior for cooperative abort, non-cooperative handler (grace expiry), stale data fencing, and backfill retry. Well-executed cancellation PR.

yujiawei (approved):
# Code Review — PR #186 (cc-channel-octo)

**Head SHA reviewed:** `301fdfd3ae5f5d1bb8c16e51fe61c8f9dbc28aad` (against merge-base `4a55e52`)

## Summary

This PR closes the deferred cancellation gap from #141/#142 (tracked in #184): a timed-out dispatch previously only released the session lock while the SDK query kept running, letting a stale turn race a newer one on shared state and keep consuming upstream resources. The change threads an `AbortController` from `runHandlerWithTimeout` into the handler, the Agent SDK query, all Octo HTTP calls, and stream delivery, then fences every late side effect behind `signal.aborted`. A bounded 3s grace preserves #141's queue-unblock guarantee for handlers that ignore cancellation.

The implementation is clean, well-commented, and the fencing is applied consistently at every persistence and I/O boundary. I verified the change end-to-end and it satisfies the acceptance criteria of #184.

## Verification performed

- `npm run type-check` — pass
- `npm run lint` — pass
- `npm run build` — pass
- Targeted tests (`session-router`, `stream-relay`, `api-timeout`, `group-md`, `thread-md`, `agent-bridge-query`, `e2e`) — 246 passing

## Correctness review

**`session-router.ts` `runHandlerWithTimeout`** — Correct. The rejection→value conversion (`handler(...).then(ok, err)`) removes the old orphan-promise unhandled-rejection risk. The timeout branch aborts with a per-invocation `Error`, warns, races the same memoized `handlerOutcome` against a 3s grace, then sends the apology. The apology (`replySafe`) intentionally carries no abort signal, so a cancelled dispatch can still deliver "⚠️ 处理超时" (it relies on `octo/api`'s own 30s bound). `timeoutHandle`/`graceHandle` are both cleared. No wedge path.

**`index.ts` handler** — Late side effects are fenced correctly:
- `store.setSdkSessionId` / `clearSdkSessionId` guarded by `!signal.aborted`, so a stale turn cannot overwrite a newer session pointer.
- `appendUser` and `appendAssistant` are both gated (append-assistant is skipped by the early `return` after `deliver`), so no out-of-order history write from a stale turn.
- roster refresh, G4 backfill, and GROUP.md/THREAD.md resolution each re-check `signal.aborted` before mutating caches/cursors; `backfilledSessions` is rolled back on abort so a cancelled backfill can be retried.
- `teeChunks` stops yielding on abort, and the SDK abort error is swallowed uniformly in `stream-relay`'s catch and the outer handler catch.

**`octo/api.ts` `withDefaultTimeout`** — `AbortSignal.any([signal, timeoutSignal])` correctly delivers the stated goal: a caller signal now *also* retains the 30s HTTP bound (previously a caller-signal call dropped the default). Behavior on Node ≥22 (enforced via `engines`) is sound.

## Findings (all advisory — none blocking)

- **P2 (nit, pre-existing)** — `src/octo/api.ts:22` `withDefaultTimeout` creates an unref'd `AbortSignal.timeout` timer per request that lives the full 30s even when the request resolves in ms. Bounded and unref'd, so not a leak, but under high concurrency it accumulates dormant timers. Pre-existing pattern; a `clearTimeout`-on-settle wrapper would tidy it if you revisit this file.
- **P2 (nit)** — `src/index.ts` (post-`appendUser` path): if the timeout fires *after* `store.appendUser` but before the assistant reply is produced, the local store keeps a user turn with no paired assistant turn. This is benign here because history is injected as a rendered **text prefix** (`renderTurn`), not as structured alternating message roles — so it cannot produce the consecutive-`user`-role API error that would occur in a role-structured backend. Worth a one-line comment noting the reliance on text-prefix rendering, in case history assembly ever moves to structured roles.

## Cross-file / coverage notes

- Confirmed the SDK (`@anthropic-ai/claude-agent-sdk`) exposes `Options.abortController` and aborts the in-flight query on `signal` — the plumbing in `agent-bridge.ts` is correct, not a no-op.
- Not assessed from the diff alone: production `dispatchTimeoutMs` tuning, the SDK's actual subprocess/subagent teardown latency on abort (relevant to whether the 3s grace is usually enough in production), and Octo server-side idempotency for segments that were delivered before the abort fired.

ningzimu:
Maintainer action is now the only remaining gate on this PR.

- head `301fdfd3` is mergeable and `reviewDecision=APPROVED`;
- there are no unresolved review threads;
- Check Sprint, PR title, label, and code-review checks pass;
- the actual fork workflows have not run: CI, Secret Scan, Dependency Review, and History Check are all `action_required`.

Please approve the pending fork workflow runs so the required code and security checks can execute against this head.

PR #193, opened 2026-08-26
fix(gateway): re-arm reconnect when a kick lands inside the refresh cooldown

king378:
## Problem

The bot can go **permanently offline** (until a manual restart) after being kicked by the IM server. Seen in production: a bot stayed offline ~2 days. The journal shows two kicks 28s apart:

```
Kicked by server → Attempting token refresh... → Token refreshed, reconnecting... → WS connected
(28s later) Kicked by server → Token refresh cooldown (33s remaining) → WS disconnected   ← never recovers
```

## Root cause

1. A `Kicked by server` disconnect sets the socket's `needReconnect = false` (`octo/socket.ts` `onDisconnect` / the `reasonCode === 0` path), so the socket will **not** self-reconnect. The only recovery path is `attemptTokenRefresh()` (register `force_refresh` + fresh socket), triggered via `onError`.
2. `attemptTokenRefresh()` has a 60s cooldown (`REFRESH_COOLDOWN_MS`). If a second kick lands inside that window, the cooldown branch logs `Token refresh cooldown (Ns remaining)` and returns **without scheduling anything**.
3. Net state: socket won't reconnect + refresh skipped + nothing pending ⇒ stuck offline forever.

The API-level heartbeat can't save it either — `/v1/bot/heartbeat` keeps succeeding while the WS is kicked, so `heartbeatFailCount` never trips the fallback refresh.

## Fix

In the cooldown branch, arm a single `refreshRetryTimer` that re-invokes `attemptTokenRefresh()` once the cooldown expires (+250ms guard). The timer is:
- deduped (only armed if not already pending),
- cleared when a real refresh proceeds (so it can't fire a redundant refresh on a fresh connection),
- cleared in `stop()` (clean shutdown, no dangling timer).

This turns a transient kick-inside-cooldown into an automatic recovery instead of a permanent outage.

## Testing

- New regression test in `gateway.test.ts`: after a second kick inside the cooldown, advancing past the cooldown reconnects with **no further kick event** (before the fix, `registerBot` stayed flat — offline forever).
- Existing `Token refresh cooldown` tests still pass (a second kick still does not refresh *immediately* — it's deferred).
- `npm run type-check` and `npm run lint` clean; full gateway suite green.


Jerry-Xin (changes requested):
## Verdict: REQUEST_CHANGES

The fix is the right shape: a deduplicated, self-clearing retry timer turns a kick-inside-cooldown from a permanent offline state into an automatic recovery. The root-cause analysis is accurate (a kick sets `needReconnect=false` on the socket, so `attemptTokenRefresh()` is the only recovery path) and the new regression test covers exactly the production scenario. One blocking race in the shutdown path.

## 🔴 Blocking on this head

### 🔴 1. Retry timer can resurrect the gateway after `stop()` completes

The new timer is armed but is only cleared in `stop()` at `src/gateway.ts:199`, which runs **after** the async handler-drain wait at `src/gateway.ts:190-192` (up to `drainTimeoutMs` = 10s). If the cooldown retry expires during that drain window, the callback invokes `attemptTokenRefresh()`, which has no `_draining` guard (only `handleMessage` checks it, `src/gateway.ts:333`). The reachable interleaving:

1. Kick inside cooldown → `refreshRetryTimer` armed.
2. `stop()` called with in-flight handlers → `_draining = true`, then awaits drain (up to 10s).
3. Timer fires mid-drain → `attemptTokenRefresh()` proceeds: nulls `this.socket` (~`src/gateway.ts:375`), then awaits `registerBot(...)`.
4. `stop()` resumes: clears the (already-fired) timer, sees `this.socket === null` so skips disconnect, releases the lock file, returns.
5. `registerBot` resolves after `stop()` returned → new socket created + `connect()` + `startHeartbeat()` (`src/gateway.ts:394-396`) ⇒ a live, connected, heartbeating gateway **after shutdown**, with the lock already released — so a fresh instance can start and double-register while the zombie stays connected.

Before this PR nothing could fire during the drain wait, so this is a regression introduced by the new timer, in the exact lifecycle domain this PR is fixing.

Suggested fix (both cheap):
- Move the `refreshRetryTimer` clear to the very top of `stop()`, before the first `await` (right after `_draining = true`).
- Add a `_draining` guard at the top of `attemptTokenRefresh()` (and/or in the timer callback) as belt-and-suspenders. Note the same guard also mitigates the pre-existing variant where a refresh is already in flight when `stop()` begins — consider re-checking `_draining` after `registerBot` resolves before connecting.

A regression test would pin this down: arm the cooldown retry, start `stop()` with an active handler so the drain wait spans the timer expiry, advance timers, and assert no `registerBot` call / socket / heartbeat is started after `stop()` resolves.

## 🟡 Non-blocking notes

- If the retry-driven refresh itself fails (`registerBot` throws), nothing is re-armed, and the API heartbeat keeps succeeding while the WS is dead, so `heartbeatFailCount` never trips. That stuck-after-failed-refresh mode predates this PR (first-kick refresh failure has the same hole), so it doesn't block here — but it's a natural follow-up: consider re-arming a bounded retry on refresh failure too.

## ✅ Highlights

- Dedup (`if (!this.refreshRetryTimer)`) guarantees exactly one retry even under repeated kicks inside the cooldown, and the earliest armed timer also expires earliest — correct ordering.
- Clearing the timer when a real refresh proceeds prevents a redundant forced refresh on top of a fresh connection.
- The regression test asserts both halves: no immediate refresh on the second kick, and self-recovery after the cooldown with **no further kick event**.
- Full gateway suite (23 tests), `type-check`, and `lint` verified green on this head.

yujiawei (approved):
# Code Review — PR #193 (cc-channel-octo)

**Verdict: Approve.** The diagnosis is correct, the fix is minimal and correctly scoped, and the new regression test genuinely proves the behavior change. Everything below is advisory (P2) or pre-existing outside this diff — none of it should block this from landing, and the availability bug it fixes is a real production outage.

---

## What I verified independently

Checked out `4dc2354` and ran the toolchain locally:

| Check | Result |
|---|---|
| `npm run type-check` | clean |
| `npm run lint` (`--max-warnings 0`) | clean |
| `npm test` | **1281 passed / 66 files**, 0 failures |
| New test against the *unpatched* `gateway.ts` | **fails** — `AssertionError: expected 2 to be 3` |

That last row matters: I reverted `src/gateway.ts` to `HEAD~1`, kept the new test, and confirmed it fails. The regression test is real, not vacuous.

⚠️ **Note for maintainers:** the `CI` workflow (type-check / lint / test) has **not run on this head SHA**. `GET /commits/4dc2354/check-runs` returns only `add-to-project`, `check-sprint`, `label`, and `pr-title-lint`. This is a cross-repo (fork) PR, so workflow runs need maintainer approval. Please approve the workflow run so CI produces an independent signal — the table above is my local run, not CI's. (`check-sprint` is also currently failing; that is a project-board field, not a code issue.)

I also read `src/octo/socket.ts` in full to check the two reconnect mechanisms don't fight.

## Scope / spec compliance

Clean. `+64 / -2` across exactly two files, and every claim in the PR description maps to code:

- single deduped timer armed in the cooldown branch — `src/gateway.ts:351-356`
- cleared when a real refresh proceeds — `src/gateway.ts:362-365`
- cleared in `stop()` — `src/gateway.ts:199-202`

No unrelated changes, no scope creep, no dead code.

## Why the state machine holds up

I tried to break it and mostly couldn't. The two things I most expected to be wrong are actually fine:

**1. The timer can't fire on a healthy socket.** This was my main worry — an armed timer surviving a self-healed connection and then tearing down a good socket with a spurious `force_refresh`. It can't happen. `createSocket`'s handler only refreshes on `Kicked` / `Connect failed` (`src/gateway.ts:322`), and *every* socket path emitting those messages sets `needReconnect = false` first:

- `socket.ts:452-455` — rapid-disconnect → `Connect failed: rapid disconnect detected`
- `socket.ts:755-760` — CONNACK `reasonCode === 0` → `Kicked by server`
- `socket.ts:762-766` — CONNACK other → `Connect failed: reasonCode=N`
- `socket.ts:857-862` — DISCONNECT frame → `Kicked by server`

The one path that *does* keep `needReconnect = true` (`socket.ts:736-738`, `CONNACK salt too short`) uses a message that matches neither filter, so it never arms the timer. So the timer is armed only when the socket is genuinely dead. The two reconnect mechanisms don't overlap.

**2. The timer can't double-fire or re-enter.** A proceeding refresh clears the timer (`:362-365`) *before* setting `isRefreshing` (`:367`), so the timer can never fire mid-refresh. The `if (!this.refreshRetryTimer)` guard keeps N kicks inside one cooldown to a single armed timer, and since `lastRefreshTime` is unchanged, the deadline stays correct regardless of which kick armed it.

---

## Findings (all P2 — advisory, non-blocking)

### P2-1 — A *failed* retry still wedges the bot offline (the same bug, one branch over)

`src/gateway.ts:376-402`. When `registerBot()` rejects, the socket has already been nulled at `:376-379` and the catch does only:

```ts
} catch (err) {
  console.error('Token refresh failed:', String(err));
  this.startHeartbeat(); // Q33: Restore heartbeat on failure for self-healing
}
```

No socket, no armed timer, nothing scheduled. Recovery now depends entirely on `heartbeatFailCount` reaching 3 — but the PR description's own root-cause analysis explains why that's unreliable: `/v1/bot/heartbeat` can keep succeeding while the WS is down. If `force_refresh` fails for a reason that doesn't also break the heartbeat endpoint (a 500 or a rate-limit on that specific path), you land in exactly the original symptom.

I confirmed this with a scratch test: kick → refresh → kick inside cooldown → retry fires → `registerBot` rejects once → then 10 minutes of healthy heartbeats. `registerBot` call count stayed flat at 3 and no new socket was ever constructed.

**This is pre-existing** — `HEAD~1` behaves identically — so it is explicitly *not* a reason to hold this PR, which strictly reduces the number of ways to get wedged. But it's the same failure class, and closing it is cheap: re-arm the timer in the catch branch (bounded, with backoff) instead of relying on the heartbeat.

Related: `if (this.isRefreshing) return;` (`:341`) also drops a recovery signal without arming anything. Benign while the in-flight refresh succeeds; combined with the above it's the path that leads there.

### P2-2 — The retry can fire during `stop()`'s drain window and resurrect the gateway

The PR says the timer is "cleared in `stop()` (clean shutdown, no dangling timer)". True for the *timer* — but `stop()` drains in-flight handlers **first** (`:188-196`, up to `drainTimeoutMs`, default 10s) and only clears the timer afterward at `:199-202`. A retry landing inside that window starts an unattended refresh that `stop()` can neither cancel nor await, because the callback nulls its own handle before doing async work:

```ts
this.refreshRetryTimer = setTimeout(() => {
  this.refreshRetryTimer = null;      // handle gone before the async work starts
  void this.attemptTokenRefresh();
}, remainingMs + 250);
```

Scratch test, driving `stop(activeHandlers, 90_000)` with a slow handler: socket constructions went **2 → 3** and `registerBot` **2 → 3** *during shutdown* — a new socket and a new heartbeat interval created after `stop()` was underway, past the point where the lock is released.

Real-world impact is low (the signal path ends in `process.exit(0)`, and the 30s heartbeat interval already has the same window), so P2. One-line hardening: `if (this._draining) return;` at the top of `attemptTokenRefresh()`.

### P2-3 — `Date.now()` is not monotonic; a backward clock step becomes a long sleep

`src/gateway.ts:343-345`:

```ts
const now = Date.now();
const remainingMs = this.REFRESH_COOLDOWN_MS - (now - this.lastRefreshTime);
```

The cooldown *comparison* was already wall-clock-based, but the PR turns skew into an actual delay. An NTP step backward of Δ makes `now - lastRefreshTime` negative, so `remainingMs = 60_000 + Δ` — a one-hour correction schedules the recovery an hour out. Cheap guard:

```ts
const remainingMs = Math.min(
  this.REFRESH_COOLDOWN_MS - (now - this.lastRefreshTime),
  this.REFRESH_COOLDOWN_MS,
);
```

(A forward jump conversely allows early churn — pre-existing. `performance.now()` / `process.hrtime` would fix both.)

### P2-4 — Two instances sharing a token now ping-pong forever instead of self-limiting

`forceRefresh: true` at `:384` with no ceiling, backoff, or jitter. If the server keeps kicking — most plausibly two deployments sharing one bot token, where each connect kicks the other — the old code was self-limiting (the loser went quiet after two kicks in 60s). Now both sides re-register and rotate their IM token every ~60s indefinitely.

For the bug being fixed this trade is clearly correct: permanent silence is worse than 1 refresh/min. But it's a behavior change worth knowing about operationally, and worth a bounded retry count or a "reset only after the connection has been stable for N seconds" condition later. The `startStableTimer()` at `socket.ts:494-502` is a natural hook.

### P2-5 — Test covers the happy path only

`src/__tests__/gateway.test.ts:329-359` proves the important thing, but `WKSocket` is fully mocked, so `needReconnect` / `scheduleReconnect` / `rapidDisconnectCount` are never exercised, `registerBot` always resolves, and the test stops right after the retry fires. Not covered: refresh rejection (P2-1), `stop()` before the timer fires, a third kick hitting the dedup guard, and "no dangling timer after `stop()`". The last two are three lines each and would lock in the behavior the PR description promises.

### Nit — the log line over-promises

`:347` unconditionally says `— scheduling retry`, but when `refreshRetryTimer` is already set (`:351`) nothing is scheduled. On the 2nd and 3rd kick inside one cooldown the log claims an action it didn't take. Move the string inside the `if`, or say `retry already pending`.

### Nit — docs

`ARCHITECTURE.md:277` still reads *"Token refresh: 60-second cooldown prevents refresh storms."* — now incomplete; the cooldown also self-schedules a recovery. One sentence.

---

## Out of scope, but flagging for the human reviewer (this PR carries `needs-human-review`)

**The bot token can reach logs on a registration failure.** This is *not* introduced here and should be a separate PR — but this change makes the code path at `gateway.ts:398` (`console.error('Token refresh failed:', String(err))`) execute more often, so it's worth raising now.

`postJson()` surfaces the raw, uncapped, server-controlled error body:

```ts
// src/octo/api.ts:79-81
const text = await response.text().catch(() => "");
throw new Error(`Octo API ${path} failed (${response.status}): ${text || response.statusText}`);
```

The repo already treats this exact threat as real — `api.ts:174-183` sanitizes it for one endpoint, with the comment *"server may echo request headers (incl. Authorization bearer token) on some error responses"*:

```ts
const sanitized = text
  .slice(0, 200)
  .replace(/Bearer\s+\S+/gi, "Bearer ***")
  .replace(/(authorization"?\s*[:=]\s*"?)[^"\s,}]+/gi, "$1***");
```

That hardening was applied to `/v1/bot/upload/credentials` only. `postJson()` — which carries `registerBot`, `sendHeartbeat`, `sendMessage`, `typing`, `readReceipt` — never got it, and neither did `getJson()` (`api.ts:283-284`). Lifting the same three lines into a shared helper closes it. Worth its own issue.

## Coverage — what I could *not* verify

- **No live verification.** Everything is fake-timer tests against a mocked `WKSocket`; the fix is unexercised against a real WuKongIM server. Given this reproduced in production over ~2 days, a staging soak with two induced kicks 28s apart would be the real confirmation.
- **CI has not run on this SHA** (see above) — my local run is the only build signal on `4dc2354`.
- **Backend behavior unknown.** Whether Octo rate-limits or throttles `force_refresh` changes the P2-4 analysis; I have no visibility into that.
- **Not reviewed:** the multi-bot orchestrator's shutdown path in `index.ts`, which is the other caller of `stop()` and therefore the other side of P2-2.

## Bottom line

Correct root cause, minimal fix, real regression test, no scope creep. Ship it. The highest-value follow-ups, in order: the token-in-logs sanitization (separate PR), then P2-1 (re-arm on refresh failure), then P2-2 and P2-3, which are one-liners.

yujiawei (changes requested):
# Code Review — PR #193 (cc-channel-octo) — revised verdict

**This supersedes my earlier Approve on the same head SHA (`4dc2354`). I am changing my own verdict to Request Changes.** A concurrent review landed while I was writing mine, flagging the shutdown race that I had filed as a non-blocking P2-2. Re-examining it, their severity call is right and mine was wrong — and I have a reproduction of my own that I under-weighted. Recording the correction explicitly rather than leaving a stale Approve on the PR.

## What changed my mind

I had rated the resurrect-after-`stop()` race P2 on the reasoning that the signal path ends in `process.exit(0)`, so a zombie socket dies with the process anyway. Two things break that reasoning:

**1. `process.exit(0)` only covers the single-bot signal path.** `stop()` is also driven by the multi-bot orchestrator with `handleSignals: false`, where the gateway does *not* own the exit. I listed that path as unreviewed in my own coverage note, then leaned on an assumption that only holds outside it. That was not a sound basis for the downgrade.

**2. The consequence is lock invalidation, not just a leaked socket.** This is the part I missed. Walking `attemptTokenRefresh()` against `stop()`:

- `src/gateway.ts:378` nulls `this.socket`, *then* awaits `registerBot()`.
- `stop()` resumes from the drain, sees `this.socket === null`, skips `disconnectAndWait()` (`:203-206`), and calls `releaseLock()` (`:207`).
- `registerBot` resolves afterwards → `:394-396` creates a socket, connects, and restarts the heartbeat.

Net: a live, connected, heartbeating gateway with the lock file **already released**. A fresh instance can then acquire the lock and register the same bot while the zombie is still connected — two live WS sessions on one uid, which is precisely the duplicate-registration condition that makes the server kick. The failure mode is the one this PR exists to fix.

**3. The race is positively correlated with the bug being fixed**, which is what tips it over the line for me. `refreshRetryTimer` is only ever pending in the minute after a kick-inside-cooldown. The operator action that follows a kick storm is a restart. So shutdown is *most* likely to happen exactly when a retry is armed — this isn't an independent-probability tail event, and reasoning about it as one is how I got the severity wrong.

## My reproduction

I had already observed this before filing it as P2 and did not weight it correctly. Scratch test on `4dc2354`: arm the cooldown retry, then call `stop(activeHandlers, 90_000)` with a handler resolving at 90s so the drain wait spans the timer expiry.

```
PROBE B sockets: 2 -> 3   registerBot: 2 -> 3
```

A third `WKSocket` was constructed and a third `registerBot` issued **during** `stop()`. The resurrection is not theoretical on this head.

## Why the ordering makes it reachable

`stop()` sets `_draining = true`, then awaits the handler drain (`:188-196`, default `drainTimeoutMs` 10s), and only clears `refreshRetryTimer` afterwards at `:199-202`. `attemptTokenRefresh()` has no `_draining` guard — `handleMessage` (`:333`) is the only `_draining` consumer. So the clear is unreachable for any timer that expires inside the drain window.

The pre-existing analogue — the 30s heartbeat interval, also still live during the drain — additionally requires three consecutive heartbeat failures to reach `attemptTokenRefresh()` (`:435-438`). The new timer calls it unconditionally. So while the lifecycle hole isn't strictly new, this change widens it from "needs a triple heartbeat failure" to "fires on schedule", in the exact subsystem the PR is repairing.

## To unblock

Either fix is sufficient; both are cheap and worth taking together:

1. Move the `refreshRetryTimer` clear to the top of `stop()`, above the first `await` (immediately after `_draining = true`).
2. Add `if (this._draining) return;` at the top of `attemptTokenRefresh()`. Also worth re-checking `_draining` after `registerBot()` resolves and before `:394-396` connects, which closes the pre-existing variant where a refresh is already in flight when `stop()` begins.

A test would pin it: arm the retry, start `stop()` with an active handler spanning the expiry, advance timers, assert no `registerBot` / socket / heartbeat after `stop()` resolves.

## Everything else in my earlier review stands

The diagnosis, the scope, and the regression test are all good, and I verified the fix's core state machine is sound — the timer cannot fire on a healthy socket, and it cannot double-fire or re-enter. My verification of `4dc2354` (`type-check` clean, `lint` clean, 1281/1281 tests passing, and the new test confirmed failing on the unpatched `gateway.ts` with `expected 2 to be 3`) is unchanged. P2-1, P2-3, P2-4, P2-5, the nits, and the out-of-diff token-in-logs issue at `api.ts:79-81` all remain as written.

The `CI` workflow still has not run on this head SHA — this is a fork PR and needs maintainer approval for workflow runs. Please approve it so CI gives an independent signal alongside my local run.

This is a two-line fix on a good PR. Once the drain-window race is closed I expect to approve.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #193 (cc-channel-octo)

## Summary

This PR fixes a liveness wedge in the gateway's token-refresh recovery. A "Kicked by server" event sets `WKSocket.needReconnect=false`, so the socket never self-reconnects — the token refresh is the only recovery path. If a second kick lands inside the 60s refresh cooldown, the refresh was deferred and nothing was rescheduled, leaving the bot offline until a manual restart. The change adds a single-shot `refreshRetryTimer` armed from the cooldown branch (`src/gateway.ts:352`), cleared in `stop()` and when a real refresh proceeds, plus a regression test that advances fake timers past the cooldown and asserts the self-scheduled reconnect.

The mechanism is sound and the fix is minimal. One non-blocking P2 below (a pre-existing shutdown race that the new timer slightly widens).

## Verification

Static analysis only at head `4dc23547f1756b8c94567693e9f9e77105830473`; build and tests not executed in this environment.

- ✅ **Core claim holds against the real socket** — both kick paths set `needReconnect=false`, null the ws, and emit exactly `"Kicked by server"` (`src/octo/socket.ts:756-761` CONNACK reasonCode 0, `:853-863` DISCONNECT packet); the close handler's stale guard (`:431`) prevents phantom reconnects. After a kick, nothing but a token refresh brings the WS back.
- ✅ **Timer lifecycle is complete** — one arm site with an idempotency guard (`src/gateway.ts:351-355`); cleared in `stop()` (`:199-202`), cleared when a real refresh proceeds (`:362-364`), self-nulled in the callback (`:353`). No other references.
- ✅ **At fire time the cooldown has always passed and no other refresh can be in flight** — delay is `remainingMs + 250`; after a kick the socket is dead, and the only other trigger (3 consecutive heartbeat failures at 30s intervals) needs ≥90s, longer than the ≤60s armed window. `registerBot`/`sendHeartbeat` are bounded by a 30s timeout (`src/octo/api.ts:19`), so `isRefreshing` cannot latch.
- ✅ **No refresh storm** — at most one armed retry per instance; sustained kicks converge to ≤1 refresh per 60s (bounded), strictly better than the pre-PR permanent wedge.
- ✅ **Regression test is genuine** — it drives the real `attemptTokenRefresh` path (only socket/API mocked), the mocked error string matches the real socket emit, and the two pre-existing cooldown tests still hold because `gw.stop()` clears the armed timer.

## Findings

No P0/P1 issues; one P2 item below.

### P2 — Retry timer (and in-flight refreshes) can resurrect a socket after stop() (`src/gateway.ts:352`)

The armed callback calls `attemptTokenRefresh()` without checking `_draining`, and `attemptTokenRefresh()` only guards on `isRefreshing` (`:341`). If the timer fires moments before `stop()` reaches the `clearTimeout` at `:199-202` — or a kick/heartbeat-initiated refresh is mid-flight during `stop()` — the refresh completes after teardown and runs `createSocket()`/`connect()`/`startHeartbeat()` on a stopped gateway: the lock is already released, and in multi-bot mode a bot removed via `BotManager.removeBot` keeps a live socket consuming messages until process restart. `registerBot` has a 30s window (`src/octo/api.ts:19`), so the overlap is not purely theoretical. This race existed before this PR for kick/heartbeat-initiated refreshes; the new timer adds one more refresh initiator.

```ts
// cheapest fix — top of attemptTokenRefresh():
if (this._draining) return;
```

Ideally also pass an `AbortSignal` into `registerBot` that `stop()` aborts, so an in-flight refresh cannot outlive shutdown.

## Human-verify

1. The server-side kick policy (why the server kicks a connected bot — e.g. duplicate-session takeover) is outside this checkout. Under a persistent duplicate using the same bot token, this fix converges to one re-register per minute instead of wedging offline; please confirm that steady state is acceptable operationally. This is **not** a merge blocker for this PR.

- `disconnectAndWait` on an already-dead socket returns immediately (`oldWs` null check, `src/octo/socket.ts:347-357`) — the timer-triggered refresh cannot hang on the dead socket.
- Heartbeat generation guard still covers timer-triggered refreshes (`startHeartbeat()` bumps `heartbeatGen`).
- No unhandled rejection from the callback — `attemptTokenRefresh` catches all refresh errors; statements before the `try` cannot throw.
- `lastRefreshTime` is not reset across `stop()`/`start()`, but production always uses a fresh `OctoGateway` per bot (`src/index.ts:267`, `BotManager.addBot` → start), so no cross-run staleness.
- Nothing parses the changed console-log string.

## Verdict: APPROVED

The fix correctly closes the kick-inside-cooldown wedge with a bounded, single-shot retry; the socket-level semantics it relies on are verified in `src/octo/socket.ts`, and the regression test exercises the real gateway path. The P2 is a pre-existing shutdown race (slightly widened, cheap to harden) and does not block.

---

## 数据流回溯（Octo 补充段）

- `remainingMs` ← `REFRESH_COOLDOWN_MS - (Date.now() - lastRefreshTime)`；`lastRefreshTime` 在**真实刷新开始时**赋值（`src/gateway.ts:368`），故定时器触发时刻 = 上次刷新开始 + 60.25s，触发时 cooldown 必然已过 ✓ 数据真流到消费点。
- 定时器回调 → `attemptTokenRefresh()` → `isRefreshing` 守卫（`:341`，触发时恒为 false，证明见 Verification 第 3 条）→ cooldown 检查（必过）→ `registerBot(forceRefresh, 同一 botToken)` → `createSocket(reg.ws_url/robot_id/im_token)` → `connect()` + `startHeartbeat()`。全链路无空值/提前 return/门控短路。
- `onError('Kicked by server')` ← `src/octo/socket.ts:760` / `:862` 两条真实 kick 路径（非测试 hand-feed）；kick 后 `needReconnect=false` 且 `this.ws=null`，close 事件被 `:431` stale guard 丢弃，确认 socket 层无任何自愈路径，刷新是唯一恢复通道——与 PR 声称一致。

## 盲点 checklist（C1–C6，security_sensitive 全项）

- **C1 双路径 parity**：arm↔clear 逐对 grep（`refreshRetryTimer` 全仓仅 4 处引用）——arm 1 处带幂等守卫；clear 3 处（stop / 真实刷新前进 / 回调自置空）。两端一致，clear。
- **C2 control-flow ordering / 复用**：`attemptTokenRefresh` 3 个 call-site（kick/Connect-failed `:323`、心跳三连败 `:438`、新定时器 `:354`）时序组合穷举：定时器触发时 isRefreshing=true 在当前触发拓扑下不可达（心跳需 ≥90s > 窗口 ≤60s）；cooldown 内二次 kick 幂等不重复 arm；定时器触发的刷新后 60s 内再 kick 正确 arm 下一窗口。clear。
- **C3 授权边界**：N/A——无新 endpoint/tool/凭证暴露，重连复用同一 `botToken`，无能力面变化。
- **C4 授权生命周期/级联**：N/A——无授权状态或容器-成员关系改动。
- **C5 build≠runtime**：已做运行时推演（定时器→刷新→重连全链），回归测试走真实 `OctoGateway` 逻辑（仅 mock socket/API 边界，非 hand-feed 绕过）；本环境未实际跑测试（审查约束），已在 Verification 诚实声明。
- **C6 治理文档自洽**：N/A——diff 不含治理/策略文档。
- **STRIDE 安全专项**：Spoofing N/A（无身份来源变化）；Tampering N/A（无解析/拼接变化）；Repudiation N/A（仅日志文案变化）；Info-disclosure N/A（无读取面扩大）；DoS——重试单次+冷却，刷新率 ≤1/60s，无放大；EoP N/A（无能力暴露变化）。无安全发现。

## 额外发现（pre-existing，非本 PR 引入，不阻塞）

刷新失败且 API 健康时（如 token 被吊销 → 401）无任何恢复调度：`src/gateway.ts:398-399` catch 只恢复心跳，而心跳自愈只在 API 不可达（心跳失败累计 3 次）时触发。kick 路径在本 PR 前即如此，本 PR 的定时器路径继承该缺口。建议后续补一个失败退避重试（与本次修复解耦）。

## 跨轮 blocker 复检（R6）

N/A——本 issue 首轮审查，无历史未解决 blocker。

[Octo-Q] verdict: APPROVE — 修复正确、有界（单次重试、≤1 次/60s）、回归测试覆盖真实路径；唯一 P2 为既有 shutdown 竞态的轻微放大，非阻塞。

yujiawei:
Addendum to my Request Changes — my verdict is unchanged, but I closed the blind spot I flagged there (I had not reviewed the multi-bot orchestrator path, the other caller of `stop()`), and what I found makes the drain-window race materially worse than I described. Posting as a comment rather than another formal review since the verdict doesn't move.

**The race is not shutdown-only. It's reachable from routine hot-reload, in a process that keeps running.**

`config-watcher.ts:51` calls `manager.removeBot(id)` when an operator edits config to drop a bot. That path is:

```
config-watcher.ts:51  manager.removeBot(id)
  → bot-manager.ts:163  await bot.shutdown()
    → index.ts:370      await gateway.stop(activeHandlers)
  → bot-manager.ts:164  crossUnregisterOnRemove(bot, remaining)
```

No `process.exit()` anywhere — the other bots stay up. So the "the process dies anyway" reasoning (which I used myself in my first review, and which underpins rating this P2) does not apply here at all. `index.ts:267` also constructs the gateway with `handleSignals: !multi`, so in multi-bot mode the gateway never registers its own signal handler and never calls `process.exit(0)`.

Full interleaving on a removed bot:

1. Bot was kicked inside the cooldown → `refreshRetryTimer` armed, ≤60s out.
2. Hot-reload removes the bot → `stop()` sets `_draining = true`, then **awaits the handler drain** (`gateway.ts:188-196`, default 10s) *before* clearing the timer at `:199-202`.
3. Timer fires mid-drain. `attemptTokenRefresh()` has no `_draining` guard, so it proceeds: nulls `this.socket` (`:378`), then awaits `registerBot()` — which has a 30s budget (`api.ts:19`).
4. `stop()` resumes, sees `this.socket === null`, skips `disconnectAndWait()` (`:203-206`), calls `releaseLock()` (`:207`), returns.
5. `crossUnregisterOnRemove` runs — siblings drop this bot's `robotUid`.
6. `registerBot` resolves → `:394-396` creates a socket, connects, restarts the heartbeat.

The removed bot ends up with a **live WebSocket consuming messages**, its lock released, and siblings no longer recognising its `robotUid`. That last part is the same failure the ordering comment at `bot-manager.ts:157-160` exists to prevent:

> If we unregistered before shutdown, the bot's in-flight handlers (still draining for up to the gateway drain timeout) could emit replies that siblings — having already dropped its robotUid — would treat as user input, reopening the very loop the guard prevents.

That comment already identifies the drain window as dangerous; the new timer puts an unguarded `registerBot` + `connect()` inside it. There is no way to stop the resulting zombie short of restarting the process.

Two things worth noting about the window itself: it isn't "moments before the `clearTimeout`" — it's the **entire drain duration**, up to `drainTimeoutMs` (10s default), because the clear is sequenced after the `await`. And it correlates with the bug being fixed: a retry is only ever pending in the minute after a kick, which is exactly when an operator is most likely to be reconfiguring or restarting.

My reproduction on `4dc2354`, driving `stop(activeHandlers, 90_000)` with a handler spanning the timer expiry:

```
PROBE B sockets: 2 -> 3   registerBot: 2 -> 3
```

The fix is unchanged and still two lines — clear the timer before the first `await` in `stop()`, and add `if (this._draining) return;` at the top of `attemptTokenRefresh()`. Given the `removeBot` path, I'd also suggest re-checking `_draining` after `registerBot()` resolves and before `:394-396` connects, since a 30s API budget can outlive a 10s drain on its own even without the new timer.

Everything else in my review stands, and the core of this PR is still good work — the retry state machine itself is sound and I couldn't break it.

king378:
Thanks for the careful review — you're right, the retry timer opened a drain-window resurrection race. Pushed `9f66c6e` addressing all three points:

- Moved the `refreshRetryTimer` clear to the **top of `stop()`**, before the first `await` (right after `_draining = true`).
- Added `if (this._draining) return;` at the top of `attemptTokenRefresh()`.
- Re-check `_draining` **after `registerBot()` resolves and before `connect()`/heartbeat**, which also closes the pre-existing in-flight-refresh variant (a refresh already running when `stop()` begins — e.g. the 30s API budget outliving a 10s drain).

Added a regression test reproducing exactly the interleaving you described (arm the retry, then `stop(activeHandlers, 90_000)` with a handler spanning the timer expiry): it **fails on the previous head with `expected 3 to be 2`** (a 3rd `registerBot`/socket created during `stop()`), and **passes after the fix**. Full gateway suite (24 tests), `type-check`, and `lint` are green.

Also confirmed the `removeBot()` hot-reload path (`config-watcher.ts:51` → `stop()` with no `process.exit`) is covered by the same guards.

Separately: `Check Sprint` is an internal-process check unrelated to this diff, and `CI` is still awaiting maintainer approval to run on this fork PR — an approval would give an independent signal alongside the local runs. Thanks again!

yujiawei (approved):
# Code Review — PR #193 (cc-channel-octo)

Re-review on head `9f66c6e` (the previous round was on `4dc2354`). Approve. All three blocking points from the last round are addressed, and I verified each one independently rather than taking the fix note at face value. Everything remaining below is P2 / pre-existing and should not hold this up — the availability bug being fixed is a real multi-day production outage.

---

## 1. Spec compliance

**Spec: ✅**

The PR description makes three claims about this head. Each maps to code, and nothing else changed (`+139 / -3`, exactly `src/gateway.ts` and `src/__tests__/gateway.test.ts`):

| Claim | Code | Status |
|---|---|---|
| Clear `refreshRetryTimer` at the top of `stop()`, before the first `await` | `src/gateway.ts:187-196` (immediately after `_draining = true` at `:185`; the drain `await` is at `:203`) | ✅ |
| `if (this._draining) return;` at the top of `attemptTokenRefresh()` | `src/gateway.ts:353` | ✅ |
| Re-check `_draining` after `registerBot()` resolves, before `connect()` / heartbeat | `src/gateway.ts:405-413` (socket creation at `:417-419`) | ✅ |

- **Missing:** none.
- **Extra / scope creep:** the only change not enumerated in the description is the `if (!this._draining)` wrap around the failure-path `startHeartbeat()` at `src/gateway.ts:422-424`. It is one line, directly implied by the third bullet (don't restart runtime services while draining), and correct. Not scope creep.
- **Divergence:** none. No behavior change outside the shutdown/refresh path.

## 2. Code quality

**Quality: Approved**

### What I verified locally at `9f66c6e`

| Check | Result |
|---|---|
| `npm run type-check` | clean |
| `npm run lint` (`--max-warnings 0`) | clean |
| `npm test` | **1282 passed / 66 files**, 0 failures |
| `gateway.test.ts` alone | 24 passed |

**Both new tests are genuine, not vacuous.** I swapped `src/gateway.ts` for older revisions while keeping the new tests:

- Against `1871c25` (pre-PR): `recovers when a second kick lands inside the cooldown` fails — `expected 2 to be 3`.
- Against `4dc2354` (previous head): `does not resurrect the gateway when the retry fires during stop() drain` fails — `expected 3 to be 2`, i.e. a third `registerBot` + socket really was constructed during `stop()`. This reproduces the interleaving raised last round, and the fix closes it.

### The three guards are correctly layered

I mutation-tested each guard in isolation. Two of the three are deliberately redundant, which is the right call here:

- Removing the `stop()` pre-drain clear (leaving both `_draining` guards) → suite still green.
- Removing the early `_draining` guard at `:353` (leaving the `stop()` clear) → suite still green.
- Removing the post-`registerBot` re-check at `:405-413` → **suite still green, but the race is back.**

That last one is the important result: `:405-413` is the only guard covering the case where a refresh was **already past** the early check (`isRefreshing = true`, awaiting `registerBot`'s 30s budget at `src/gateway.ts:393-399`) when `stop()` begins. Neither the pre-drain clear nor the `:353` guard helps there. See the P2-1 coverage note below.

### The state machine holds up

- **Timer can only arm on a genuinely dead socket.** `createSocket`'s `onError` refreshes on `Kicked` / `Connect failed` (`src/gateway.ts:329`), and every socket path emitting those strings sets `needReconnect = false` first — `src/octo/socket.ts:453-455` (rapid disconnect), `:758-760` (CONNACK reasonCode 0), `:765-766` (CONNACK other), `:858-862` (DISCONNECT frame). The one path that keeps `needReconnect = true` (`:737-738`, `CONNACK salt too short`) emits a message matching neither filter, so it can never arm the timer. The two reconnect mechanisms do not overlap, and a spurious `force_refresh` cannot land on a healthy socket.
- **No new timer can be armed during drain.** Arming happens only inside `attemptTokenRefresh()`, which now returns at `:353` while draining. So the pre-drain clear at `:187-196` is genuinely final, not a best-effort.
- **Dedup + self-correction.** One arm site guarded by `if (!this.refreshRetryTimer)` (`:363`), nulled in the callback (`:365`), cleared when a real refresh proceeds (`:372-377`). If the timer somehow fired early (timer coarseness under the `+250ms` guard at `:367`), the callback re-enters the cooldown branch and re-arms rather than wedging. Sustained kicks converge to ≤1 refresh per 60s — bounded, and strictly better than the pre-PR permanent wedge.
- **`_draining` is never reset, and that is fine.** No `OctoGateway` instance is reused after `stop()`: `bot-manager.ts` `removeBot` (`:150-164`) drops the bot from the map, and every re-add goes through `StartFn` → a fresh `new OctoGateway(...)`. The hot-reload path called out last round (`config-watcher.ts:51` → `bot-manager.ts:161` → `index.ts:370`, no `process.exit`) is covered by the same guards, as claimed.
- **`isRefreshing` cannot swallow the armed retry.** The retry fires ≤60.25s after the refresh that reset the heartbeat counter, and the only other path into refresh needs 3 consecutive heartbeat failures at 30s intervals (≥90s, `src/gateway.ts:436-464`). No overlap.

### Findings

No P0 / P1. Four P2s, all advisory; three predate this PR.

**P2-1 — the load-bearing guard at `src/gateway.ts:405-413` has zero test coverage.** Deleting that block leaves all 1282 tests green (verified). It is the only guard for the already-in-flight-refresh variant. A scratch test I wrote reproduces it in ~30 lines: make the refresh's `registerBot` hang, kick the socket, `await gw.stop()` with no active handlers (so there is no drain `await` at all), then resolve `registerBot`. With the block: sockets `1 -> 1`. Without it: `1 -> 2` — a socket connected after `stop()` resolved and released the lock. Worth pinning down alongside the drain-window test at `src/__tests__/gateway.test.ts:362`, since the two tests cover different windows and only one of them exists today.

**P2-2 — a refresh that fails is still not re-armed** (`src/gateway.ts:420-424`). `lastRefreshTime` is set at `:380` *before* the attempt, so a transient `registerBot` failure burns the cooldown and schedules nothing; the API heartbeat keeps succeeding while the WS is dead, so `heartbeatFailCount` never trips the fallback. The bot is back to offline-until-restart — the exact failure mode this PR exists to eliminate, one step removed. Pre-existing (the first-kick path had it too), and this PR is strictly an improvement, so it does not block. But it does mean the recovery guarantee is "one retry, if the network cooperates." A bounded re-arm on the failure path would close the loop properly.

**P2-3 — an aborted refresh still leaves a completed server-side `force_refresh`.** When `:405-413` bails, `registerBot(..., forceRefresh: true)` has already run and `:401-403` already overwrote `robotId` / `_ownerUid` / `registration`. If an operator's config edit removes and re-adds a bot, the departing instance's in-flight `force_refresh` can resolve *after* the replacement registered, rotating the token out from under the live instance. Narrow window and pre-existing, but an `AbortSignal` threaded from `stop()` into `registerBot` (`postJson` already accepts one, `src/octo/api.ts:62-67`) would make the abort real rather than cosmetic.

**P2-4 (nit) — the drain-timeout timer is never cleared.** `src/gateway.ts:202` races a `setTimeout(r, drainTimeoutMs)` that `Promise.race` cannot cancel. Single-bot mode follows with `process.exit(0)` so it is invisible there, but on the `removeBot` hot-reload path it holds an unref'd 10s timer after shutdown. Pre-existing and cosmetic.

## 3. Overall verdict

**APPROVE** (Spec ✅ AND Quality Approved).

## 4. Suggested follow-ups

Not blocking, in priority order:

1. Add the in-flight-refresh test for `:405-413` (P2-1) — the sketch above is close to drop-in. Right now a future refactor can delete that block with a green suite.
2. Re-arm a bounded retry when the refresh itself fails (P2-2) — this is the remaining path to the same permanent-offline outcome.
3. Thread an `AbortSignal` from `stop()` through `registerBot` (P2-3).

## 5. Items worth a human look

This PR carries the `needs-human-review` label. Three things I would want a maintainer to confirm rather than take from a static read:

1. **CI has still not run on this head.** `GET /commits/9f66c6e/check-runs` returns only `pr-title-lint`, `label`, and `check-sprint`. The type-check / lint / test table above is my local run, not CI's — this is a fork PR, so the workflow needs maintainer approval to produce an independent signal. (`check-sprint` is failing; it is a project-board field, unrelated to this diff.) **Please approve the workflow run before merging** — a fix in the shutdown path is precisely where you want a second, non-local execution.
2. **Server-side semantics of `force_refresh`.** P2-3 assumes a later `force_refresh` invalidates an earlier-issued `im_token`. If that is true, the abort-after-register window is real; if the server keys tokens per session instead, it is inert. That answer lives in the Octo service, not this repo.
3. **Refresh path does not apply the WS URL policy.** `connect()` refuses a non-loopback `ws://` endpoint at `src/gateway.ts:135-141`, with an explicit rationale (the AES-CBC payload layer is unauthenticated, so transport is the only tamper guarantee). The refresh path at `src/gateway.ts:417` constructs the socket straight from `reg.ws_url` with no such check. **Pre-existing and identical before this PR** — not introduced here and not a reason to block — but this PR makes the refresh path fire automatically where it previously required a manual restart, so the unvalidated branch is now reached more often. A one-line `isAllowedWsUrl` check at `:417` would make the two paths symmetric; worth a separate issue.

---

## Coverage note

Read in full: `src/gateway.ts`, the new tests in `src/__tests__/gateway.test.ts`, `src/bot-manager.ts`, `src/config-watcher.ts`, the gateway lifecycle section of `src/index.ts`, and the `needReconnect` / `onError` paths in `src/octo/socket.ts`. Executed: type-check, lint, full suite, plus three gateway.ts mutations and one scratch reproduction.

Not covered: the live Octo service's `force_refresh` behavior (item 2 above); real-network timing under a genuine kick storm — all timer reasoning here is against vitest fake timers; and `crossUnregisterOnRemove` ordering under a concurrent hot-reload of more than one bot, which I reasoned about but did not exercise.

Jerry-Xin (approved):
Re-review at 9f66c6ed: the shutdown-resurrection race flagged on 4dc23547 is fixed correctly, and the fix is regression-tested. APPROVE.

## Previous-round findings — status

- 🔴 **FIXED** — `stop()` timer-clear timing: `refreshRetryTimer` is now cancelled at the TOP of `stop()`, immediately after `_draining = true` and **before the first await** (`src/gateway.ts:193-196`), so the drain wait can no longer span the timer's expiry.
- 🔴 **FIXED** — `attemptTokenRefresh()` re-entry: a `_draining` guard now bails at the top of the method (`src/gateway.ts:353`), covering the heartbeat-driven path and any timer callback already queued when `stop()` runs.
- 🟡 **ADDRESSED** — in-flight `registerBot()` during shutdown: after registration returns, the refresh re-checks `_draining` and aborts before creating/connecting a socket or restarting the heartbeat (`src/gateway.ts:410-413`); the failure path also skips `startHeartbeat()` while draining (`src/gateway.ts:422-424`). Functionally equivalent to the suggested AbortSignal — an unabortable REST call completing server-side is harmless since normal `stop()` also leaves server-side registration in place.
- 🟡 **UNCHANGED (pre-existing, non-blocking)** — after a *failed* retry refresh the bot still relies on heartbeat failure counting to recover; this gap predates the PR and doesn't block it.

## Verification at this head

- Delta vs 4dc23547 is exactly the fix + one regression test (`src/gateway.ts` +30/-5, `src/__tests__/gateway.test.ts` +49).
- New test reproduces the exact race: retry armed → `stop()` called with an in-flight handler and a 90s drain budget → timers advanced past the retry's expiry mid-drain → asserts no new socket and no re-registration, mid-drain or after `stop()` resolves.
- Race audit at this head: no new windows found. The clear+guard pair is airtight in a single-threaded loop (no await between `_draining = true` and `clearTimeout`); a kick landing mid-drain cannot re-arm the timer (guarded); `_draining` never needs resetting because no code path restarts a stopped instance (signal handlers exit; hot-reload `removeBot()` discards the instance and `addBot()` constructs a fresh one).
- Local run: full test suite 66 files / 1282 tests passed; `tsc --noEmit` clean.

✅ Nice work closing the window with both halves (pre-await clear + draining guard) rather than just one — either alone would have left an edge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #193 (cc-channel-octo)
PR: Mininglamp-OSS/cc-channel-octo#193 — fix(gateway): re-arm reconnect when a kick lands inside the refresh cooldown
Head: `9f66c6edfa5680e0d6023031bd82044dd84d346c` · Base: `main` (merge-base `1871c25`) · Round: re-review (prior pass at head `4dc23547`)

## Summary

A "Kicked by server" event sets the socket's `needReconnect=false`, so token refresh is the only recovery path. If a second kick landed inside the 60s refresh cooldown, the refresh was skipped and nothing rescheduled — the bot stayed offline until a manual restart. This PR arms a single `refreshRetryTimer` in the cooldown branch so exactly one retry fires once the cooldown expires, and (in the follow-up commit) hardens shutdown: `stop()` cancels the timer before its first await, `attemptTokenRefresh()` bails while `_draining`, a refresh whose `registerBot` outlives `stop()` no longer creates a socket, and the failure path skips heartbeat restore while draining. The scheduling logic is sound and the prior round's P2 is implemented; three non-blocking P2 items remain below.

## Verification

Static analysis only at head `9f66c6ed`; build and tests not executed in this environment.

- ✅ **Kick premise holds** — `src/octo/socket.ts:756-760` and `:858-862` set `needReconnect=false` before `onError("Kicked by server")`, and the close handler's stale guard (`:431`) prevents phantom reconnects, so token refresh really is the only recovery path.
- ✅ **Exactly-one-retry invariant** — the timer is armed only when `refreshRetryTimer==null` (`src/gateway.ts:363`) and is cleared on all three exit paths: self-fire (`:364-367`), cooldown-passed passthrough (`:372-378`), and `stop()` pre-await (`:193-195`). The fire time `remainingMs+250` guarantees the cooldown has expired; the 250ms window is covered by the passthrough clear.
- ✅ **No zombie after stop()** — from the post-`registerBot` `_draining` check (`:404-413`) to `startHeartbeat()` the code is synchronous, so `stop()` can only interleave at an await; every continuation after an await re-checks `_draining` or is gated by it. `disconnectAndWait` is idempotent (`src/octo/socket.ts:347-360`), so the narrow timer-fires-just-before-`stop()` race double-disconnects harmlessly and still aborts before creating a socket.
- ✅ **No instance reuse after stop()** — `_draining` is never reset, but every production path constructs a fresh `OctoGateway` (`src/index.ts:267`; bot-manager remove→add creates a new bot), so a stuck `_draining=true` cannot silence a restarted gateway.

## Findings

No P0/P1 issues; three P2 items below.

### P2 — Refresh reconnect bypasses the `isAllowedWsUrl()` transport policy (`src/gateway.ts:417`)

`connect()` refuses plaintext `ws://` endpoints for non-loopback hosts (`src/gateway.ts:135`) because the AES-CBC payload layer is unauthenticated and transport integrity is the only tamper guarantee. The refresh path constructs the socket directly via `createSocket(reg.ws_url, ...)` at `:417` with no such check, so any kick/heartbeat/timer-driven refresh reconnects to a `ws://` URL the startup path would reject. Diff-scope: pre-existing at base (base `gateway.ts:365`), not amplified in severity, but this PR inserts the draining guard immediately above and adds a new autonomous trigger into this path, so the hunk is in scope. Fix: apply `isAllowedWsUrl(reg.ws_url)` here and abort like `connect()` does.

### P2 — New post-`registerBot` draining guard has no regression test (`src/gateway.ts:411`)

The "registerBot resolves after `stop()` began → do not create a socket" branch is production-reachable per the PR's own comment (registerBot can outlive the drain window on its own 30s budget), but both new tests cover only the timer-firing-mid-drain case. Suggested test: kick → refresh with `registerBot` mocked to a controllable pending promise → `gw.stop(activeHandlers, large)` → resolve `registerBot` → assert no new `WKSocket` and no heartbeat after `stop()` resolves.

### P2 — Refresh failure leaves no scheduled recovery while the REST heartbeat is healthy (`src/gateway.ts:422`)

If the (possibly timer-driven) refresh fails, the catch branch only restores the heartbeat. When `registerBot` fails transiently but `sendHeartbeat` keeps succeeding (same `apiUrl`; e.g. a forceRefresh-specific server error), nothing re-triggers `attemptTokenRefresh`: the socket is null, heartbeat stays green, and the bot is wedged offline until manual restart or an external kick. Diff-scope: pre-existing (base had the identical catch-without-retry), not amplified — the PR strictly reduces wedge probability — but it is the same wedge class this PR targets. Fix: on failure (when `!_draining`), arm `refreshRetryTimer` with the same cooldown-based scheduling.

## Human-verify

1. Server-side kick semantics: the client assumes a fresh `registerBot(forceRefresh=true)` invalidates/kicks the previous session and that a registration with no following WS connect simply leaves the bot offline. Outside this checkout — please confirm. **Not a merge blocker for this PR.**
2. Whether the registration API can ever return a non-loopback `ws://` (misconfiguration/failover). If it cannot, finding P2#1 is defense-in-depth only. **Not a merge blocker for this PR.**

- Existing cooldown tests stay compatible: the timer armed by a second kick is cancelled by `gw.stop()`, and fake timers/mocks are cleaned up in before/afterEach.
- All three `attemptTokenRefresh` call sites (socket `onError`, heartbeat-failure streak, retry timer) pass through the same `isRefreshing` + `_draining` guards; nested calls while `isRefreshing` return early without re-arming (single-threaded, no double effects).
- Multi-bot mode: per-instance timers/state; `removeBot()`/`shutdownAll()` both route through `stop()`, so the timer is cancelled on every teardown path.
- Kicks arriving mid-drain are absorbed by the top-of-function `_draining` guard; no new arming is possible once draining.

## Verdict: APPROVED

The scheduling invariant (exactly one retry after cooldown), the shutdown hardening, and the prior round's recommended guard are all correctly implemented with matching regression tests. The three P2 items are non-blocking: one pre-existing transport-policy parity gap on the refresh path, one missing test for the new in-flight-abort guard, and one pre-existing liveness gap on refresh failure.

---

## 数据流回溯（Octo 附加段）

| 被消费数据 | 上游来源 | 是否真流到消费点 |
|---|---|---|
| `refreshRetryTimer` | 仅 cooldown 分支在 `!_draining && !isRefreshing && timer==null` 时置位（`src/gateway.ts:363-369`） | ✅ 三个消费点全覆盖：自触发置空、cooldown 通过时清除、`stop()` 首 await 前清除；无泄漏路径 |
| `remainingMs`（定时延时） | `REFRESH_COOLDOWN_MS - (now - lastRefreshTime)`，`lastRefreshTime` 在刷新开始时写入（`:379-380`） | ✅ 触发时刻恒满足 cooldown 已过（+250ms 余量）；250ms 窗口内其他触发进入真实刷新时会清掉定时器（`:372-378`） |
| `_draining` | `stop()` 第一行置 true（`:184`），从不复位 | ✅ 消费点：`handleMessage`、`attemptTokenRefresh` 顶部守卫、registerBot 后中止、catch 心跳门控；生产无实例复用（`src/index.ts:267` 每次新建），不复位安全 |
| kick 事件 → 刷新 | `socket.ts:756-760 / :858-862` `needReconnect=false` + `onError` → `createSocket` 注入的 `onError` 回调（`gateway.ts:331-335`） | ✅ 前提成立：kick 后 socket 无自愈，刷新是唯一恢复路径 |
| `reg.ws_url`（刷新路径） | `registerBot` 响应 | ⚠️ 直接进 `createSocket` 未经 `isAllowedWsUrl`（见 P2#1） |
| stop()/drain 交互 | `index.ts` shutdown 回调、bot-manager `removeBot`/`shutdownAll` | ✅ 全部经由 `stop()`，定时器必被清除 |

## 盲点 Checklist（security_sensitive，全项）

- **C1 双路径 parity** — HIT：启动 `connect()` 与刷新重连这对对称路径上 `isAllowedWsUrl` 执行不一致（P2#1）。arm↔clear 对称性已逐对核验：三处 clear 覆盖全部退出路径，CLEAR。
- **C2 control-flow ordering / 复用** — CLEAR：`attemptTokenRefresh` 三个 call-site 共用同一守卫序列；`isRefreshing` 检查到置位之间无 await，单线程下无交错；嵌套调用早退不重复 arm。
- **C3 授权边界** — CLEAR：无新 endpoint/tool/凭证暴露；定时器纯内部，仍用既有 botToken 走原 `registerBot`。
- **C4 授权生命周期级联** — CLEAR：本 diff 正是生命周期级联加固（stop → 刷新中止）；无父容器/成员状态级联语义。
- **C5 build/测试通过 ≠ 运行期正确** — 已做运行期推演（drain 竞态、250ms 窗口、定时器先于 stop 触发的窄竞态均逐帧分析）；测试经 mock seam 驱动的是真实调度逻辑；未覆盖的 in-flight-abort 守卫已作为 P2#2 提出。本环境未实际跑测试（静态审查）。
- **C6 治理/策略文档自洽** — N/A：本 PR 未改任何文档/治理文件。

## 跨轮 Blocker 复检（R6）

上一轮（[automated review](mention://issue/84dcfd73-3a78-442b-94fc-f40ccbfa9eee)，head `4dc23547`）：0 P1；1 P2 建议——「refreshRetryTimer/在途刷新可在 stop() 后复活 socket：attemptTokenRefresh 顶部加 `_draining` 守卫 + stop() abort registerBot」+ 备注 pre-existing「刷新失败且 API 健康时无恢复调度」。本轮逐条核验（新 commit `9f66c6e`）：

1. `_draining` 顶部守卫 — **已修**（`src/gateway.ts:353`，新增）。
2. `stop()` 首 await 前清定时器 — **已修**（`src/gateway.ts:187-195`，新增）。
3. in-flight registerBot 中止（不建 socket）— **已修**（`src/gateway.ts:404-413`，新增；但缺回归测试 → 本轮 P2#2）。
4. catch 心跳恢复加 `!_draining` 门控 — **已修**（`src/gateway.ts:422-424`，新增）。
5. pre-existing「刷新失败无恢复调度」— **仍在**，与上轮定性一致，pre-existing 未放大 → 本轮 P2#3，不阻塞。

上轮无未解决 P0/P1；其 P2 已全部落地。本轮 0 P0/P1 → 按 R4 建议 APPROVED。

[Octo-Q] verdict: APPROVE — 0 P0/P1；cooldown 重试的 exactly-one 不变量、stop() 竞态硬化、上轮 P2 守卫均验证成立且有针对性回归测试。3 个 P2 均非阻塞（刷新路径缺 `isAllowedWsUrl` parity【pre-existing，本 hunk 触碰】、新中止守卫缺回归测试、刷新失败后无恢复调度【pre-existing】），供终审参考。

